PR #168, opened 2023-03-22
[bloqs] Auto-dagger

mpharrigan:
Issue #172 

Automatically compute the adjoint of a composite bloq by reversing the order of operations and getting the adjoint of all subbloqs.


mpharrigan:
closing in favor of #565

PR #190, opened 2023-04-10
Clean notebooks

mpharrigan:
Example script that can strip notebook metadata and optionally be used as a CI check. I didn't actually commit the results in this PR


mpharrigan:
I came back to this old PR. I updated it to use the officially-supported nbconvert tooling for stripping details from notebooks

PR #300, opened 2023-07-11
Qubitization top-down skeleton

### LINKED ISSUE #272: Implement high-level qubitization primitives in bloqs
Implement phase estimation of a walk operator made up of select and prepare in bloqs.

mpharrigan:
image

This implements a general skeleton for doing a qubitized walk.

 - `PhaseEstimation` will set up a phase register and do controlled powers of `Walk`
 - `Walk` will do `Select` and `Reflect(Prepare)`

------

`BlackBoxSelect` and `BlackBoxPrepare` are meta-bloqs that expose a consistent signature of "selection" and "system" . They decompose to partitioning primitives that match up to more expressive implementations of the `Select` and `Prepare` interface which can have many selection registers. In this PR, there's just one system register but you can see how it would be extended in the same way. 

-------

This PR lets bloqs express how they're controlled by overriding either `Bloq.controlled()` or `Bloq.add_controlled(...)`. You can see this in action: `Reflect.controlled()` updates its `cv` parameter while `Partition.add_controlled(...)` will ignore the control line since partitioning is a bookkeeping operation. 

Fixes #272


mpharrigan:
Closing. The spiritual successor is #508

PR #327, opened 2023-07-19
Add Azure Model

NoureldinYosri:
Reproduction of the estimates in https://arxiv.org/abs/2211.07629

### DIFF
--- a/qualtran/surface_code/algorithm_specs.py
+++ b/qualtran/surface_code/algorithm_specs.py
@@ -1,0 +1,40 @@
+#  Copyright 2023 Google LLC
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+
+import attr
+from attrs import frozen
+
+_PRETTY_FLOAT = attr.ib(type=float, default=0.0, repr=lambda x: f'{x:g}')
+
+
+@frozen
+class AlgorithmSpecs:
+    """AlgorithmSpecs is a summary of a quantum algorithm/circuit.
+
+    Attributes:
+        algorithm_qubits: Number of qubits used by the algorithm.
+        measurements: Number of Measurements.
+        t_gates: Number of T gates.
+        toffoli_gates: Number of Toffoli gates.
+        rotation_gates: Number of Rotations.
+        rotation_circuit_depth: Depth of rotation circuit.
+    """
+
+    algorithm_qubits = _PRETTY_FLOAT  # Number of algorithm qubits $Q_{alg}$
+    measurements = _PRETTY_FLOAT  # Number of measurements $M_R$.
+    t_gates = _PRETTY_FLOAT  # Number of T gates $M_T$.
+    toffoli_gates = _PRETTY_FLOAT  # Number of Toffoli gates $M_{Tof}$.
+    rotation_gates = _PRETTY_FLOAT  # Number of Rotations $M_R$.
+    rotation_circuit_depth = _PRETTY_FLOAT  # Depth of rotation circuit $D_R$.
--- a/qualtran/surface_code/azure_cost_model.ipynb
+++ b/qualtran/surface_code/azure_cost_model.ipynb
@@ -1,0 +1,726 @@
+{
+ "cells": [
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Azure Cost Model\n",
+    "In this notebook, we go through the cost model in \"Assessing requirements to scale to practical quantum advantage\" by [Beverland et al](https://arxiv.org/abs/2211.07629) to reproduce the estimated costs in table IV."
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The Azure cost model has 5 inputs, these in no particular order are:\n",
+    "1. The specs of the circuit/algorithm to execute (e.g. number of T gates, ..etc).\n",
+    "1. The physical assumptions about the hardware (e.g. error rate, the latency of Clifford and measurement operations, ...etc).\n",
+    "1. The magic state factory (e.g. number of qubits consumed by it, its error rate, ...etc)\n",
+    "1. Cost of Approximating rotations up to error $\\epsilon$\n",
+    "1. The quantum error correction scheme.\n",
+    "\n",
+    "We take a look at each of these and then reproduce the results of appendix F."
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## The Inputs\n",
+    "### Algorithm Specs\n",
+    "The first input to the estimation is a summary of the circuit or algorithm to execute. This summary is a couple of simple counts:\n",
+    "- `algorithm_qubits`: Number of algorithm qubits.\n",
+    "- `measurements`: Number of Clifford measurements ($M_R$).\n",
+    "- `t_gates`: Number of T gates ($M_T$).\n",
+    "- `toffoli_gates`: Number of Toffoli gates ($M_{Tof}$).\n",
+    "- `rotation_gates`: Number of Rotations ($M_R$).\n",
+    "- `rotation_circuit_depth`: Depth of rotation circuit  ($D_R$).\n",
+    "\n",
+    "Note: where we add the symbol used to refer to the quantity in the paper in parentheses.\n",
+    "\n",
+    "In Qualtran the algorithm specs are represented by the data class `qualtran.surface_code.class AlgorithmSpecs`."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from qualtran.surface_code import AlgorithmSpecs\n",
+    "\n",
+    "shor_specs = AlgorithmSpecs(\n",
+    "    algorithm_qubits=12581,\n",
+    "    rotation_gates=12,\n",
+    "    measurements=1.08e9,\n",
+    "    rotation_circuit_depth=12,\n",
+    "    toffoli_gates=3.73e10,\n",
+    "    t_gates=12,\n",
+    ")\n",
+    "shor_specs"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Physical Parameters\n",
+    "These are assumptions about the quantum hardware and are:\n",
+    "- `t_gate`: Latency of Clifford gates. \n",
+    "- `t_meas`: Latency of Measurement operations.\n",
+    "- `physical_error_rate`: Physical error rate ($p$)."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from qualtran.surface_code.physical_parameters import BeverlandEtAl as physical_parameters\n",
+    "\n",
+    "physical_parameters"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Magic State Factory\n",
+    "\n",
+    "The magic state factory in our case is a T-state factory and has parameters:\n",
+    "- `num_qubits`: Number of qubits consumed by the factory.\n",
+    "- `t_states_rate`: Number of T states produces per factory cycle.\n",
+    "- `duration`: Duration of the factory cycle."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from qualtran.surface_code import TFactory\n",
+    "\n",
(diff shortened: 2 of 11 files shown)


mpharrigan:
Let me know when you'd like a full review, but please be advised

 - it's "Azure"
 - it's "Gidney"
 - Pick either "Azure model" or "Beverland et al" model and be consistent with the names of things

qualtran/surface_code/algorithm_specs.py
```diff
@@ -0,0 +1,29 @@
...
+
+import attr
+from attrs import frozen
+
+_PRETTY_FLOAT = attr.ib(type=float, default=0.0, repr=lambda x: f'{x:g}')
+
+
+@frozen
+class AlgorithmSpecs:
+    algorithm_qubits = _PRETTY_FLOAT  # Number of algorithm qubits $Q_{alg}$
```
mpharrigan, line 24: please document with a standard class docstring. Document each member attribute in the `Attributes` section. (The doc system expects this)

qualtran/surface_code/algorithm_specs.py
```diff
@@ -0,0 +1,29 @@
...
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+
+import attr
+from attrs import frozen
+
+_PRETTY_FLOAT = attr.ib(type=float, default=0.0, repr=lambda x: f'{x:g}')
```
mpharrigan, line 19: so all of these are optional? Why are they all floats? I would expect most of the "counts" to be integers

qualtran/surface_code/azure_cost_model.py
```diff
@@ -0,0 +1,78 @@
...
+from qualtran.surface_code.algorithm_specs import AlgorithmSpecs
+from qualtran.surface_code.magic_state_factory import MagicStateCount
+from qualtran.surface_code.physical_parameters import PhysicalParameters
+from qualtran.surface_code.quantum_error_correction_scheme import QuantumErrorCorrectionScheme
+from qualtran.surface_code.rotation_cost_model import RotationCostModel
+from qualtran.surface_code.t_factory import TFactory
+
+
+def logical_qubits(algorithm_specs: AlgorithmSpecs) -> int:
+    """Equation D1 in https://arxiv.org/abs/2211.07629"""
```
mpharrigan, line 28: can you document why this is the formula in the docstring and then provide the equation reference as a reference? Ideally this file could be a pretty self-contained description of a cost model

qualtran/surface_code/azure_cost_model.py
```diff
@@ -0,0 +1,78 @@
...
+    q_alg = algorithm_specs.algorithm_qubits
+    return 2 * q_alg + math.ceil(math.sqrt(8 * q_alg)) + 1
+
+
+def minimum_time_steps(
+    error_budget: float | np.ndarray, alg: AlgorithmSpecs, rotation_cost: RotationCostModel
+) -> int | np.ndarray:
```
mpharrigan, line 35: document and say what this corresponds to in the paper

qualtran/surface_code/azure_cost_model.py
```diff
@@ -0,0 +1,78 @@
...
+    return c_min
+
+
+def code_distance(
+    error_budget: float | np.ndarray,
+    time_steps: float | np.ndarray,
+    alg: AlgorithmSpecs,
+    qec_scheme: QuantumErrorCorrectionScheme,
+    physical_parameters: PhysicalParameters,
+) -> int:
```
mpharrigan, line 50: document and consider changing the name of the function to be more specific. 

This is the particular heuristic that the Azure paper uses for choosing a code distance, yes? Since this is a particular choice for the backwards, design problem -- we should use a particularly explicit function name

qualtran/surface_code/azure_cost_model.py
```diff
@@ -0,0 +1,78 @@
...
+    d = np.clip(np.ceil(d), a_min=3, a_max=None)
+    if isinstance(d, float):
+        d = int(d)
+    else:
+        d = np.array(d, dtype=np.uint64)
+    d += 1 - (d & 1)
+    return d
+
+
+def t_states(
```
mpharrigan, line 64: document and consider a more explicit function name

qualtran/surface_code/azure_cost_model.py
```diff
@@ -0,0 +1,78 @@
...
+) -> float | np.ndarray:
+    eps_syn = error_budget / 3
+    return (
+        alg.t_gates
+        + 4 * alg.toffoli_gates
+        + alg.rotation_gates * np.ceil(rotation_cost.mean_cost(eps_syn / alg.rotation_gates))
+    )
+
+
+def number_t_factories(
```
mpharrigan, line 75: document

why are all these things `float | np.ndarray`? what happens if I pass an array of numbers?

qualtran/surface_code/azure_cost_model_test.py
```diff
@@ -0,0 +1,176 @@
...
+from qualtran.surface_code import azure_cost_model
+from qualtran.surface_code.algorithm_specs import AlgorithmSpecs
+from qualtran.surface_code.physical_parameters import BeverlandEtAl as physical_parameters
+from qualtran.surface_code.quantum_error_correction_scheme import GateBasedSurfaceCode
+from qualtran.surface_code.rotation_cost_model import BeverlandEtAl as rotation_cost_model
+from qualtran.surface_code.t_factory import TFactory
+
+
+@frozen
+class Test:
```
mpharrigan, line 28: please use a more specific name

qualtran/surface_code/azure_cost_model_test.py
```diff
@@ -0,0 +1,176 @@
...
+        c_min=1.23e11,
+        time_steps=1.23e10,
+        code_distance=13,
+        t_states=1.49e11,
+    ),
+]
+
+
+@pytest.mark.parametrize('test', _TESTS)
+def test_logical_qubits(test: Test):
```
mpharrigan, line 92: instead of using `pytest.mark.parameterize` this calls for a pytest test fixture

qualtran/surface_code/physical_parameters.py
```diff
@@ -0,0 +1,43 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import attr
+from attrs import frozen
+
+
+@frozen
+class PhysicalParameters:
```
mpharrigan, line 20: document

qualtran/surface_code/physical_parameters.py
```diff
@@ -0,0 +1,43 @@
...
+    t_gate = attr.ib(type=float, default=1e-6, repr=lambda x: f'{x:g}')  # 1us
+    t_meas = attr.ib(type=float, default=1e-6, repr=lambda x: f'{x:g}')  # 1us
```
mpharrigan, line 22: please use a variable name suffix to imply the unit. i.e. `t_gate_us`. Also consider using microseconds as the base unit

qualtran/surface_code/physical_parameters.py
```diff
@@ -0,0 +1,43 @@
...
+BeverlandEtAl = PhysicalParameters(
+    t_gate=50 * 1e-9,
+    t_meas=100 * 1e-9,
+    physical_error_rate=1e-4,
+    reference='https://arxiv.org/abs/2211.07629',
+)
+
+
+# error_rate_scaler=0.03,
+# error_rate_threshold=0.01,
```
mpharrigan, line 43: remove

qualtran/surface_code/quantum_error_correction_scheme.py
```diff
@@ -0,0 +1,72 @@
...
+from qualtran.surface_code.physical_parameters import PhysicalParameters
+
+
+@frozen
+class QuantumErrorCorrectionScheme:
+    error_rate_scaler = attr.ib(type=float, default=0.1, repr=lambda x: f'{x:g}')
+    error_rate_threshold = attr.ib(type=float, default=0.01, repr=lambda x: f'{x:g}')
+    reference = attr.ib(type=str, default='')
+
+    def logical_error_rate(
```
mpharrigan, line 28: there's a function elsewhere that does this. If you intend to replace it, please remove the existing function and update usages

qualtran/surface_code/quantum_error_correction_scheme.py
```diff
@@ -0,0 +1,72 @@
...
+        self, code_distance: int | np.ndarray, physical_error_rate: float | np.ndarray
+    ) -> float | np.ndarray:
+        """Computes the logical error rate."""
+        return self.error_rate_scaler * np.power(
+            physical_error_rate / self.error_rate_threshold, (code_distance + 1) / 2
+        )
+
+    def physical_qubits(self, code_distance: int | np.ndarray) -> int | np.ndarray:
+        """Computes number of physical qubits"""
```
mpharrigan, line 37: mark as abstract or raise NotImplementedError

qualtran/surface_code/quantum_error_correction_scheme_test.py
```diff
@@ -0,0 +1,77 @@
...
+        [
+            qecs.MeasurementBasedHastingsHaahCode(
+                error_rate_scaler=0.05, error_rate_threshold=0.06
+            ),
+            564,
+        ],
+    ],
+)
+def test_physical_qubits(qec: qecs.QuantumErrorCorrectionScheme, want: int):
+    assert qec.physical_qubits(11) == want
```
mpharrigan, line 55: I guess my question is why bother going through the parameterization rigamarole when you're just factoring out `assert ==`

DRY does not apply to tests. https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html

qualtran/surface_code/rotation_cost_model.py
```diff
@@ -0,0 +1,70 @@
...
+    def mean_cost(self, error_budge: float | np.ndarray) -> float:
+        """Returns the mean number of T gates needed to approx a rotation."""
+
+    @abc.abstractmethod
+    def max_cost(self, error_budge: float | np.ndarray) -> float:
+        """Returns the max number of T gates needed to approx a rotation."""
+
+
+@frozen
+class RotationCostLinearModel(RotationCostModel):
```
mpharrigan, line 35: document

qualtran/surface_code/azure_cost_model.ipynb
```diff
@@ -0,0 +1,726 @@
...
+    "### Algorithm Specs\n",
+    "The first input to the estimation is a summary of the circuit or algorithm to execute. This summary is a couple of simple counts:\n",
+    "- `algorithm_qubits`: Number of algorithm qubits.\n",
+    "- `measurements`: Number of Clifford measurements ($M_R$).\n",
+    "- `t_gates`: Number of T gates ($M_T$).\n",
+    "- `toffoli_gates`: Number of Toffoli gates ($M_{Tof}$).\n",
+    "- `rotation_gates`: Number of Rotations ($M_R$).\n",
+    "- `rotation_circuit_depth`: Depth of rotation circuit  ($D_R$).\n",
+    "\n",
+    "Note: where we add the symbol used to refer to the quantity in the paper in parentheses.\n",
```
mpharrigan, line 42: grammar (?)

> Note: The symbol in parentheses corresponds to the notation in the paper

qualtran/surface_code/azure_cost_model.ipynb
```diff
@@ -0,0 +1,726 @@
...
+    "\n",
+    "In Qualtran the algorithm specs are represented by the data class `qualtran.surface_code.class AlgorithmSpecs`."
```
mpharrigan, line 44: `.class`?

qualtran/surface_code/azure_cost_model.ipynb
```diff
@@ -0,0 +1,726 @@
...
+   "source": [
+    "from qualtran.surface_code import AlgorithmSpecs\n",
+    "\n",
+    "shor_specs = AlgorithmSpecs(\n",
+    "    algorithm_qubits=12581,\n",
+    "    rotation_gates=12,\n",
+    "    measurements=1.08e9,\n",
+    "    rotation_circuit_depth=12,\n",
+    "    toffoli_gates=3.73e10,\n",
+    "    t_gates=12,\n",
```
mpharrigan, line 61: what are these numbers?

qualtran/surface_code/azure_cost_model.ipynb
```diff
@@ -0,0 +1,726 @@
...
+    "- `physical_error_rate`: Physical error rate ($p$)."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from qualtran.surface_code.physical_parameters import BeverlandEtAl as physical_parameters\n",
```
mpharrigan, line 84: there's some weird pep8 things going on here. I think these physical parameter instantiations should be in all caps because they're a constant. I would try to avoid assigning them to a different name using "as". Use `=` assignment in a different line or (even better:) just use the explicit name

mpharrigan (commented):
nice notebook! I'm leaving some quick comments on the overall PR.

I strongly suggest you split this into multiple, small PRs so we can quickly iterate on the parts. I think it should be straightforward to factor: let's start with the new data classes. 

Please make sure everything is documented. The actual formulae (as you know) are pretty simple arithmetic computations. The challenge is understanding what all the moving parts actually are! The notebook is good; but lets make sure we have the functions and data documented as well

NoureldinYosri:
@mpharrigan I created #356 for the creation of the data classes. the reason the numbers are float is that as long as they are not small they are usually given as estimates (e.g. 1e10 measurements) and because sometimes they are estimates themselves (e.g. number of T gates needed to approximate a rotation)

the type is `float | np.ndarray` is to allow calculating the estimates fo several values at the same time since that doesn't change the code. for example to get the code distance given different time duration we can do
`code_distance(error_budget, durations, ...etc)` which will return the code distance for each duration as an `np.ndarray`

PR #331, opened 2023-07-25
Release v0.1

mpharrigan:
- Set the version to a released version. I'll tag this commit when merged to `main`. 
 - Need a manifest file for the sdist to work, so that's also included
 - Then I'll open a PR to bump the version spec to `"0.2.0.dev0"`

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include dev_tools/requirements/deps/runtime.txt
--- a/qualtran/_version.py
+++ b/qualtran/_version.py
@@ -12,4 +12,4 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-__version__ = "0.1.0dev0"
+__version__ = "0.1.0"


tanujkhattar:
The way we do releases in Cirq is that we create a separate release branch (so this would not be a PR but a commit to a branch named `v0.1.dev`) and tag this commit with a `v0.1.0` tag and attach it to the github release when you create it. 

This model is useful because if you want to do a patch release to a minor release, you can cherry pick commits from the main branch and apply them to your release branch and do a patch release without necessarily including all the changes done to master till that point; which can become part of the follow-up minor release. 

See https://github.com/quantumlib/Cirq/blob/master/release.md#release-procedure


So, for this PR, let's do the following:

1) Include only `MANIFEST.in` as part of this PR, since it's required to do the release. 
2) Create a new `v0.1.0dev` branch where that updates the `qualtran/_version.py`. Create a release using this branch
3) Send a PR to update the version on main to `0.2.0.dev`

mpharrigan:
That seems like overkill (especially pre-1.0) and breaks the github "changes since this release" feature. Let's stick to `main` and create release branches only in the (rare) case we need a patch release.

tanujkhattar:
What's the `github "changes since this release" feature` that you are talking about? The github notes for cirq releases are auto generated using the "previous tag: auto" feature, so it is able to infer changes since the previous release.

mpharrigan:
image

tanujkhattar (approved):

PR #332, opened 2023-07-25
Bump version to 0.2.0.dev0

mpharrigan:
Following #331 and https://github.com/quantumlib/Qualtran/releases/tag/v0.1.0

### DIFF
--- a/qualtran/_version.py
+++ b/qualtran/_version.py
@@ -12,4 +12,4 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-__version__ = "0.1.0"
+__version__ = "0.2.0.dev0"


qualtran/_version.py
```diff
@@ -12,4 +12,4 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-__version__ = "0.1.0"
+__version__ = "0.2.0.dev0"
```
tanujkhattar, line 15: Why `0.2.0.dev0` and not `0.2.0.dev` ?

tanujkhattar (approved):

qualtran/_version.py
mpharrigan: xref https://peps.python.org/pep-0440/#implicit-development-release-number

Issue #149, opened 2023-03-13
ControlledBloq features


mpharrigan:
- [ ] multiple control bits
 - [ ] inverted controls ("open circle" controls)
 - [ ] Flattening with existing control register

I propose we add an additional argument that gives the name of the desired control register so that the user can have control (heh) over the flattening. Example:

```python
ca = Controlled(Atom(), ctrl_reg_name='control')
cca = Controlled(ca, reg_name='control')  # `control` register flattened

ca = Controlled(Atom(), 'control1')
cca = Controlled(Atom(), 'control2')  # two regs, not flattened
```

mpharrigan:
This is done

PR #337, opened 2023-07-30
QROM bloq

fdmalone:
Attempt at reusing cirq_ft primitives in bloqs. This is mostly a request for feedback if this is the preferred approach.

In particular:

- QROM is a bit of a special case as it requires data to load. This would make the bloq unhashable which was an issue we hit in the cirq_ft variant. Here I spoof the cirq QROM build with zeros, with the target_bitsizes / data_bitsizes and data shape expected as input on the bloqs side.
- How should testing work in this model. i.e. do I need to retest the qrom circuit given it is carefully tested in cirq land? Maybe QROM is a bad first example due to needing data, as quimb simulation could be helpful too.

Anyway, I need several cirq_ft primitives (SelectedMajorana, Uniform preparation, ....) for the chemistry block encodings, so can just keep adding as I go.


fdmalone:
@mpharrigan @tanujkhattar thoughts?

qualtran/bloqs/qrom.py
```diff
@@ -0,0 +1,71 @@
...
+        regs = [
+            Register(f"selection{i}", bitsize=bs) for i, bs in enumerate(self.selection_bitsizes)
+        ]
+        regs += [Register(f"target{i}", bitsize=bs) for i, bs in enumerate(self.data_bitsizes)]
+        if len(self.cvs) > 0:
+            regs += [Register("control", bitsize=1, shape=(len(self.cvs),))]
+        return Signature(regs)
+
+    def decompose_bloq(self) -> 'CompositeBloq':
+        return decompose_from_cirq_op(self)
```
mpharrigan, line 60: @tanujkhattar the fact that `decompose_from_cirq_op()` doesn't take a cirq op as its argument is a bit counterintuitive

qualtran/bloqs/qrom.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def as_cirq_op(
+        self, qubit_manager: 'cirq.QubitManager', **cirq_quregs: 'CirqQuregT'
+    ) -> Tuple[Union['cirq.Operation', None], Dict[str, 'CirqQuregT']]:
```
mpharrigan, line 64: I think the first return value type can be `'cirq.Operation'` instead of Union[op, None] since we know it's not None

qualtran/bloqs/qrom_test.py
```diff
@@ -0,0 +1,29 @@
...
+from cirq_ft.infra import t_complexity
+
+from qualtran.bloqs.qrom import QROM
+
+
+def test_qrom():
+    data = np.zeros((10, 10))
+    sel_bitsizes = tuple((s - 1).bit_length() for s in data.shape)
+    qrom = QROM(data_bitsizes=(4,), selection_bitsizes=sel_bitsizes, data_ranges=data.shape)
+    cbloq = qrom.decompose_bloq()
```
mpharrigan, line 27: In tests, you can replace `cbloq = qrom.decompose_bloq()` with

```python
import qualtran.testing as qlt_testing

cbloq = qlt_testing.assert_valid_bloq_decomposition(qrom)
```

which will do a whole bunch of checks that you have created a well-formed bloq

mpharrigan (commented):
Can you provide more context on the data/hashability thing? Presumably the cirq version of the gate takes in data and is hashable and doesn't just use a bunch of zero placeholders :)

mpharrigan:
I don't think reproducing all the tests from cirq is a prerequisite for including this

qualtran/bloqs/qrom.py
tanujkhattar: That's because `decompose_from_cirq_op` is primarily responsible for constructing the cirq operation from the corresponding bloq and then delegating to `cirq_optree_to_cbloq`.

qualtran/bloqs/qrom.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @cached_property
+    def signature(self) -> Signature:
+        regs = [
+            Register(f"selection{i}", bitsize=bs) for i, bs in enumerate(self.selection_bitsizes)
+        ]
+        regs += [Register(f"target{i}", bitsize=bs) for i, bs in enumerate(self.data_bitsizes)]
+        if len(self.cvs) > 0:
+            regs += [Register("control", bitsize=1, shape=(len(self.cvs),))]
+        return Signature(regs)
```
tanujkhattar, line 57: @mpharrigan I think we should also have a `signature_from_cirq_op` conversion method for convenience here, especially for cases when the signature matches 1:1 

Also, I'm considering to add a `bitsize` argument to cirq registers, even though it would essentially just expand out everything into a flat qubit list API for now. This would make sure the traditional case of only having thru registers has 1:1 correspondence with cirq registers and the additional structure is not lost with a cirq conversion

fdmalone:
@mpharrigan  There was some discussion at the time in #242. Essentially the problem will be the QROM bloq can't have a frozen attribute, and in cirq land we needed to use cirq.value_equality (I think the TComplexity protocol requires hashability?) Maybe this is fixed now with bloq_counts taking precedence?

qualtran/bloqs/qrom.py
```diff
@@ -0,0 +1,71 @@
...
+    cvs: Tuple[int, ...] = ()
+
+    @cached_property
+    def signature(self) -> Signature:
+        regs = [
+            Register(f"selection{i}", bitsize=bs) for i, bs in enumerate(self.selection_bitsizes)
+        ]
+        regs += [Register(f"target{i}", bitsize=bs) for i, bs in enumerate(self.data_bitsizes)]
+        if len(self.cvs) > 0:
+            regs += [Register("control", bitsize=1, shape=(len(self.cvs),))]
```
tanujkhattar, line 56: @mpharrigan I'm not sure if `control` register should be `bitsize=1, shape=(len(self.cvs),)` or `bitsize=len(self.cvs)`; what do you think?

qualtran/bloqs/qrom.py
```diff
@@ -0,0 +1,71 @@
...
+            registers. This can be deduced from the maximum element of each of the
+            datasets. Should be of length len(data), i.e. the number of datasets.
+        data_ranges: Tuple specifying the ranges of each axis of the data to be loaded.
+        cvs: The control values for the gate. Defaults to no controls.
+    """
+
+    selection_bitsizes: Tuple[int, ...]
+    data_bitsizes: Tuple[int, ...]
+    data_ranges: Tuple[int, ...]
+    cvs: Tuple[int, ...] = ()
```
tanujkhattar, line 47: QROM should take the data to be loaded as input? What's the logic for not specifying the data? You want to create a placeholder bloq that has the right signature but don't care too much about the actual underlying data?

tanujkhattar (changes requested):

qualtran/bloqs/qrom.py
fdmalone: Right yeah, maybe I'm working off old information, but do we expect bloqs to be hashable?

fdmalone: Ideally this wouldn't be the case and we could test the correctness on the qualtran side too. Given the importance of qrom for other subroutines it's better to be able to check for correctness, so I'll make it mirror cirq_ft (in that it expects the actual data).

tanujkhattar: Yes, Bloqs should be hashable.

fdmalone: numpy arrays not being hashable lead us to use cirq.value_equality in cirq_ft, I wasn't sure what the appropriate thing to do in qualtran.

qualtran/bloqs/qrom.py
mpharrigan: > we should also have a signature_from_cirq_op conversion method 

+1

>  I'm considering to add a bitsize argument to cirq registers

That sounds ok a priori. We just need to make sure that as the distinction between cirq_ft.Register and qualtran.Register becomes more subtle it doesn't lead to unnecessary confusion

qualtran/bloqs/qrom.py
mpharrigan: I think we'll have a clearer answer to this after fleshing out the `ControlledBloq` protocol (see #149 and the controlled changes in #300 ) 

Barring that, I think an array of bools represents the intent here. Are we trying to express

```python
if all(cond for cond in conditions):
 active()
```

or 

```python
if ctrl_val == 32:
  active()
```

qualtran/bloqs/qrom.py
fdmalone: Ok, after talking to Matt I think it's best to go the cirq.value_equality route (implement the necessary __hash__ etc routines and don't compare on the data for the moment), just to make it a bit more transparent. I will open an issue to track this. The alternative would be to just have a bloq to count resources, but I think it would be nice to be able to rely on qrom for some testing.

mpharrigan (approved):
looks good for a start

fdmalone:
@tanujkhattar I think I addressed your commend (qrom now expects data).

qualtran/bloqs/qrom.py
```diff
@@ -0,0 +1,87 @@
...
+        return decompose_from_cirq_op(self)
+
+    def as_cirq_op(
+        self, qubit_manager: 'cirq.QubitManager', **cirq_quregs: 'CirqQuregT'
+    ) -> Tuple['cirq.Operation', Dict[str, 'CirqQuregT']]:
+        qrom = CirqQROM(
+            data=self.data,
+            selection_bitsizes=self.selection_bitsizes,
+            target_bitsizes=self.data_bitsizes,
+            num_controls=self.cvs,
```
tanujkhattar, line 74: `CirqQROM`'s `num_controls` currently expects an `int` and not a `Tuple[int, ...]`, we should add support for arbitrary `cvs: Tuple[int, ...]` to the Cirq QROM but right now, this code should result in  an error. 

Can we add tests for the controlled version of `QROM` Bloq as well and fix the bug and raise an issue to track `cvs:Tuple[int, ...]` improvements to the Cirq / Qualtran QROM?

tanujkhattar (changes requested):
LGTM % a bug in the conversion.

qualtran/bloqs/qrom.py
fdmalone: @tanujkhattar should be done now!

tanujkhattar (approved):

PR #340, opened 2023-08-01
Use Docker to pin dependencies

### LINKED ISSUE #308: `pip-compile` within a consistent environment
We've been running pip-compile on our machines. The behavior of pip-compile depends on your platform and python version. I've been running it from a linux machine with a Python 3.10 install. We should do this from a docker image or something more reproducible. 

cc @tanujkhattar

mpharrigan:
`pip-tools`'s output depends on the environment in which it was run. Control the environment by generating dependencies from a Python 3.10 linux docker image.

This also moves the "environment" files to a subdirectory. This cleans things up and has the side benefit of making it easier to copy just those files _out_ of the docker image. 

Note: that there's a new, tiny env just for `pip-tools` to bootstrap the docker container. I think the latest pip-tools removes the fact that it was called with `resolver=backtracking` from the generated output. I didn't remove that argument (but we probably could because I know it was going to be the default at some point)

Fixes #308


dev_tools/requirements/re-pip-compile-in-docker.sh
```diff
@@ -0,0 +1,31 @@
...
+
+set -ex
+
+# We use docker to run the actual `pip-compile` command because its
+# behavior depends on the platform from which it is run. Please see
+# ./Dockerfile for the list of `pip-compile` commands that are run.
+docker build -t qualtran-pip-compile .
+
+# Create a container from the image so we can copy out the outputs
+id=$(docker create qualtran-pip-compile)
```
joshgc, line 25: FWIW `id` is a bash command already. What about image_id?

dev_tools/requirements/re-pip-compile.sh
```diff
@@ -17,15 +17,16 @@
...
-# has only one line: "-c ../dev.env.txt", which will use the above rendered dev.env.txt as a
+# has only one line: "-c envs/dev.env.txt", which will use the above rendered dev.env.txt as a
 # "constraints file", which is exactly what we want.
-pip-compile --output-file=runtime.env.txt --resolver=backtracking constrain.txt deps/runtime.txt
-pip-compile --output-file=format.env.txt  --resolver=backtracking constrain.txt deps/runtime.txt deps/format.txt
-pip-compile --output-file=pylint.env.txt  --resolver=backtracking constrain.txt deps/runtime.txt deps/pylint.txt
-pip-compile --output-file=pytest.env.txt  --resolver=backtracking constrain.txt deps/runtime.txt deps/pytest.txt
-pip-compile --output-file=docs.env.txt    --resolver=backtracking constrain.txt deps/runtime.txt deps/docs.txt
-
+pip-compile --output-file=envs/runtime.env.txt   --resolver=backtracking constrain.txt deps/runtime.txt
```
joshgc, line 27:
```
for pip_env in runtime format pylint pytest docs pip-tools; do 
  pip-compile --output-file "envs/${pip_env}.env.txt" --resolver backtracing constrain.txt deps/runtime.txt "deps/${pip_env}.txt
  done
  ```
  WDYT?

File: dev_tools/requirements/re-pip-compile-in-docker.sh (hunk 1 above)
Comment (mpharrigan) [reply dev_tools/requirements/re-pip-compile-in-docker.sh:25]: fun fact this is a container id and _not_ an image id (I learned that those are different things and why this line is required at all :) )

File: dev_tools/requirements/re-pip-compile-in-docker.sh [hunk 3]
```diff
@@ -0,0 +1,31 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+set -ex
+
+# We use docker to run the actual `pip-compile` command because its
+# behavior depends on the platform from which it is run. Please see
+# ./Dockerfile for the list of `pip-compile` commands that are run.
+docker build -t qualtran-pip-compile .
```
joshgc, line 22: I make pretty liberal use of 

```
# go/soverflow/59895/how-do-i-get-the-directory-where-a-bash-script-is-located
SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
readonly SCRIPT_DIR
```
In my build scripts so that I don't depend on the script being called from a specific directory. This SCRIPT_DIR can then inform the builds `.` and the cp's `./`

dev_tools/requirements/envs/dev.env.txt
```diff
@@ -2,18 +2,14 @@
 # This file is autogenerated by pip-compile with Python 3.10
 # by the following command:
 #
-#    pip-compile --output-file=dev.env.txt --resolver=backtracking deps/dev-tools.txt deps/runtime.txt
+#    pip-compile --output-file=envs/dev.env.txt deps/dev-tools.txt deps/runtime.txt
```
joshgc, line 5: Does pip-tools allow you to assert hashes?

dev_tools/requirements/re-pip-compile-in-docker.sh
joshgc: lol great point!! an even better reason to name it =D

dev_tools/requirements/re-pip-compile.sh
mpharrigan: My personal rule is "never write a loop in a bash script"

The idea being once you do that you're on the path of no return to an incomprehensible mess of bash when you probably should have switched to a real ™️  scripting language. I'll keep this unrolled for now, but will keep in mind going forward 

Also I will point out that we don't want `runtime.txt` included in the pip-tools env

dev_tools/requirements/envs/dev.env.txt
mpharrigan: it can generate hashes

dev_tools/requirements/re-pip-compile-in-docker.sh
mpharrigan: done

dev_tools/requirements/re-pip-compile-in-docker.sh
mpharrigan: done

mpharrigan:
@tanujkhattar ptal

fdmalone:
What is the workflow if a new dependency gets added? Just add to deps and run re-pip-compile-in-docker.sh and git add the updated *.env.txt?

tanujkhattar (approved):

mpharrigan:
> What is the workflow if a new dependency gets added? Just add to deps and run re-pip-compile-in-docker.sh and git add the updated *.env.txt?

yes

PR #343, opened 2023-08-10
Add hadamard gate.

fdmalone:


qualtran/bloqs/basic_gates/hadamard.py
```diff
@@ -0,0 +1,71 @@
...
+
+from qualtran import Bloq, Signature, SoquetT
+from qualtran.cirq_interop import CirqQuregT
+
+_HADAMARD = np.array([[1.0, 1.0], [1.0, -1.0]]) / (2**0.5)
+
+
+@frozen
+class Hadamard(Bloq):
+    """The Hadamard gate.
```
mpharrigan, line 31: provide a basic intro to what this is in case someone is living under a rock.

qualtran/bloqs/basic_gates/hadamard.py
```diff
@@ -0,0 +1,71 @@
...
+
+    Args:
+        bitsize: The number qubits to add a Hadamard gate to.
```
mpharrigan, line 34: be explicit that multi-qubit H is the same as H on each qubit.

---

I'm torn about whether we should do it this way or introduce an `OnEach` meta-bloq. The reason why having bitsize be acceptable here is because "hadamard matrix" in math is a family of different sized matrices (although they don't l2-normalize theirs).

qualtran/bloqs/basic_gates/hadamard.py
```diff
@@ -0,0 +1,71 @@
...
+    def as_cirq_op(
+        self, qubit_manager: 'cirq.QubitManager', q: 'CirqQuregT'
+    ) -> Tuple['cirq.Operation', Dict[str, 'CirqQuregT']]:
+        import cirq
+
+        if len(q) == 1:
+            (q,) = q
+            return cirq.H(q), {'q': [q]}
+        else:
+            return cirq.H.on_each(q), {'q': q}
```
mpharrigan, line 55: This would return a list of operations instead of one operation. Currently in qualtran, I've been a stickler for this returning exactly one operation so we don't start writing decompositions that are accessible during cirq conversion but which are not accessible via `decompose_bloq()`. In #344 (for unrelated state fixes) -- in the case where there's literally not a single cirq operation but we still want cirq conversion I use `CircuitOperation` so the structure isn't (automatically) lost. The cirq user can flatten it or whatever after the fact

qualtran/bloqs/basic_gates/hadamard.py
```diff
@@ -0,0 +1,71 @@
...
+        self,
+        tn: qtn.TensorNetwork,
+        binst,
+        *,
+        incoming: Dict[str, SoquetT],
+        outgoing: Dict[str, SoquetT],
+    ):
+        tn.add(
+            qtn.Tensor(
+                data=reduce(np.kron, (_HADAMARD,) * self.bitsize),
```
mpharrigan, line 67: The opposite of my other comment:

It hurts my brain that you are `np.kron`ing up a big matrix to add to the tensor network :) consider adding `n` 2x2 tensors

fdmalone: ugh yeah, I couldn't figure out the quimb syntax quickly or how to split register into n registers for the inds argument.

fdmalone:
closing this in favor of #344

PR #344, opened 2023-08-10
Add Hadamard Bloq

mpharrigan:
Pretty important :p

### DIFF
--- a/qualtran/bloqs/basic_gates/x_basis.py
+++ b/qualtran/bloqs/basic_gates/x_basis.py
@@ -84,7 +84,13 @@ class _XVector(Bloq):
         import cirq
 
         (q,) = qubit_manager.qalloc(self.n)
-        return cirq.H(q), {'q': np.array([q])}
+
+        if self.bit:
+            op = cirq.CircuitOperation(cirq.FrozenCircuit(cirq.X(q), cirq.H(q)))
+        else:
+            op = cirq.H(q)
+
+        return op, {'q': np.array([q])}
 
     def pretty_name(self) -> str:
         s = self.short_name()
--- a/dev_tools/autogenerate-bloqs-notebooks.py
+++ b/dev_tools/autogenerate-bloqs-notebooks.py
@@ -52,6 +52,7 @@ import qualtran.bloqs.and_bloq_test
 import qualtran.bloqs.arithmetic
 import qualtran.bloqs.arithmetic_test
 import qualtran.bloqs.basic_gates.cnot_test
+import qualtran.bloqs.basic_gates.hadamard_test
 import qualtran.bloqs.basic_gates.rotation_test
 import qualtran.bloqs.basic_gates.swap_test
 import qualtran.bloqs.basic_gates.t_gate_test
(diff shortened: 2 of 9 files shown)


fdmalone:
Should probably accept yours as you fixed something else along the way. I added support for H on n qubits (through a bitsize attribute). Do we want to add that here too? Hadamard on N qubits is quite frequently used.

fdmalone (approved):

mpharrigan:
>  Do we want to add that here too? Hadamard on N qubits is quite frequently used.

See my comments on the other PR. I'm torn philosophically.

Pro: multi-bit hadamard is a real thing (https://en.wikipedia.org/wiki/Hadamard_matrix)
Con: the concept of `OnEach` should probably be a meta-bloq so you could do this for all single-qubit operations
Other considerations: 1) gotta make sure tensor simulation is still efficient 2) there's no equivalent cirq gate for multi-qubit H. `on_each()` returns a list

qualtran/bloqs/basic_gates/x_basis.py
```diff
@@ -84,7 +84,13 @@ def as_cirq_op(
         import cirq
 
         (q,) = qubit_manager.qalloc(self.n)
-        return cirq.H(q), {'q': np.array([q])}
+
+        if self.bit:
+            op = cirq.CircuitOperation(cirq.FrozenCircuit(cirq.X(q), cirq.H(q)))
```
tanujkhattar, line 89: @mpharrigan `cirq.CircuitOperation(cirq.FrozenCircuit())` is essentially just an unnecessary boiler place here since it can be called in whatever utility method that calls `as_cirq_op`. This goes back to my original proposal of letting `as_cirq_op` be `as_cirq_optree` and wrap the yielded op-tree in a circuit operation at the call site when decomposing into a composite bloq. 

Can you clarify the benefits of this verbosity instead of the simplicity of my approach?

tanujkhattar (changes requested):

qualtran/bloqs/basic_gates/x_basis.py
tanujkhattar: I want to make sure we make it easy for users to express the fact that many times (like this one) you'd want to write a cirq conversion which is not 1:1. 

I tried to argue in my original comment https://github.com/quantumlib/Qualtran/pull/310#issuecomment-1634955959) that users can always wrap this conversion in a circuit operation, so there's really no way for us to enforce that we have a 1:1 mapping with a cirq (gate-)operation. It's good to see that now you also agree with me that the use case is a valid one and essentially end up doing the same thing.

Therefore, instead of pushing the burden of wrapping the OP-TREE in a circuit operation, we should start supporting an `as_cirq_optree` conversion.  The technical details of why we need to wrap the op-tree in a circuit operation should be handled by the library transparently from the user.

mpharrigan: I don't want decompositions hiding in the cirq-conversion code

tanujkhattar: I dont think you can stop that from happening by requiring an extra boilerplate. The boilerplate does add friction to legit use cases like this one, where the decomposition is conceptually 1:1, but not exactly.

fdmalone:
Can we me merge this @tanujkhattar

tanujkhattar (approved):
LGTM! I'll follow up on the `as_cirq_op` in a separate issue.

PR #346, opened 2023-08-11
Add OnEach bloq

fdmalone:
It's convenient to have, say, a Hadamard on log N qubits. This bloq should do that. Would be nice to replace ^ with $^{\otimes n}$

### DIFF
--- a/qualtran/bloqs/on_each.py
+++ b/qualtran/bloqs/on_each.py
@@ -17,9 +17,8 @@ from functools import cached_property
 from typing import Dict
 
 import attrs
-import quimb.tensor as qtn
 
-from qualtran import Bloq, BloqBuilder, CompositeBloq, Register, Signature, SoquetT
+from qualtran import Bloq, BloqBuilder, Register, Signature, SoquetT
 
 
 @attrs.frozen
@@ -33,6 +32,7 @@ class OnEach(Bloq):
     Registers:
      - q: an n-qubit register.
     """
+
     n: int
     gate: Bloq
 
@@ -44,13 +44,9 @@ class OnEach(Bloq):
     def short_name(self) -> str:
         return f'{self.gate.short_name()}^{self.n}'
 
-    def build_composite_bloq(
-            self, bb: BloqBuilder, *, q: SoquetT,
-    ) -> Dict[str, SoquetT]:
+    def build_composite_bloq(self, bb: BloqBuilder, *, q: SoquetT) -> Dict[str, SoquetT]:
 
         qs = bb.split(q)
         for i in range(self.n):
             qs[i] = bb.add(self.gate, q=qs[i])
-        return {
-            'q': bb.join(qs),
-        }
+        return {'q': bb.join(qs)}
--- a/qualtran/bloqs/on_each.py
+++ b/qualtran/bloqs/on_each.py
@@ -1,0 +1,56 @@
+#  Copyright 2023 Google LLC
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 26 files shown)


qualtran/bloqs/on_each.py
```diff
@@ -0,0 +1,52 @@
...
+class OnEach(Bloq):
+    """Add a single-qubit (unparameterized) bloq on each of n qubits.
+
+    Args:
+        n: the number of qubits to add the bloq to.
+        gate: A single qubit gate
+
+    Registers:
+     - q: an n-qubit register.
+    """
```
tanujkhattar, line 34: Do we really need a separate Bloq for `OnEach`? Can it be a convenience method instead?

qualtran/bloqs/on_each.py
```diff
@@ -0,0 +1,52 @@
...
+
+    n: int
+    gate: Bloq
+
+    @cached_property
+    def signature(self) -> Signature:
+        reg = Register('q', bitsize=self.n)
+        return Signature([reg])
```
tanujkhattar, line 42: Does this assume that `self.gate` is a `Bloq` that acts only on single qubits? If yes, we add a post init check or extend the functionality to general Bloqs.

qualtran/bloqs/on_each.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def short_name(self) -> str:
+        return f'{self.gate.short_name()}^{self.n}'
+
+    def build_composite_bloq(self, bb: BloqBuilder, *, q: SoquetT) -> Dict[str, SoquetT]:
+
+        qs = bb.split(q)
+        for i in range(self.n):
+            qs[i] = bb.add(self.gate, q=qs[i])
```
tanujkhattar, line 51: This assumes `self.gate` expects a register named `q`. Do we need this assumption?

tanujkhattar (changes requested):
What's the motivation behind having an explicit `OnEach` Bloq ? Can we instead just have a method?

qualtran/bloqs/on_each.py
fdmalone: a convenience method of Bloq?

fdmalone: I wanted to be able to have a highlevel H on n qubits without the decomposition, as well as the decomposition for testing, which sounds like we need a separate bloq? A meta bloq was suggested during #343

qualtran/bloqs/on_each.py
fdmalone: yeah, I'm not sure if there's a way around this. All the single qubit gates in bloqs have a register q, but that's a little dangerous.

qualtran/bloqs/on_each.py
```diff
@@ -0,0 +1,52 @@
...
+    n: int
+    gate: Bloq
+
+    @cached_property
+    def signature(self) -> Signature:
+        reg = Register('q', bitsize=self.n)
+        return Signature([reg])
+
+    def short_name(self) -> str:
+        return f'{self.gate.short_name()}^{self.n}'
```
mpharrigan, line 45: This could be confusing vs repeated application. We need the otimes thing somewhere

qualtran/bloqs/on_each.py
mpharrigan: Please document this restriction and perform a check

mpharrigan (commented):
This looks mostly good.

I agree that we need more validation
 - check that `gate` has one register named "q" of bitsize 1

qualtran/bloqs/on_each.py
mpharrigan: Or you could query the sub-gate's signature, assert that it has one register of bitsize 1, figure out its name, save its name to a member variable, and use it in the decomposition. I personally think that's overkill at this point

fdmalone: I'll just use the asserts in post init to enforce this for the moment.

fdmalone:
@tanujkhattar can this go in?

mpharrigan (approved):

tanujkhattar (approved):

Issue #335, opened 2023-07-30
List of Bloqs needed for THC


fdmalone:
Enumerating gates needed for THC and some thoughts on how I will structure the tutorial.

### Uniform Prep
- [ ] LessThanEqualGate (from cirq_ft)
- [ ] Hadamard on_each (from cirq)
- [ ] Ry rotation (phase gradient gate not 100% required)
- [ ] Multi-controlled paulis / reflection about zero (from cirq?)

### PREPARE
- [ ]  QROM / QRO(A)M  (from cirq_ft)
- [ ] CSWAP with multi-controls (possibly just add some logic before swap).

## Higher level Tutorial
- [ ] Prepare on two registers given some non trivial logic for ranges with amplitude amplification
- [ ] Amplitude amplification (maybe separate tutorial based off of Tanuj's talk) with examples in bloqs.
- [ ] SELECT self inverse?
- [ ] Controlled rotations and simplification from DF.

fdmalone:
@tanujkhattar fyi

PR #349, opened 2023-08-14
Multi controlled pauli

fdmalone:
Wrap cirq fts MultiControlledPauli. Needed for #335. The interface is modified slightly to allow for multiple named control registers, the size of which is determined by the length of the cv tuple of control values. For example, cvs=((0,0), (1), (1,1,1)) would lead to 3 registers of size 2, 1, and 3 bits respectively

### DIFF
--- a/qualtran/bloqs/multi_control_multi_target_pauli.py
+++ b/qualtran/bloqs/multi_control_multi_target_pauli.py
@@ -1,0 +1,63 @@
+#  Copyright 2023 Google LLC
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+from functools import cached_property
+from typing import Dict, Tuple
+
+import cirq
+from attrs import frozen
+from cirq_ft import MultiControlPauli as CirqMultiControlPauli
+
+from qualtran import Bloq, CompositeBloq, Signature
+from qualtran.cirq_interop import CirqQuregT, decompose_from_cirq_op
+
+
+@frozen
+class MultiControlPauli(Bloq):
+    """Implements multi-control, single-target C^{n}P gate.
+
+    Implements $C^{n}P = (1 - |1^{n}><1^{n}|) I + |1^{n}><1^{n}| P^{n}$ using $n-1$
+    clean ancillas using a multi-controlled `AND` gate.
+
+    References:
+        [Constructing Large Controlled Nots]
+        (https://algassert.com/circuits/2015/06/05/Constructing-Large-Controlled-Nots.html)
+
+    Args:
+        cvs: Control values.
+        target_gate: Pauli gate to apply to target register.
+    """
+
+    cvs: Tuple[int, ...]
+    target_gate: Bloq
+
+    @cached_property
+    def signature(self) -> Signature:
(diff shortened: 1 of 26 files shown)


mpharrigan:
I feel like this should be in a submodule. What are your thoughts on organization?

qualtran/bloqs/multi_control_multi_target_pauli.py
```diff
@@ -0,0 +1,68 @@
...
+
+    Implements $C^{n}P = (1 - |1^{n}><1^{n}|) I + |1^{n}><1^{n}| P^{n}$ using $n-1$
+    clean ancillas using a multi-controlled `AND` gate.
+
+    References:
+        [Constructing Large Controlled Nots]
+        (https://algassert.com/circuits/2015/06/05/Constructing-Large-Controlled-Nots.html)
+
+    Args:
+        cvs: Control values.
```
mpharrigan, line 38: can you document this more? Specifically how the nested cvs turn into the control registers.

Note that the bra-ketty expresssion above assumes cv's of "1". For simplicity, we should keep that but make a note that `cvs` changes the projectors

qualtran/bloqs/multi_control_multi_target_pauli.py
```diff
@@ -0,0 +1,68 @@
...
+        target_gate: Pauli gate to apply to target register.
+    """
+
+    cvs: Tuple[Tuple[int, ...], ...]
+    target_gate: Bloq
```
mpharrigan, line 43: should this be called `pauli`?

fdmalone:
Should we move all controlled bloqs into bloqs/controlled/...?

tanujkhattar:
> bloqs/controlled/.

I don't think so. We'll need controlled versions of almost all primitives (eg: arithmetic, unary iteration based primitves etc.) so having a separate module for `controlled` gates doesn't sound right to me.

tanujkhattar:
Sorry, closed by mistake.

fdmalone:
right then I don't really know how else to organize this. I can see submodules arising for say the chemistry algorithms as there are already for factoring etc, but not sure here.

qualtran/bloqs/multi_control_multi_target_pauli.py
```diff
@@ -0,0 +1,76 @@
...
+        cvs: Control values. Expect a tuple of tuples of control values, where
+            each tuple of control values should be the same length as the number
+            of bits in the corresponding control register. E.g. cvs = ((1, 1),
+            (1,), (1,1,1)) would mean we have 3 control registers of sizes (2,
+            1, 3). Currently we assume only control values of 1, a zero would
+            signify anti controls.
+        target_pauli: The name of the Pauli gate ("X", "Y", or "Z")
+    """
+
+    cvs: Tuple[Tuple[int, ...], ...]
```
tanujkhattar, line 47: I think at this point it's worth spending the additional time to create a `ControlRegister` object and expect a `Sequence[ControlRegister]` instead of `cvs: Tuple[Tuple[int, ...], ...]`. 

Specifically, control values can be more than just `Tuple[int, ...]`. We've run into this before in Cirq and as a result, we ended up creating the control_values class that represents a combination of control values. 

https://github.com/quantumlib/Cirq/blob/64e1a7fda696af542fb32c4a4c664f1c5d51546f/cirq-core/cirq/ops/control_values.py#L25


For backwards compatibility, we couldn't get rid of the `Tuple[int, ...]` interface and it made the signature of `gate.controlled()` and `operation.controlled_by()` an absolute nightmare. 

Since we have the time and flexibility to do things right this time, I'd suggest let's create the right design to represent control values and then use that everywhere in classes that expect a `cvs` (like this one).

fdmalone: Is the cirq control_values design something you would like to see again here?

mpharrigan: See #300 -- a controlregister class is on the roadmap. I don't think it should be blocking implementation of this bloq

mpharrigan (approved):
we'll defer organization until later

PR #353, opened 2023-08-16
Use Cirq-FT's multi-dimensional registers directly in BloqAsCirqGate

tanujkhattar:
This gets rid of the conversion to/from a flat list of API when wrapping a Bloq as a Cirq gate. Part of a larger series of PRs to improve Cirq interop. 

cc @mpharrigan


fdmalone (approved):
LGTM

PR #355, opened 2023-08-21
Support cirq registers in `CirqGateAsBloq` conversion

tanujkhattar:
This PR improves the `CirqGateAsBloq` such that now we can
- wrap Cirq-FT's `GateWithRegisters` without loosing the multi-dimensional register information
- Fallback on `cirq.decompose_once` to decompose the corresponding `CirqGateAsBloq`. 
- Any ancillas introduced in the decomposition are represented as explicit allocation/ deallocation nodes in the resulting composite graph (see example below for prepare hubbard)

These improvements will now enable importing all the nice primitives written in Cirq-FT with minimal overhead, so that they can be used directly as Bloqs. The modified notebook shows example code for importing the SELECT and PREARE circuits for 2D Hubbard Model as Bloqs using `CirqGateAsBloq`. Also see the following examples

### `cirq_ft.SwapWithZeroGate`  for multi-dimensional cirq registers
image

### `cirq_ft.PrepareHubbard` for decomposition with ancilla allocations
image

### `show_bloq(cbloq)` for  Test-2 of `test_cirq_optree_to_cbloq`
The diagrams below is for a scaled down version which uses 12 qubits instead of 24, as used in the tests. 

#### Cirq circuit
image

#### Corresponding Composite Bloq created using `cirq_optree_to_cbloq`
image

### DIFF
--- a/qualtran/cirq_interop/_cirq_interop.py
+++ b/qualtran/cirq_interop/_cirq_interop.py
@@ -24,8 +24,6 @@ import networkx as nx
 import numpy as np
 import quimb.tensor as qtn
 from attrs import frozen
-from cirq_ft import Register as LegacyRegister
-from cirq_ft import Registers as LegacyRegisters
 from numpy.typing import NDArray
 
 from qualtran import (
@@ -135,7 +133,7 @@ class CirqGateAsBloq(Bloq):
 
         wire_symbols = cirq.circuit_diagram_info(self.gate).wire_symbols
         begin = 0
-        symbol = soq.pretty()
+        symbol: str = soq.pretty()
         for reg in self.signature:
             finish = begin + np.product(reg.shape)
             if reg == soq.reg:
@@ -144,7 +142,10 @@ class CirqGateAsBloq(Bloq):
         return directional_text_box(text=symbol, side=soq.reg.side)
 
 
-def split_qvars(qvars: Sequence[Soquet], signature: Signature) -> Dict[str, NDArray[Soquet]]:
+def _split_qvars_for_regs(
+    qvars: Sequence[Soquet], signature: Signature
+) -> Dict[str, NDArray[Soquet]]:
+    """Split a flat list of soquets into a dictionary corresponding to `signature`."""
     qvars_regs = {}
     base = 0
     for reg in signature:
@@ -172,10 +173,9 @@ def cirq_optree_to_cbloq(
     OP-TREE corresponding to the `signature`. If `signature` has registers with entry
         - `Register('x', bitsize=2, shape=(3, 4))` and
         - `Register('y', bitsize=1, shape=(10, 20))`
-    then `cirq_quregs` should one entry corresponding to each register as follows:
+    then `cirq_quregs` should have one entry corresponding to each register as follows:
         - key='x'; value=`np.array(cirq_qubits_used_in_optree, shape=(3, 4, 2))` and
-        - key='y'; value=`np.array(cirq_qubits_used_in_optree, shape=(10, 20))`.
-
+        - key='y'; value=`np.array(cirq_qubits_used_in_optree, shape=(10, 20, 1))`.
 
     If `signature` is None, the resultant composite bloq will have one thru-register named "qubits"
     of shape `(n_qubits,)`.
@@ -198,7 +198,7 @@ def cirq_optree_to_cbloq(
         assert reg.name in cirq_quregs
         soqs = initial_soqs[reg.name]
         if isinstance(soqs, Soquet):
-            soqs = np.array([soqs])
+            soqs = np.asarray(soqs)[np.newaxis, ...]
         if reg.bitsize > 1:
             soqs = np.array([bb.split(soq) for soq in soqs.flatten()])
         soqs = soqs.reshape(reg.shape + (reg.bitsize,))
@@ -217,7 +217,7 @@ def cirq_optree_to_cbloq(
                 allocated_qubits.add(q)
 
         qvars_in = [qubit_to_qvar[qubit] for qubit in op.qubits]
-        qvars_out = bb.add_t(bloq, **split_qvars(qvars_in, bloq.signature))
+        qvars_out = bb.add_t(bloq, **_split_qvars_for_regs(qvars_in, bloq.signature))
         qubit_to_qvar |= zip(
             op.qubits, itertools.chain.from_iterable([arr.flatten() for arr in qvars_out])
         )
@@ -225,7 +225,7 @@ def cirq_optree_to_cbloq(
     for q in allocated_qubits:
         bb.add(Free(1), free=qubit_to_qvar[q])
 
-    qvars = [*qubit_to_qvar.values()]
+    qvars = np.array([*qubit_to_qvar.values()])
     final_soqs = {}
     idx = 0
     for reg in signature.rights():
@@ -236,16 +236,19 @@ def cirq_optree_to_cbloq(
         if reg.bitsize > 1:
             # Need to combine the soquets here.
             if len(soqs) == reg.bitsize:
-                final_soqs[name] = bb.join(np.array(soqs))
+                final_soqs[name] = bb.join(soqs)
             else:
                 final_soqs[name] = np.array(
-                    bb.join(np.array(subsoqs)) for subsoqs in soqs[:: reg.bitsize]
+                    [
+                        bb.join(soqs[st : st + reg.bitsize])
+                        for st in range(0, len(soqs), reg.bitsize)
+                    ]
                 ).reshape(reg.shape)
         else:
             if len(soqs) == 1 and reg.shape == ():
                 final_soqs[name] = soqs[0]
             else:
-                final_soqs[name] = np.array(soqs).reshape(reg.shape)
+                final_soqs[name] = soqs.reshape(reg.shape)
     return bb.finalize(**final_soqs)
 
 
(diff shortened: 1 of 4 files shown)


fdmalone (approved):
LGTM

qualtran/cirq_interop/_cirq_interop.py
```diff
@@ -43,11 +44,16 @@
     SoquetT,
 )
 from qualtran._infra.composite_bloq import _binst_to_cxns
+from qualtran.bloqs.util_bloqs import Allocate, Free
 
 CirqQuregT = NDArray[cirq.Qid]
 CirqQuregInT = Union[NDArray[cirq.Qid], Sequence[cirq.Qid]]
 
 
+def signature_from_cirq_registers(registers: Iterable[cirq_ft.Register]) -> 'Signature':
```
mpharrigan, line 53: elsewhere in this module we import `cirq_ft.Register` as `LegacyRegister`. Please be consistent throughout

mpharrigan: add to `cirq_interop/__init__.py`

qualtran/cirq_interop/_cirq_interop.py
```diff
@@ -81,28 +100,65 @@ def add_my_tensors(
...
+    def wire_symbol(self, soq: 'Soquet') -> 'WireSymbol':
+        from qualtran.drawing import directional_text_box
+
+        wire_symbols = cirq.circuit_diagram_info(self.gate).wire_symbols
+        begin = 0
+        symbol = soq.pretty()
+        for reg in self.signature:
+            finish = begin + np.product(reg.shape)
+            if reg == soq.reg:
+                symbol = np.array(wire_symbols[begin:finish]).reshape(reg.shape)[soq.idx]
```
mpharrigan, line 142: wait .. what is happening here? since bitsize is always `1`, we're guaranteed that symbol is one string, yes? maybe add a little `assert` just to elucidate the code

qualtran/cirq_interop/_cirq_interop.py
```diff
@@ -67,11 +73,24 @@ def short_name(self) -> str:
...
-    def n_qubits(self):
-        return cirq.num_qubits(self.gate)
+    def cirq_registers(self) -> cirq_ft.Registers:
+        if isinstance(self.gate, cirq_ft.GateWithRegisters):
+            return self.gate.registers
+        else:
+            return cirq_ft.Registers.build(qubits=cirq.num_qubits(self.gate))
+
+    def decompose_bloq(self) -> 'CompositeBloq':
+        quregs = self.signature.get_cirq_quregs()
```
mpharrigan, line 86: what do you think about moving `get_cirq_quregs` out of `Signature` and into this file?

qualtran/cirq_interop/_cirq_interop.py
```diff
@@ -81,28 +100,65 @@ def add_my_tensors(
...
+        symbol = soq.pretty()
+        for reg in self.signature:
+            finish = begin + np.product(reg.shape)
+            if reg == soq.reg:
+                symbol = np.array(wire_symbols[begin:finish]).reshape(reg.shape)[soq.idx]
+            begin = finish
+        return directional_text_box(text=symbol, side=soq.reg.side)
+
+
+def split_qvars(qvars: Sequence[Soquet], signature: Signature) -> Dict[str, NDArray[Soquet]]:
```
mpharrigan, line 147: make private, little docstring, maybe rename `qvars_regs` to something more suggestive

qualtran/cirq_interop/_cirq_interop.py
```diff
@@ -111,67 +167,85 @@ def cirq_optree_to_cbloq(
     If `signature` is not None, the signature of the resultant CompositeBloq is `signature`. For
     multi-dimensional registers and registers with > 1 bitsize, this function automatically
     splits the input soquets into a flat list and joins the output soquets into the correct shape
-    to ensure compatibility with the flat API expected by Cirq.
+    to ensure compatibility with the flat API expected by Cirq. When specifying a signature, users
+    must also specify the `cirq_quregs` argument, which is a mapping of cirq qubits used in the
+    OP-TREE corresponding to the `signature`. If `signature` has registers with entry
+        - `Register('x', bitsize=2, shape=(3, 4))` and
+        - `Register('y', bitsize=1, shape=(10, 20))`
+    then `cirq_quregs` should one entry corresponding to each register as follows:
```
mpharrigan, line 175: should *have* one entry

qualtran/cirq_interop/_cirq_interop.py
```diff
@@ -111,67 +167,85 @@ def cirq_optree_to_cbloq(
...
+        - key='x'; value=`np.array(cirq_qubits_used_in_optree, shape=(3, 4, 2))` and
+        - key='y'; value=`np.array(cirq_qubits_used_in_optree, shape=(10, 20))`.
+
```
mpharrigan, line 178: extra newline

qualtran/cirq_interop/_cirq_interop.py
```diff
@@ -111,67 +167,85 @@ def cirq_optree_to_cbloq(
...
     bb, initial_soqs = BloqBuilder.from_signature(signature, add_registers_allowed=False)
 
     # Magic to make sure signature of the CompositeBloq matches `Signature`.
-    qvars = {}
+    qubit_to_qvar = {}
     for reg in signature.lefts():
+        assert reg.name in cirq_quregs
         soqs = initial_soqs[reg.name]
+        if isinstance(soqs, Soquet):
+            soqs = np.array([soqs])
```
mpharrigan, line 201: `np.asarray(soqs)[np.newaxis, ...]` (literally use the elipses)

qualtran/cirq_interop/_cirq_interop.py
```diff
@@ -111,67 +167,85 @@ def cirq_optree_to_cbloq(
...
     circuit = cirq.Circuit(optree)
-    all_qubits = sorted(circuit.all_qubits())
+    # "qubits" means cirq qubits | "qvars" means bloq Soquets
     if signature is None:
+        assert cirq_quregs is None
+        all_qubits = sorted(circuit.all_qubits())
         signature = Signature([Register('qubits', 1, shape=(len(all_qubits),))])
+        cirq_quregs = {'qubits': np.array(all_qubits).reshape(len(all_qubits), 1)}
+
+    assert signature is not None and cirq_quregs is not None
```
mpharrigan, line 191: can you tell me (in this PR) what used to happen if you provided signature and not `cirq_quregs` (because that argument didn't exist) and why you'd like to change that behavior by requiring explicit cirq_quregs

qualtran/cirq_interop/_cirq_interop.py
```diff
@@ -111,67 +167,85 @@ def cirq_optree_to_cbloq(
...
-        soqs = qvar_vals_out[idx : idx + len(qvars[name])]
-        idx = idx + len(qvars[name])
+        assert name in cirq_quregs
+        soqs = qvars[idx : idx + np.product(cirq_quregs[name].shape)]
+        idx = idx + np.product(cirq_quregs[name].shape)
         if reg.bitsize > 1:
             # Need to combine the soquets here.
             if len(soqs) == reg.bitsize:
-                final_soqs[name] = bb.join(soqs)
+                final_soqs[name] = bb.join(np.array(soqs))
```
mpharrigan, line 239: change qvars to a numpy array during construction and then these castings are unnecessary?

mpharrigan (commented):
initial comments.

I didn't dig through the `cirq_optree_to_cbloq` logic changes with a fine-toothed comb and/so tests would be appreciated here!

qualtran/cirq_interop/_cirq_interop.py
tanujkhattar: Yes, `symbol` (now typed as `str`) is always expected to be a single string. The `reshaping` is to make it convenient to convert from flat list of symbols, one for each qubit, to a multi-dimensional representation so we can use the composite index of `soq.idx` directly. 

`reg.bitsize` is 1 but `reg.shape` can be arbitrary, and we have one soquet for each `reg.shape` and one wire symbol for each soquet.

qualtran/cirq_interop/_cirq_interop.py
tanujkhattar: changed to `cirq_ft.Register` throughout.

mpharrigan:
No rush and apologies if you already know this and intended it, but I see you've resolved some of the suggestions but haven't pushed any changes :)

Ping me when ready for re-review

tanujkhattar:
@mpharrigan Addressed your comments and added tests, this is ready for a review!

qualtran/cirq_interop/_cirq_interop.py
tanujkhattar: Let's do that in a follow-up PR since `get_cirq_quregs` is also used in a few other places like tests basic gates / composite gates etc.

qualtran/cirq_interop/_cirq_interop.py
tanujkhattar: When converting a cirq op-tree to Bloqs, we need to have a mapping between qubits used in the cirq op-tree to the corresponding named soquets in the Bloqs. Prior to this PR, the mapping was created by creating a flat list of qubits (via `all_qubits = sorted(circuit.all_qubits())`) and a flat list of soquets (by iterating over the left registers in `signature` and flattening out every multi-dimensional collection of soquets) and then asserting that the two lists are of the same length. 

As you can imagine, this brute force would work only for the most trivial cases and quickly breaks down the 1:1 mapping between the corresponding sorted lists does not hold true. 

Now, forcing users to provide the cirq qubits in the `cirq_quregs` format enables us to wire up the `CirqGateAsBloq`s correctly and note that this information already exists at all the call sites. The trivial case, where the signature of composite bloq is a single register called `qubits` of shape `(len(all_qubits),)` and `bitsize=1` is still supported. 

See the tests that I've added for more context.

mpharrigan (approved):
cool stuff :sunglasses:

PR #364, opened 2023-09-22
Add missing bloq_counts and t_complexity for some basic gates.

fdmalone:


mpharrigan:
I'm -1 on some of these changes.

The `bloq_counts` should match the decomposition. Philosophically: H isn't composed of one arbitrary clifford.

The way to handle this is in the `generalizer` argument of `get_bloq_counts_graph`. 

```python
def ignore_split_join(b: Bloq):
  if isinstance(b, (Split, Join)):
    return None
  return b

def group_cliffords(b: Bloq):
  if isinstance(b, (Hadamard, CNOT, ...)):
     return ArbitraryClifford(n=2)
  return b

get_bloq_counts_graph(thing, generalizer = lambda b: group_cliffords(ignore_split_join(b)))
```

Issue #360, opened 2023-09-19
Cleaner way to add CirqGateAsBloq to bloq decompsition?


fdmalone:
I found in #350 a few places where it was initially a bit awkward to map bloq registers to cirq registers and back again. The first case was for arithmetic gates which expect a register called `qubits` of bitsize=1, and shape=(n,), (rather than bitsize=n in bloqs land) 

Converter
``` python

def add_from_bloq_register_flat_qubits(
    bb: 'BloqBuilder', cirq_bloq: Bloq, **regs: SoquetT
) -> Tuple[SoquetT, ...]:
    """Helper function to split / join registers for cirq gates expeciting single 'qubits' register.
    Args:
        bb: Bloq builder used during decompostion.
        cirq_bloq: A CirqGateAsBloq wrapped arithmetic gate.
        regs: bloq registers we wish to use as flat list of qubits for cirq gate.
    Returns:
        regs: bloq registers appropriately rejoined following split.
    """
    flat_regs = []
    for _, v in regs.items():
        if v.reg.bitsize == 1:
            flat_regs.append([v])
        else:
            flat_regs.append(bb.split(v))
    qubits = np.concatenate(flat_regs)
    qubits = bb.add(cirq_bloq, qubits=qubits)
    out_soqs = {}
    start = 0
    for _, v in regs.items():
        if v.reg.bitsize == 1:
            end = start + 1
            out_soqs[v] = qubits[start:end][0]
            start += 1
        else:
            end = start + v.reg.bitsize
            out_soqs[v] = bb.join(qubits[start:end])
            start += v.reg.bitsize
    return tuple(s for _, s in out_soqs.items()) 
```

The second case is being addressed in quantumlib/cirq#6286 I think, but occurs if bitsize=n, shape=(), then cirq expects bitsize=1 and shape=(n,)

Converter:

``` python
def add_from_bloq_registers(
    bb: 'BloqBuilder', cirq_bloq: Bloq, **bloq_regs: SoquetT
) -> Tuple[SoquetT, ...]:
    """Shift from bitsize=n, shape=() to bitsize=1, shape=(n,) and back again"""
    cirq_regs = {}
    for reg_name, soq in bloq_regs.items():
        cirq_regs[reg_name] = bb.split(soq)
    cirq_regs = bb.add(cirq_bloq, **cirq_regs)
    out_soqs = {}
    for ix, (reg_name, soq) in enumerate(bloq_regs.items()):
[...]

PR #366, opened 2023-09-24
Preserve identical signature with bitsize information during Cirq interop

tanujkhattar:
This is built on top of https://github.com/quantumlib/Cirq/pull/6286 and https://github.com/quantumlib/Cirq/pull/6298 and should address parts of https://github.com/quantumlib/Qualtran/issues/360

This is the last big change to support seamless Cirq-FT interop. See example below for updated usage. 

```python
bloq = CirqGateAsBloq(cirq_ft.SwapWithZeroGate(2, 3, 4))
cbloq = bloq.decompose_bloq()
fig, ax = draw_musical_score(get_musical_score_data(cbloq))
fig.set_size_inches(11, 4)
```

image

cc @fdmalone @mpharrigan 

Note: Failing tests will be fixed once https://github.com/quantumlib/Cirq/pull/6298 is merged.


dev_tools/requirements/envs/dev.env.txt
```diff
@@ -10,6 +10,10 @@ accessible-pygments==0.0.4
     # via pydata-sphinx-theme
 alabaster==0.7.13
     # via sphinx
+appnope==0.1.3
```
mpharrigan, line 13: how hard would it be to re-generate from a linux box so we don't get this platform specific churn

mpharrigan: oh wait I forgot the docker thing is already merged. Please use that to generate these files (ping me if you need help or want me to do it)

tanujkhattar: Regenerated using the docker script.

qualtran/_infra/registers_test.py
```diff
@@ -103,10 +103,10 @@ def test_agg_split():
 
 
 def test_get_named_qubits_multidim():
-    regs = Signature([Register('matt', shape=(2, 3), bitsize=4)])
```
mpharrigan, line 106: booooooo haha

qualtran/cirq_interop/_cirq_interop.py
```diff
@@ -13,11 +13,11 @@
 #  limitations under the License.
 
 """Functionality for the `Bloq.as_cirq_op(...)` protocol"""
-
 import itertools
 from functools import cached_property
 from typing import Any, Callable, Dict, Iterable, List, Mapping, Optional, Sequence, Tuple, Union
 
+import attr
```
mpharrigan, line 20: annoying nit: can you use attrs instead of attr. It's their cleaned-up API (but the other one will be around forever)

qualtran/cirq_interop/_cirq_interop.py
```diff
@@ -135,7 +144,7 @@ def wire_symbol(self, soq: 'Soquet') -> 'WireSymbol':
         begin = 0
         symbol: str = soq.pretty()
         for reg in self.signature:
-            finish = begin + np.product(reg.shape)
+            finish = begin + int(np.product(reg.shape))
```
mpharrigan, line 147: is the int cast here and above necessary?

qualtran/cirq_interop/_cirq_interop.py
tanujkhattar: Cirq can't use attrs because of a dependency issue due to external vendors. So cirq-ft defaults to using `attr` and so I end up defaulting to using `attr`. I think it should be fine to using `attrs` here since we only depend upon cirq-ft & cirq-core and don't pin attr version dependency. 

Tl;Dr - I'll change it to `attrs`.

qualtran/cirq_interop/_cirq_interop.py
tanujkhattar: Yes, `np.product` returns a float instead of an `int` so the cast is necessary. I should probably also replace `np.product` with `np.prod` instead. Will do that.

qualtran/cirq_interop/_cirq_interop.py
```diff
@@ -155,6 +164,13 @@ def _split_qvars_for_regs(
     return qvars_regs
 
 
+@attr.frozen
+class QReg:
```
mpharrigan, line 168: consider making this private (?)

If private: add a short docstring about how it's used.

If public: add a long docstring about how it's meant to be used

tanujkhattar: SG, will make it private.

qualtran/cirq_interop/cirq_interop.ipynb
mpharrigan, line 1: please clear outputs

qualtran/cirq_interop/_cirq_interop.py
mpharrigan: interesting. I thought installing attr or attrs would get you both `import attr` and `import attrs`

tanujkhattar: `attrs` was introduced in a specific version of `attr` and one of the cirq vendors has their `attr` dependency pinned to an older version so the `attrs` namespace does not exist.

qualtran/cirq_interop/_cirq_interop.py
```diff
@@ -188,67 +204,97 @@ def cirq_optree_to_cbloq(
...
-        soqs = soqs.reshape(reg.shape + (reg.bitsize,))
-        assert cirq_quregs[reg.name].shape == soqs.shape
-        qubit_to_qvar |= zip(cirq_quregs[reg.name].flatten(), soqs.flatten())
-
-    allocated_qubits = set()
+            soqs = np.asarray(soqs)
+        assert cirq_quregs[reg.name].shape == soqs.shape, (
+            f'{reg.name=}, {cirq_quregs[reg.name]=}, {soqs=},'
+            f'{cirq_quregs[reg.name].shape=}, {soqs.shape=}'
+        )
```
mpharrigan, line 263: please replace with a helpfully worded ValueError

qualtran/cirq_interop/_cirq_interop.py
```diff
@@ -188,67 +204,97 @@ def cirq_optree_to_cbloq(
         signature = Signature([Register('qubits', 1, shape=(len(all_qubits),))])
         cirq_quregs = {'qubits': np.array(all_qubits).reshape(len(all_qubits), 1)}
 
+    cirq_quregs = {k: np.apply_along_axis(QReg, -1, v) for k, v in cirq_quregs.items()}
```
mpharrigan, line 207: there's a comment above

> "qubits" means cirq qubits | "qvars" means bloq Soquets

does this still apply? update or delete

qualtran/cirq_interop/_cirq_interop.py
```diff
@@ -188,67 +204,97 @@ def cirq_optree_to_cbloq(
...
+        for qreg, soq in new_qreg_to_qvar.items():
+            if len(in_reg_qubits) > 1 and qreg.qubits and qreg.qubits[0] in in_reg_qubits:
+                assert len(qreg.qubits) == 1, "Individual qubits should have been split by now."
+                soqs_to_join.append(soq)
+            else:
+                qreg_to_qvar[qreg] = soq
+        if soqs_to_join:
+            qreg_to_qvar[in_reg] = bb.join(np.array(soqs_to_join))
+
+    def _find_input_soqs(
```
mpharrigan, line 241: suggest `_gather_input_soqs`.

qualtran/cirq_interop/_cirq_interop.py
```diff
@@ -188,67 +204,97 @@ def cirq_optree_to_cbloq(
...
+    cirq_quregs = {k: np.apply_along_axis(QReg, -1, v) for k, v in cirq_quregs.items()}
+
     assert signature is not None and cirq_quregs is not None
 
     bb, initial_soqs = BloqBuilder.from_signature(signature, add_registers_allowed=False)
 
-    # Magic to make sure signature of the CompositeBloq matches `Signature`.
-    qubit_to_qvar = {}
+    # 0. Helper functions
+    def _update_qreg_to_qvar(in_reg: QReg, qreg_to_qvar: Dict[QReg, Soquet]) -> None:
```
mpharrigan, line 214: suggest `_ensure_in_reg_exists` or something to that effect

qualtran/cirq_interop/_cirq_interop.py
```diff
@@ -188,67 +204,97 @@ def cirq_optree_to_cbloq(
...
+        """Takes care of splits and joins to make sure `qreg_to_qvar[in_reg]` exists."""
+        if in_reg in qreg_to_qvar:
+            # This is the easy case when no split / joins are needed.
+            return
+
+        # Split everything and join s.t. joined register corresponds to `in_reg`.
+        # 1. Split all registers containing at-least one qubit corresponding to `in_reg`.
```
mpharrigan, line 221: vestigial comment? these numbered things should be outside the closure

qualtran/cirq_interop/_cirq_interop.py
```diff
@@ -188,67 +204,97 @@ def cirq_optree_to_cbloq(
...
-    # Magic to make sure signature of the CompositeBloq matches `Signature`.
-    qubit_to_qvar = {}
+    # 0. Helper functions
+    def _update_qreg_to_qvar(in_reg: QReg, qreg_to_qvar: Dict[QReg, Soquet]) -> None:
+        """Takes care of splits and joins to make sure `qreg_to_qvar[in_reg]` exists."""
+        if in_reg in qreg_to_qvar:
+            # This is the easy case when no split / joins are needed.
+            return
+
+        # Split everything and join s.t. joined register corresponds to `in_reg`.
```
mpharrigan, line 220: I suggest turning this into two comments. First: we split everything that needs splitting

Then: we join everything that needs joining

mpharrigan (commented):
Super cool stuff. The logic -- while complex -- looks to be organized quite well in this PR. And of course the users will see a seamless api!

I have two requests
 - Asserts should be used such that if one is violated it means the developer has made a mistake. Some of these assertions made sense as assertions when we were always making quregs ourselves, for example. But now there are places where user-provided input can trigger an assertion failure including those without accompanying messages. Can you replace data input validation insertions with `if not x: raise ValueError(f"Helpful error message")`?
 - The closures (ie nested function definitions) inside this large and complex function are too large and complex. It is not apparent what is being closed over (which is `bb`). And since we already have a ton of variable names which are all variants of "qvar" or "qureg" there's a non-zero chance of accidentally shadowing the name of an outer scope variable inside the closure accidentally. Can you please move these to free functions.


```patch
From 654e8654a6fbd62bd714ed60a6f2bcd8b2f47eea Mon Sep 17 00:00:00 2001
From: Matthew Harrigan <mpharrigan@google.com>
Date: Mon, 25 Sep 2023 17:14:08 -0700
Subject: [PATCH 2/3] docstring

---
 qualtran/cirq_interop/_cirq_interop.py | 22 +++++++++++-----------
 1 file changed, 11 insertions(+), 11 deletions(-)

diff --git a/qualtran/cirq_interop/_cirq_interop.py b/qualtran/cirq_interop/_cirq_interop.py
index d1b92ad..208364d 100644
--- a/qualtran/cirq_interop/_cirq_interop.py
+++ b/qualtran/cirq_interop/_cirq_interop.py
@@ -180,21 +180,21 @@ def cirq_optree_to_cbloq(
     """Convert a Cirq OP-TREE into a `CompositeBloq` with signature `signature`.
     
     Each `cirq.Operation` will be wrapped into a `CirqGateAsBloq` wrapper.
+    The signature of the resultant CompositeBloq is `signature`, if provided. Otherwise, use
+    one thru-register named "qubits" of shape `(n_qubits,)`.
 
-    If `signature` is not None, the signature of the resultant CompositeBloq is `signature`. For
-    multi-dimensional registers and registers with > 1 bitsize, this function automatically
-    splits the input soquets into a flat list and joins the output soquets into the correct shape
-    to ensure compatibility with the flat API expected by Cirq. When specifying a signature, users
-    must also specify the `cirq_quregs` argument, which is a mapping of cirq qubits used in the
-    OP-TREE corresponding to the `signature`. If `signature` has registers with entry
+    For multi-dimensional registers and registers with bitsize>1, this function automatically
+    splits the input soquets and joins the output soquets to ensure compatibility with the
+    flat-list-of-qubits expected by Cirq.
+
+    When specifying a signature, users must also specify the `cirq_quregs` argument, which is a
+    mapping of cirq qubits used in the OP-TREE corresponding to the `signature`. If `signature`
+    has registers with entry
         - `Register('x', bitsize=2, shape=(3, 4))` and
         - `Register('y', bitsize=1, shape=(10, 20))`
     then `cirq_quregs` should have one entry corresponding to each register as follows:
-        - key='x'; value=`np.array(cirq_qubits_used_in_optree, shape=(3, 4, 2))` and
-        - key='y'; value=`np.array(cirq_qubits_used_in_optree, shape=(10, 20, 1))`.
-
-    If `signature` is None, the resultant composite bloq will have one thru-register named "qubits"
-    of shape `(n_qubits,)`.
+        - key='x'; value=`np.array(cirq_qubits_used_for_x, shape=(3, 4, 2))` and
+        - key='y'; value=`np.array(cirq_qubits_used_for_y, shape=(10, 20, 1))`.
     """
     circuit = cirq.Circuit(optree)
     # "qubits" means cirq qubits | "qvars" means bloq Soquets
-- 
2.42.0.515.g380fc7ccd1-goog

```

```patch
From f43cbcb138314dadcdff4b2126baea98bb834c5f Mon Sep 17 00:00:00 2001
From: Matthew Harrigan <mpharrigan@google.com>
Date: Mon, 25 Sep 2023 17:14:42 -0700
Subject: [PATCH 3/3] value error 2

---
 qualtran/cirq_interop/_cirq_interop.py | 10 +++++-----
 1 file changed, 5 insertions(+), 5 deletions(-)

diff --git a/qualtran/cirq_interop/_cirq_interop.py b/qualtran/cirq_interop/_cirq_interop.py
index 208364d..1d54dee 100644
--- a/qualtran/cirq_interop/_cirq_interop.py
+++ b/qualtran/cirq_interop/_cirq_interop.py
@@ -197,17 +197,17 @@ def cirq_optree_to_cbloq(
         - key='y'; value=`np.array(cirq_qubits_used_for_y, shape=(10, 20, 1))`.
     """
     circuit = cirq.Circuit(optree)
-    # "qubits" means cirq qubits | "qvars" means bloq Soquets
     if signature is None:
-        assert cirq_quregs is None
+        if cirq_quregs is not None:
+            raise ValueError("`cirq_quregs` requires specifying `signature`.")
         all_qubits = sorted(circuit.all_qubits())
         signature = Signature([Register('qubits', 1, shape=(len(all_qubits),))])
         cirq_quregs = {'qubits': np.array(all_qubits).reshape(len(all_qubits), 1)}
+    else:
+        if cirq_quregs is None:
+            raise ValueError("`signature` requires specifying `cirq_quregs`.")
 
     cirq_quregs = {k: np.apply_along_axis(QReg, -1, v) for k, v in cirq_quregs.items()}
-
-    assert signature is not None and cirq_quregs is not None
-
     bb, initial_soqs = BloqBuilder.from_signature(signature, add_registers_allowed=False)
         
     # 0. Helper functions
-- 
2.42.0.515.g380fc7ccd1-goog

```

mpharrigan (approved):
go 4 it

qualtran/cirq_interop/_cirq_interop.py
```diff
@@ -188,67 +204,97 @@ def cirq_optree_to_cbloq(
...
         if isinstance(soqs, Soquet):
-            soqs = np.asarray(soqs)[np.newaxis, ...]
-        if reg.bitsize > 1:
-            soqs = np.array([bb.split(soq) for soq in soqs.flatten()])
-        soqs = soqs.reshape(reg.shape + (reg.bitsize,))
-        assert cirq_quregs[reg.name].shape == soqs.shape
-        qubit_to_qvar |= zip(cirq_quregs[reg.name].flatten(), soqs.flatten())
-
-    allocated_qubits = set()
+            soqs = np.asarray(soqs)
```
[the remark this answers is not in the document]
tanujkhattar: In this case, `np.asarray(soqs)` works similar to `np.array(soqs)` and returns a `array(soquet, dtype=object)` with shape `()`. `np.asarray([soqs])` would instead return an array with shape `(1,)`, which is not what we want. 

Replaced with `np.array(soqs)` to avoid the confusion.

qualtran/cirq_interop/_cirq_interop.py
tanujkhattar: It still applies, but I've deleted the line anyway since there are 3 things at play now (Qubits, _QReg's and Soquets)

qualtran/cirq_interop/_cirq_interop.py
tanujkhattar: done

qualtran/cirq_interop/_cirq_interop.py
tanujkhattar: Changed numbering of comments inside the closure to a/b instead of 1/2

tanujkhattar:
@mpharrigan Address all comments.

PR #372, opened 2023-09-28
[jupyter_autogen] support "Registers" section

mpharrigan:
Render "Registers:" in docstrings like the "Args:" section


fdmalone (approved):
awesome

Issue #375, opened 2023-10-05
Cost of out of place addition?


fdmalone:
We can have |a>|b> -> |a>|a+b> or |a>|b|0> -> |a>|b>|a+b>. Currently these both have the same 4n - 4 T cost. Is this correct?

fdmalone:
Extra clifford cost of bitsize from CNOTs, thanks @NoureldinYosri!

PR #359, opened 2023-09-08
Trotter Costs

fdmalone:
I'm opening this as a draft at the moment for some feedback, but it may be good to merge and follow up with some issues later:

Issues:
1. The QVR is the naive way which I think can be improved from n log (1/e) to log (1/e).
2. We had some questions whether the  potential was suboptimally implemented (c.f. the first quantized block encoding potential doesn't have this eta^2 cost?
3. I started to go down a bit of a rabbit hole with fixed point arithmetic. I would like to revisit the interpolation section in the notebook to use fixed point and a proxy for QROM so that we check the quality of interpolation directly. Could be a nice demonstration of classical_vals
4. I have a huge hack to render the bloq counts for QROM which I think @tanujkhattar had a better solution to.

I think 4 should at least be resolved before merging. But yeah probably good to get some feedback before it gets any bigger.

PR #376, opened 2023-10-05
Add more arithmetic bloqs.

fdmalone:
* Square real
* Scale int by real
* Multiply two reals

Fix bug with bitsize for sum of squares.


fdmalone:
Pulled out from #359

fdmalone:
Oh and an out of place adder which raised #375

qualtran/bloqs/arithmetic.py
```diff
@@ -50,12 +58,58 @@ def t_complexity(self):
...
+
+    @property
+    def signature(self):
+        return Signature.build(a=self.bitsize, b=self.bitsize, c=self.bitsize)
+
+    def pretty_name(self) -> str:
+        return "c = a + b"
+
+    def t_complexity(self):
+        # TODO: This is just copied from adder above, not sure if correct.
```
NoureldinYosri, line 96: T count is correct. Clifford count should be inceased by `bitsize`.

the decomposition would be

```py3
yield CNOT.on_each(zip(b, c))
yield Add(a, c)
```

fdmalone: Awesome thanks!

mpharrigan: can we capture this somehow? i.e. by using bloq_counts of `n` cnots and an `Add`

mpharrigan (approved):
giving approval, but try to incorporate nour's comment

qualtran/bloqs/arithmetic.py
fdmalone: Done

PR #379, opened 2023-10-06
Support IntEffect.decompose() and fix bug in final_soqs

mpharrigan:
- [Bugfix: fix final_soqs for LEFT bloqs](https://github.com/quantumlib/Qualtran/commit/1ab6a7d0d05a49cd84da3521756f1abe457b6aaa)
 - [Enable decomposition for IntEffect too](https://github.com/quantumlib/Qualtran/commit/df6fa314ccbb1b23bb5d2d727784b74f6184651e)
 - [use IntEffect to fully test bugfix](https://github.com/quantumlib/Qualtran/commit/47eb53067959481bfb4b1ffd711fffa800f8105e)


mpharrigan:
@tanujkhattar can you ptal

qualtran/bloqs/basic_gates/z_basis.py
```diff
@@ -230,15 +230,26 @@ def signature(self) -> Signature:
...
+            xs = np.array(xs)
+
+            return {'val': bb.join(xs)}
+
+        val = val['val']
+        xs = bb.split(val)
+        effects = [ZeroEffect(), OneEffect()]
+        for i, bit in enumerate(bits):
+            bb.add(effects[bit], q=xs[i])
+        return {}
```
tanujkhattar, line 252: I have a small preference to split the two decompositions into `_build_composite_bloq_effect` and `_build_composite_bloq_state` and then call either of the two methods using an if / else on `self.state` from `_build_composite_bloq`. 

I know maybe initially I proposed that we should have a common base class for `_IntVector` but at this point, I'm also fine if we want to split it into `_IntEffect` and `_IntState`.

tanujkhattar (approved):
LGTM % one nit

qualtran/bloqs/basic_gates/z_basis.py
mpharrigan: I did the refactor

Issue #373, opened 2023-09-28
Separate methods for overriding vs calling in bloq counts


mpharrigan:
It's really useful to have different methods for overriding by bloq authors versus methods designed to be called by a user. For example: A user calls `bloq.decompose_bloq()` which is easy because it has no arguments and a bloq author overrides `build_composite_bloq` which is great because the system can pass in a bloq builder, input soquets, and munge the return value. In fact, we do this for most of our "protocols". Taken from the tutorial notebook:

| What                 | Call this                   | Override this               | Docs                                                    |
|----------------------|-----------------------------|-----------------------------|---------------------------------------------------------|
| Bloq decomposition   | `decompose_bloq()`          | `build_composite_bloq(...)` | [Composite bloq](/_infra/composite_bloq.ipynb)          |
| Numerical simulation | `tensor_contract()`         | `add_my_tensors(...)`       |                                                         |
| Classical simulation | `call_classically(**vals)`  | `on_classical_vals(...)`    | [Classical simulation](/simulation/classical_sim.ipynb) |
| Resource counting    | `bloq_counts()`             | `bloq_counts()`             | [Bloq counts](/resource_counting/bloq_counts.ipynb)     |
| Conversion to Cirq   | `to_cirq_circuit(**quregs)` | `as_cirq_op(...)`           | [Cirq Interop](/cirq_interop/cirq_interop.ipynb)        |

I don't know why I didn't follow this from the start with `bloq_counts` but I want to now.

 - override method is passed a `SympySymbolAllocator` by the library so you can integrate out parameters you don't care about
 - user-callable method lets you pass a generalizer or info on when to stop descending into the call graph.

I need some help with names. It would be nice if you could tell via grammar or something which methods are which flavor, but I don't know exactly what that would look like. So far: the override names are generally more words that contain more prepositions.

Some options

 - `Bloq.get_counts_graph() --> `def my_bloq_counts(self, ssa)`
 - `Bloq.counts_graph()` --> `count_my_bloqs(self, ...)`
 - ??

tanujkhattar:
My recommendation would be to rename to a more general `call_graph` terminology, since that's what everyone was calling it whenever I showed the diagram to new people at IEEE Quantum Week. It's also good to not have the word "bloq" in the name of the protocol so that we can also add the same protocol to Cirq's GateWithRegisters class once the code is moved over the qualtran repo. 

Tl;Dr my suggestion is to have 
[...]

PR #380, opened 2023-10-06
Move drawing code to `drawing`

mpharrigan:
- move bloq counts drawing functionality to `drawing` package
 - `show_xxx` jupyter helper functions for all the different things
 - organize `qualtran.drawing` submodules: split classical sim drawer and bloq counts graph into their own modules


qualtran/bloqs/basic_gates.ipynb
```diff
@@ -418,9 +418,12 @@
...
    "outputs": [],
    "source": [
-    "from qualtran.resource_counting import get_bloq_counts_graph, GraphvizCounts\n",
+    "from qualtran.resource_counting import get_bloq_counts_graph\n",
+    "from qualtran.drawing import show_counts_graph, show_counts_sigma\n",
+    "\n",
     "g, sigma = get_bloq_counts_graph(bloq)\n",
-    "GraphvizCounts(g).get_svg()"
+    "show_counts_graph(g)\n",
+    "show_counts_sigma(sigma)"
```
tanujkhattar, line 426: Can we have a single helper to show both `counts_graph` and `counts_sigma` ? It would be nice to have a single statement, maybe like `show_bloq_counts(counts)`, that displays both of these quantities instead of the current 3 line workflow where we first get the graph and sigma and then display each of them separately.

mpharrigan: I don't want to merge the count-graph-getting function with the showing function; but I don't think that's what you're proposing.

I _do_ want to put the count-graph-getting function as a method on `Bloq` cc #373 

showing two things with one line? maybe. Can we punt to later?

mpharrigan:
@tanujkhattar anything else?

tanujkhattar (approved):

Issue #219, opened 2023-05-09
Bloqs Primitives Implementation Details


fdmalone:
Several TODOs from #216. Many of the primitives are described in words without some gate decomposition. Will need to be worked out.

- [ ] Implement the bitonic sort algorithm using comparitor bloq.
- [ ] Bitonic sort: Sympy big-O complexity for T cost?
- [ ] Comparator: complexity is from less than on two n qubit numbers + controlled swap. #176 
- [ ] Squaring: missing clifford count
- [ ] Greater than cliffords
- [ ] Cliffords for synthesizing rotations

PR #381, opened 2023-10-08
Add more arithmetic gates.

fdmalone:
Question: How should greater than, greater than constant and equals a constant cost. Is it just the same as <=  and < given in cirq_ft?


fdmalone:
Pulled from #350

fdmalone:
Also, why Toffoli more appropriate than T for contiguous index?

fdmalone:
@NoureldinYosri do you know what the appropriate costs are for equality, greater than a constant and greater than (for two quantum registers)

NoureldinYosri:
@fdmalone yes the number of T gates is the same. The number of cliffords is at most 2 away from its less than counterpart.

A new clean decomposition will have the same clifford count. A decomposition that uses the appropiate less than gate will use either an `X` or a `CNOT` or both. For example greater than is just LessThanOrEqual followed/proceeded by an `X` on the target qubit.

qualtran/bloqs/arithmetic.ipynb
```diff
@@ -361,6 +360,118 @@
...
+    "## `GreaterThanConstant`\n",
+    "Implements $U_a|x\\rangle = U_a|x\\rangle|z\\rangle = |x\\rangle |z ^ (x > a)\\rangle\"\n",
+    "\n",
+    "#### Parameters\n",
+    " - `bitsize`: bitsize of x register.\n",
+    " - `val`: integer to compare x against (a above.) \n",
+    "\n",
+    "#### Registers\n",
+    " - `- x`: Register to compare against val.\n",
+    " - `- result`: Register to hold result of comparison.\n"
```
mpharrigan, line 380: following the PR to appropriately parse the "Registers:" section of a docstring, can you update the registers section of the docstrings to be structured like the Args section. In particular: the docstring should not contain the `-`; that will be added by the system.

Then please re-generate :)

mpharrigan (commented):
can you update the registers section and re-generate the notebooks.

@NoureldinYosri can you provide a review of this PR?

qualtran/bloqs/arithmetic.ipynb
fdmalone: Done.

fdmalone:
Ok, I reverted back to what I did originally in #350 . I use the t_complexity of the appropriate cirq_ft gate to determine the T complexity for the bloq_counts method. I'm not really sure what to do about cliffords in the bloq_counts case as cirq_ft's t_complexity protocol doesn't distinguish between 1, and 2 qubit cliffords.

 For the moment I'm just ignoring the clifford counts (which is how it was before, see #219), but using the more accurate T counts from cirq_ft.

qualtran/bloqs/arithmetic.py
```diff
@@ -444,11 +446,132 @@ def signature(self):
...
+
+    bitsize: int
+    val: int
+
+    @cached_property
+    def signature(self) -> Signature:
+        return Signature.build(x=self.bitsize, result=1)
+
+    def t_complexity(self) -> 'TComplexity':
+        return LessThanGate(self.bitsize, val=self.val).t_complexity()
```
NoureldinYosri, line 516: `LessThanGate` is not the correct one here. Equality is always $4 (bitsize - 1)$ regardlress of whether it's between a register and a constant or 2 registers. see https://qualtran.readthedocs.io/en/latest/bloqs/comparison_gates.html#equality-as-a-special-case

NoureldinYosri (changes requested):
> cirq_ft's t_complexity protocol doesn't distinguish between 1, and 2 qubit cliffords.

cirq_ft counts single qubit cliffords + CNOT. 

Exact count of clifford is not a precise statement since  product of two cliffords is still a clifford. e.g. a decomposition that has `X` followed by a `Y` can later be optimized into just a `Z` up to a phase.

so for now lets report the same clifford counts that cirq_ft reports.

fdmalone:
I meant more that in the bloq counts one can return ArbitraryClifford(n), where n in the number of qubits the gate acts on. It feels a little misleading (or at least maybe surprising) to return say (t_comp.clifford, ArbitraryClifford(n=2)), when this clifford count has n=1, and n=2 qubit gates no?

qualtran/bloqs/arithmetic.ipynb
```diff
@@ -361,6 +360,119 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "c63ef890",
+   "metadata": {
+    "cq.autogen": "_make_greater_than_constant.md"
+   },
+   "source": [
+    "## `GreaterThanConstant`\n",
+    "Implements $U_a|x\\rangle = U_a|x\\rangle|z\\rangle = |x\\rangle |z ^ (x > a)\\rangle\"\n",
```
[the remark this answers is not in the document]
fdmalone: Done.

qualtran/bloqs/arithmetic.py
fdmalone: Done.

qualtran/bloqs/arithmetic.py
```diff
@@ -444,11 +446,132 @@ def signature(self):
...
+        x: Register to compare against val.
+        result: Register to hold result of comparison.
+    """
+
+    bitsize: int
+    val: int
+
+    @cached_property
+    def signature(self) -> Signature:
+        return Signature.build(x=self.bitsize, result=1)
```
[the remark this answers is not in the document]
fdmalone: Done.

NoureldinYosri (approved):

PR #392, opened 2023-10-11
Block Encoding Bloqs RFC

fdmalone:
Following from the discussion at the recent sync: For some of the second quantized chemistry algorithms we need to implement the square of a one-body operator. E.g., the single factorized and double factorized hamiltonians:

Screenshot 2023-10-11 at 12 31 12 PM

These typically involve an outer prepare followed by an inner prepare and select, a reflection about the state preparation ancilla in the inner prepare (controlled on some registers from the outer prepare), and then the inner block is repeated again.

This structure follows from the following theorem in [ref](https://arxiv.org/pdf/2007.14460.pdf)

Screenshot 2023-10-11 at 1 20 43 PM

For SF and DF we need the square, which is the second Chebyshev polynomial (2x^2 - 1) up to some factors of 2 and an identity term.*

This WIP PR defines a block encoding abstract base class and a bloq to implment the above theorem for T_j[H] through a series of reflections and block encodings of H. To complete the walk we need to reflect about the inner and outer state preparation ancilla.

This is mostly a sketch at this time and I will split it into 3 PRs (1. block encoding, 2. chebyshev, 3. single factorization) when ready. I included them all just to see a complete workthrough.

## Block(q) encoding bloq

This just defines control, selection, and target registers, from which the signature is derived. Not sure if control_registers are necessary or this will come from a controlled_by method in the future.

## ChebyshevPolynomial

This accepts a block encoding of some Hamiltonian and will build the circuit from Theorem 1 above. I just define bloq_counts at the moment and will add the decomposition shortly.

## SingleFactorization

This implements the circuit from the first reference as a prototype (i.e. there is no decomposition just costs). Going forward I'm going to add chemistry bloqs like this and iteratively implement the sub components. So I wouldn't worry too much about the interfaces which will likely change (e.g. alt / keep values will crop up and qrom blocking factors will disappear in favour of selectswapqrom cirq_ft bloqs.) This top down approach is better to get a rough idea of where the gaps are and also helps document the expected costs from the papers.

* there are some subleties for DF with state preparation discussed in the THC paper in addition to there being some conditional controls in the second application of the one-body hamiltonian to select for two-body only terms (i.e. the one- and two-body Hamiltonians are implemented in one go controlled on a flag register)


fdmalone:
The bloq_counts looks like this:

Screenshot 2023-10-11 at 1 38 17 PM

fdmalone:
cc @mpharrigan @tanujkhattar

fdmalone:
I should say this is very much inspired by the SELECT / PREPARE oracles in cirq_ft.

fdmalone:
Also for DF
Screenshot 2023-10-13 at 2 38 22 PM

mpharrigan:
I encourage you to take a look at #300 

Using python to meta-program the grouping of the various selection, target (, and junk) registers is hiding structure in the python code rather than the compute graph. As a thought exercise: how would you make the system generate a diagram like (33) that could be further decomposed to your first extracted figure

PR #397, opened 2023-10-13
[autotest] Basic structure for automatic testing

mpharrigan:
- `BloqExample` for encapsulating instantiations of bloqs on which we can perform tests (and eventually: document). Consider this an extension of the `_make_xx()` functions currently used in `dev_tools/autogenerate-bloqs-notebooks.py`. Follow-on work will port all these make functions and the autodoc infrastructure to use `BloqExample`.
 - "check" functions that return not only "pass" or "fail" but also "na", "missing", or "unverified" to form the basis of a bloq "report card": a matrix of bloq examples on one axis and checks on the other
 - wrap the check functions in a `pytest` fixture so all the checks can be run on a given bloq example, simplifying running unit tests. Just like the notebook checks: I imagine that each bloq will have a unit test that runs the checks on that bloq for interactive or incremental developing and debugging. Additionally: there will be a script that will run all the tests on all the bloqs. 
 - use bespoke exceptions to indicate failure modes. For example: `DecomposeNotImplementedError` is different than a general exception. 

ptal. I'll add some more docstrings and tester tests. Next PR will make bloq examples for each bloq

### DIFF
--- a/qualtran/_infra/bloq_example.py
+++ b/qualtran/_infra/bloq_example.py
@@ -1,0 +1,75 @@
+#  Copyright 2023 Google LLC
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+from typing import Callable, Optional, Type
+
+from attrs import field, frozen
+
+from .bloq import Bloq
+
+
+@frozen
+class BloqExample:
+    _func: Callable[[], Bloq] = field(repr=False, hash=False)
+    name: str
+    bloq_cls: Type[Bloq]
+    generalizer: Callable[[Bloq], Optional[Bloq]] = lambda x: x
+
+    def make(self) -> Bloq:
+        """Make the bloq."""
+        return self._func()
+
+    def __call__(self) -> Bloq:
+        """This class is callable: it will make the bloq.
+
+        This makes the `bloq_example` decorator make sense: we wrap a function, so this
+        callable should do the same thing when called.
+        """
+        return self.make()
+
+
+def _name_from_func_name(func: Callable[[], Bloq]):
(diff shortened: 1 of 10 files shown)


mpharrigan:
@tanujkhattar can you PTAL

qualtran/testing.py
```diff
@@ -204,3 +217,86 @@ def execute_notebook(name: str):
         nb = nbformat.read(f, as_version=4)
     ep = ExecutePreprocessor(timeout=600, kernel_name="python3")
     ep.preprocess(nb)
+
+
+class BloqCheckResult(Enum):
```
tanujkhattar, line 222: Do you think it's time to create a folder for `testing/` and have separate files for the separate logics? I suspect this file will continue to grow and become hard to manage at some point, but we can punt it for later as well. Good to keep this at the back of the mind as we continue to add more checks.

qualtran/_infra/bloq_example.py
```diff
@@ -0,0 +1,92 @@
...
+
+    Args:
+        func: The function that returns the bloq instantiation. Calling the `BloqExample` instance
+            will call this function.
+        name: A name for the bloq instantiation.
+        bloq_cls: The `Bloq` class that this instantiation is an instance of.
+        generalizer: Passed to `get_bloq_counts_graph` calls for bloq-counts equivalence checking.
+    """
+
+    _func: Callable[[], Bloq] = field(repr=False, hash=False)
```
tanujkhattar, line 38: Can we also add a convertor so that if user directly provides a `bloq` then we can wrap it in a trivial lambda?  I can imagine calling `bloq_tester(BloqExample(bloq))` to make sure the current bloq passes all checks?

tanujkhattar (approved):
LGTM % two nits

qualtran/testing.py
mpharrigan: Sorry, what's the separate logic? I think this one file is pretty cohesive for now (and only 300 lines)

qualtran/_infra/bloq_example.py
mpharrigan: Fair question, but I plan to use these functions in place of the current `xx_test._make_xx()` jupyter notebook autogen functions where we secretly extract the source code from it. It adds a little boilerplate for developers but will unify the examples that we document with the examples that we (auto)test.

qualtran/testing.py
tanujkhattar: a bunch of functions to test consistency of composite bloqs, which is in some sense different from consistency checks for a Bloq in general. I imagine going forward this would evolve into a `testing/` directory like cirq and each consistency check would have it's own file. But we don't need to do that right now

PR #407, opened 2023-10-15
Generalized QSP

anurudhp:
An implementation of [Generalized Quantum Signal Processing (arxiv:2308.01501)](https://arxiv.org/abs/2308.01501).


anurudhp:
Hey @tanujkhattar, could you take a look at this?

qualtran/bloqs/qsp_bloq.py
```diff
@@ -0,0 +1,149 @@
...
+            S = _arbitrary_SU2_rotation(theta[d], phi[d], 0) @ S
+            S = np.array([S[0][1:d], S[1][0 : d - 1]])
+
+    return {'theta': theta, 'phi': phi, 'lambda': lambd}
+
+
+@frozen
+class QEVTCircuit(GateWithRegisters):
+    r"""Applies a QSP sequence described by a pair of polynomials $P, Q$, to a unitary $U$ to obtain a block-encoding of $P(U)$.
+    The exact circuit is described in Figure 2.
```
tanujkhattar, line 91: nit: Each line in the docstring should be < 100 characters. Also, the summary line should be a single line followed by an empty line followed by a detailed description. 

See https://peps.python.org/pep-0257/#multi-line-docstrings

qualtran/bloqs/qsp_bloq.py
```diff
@@ -0,0 +1,149 @@
...
+
+    When $U$ encodes a Hermitian matrix $H$ as $U = e^{iH}$, then this bloq encodes the following unitary:
+        $\begin{matrix} P(U) & \cdot \\ Q(U) & \cdot \end{matrix}$.
```
tanujkhattar, line 94: Is there a reason we mention "$U$ encodes a Hermitian matrix $H$" ? Every unitary $U$ encodes some hermitian matrix $H$ s.t. $U = e^{iH}$. Also, the unitary of the final block encoding does not depend upon what $H$ is, correct?

qualtran/bloqs/qsp_bloq.py
```diff
@@ -0,0 +1,149 @@
...
+            Motlagh and Wiebe. (2023). Theorem 3; Figure 2.
+    """
+
+    U: GateWithRegisters
+    P: Sequence[float]
+    Q: Sequence[float]
+
+    @property
+    def signature(self) -> Signature:
+        return Signature((Register(name='signal', bitsize=1),) + tuple(self.U.signature))
```
tanujkhattar, line 115:
```suggestion
        return Signature([Register(name='signal', bitsize=1), *self.U.signature])
```

qualtran/bloqs/qsp_bloq_test.py
```diff
@@ -0,0 +1,24 @@
...
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import numpy as np
+
+from qualtran.bloqs.qsp_bloq import qsp_phase_factors
+
+
```
tanujkhattar, line 19: Can we add a test that checks for unitary correctness of the final `QSP` block? I'm thinking of a parameterized test that takes in:
1) `U` : A random single / two qubit unitary generated using `cirq.testing.random_unitary`. You can create a test class that derives from `GateWithRegisters` and has this unitary.  
2) `P`, `Q`: Coefficients (both real / complex?) for polynomials with degree 2, 3 & 4

and then assert that top-left and bottom-left blocks of `cirq.unitary(QSP(U, P, Q))` correspond to `P(U)` `Q(U)` respectively?

qualtran/bloqs/qsp_bloq.py
```diff
@@ -0,0 +1,149 @@
...
+        Q: Co-efficients of a complex polynomial.
+
+    References:
+        [Generalized Quantum Signal Processing](https://arxiv.org/abs/2308.01501)
+            Motlagh and Wiebe. (2023). Theorem 3; Figure 2.
+    """
+
+    U: GateWithRegisters
+    P: Sequence[float]
+    Q: Sequence[float]
```
tanujkhattar, line 111: P & Q can be complex, correct?

tanujkhattar (changes requested):
@anurudhp Thanks for opening the PR! This is a great start!  I've left a few comments. 

Also, I think it'd be great if we can implement Corollary 5 and Corollary 7 as well; but we can open an issue to track it and this can be done in follow up PRs.

qualtran/bloqs/qsp_bloq.py
```diff
@@ -0,0 +1,149 @@
...
+
+    @property
+    def signature(self) -> Signature:
+        return Signature((Register(name='signal', bitsize=1),) + tuple(self.U.signature))
+
+    @cached_property
+    def _qsp_phases(self):
```
tanujkhattar, line 118: Please add return type annotations to all methods

qualtran/bloqs/qsp_bloq.ipynb
```diff
@@ -0,0 +1,99 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "initial_id",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "\n",
+    "import cirq\n",
+    "from cirq_ft.algos.qubitization_walk_operator_test import get_walk_operator_for_1d_Ising_model\n",
```
tanujkhattar, line 13: This has been migrated over to qualtran, so please rebase and use the qualtran bloqs.
```suggestion
    "from qualtran.bloqs.qubitization_walk_operator_test import get_walk_operator_for_1d_Ising_model\n",
```

qualtran/bloqs/qsp_bloq.py
anurudhp: Ah yes, you are correct. (I got it mixed up with the later corollaries.) Fixed now.

qualtran/bloqs/qsp_bloq_test.py
anurudhp: Yep will add this. Unfortunately there are no examples in the paper for computing Q from P, so I'll hand-craft some polynomials for now.

qualtran/bloqs/qsp_bloq.py
```diff
@@ -0,0 +1,165 @@
...
+            lambd = np.angle(b)
+        else:
+            S = _arbitrary_SU2_rotation(theta[d], phi[d], 0) @ S
+            S = np.array([S[0][1:d], S[1][0 : d - 1]])
+
+    return theta, phi, lambd
+
+
+@frozen
+class GeneralizedQSP(GateWithRegisters):
```
mpharrigan, line 103: Generally, we'll try to match the module name with the bloq class name. You don't need `_bloq` in the module name.

`generalized_qsp.py`

qualtran/bloqs/qsp_bloq.py
```diff
@@ -0,0 +1,165 @@
...
+        Q: Co-efficients of a complex polynomial.
+
+    References:
+        [Generalized Quantum Signal Processing](https://arxiv.org/abs/2308.01501)
+            Motlagh and Wiebe. (2023). Theorem 3; Figure 2.
+    """
+
+    U: GateWithRegisters
+    P: Sequence[complex]
+    Q: Sequence[complex]
```
mpharrigan, line 127: can we use more meaningful and pep8 compliant variable names for these?

qualtran/bloqs/qsp_bloq.py
anurudhp: renamed to `generalized_qsp.py`

mpharrigan:
What's the status of this? cc @tanujkhattar  ?

anurudhp:
Hey @mpharrigan, @tanujkhattar, sorry for the delay, I did not get much time during the week.

So far:
- I've implemented symbolic tests to verify the QSP sequence is correct
- Tested for random unitaries (sanity check)

TODO:
- Need more example polynomials to test against.
- Implement the optimization algorithm to compute `Q` from `P`

Both these might take a bit of time and effort, so perhaps its better to track this in separate issues/PRs?

qualtran/bloqs/qsp_bloq.py
anurudhp: Any suggestions? For `U`, Is there some naming convention for nested bloqs?

As for `P, Q`, I'm not really sure what to name them (as the literature also doesn't explicitly name them). Perhaps I could call them qsp_polynomials and store them as a tuple instead?

tanujkhattar (approved):
LGTM! Let's get this merged and continue iterating to add more features as described. Thanks Anurudh!

PR #401, opened 2023-10-13
Double factorization

fdmalone:
Add double factorization bloq counts. Dependent on #392.

Issue #395, opened 2023-10-13
Add openfermion dependency?


fdmalone:
It can be convenient for testing to compare to openfermion's existing cost estimates. yay/nay on adding it as a dependency?

mpharrigan:
We have a notion of test dependencies vs runtime dependencies. If it's just used for tests then go 4 it

mpharrigan:
If adding it only as a test dependency or doc-dependency is sufficient, I'm going to close this issue and you can open your PR at your leisure.

If you need to take a runtime dependency on openfermion, let's discuss what is specifically needed.

Issue #390, opened 2023-10-11
THC cost differences


fdmalone:
## Uniform state preparation 
1. inverting inequality tests should not need more toffolis.
2. We are not using phase-gradient gate toffoli cost for Ry rotations
3. Small differences in quoted vs implemented comparator costs.

## Prepare
1. Comparator costs differences.
2. QROAM variations #368

mpharrigan:
I'm trying to port `test_outerprep_t_counts` to the `QECGatesCost` infrastructure. I was hoping to directly use the toffoli count for the variable `toff`, but the counts include rotations. I assume this is the same problem as point (2)

fdmalone:
That's right, although I think rotations via phase gradient have been added?

mpharrigan:
```patch
diff --git a/qualtran/bloqs/chemistry/sf/prepare_test.py b/qualtran/bloqs/chemistry/sf/prepare_test.py
index b87ddb7b..5a64c5de 100644
--- a/qualtran/bloqs/chemistry/sf/prepare_test.py
+++ b/qualtran/bloqs/chemistry/sf/prepare_test.py
@@ -14,7 +14,7 @@
 
 from openfermion.resource_estimates.utils import power_two, QI, QI2, QR, QR2
 
-from qualtran.bloqs.basic_gates import TGate
+from qualtran.bloqs.basic_gates import TGate, Toffoli
 from qualtran.bloqs.chemistry.sf.prepare import (
     _prep_inner,
     _prep_outer,
@@ -24,6 +24,7 @@ from qualtran.bloqs.chemistry.sf.prepare import (
 from qualtran.bloqs.state_preparation.prepare_uniform_superposition import (
     PrepareUniformSuperposition,
 )
+from qualtran.resource_counting import get_cost_value, QECGatesCost
 
 
 def test_prep_inner(bloq_autotester):
@@ -43,21 +44,18 @@ def test_outerprep_t_counts():
     outer_prep = OuterPrepareSingleFactorization(
         num_aux, num_bits_state_prep=num_bits_state_prep, num_bits_rot_aa=num_bits_rot_aa
     )
-    _, counts = outer_prep.call_graph()
-    toff = counts[TGate()]
+    # Note: this contains rotations, but we just divide the total T count by 4 to approximate
+    # the toffoli count: https://github.com/quantumlib/Qualtran/issues/390
+    toff = get_cost_value(outer_prep, QECGatesCost()).total_t_count(ts_per_cswap=4)
[...]

Issue #368, opened 2023-09-25
Many Variations of QROAM


fdmalone:
I was a little confused about the expected QROM costs when implementing something for THC.

There are at least four possible expected (Toffoli) costs:
1. No dirty ancilla [1] : d / k + M * (k-1) (this is used for THC prepare)
2. Dirty Ancilla [1]: 2 d / k + 4 * M * (k - 1)
3. No dirty ancilla [2] : d / k + 2 * M * k
4. Dirty ancilla [2] : 2 d / k + 4 M * k

d = data size, k = block size, M = number of bits of output.

[1]: https://quantum-journal.org/papers/q-2019-12-02-208/pdf/ (page 9)
[2] https://arxiv.org/pdf/1812.00954.pdf Table II, rows 3 and 4, 4th column

fdmalone:
Relevant Screenshot from [1] 
![Screenshot 2023-09-25 at 3 58 11 PM](https://github.com/quantumlib/Qualtran/assets/12097876/869d8d44-ed37-4c8a-a3f2-2e32e66af2f1)

fdmalone:
Some code

``` python 

from qualtran.cirq_interop import CirqGateAsBloq
from cirq_ft.algos.select_swap_qrom import SelectSwapQROM
num_bits = 15
data_size = 230
num_blocks = 2
sel_swap = SelectSwapQROM(
        (10,)*data_size,
        target_bitsizes=(15,),
        block_size=2**num_blocks,
    )
qroam = CirqGateAsBloq(
    sel_swap
)
print(qroam.t_complexity())
cost_1 = int(np.ceil(data_size/2**num_blocks))
cost_2 = num_bits * (2**num_blocks - 1)
print(f"THC cost (no dirty ancilla) = {4*(cost_1 + cost_2)}")
print(f"THC cost (dirty ancilla) = {4*(2* cost_1 + 4*cost_2)}")
cost_1 = int(np.ceil(data_size/2**num_blocks))
cost_2 = num_bits * (2**num_blocks)
print(f"Select Swap QROM (no dirty ancilla) = {4*(cost_1 + 2 * cost_2)}")
cost_2 = num_bits * (2**num_blocks)
print(f"Select Swap QROM (dirty ancilla) = {4*(2 * cost_1 + 8 * cost_2)}")
```

fdmalone:
[...]

PR #409, opened 2023-10-16
Second quantized sparse hamiltonian Select and Prepare

fdmalone:
Adds the second quantized sparse hamiltonian select and prepare bloqs. Currently no decompositions but following strategy of #392. Costs match those from openfermion and do not require the block encoding bloq so should be good to go.


fdmalone:
Currently blocking myself with #395

qualtran/bloqs/chemistry/sparse.py
```diff
@@ -0,0 +1,208 @@
...
+from qualtran import Bloq, Register, Signature
+from qualtran.bloqs.basic_gates import TGate
+from qualtran.bloqs.swap_network import CSwapApprox
+
+if TYPE_CHECKING:
+    from qualtran.resource_counting import SympySymbolAllocator
+
+
+@frozen
+class PrepareUniformSuperposition(Bloq):
```
mpharrigan, line 37: What's the naming scheme here? In the THC you have

`UniformSuperpositionTHC` and `PrepareTHC`. The name should imply where in the matrix this bloq resides

fdmalone: Should I continue with UniformSuperpositionSparse, etc. They all have potentially slightly different variants of this for each of the algorithms. UniformSuperposition is a subroutine of Prepare typically.

fdmalone: I was starting to think that the *Sparse/SF/DF/THC part of the name became somewhat redundant since it's in the sparse module?

mpharrigan: I'll defer to you for the actual naming scheme; but it should be consistent and explainable.

Relying (solely) on the module to disambiguate may present problems down the line. Elsewhere in qualtran, we use `from module import TheBloq` as a convention because 1) pycharm will add the imports automatically for you 2) I personally find the usages of things more readable if there aren't varying levels of namespacing and abbreviation around them. We could use `sparse.UniformSuperposition` everywhere but might still be easy to mess up. The second reason is that the bloq's name shows up in diagrams. The `short_name` may be shared and ambiguous but it's probably useful for the repr at least to indicate which one is being used.

tanujkhattar: @fdmalone Isn't this exactly same as https://github.com/quantumlib/Qualtran/blob/dc4fecf9d0b761f30d4f896fb49558d8591651d1/qualtran/bloqs/prepare_uniform_superposition.py#L29 ?

tanujkhattar: Oh, here the `d` basis states can by `_any_` `d` basis states and not just the first `d` basis states? The description of the algorithm looks very similar to the what we do for the first `d` basis states, so I would personally want to just generalize the implementation of `PrepareUniformSuperposition` so it takes `d` basis states as input with the default being `[0, d-1]`.

fdmalone: It is, but I want to decouple the counts from the implementation as a first pass (see #392). That way I can come back and systematically compare the costs listed in the papers to the implementation. Uniform superposition crops up everywhere in various forms but the costs don't necessarily match those in the literature due to say, inequalities being invertable for free (#390), and so on. Similarly QROAM costs differ from those in the paper (#368), inequalities also differ, rotations cost might differ etc etc. Having a ground "truth" first before implementing a decomposition is easier IMO to track differences.

This way I can come back and substitute out one subbloq at a time with the decomposition and I can more easily track any sources of disagreement. I think this is the best way to go rather than getting bogged down like I did with #350

fdmalone: All of that is to say #401 #416 and this will contain redundant code, the main idea is to adopt and correclty partition the costs from openfermion so we can then carefully and systematically make comparisons.

tanujkhattar: Okay, that makes sense. I would suggest we rename it to `UniformSuperpositionSparse` then and open a separate issue (or add a note if there's already an existing issue) documenting that we can probably make incremental changes to `PrepareUniformSuperposition` that can help replace some / all of these new uniform state prep bloqs. 

The incremental change to `PrepareUniformSuperposition`'s implementation can be a good 20% / community contribution task so I think it's worth tracking. 

LGTM for having separate bloqs for now.

fdmalone: Cool. Yeah I actually wrote several variations of state preparation when I was getting lost in the weeds in THC (signed state, two registers, and upper triangular (buggy) as cirq circuits) which maybe we could add as a notebook (I think I need your amplitude amplification expertise again).

The lesson here was that it was probably better for me not to do it this way lol. Top down is the way to go, there are too many little details.

fdmalone:
@mpharrigan / @tanujkhattar This is now passing CI, I'll need to make some updates following #413

mpharrigan (approved):
some of the tests are a bit (removes sunglasses) ... sparse.

But I'm good to merge and refine

fdmalone:
It's sort of hard to test without decompositions...

PR #410, opened 2023-10-16
Add openfermion to pytest env.

fdmalone:
Add openfermion to the pytest environment #395 and needed for #409

### DIFF
--- a/dev_tools/requirements/deps/pytest.txt
+++ b/dev_tools/requirements/deps/pytest.txt
@@ -9,3 +9,6 @@ pytest-xdist
 nbformat
 nbconvert
 ipykernel
+
+# chemistry resource estimates
+openfermion[resources]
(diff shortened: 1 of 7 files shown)


fdmalone:
I noticed this warning when re pip compiling in the docker container:

```
# The following packages are considered to be unsafe in a requirements file:
# pip
# setuptools
```

fdmalone:
Actually hold off on this for the moment. @mpharrigan What's the best way to pull in the additional requirements required for resource estimates. I also need to cut an openfermion release for pbc resources.

mpharrigan:
> I noticed this warning when re pip compiling in the docker container:

this isn't a warning so much as a note it puts in

mpharrigan:
> What's the best way to pull in the additional requirements required for resource estimates. I also need to cut an openfermion release for pbc resources.

I'd say this current approach is correct. Mark the PR as draft if it's not ready to merge

mpharrigan (approved):

PR #413, opened 2023-10-17
Bloq.call_graph instead of "bloq_counts"

### LINKED ISSUE #373: Separate methods for overriding vs calling in bloq counts

mpharrigan:
Fixes #373 

 - override `build_call_graph(...)`
 - call `call_graph()`
 - can also call `bloq_counts` for a dictionary of one level of counts.

This obviates most usages of the free-functions in `qualtran.resource_counting.bloq_counts`. 

I also added some fun helpers to the dev_tools. I plan on adding more codebase-wide auditing scripts here in the future.


mpharrigan:
@tanujkhattar what do you think cc #373

qualtran/_infra/bloq.py
```diff
@@ -231,16 +233,36 @@ def add_my_tensors(
...
-        return self.decompose_bloq().bloq_counts(ssa)
+        return self.decompose_bloq().build_call_graph(ssa)
+
+    def call_graph(
+        self,
+        generalizer: Callable[['Bloq'], Optional['Bloq']] = None,
+        keep: Optional[Sequence['Bloq']] = None,
+        max_depth: Optional[int] = None,
+    ) -> Tuple['nx.DiGraph', Dict['Bloq', Union[int, 'sympy.Expr']]]:
+        from qualtran.resource_counting.bloq_counts import get_bloq_call_graph
```
tanujkhattar, line 257: please add docstrings for the arguments.

qualtran/_infra/bloq.py
```diff
@@ -231,16 +233,36 @@ def add_my_tensors(
...
+
+        return get_bloq_call_graph(self, generalizer=generalizer, keep=keep, max_depth=max_depth)
+
+    def bloq_counts(
+        self, generalizer: Callable[['Bloq'], Optional['Bloq']] = None
+    ) -> Dict['Bloq', Union[int, 'sympy.Expr']]:
+        _, sigma = self.call_graph(generalizer=generalizer, max_depth=1)
```
tanujkhattar, line 264: Please add docstrings.

qualtran/_infra/bloq.py
```diff
@@ -231,16 +233,36 @@ def add_my_tensors(
...
         """
-        return self.decompose_bloq().bloq_counts(ssa)
+        return self.decompose_bloq().build_call_graph(ssa)
+
+    def call_graph(
+        self,
+        generalizer: Callable[['Bloq'], Optional['Bloq']] = None,
+        keep: Optional[Sequence['Bloq']] = None,
+        max_depth: Optional[int] = None,
+    ) -> Tuple['nx.DiGraph', Dict['Bloq', Union[int, 'sympy.Expr']]]:
```
tanujkhattar, line 256: It's a little confusing that `build_call_graph` is supposed to return `(n, bloq)` tuples but the `call_graph` returns `(bloq, n)`. Maybe let's change to `(bloq, n)` at both the places to keep things consistent?

tanujkhattar (commented):
Looking good overall, left a few comments

qualtran/_infra/bloq.py
mpharrigan: yeah good point.

The first is tuples so it doesn't really matter and the latter is a dictionary where it does matter and bloq should be key; so sadly we have to make a change. But I agree that they should be consistent.

mpharrigan:
@tanujkhattar ptal

qualtran/_infra/bloq.py
```diff
@@ -249,16 +251,73 @@ def add_my_tensors(
...
+                the number of times successor bloq is called via its predecessor.
+            sigma: Call totals for "leaf" bloqs. We keep a bloq as a leaf in the call graph
+                according to `keep` and `max_depth` (if provided) or if a bloq cannot be
+                decomposed.
+        """
+        from qualtran.resource_counting.bloq_counts import get_bloq_call_graph
+
+        return get_bloq_call_graph(self, generalizer=generalizer, keep=keep, max_depth=max_depth)
+
+    def bloq_counts(
```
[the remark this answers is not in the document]
tanujkhattar: Also, I see that most of the usages are either in tests or notebooks (or `t_complexity` method, which looks redundant in the first place), so that's further evidence that we can get rid of the method without compromising code readability.

tanujkhattar (commented):
One main suggestion to get rid of the `bloq_counts` method. LGTM otherwise.

qualtran/_infra/bloq.py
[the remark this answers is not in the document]
mpharrigan: you'd need `call_graph(max_depth=1)[1]`

yes, usages are primarily in "user code". This method is to prevent users from calling `build_call_graph` directly. There is a fundamental difference between finding the immediate children bloqs vs the recursive call graph

tanujkhattar (approved):

fdmalone:
This should probably go in soon to avoid more merge pain with?

Issue #414, opened 2023-10-17
Tensor contraction fails for composite bloqs where a register is not acted upon by any Bloq


tanujkhattar:
```python
from qualtran import BloqBuilder, Signature, SelectionRegister
from qualtran.bloqs.util_bloqs import Split, Join
from qualtran.drawing import show_bloq
from qualtran.simulation.quimb_sim import cbloq_to_quimb, _cbloq_as_contracted_tensor_data_and_inds
signature = Signature.build(selection=3, target=1)
bb, soqs = BloqBuilder().from_signature(signature=signature)
selection = soqs['selection']
selection = bb.add(Split(3), split=selection)
selection = bb.add(Join(3), join=selection)
cbloq = bb.finalize(selection=selection, target=soqs['target'])
show_bloq(cbloq)
cbloq.tensor_contract()
```

<img width="425" alt="image" src="https://github.com/quantumlib/Qualtran/assets/7863287/e3b2d5cf-f94a-4e0e-88c4-d98c5fb825d6">

<details>

```
File ~/opt/anaconda3/envs/qualtran/lib/python3.11/site-packages/quimb/tensor/tensor_core.py:7979, in TensorNetwork.to_dense(self, to_qarray, *inds_seq, **contract_opts)
   7974 """Convert this network into an dense array, with a single dimension
   7975 for each of inds in ``inds_seqs``. E.g. to convert several sites
   7976 into a density matrix: ``TN.to_dense(('k0', 'k1'), ('b0', 'b1'))``.
   7977 """
   7978 tags = contract_opts.pop('tags', all)
-> 7979 t = self.contract(
   7980     tags,
   7981     output_inds=tuple(concat(inds_seq)),
   7982     preserve_tensor=True,
   7983     **contract_opts
   7984 )
   7985 return t.to_dense(*inds_seq, to_qarray=to_qarray)

File ~/opt/anaconda3/envs/qualtran/lib/python3.11/site-packages/quimb/tensor/tensor_core.py:7774, in TensorNetwork.contract(self, tags, output_inds, optimize, get, backend, preserve_tensor, max_bond, inplace, **opts)
   7772 # contracting everything to single output
   7773 if all_tags and not inplace:
-> 7774     return tensor_contract(*self.tensor_map.values(), **opts)
   7776 # contract some or all tensors, but keeping tensor network
   7777 return self.contract_tags(tags, inplace=inplace, **opts)

[...]

Issue #398, opened 2023-10-13
Track short lived code that should be removed after migrating Cirq-FT code to Qualtran


tanujkhattar:
As part of [Qualtran & Cirq-FT Integration](https://docs.google.com/document/d/1miypjcCNt6JV_qAypUVPy5eH3EmKadLqeZomA5QoWK0/edit?usp=sharing&resourcekey=0-ciQ8jPfQmK8gtey2lNKesw), we need to move all code from Cirq-FT to Qualtran. During this transition, Qualtran will contain redundant code to support both - the code that have been moved to Qualtran and the code that is yet to be moved

This issue is to track all the short lived glue code which will exist during this transition period and should be removed once the transition is completed.

PR #415, opened 2023-10-17
Port `swap_network` from Cirq-FT and make further improvements to `GateWithRegisters`

tanujkhattar:
Part of [Qualtran & Cirq-FT Integration](https://docs.google.com/document/d/1miypjcCNt6JV_qAypUVPy5eH3EmKadLqeZomA5QoWK0/edit?usp=sharing&resourcekey=0-ciQ8jPfQmK8gtey2lNKesw)

This PR makes a bunch of improvements

- Fixes https://github.com/quantumlib/Qualtran/issues/414 by adding an identity tensor for unused variables in the Tensor Network corresponding to `CompositeBloq`.
- Adds fallbacks to `GateWithRegisters` so that even if the users override the `build_decompose_bloq` method on a `GateWithRegisters`, they can get `decompose_from_registers` for free without a wrapper. See implementation and tests of `SwapWithZero` and `CSwap` for more details. 
- In this new workflow, the guidance for users would be to define `decompose_from_registers` or `build_decomposite_bloq` for composite bloqs that have a decomposition; and override `as_cirq_op` only for the primitive gates (for example: cirq.CSWAP and `TwoBitCSwap`). As we've discussed before, `as_cirq_op` should be used to provide a 1:1 correspondence with a corresponding Cirq operation whereas `decompose_from_registers` should be used to provide a cirq-style decomposition. 

Get's rid of a few TODOs linking to https://github.com/quantumlib/Qualtran/issues/398


qualtran/_infra/gate_with_registers.py
```diff
@@ -157,12 +157,15 @@ def decompose_bloq(self) -> 'CompositeBloq':
             A CompositeBloq containing the decomposition of this Bloq.
 
         Raises:
-            NotImplementedError: If there is no decomposition defined; namely: if
+            DecomposeNotImplementedError: If there is no decomposition defined; namely: if
```
mpharrigan, line 160: the rest of this docstring is no longer accurate

qualtran/_infra/gate_with_registers.py
```diff
@@ -157,12 +157,15 @@ def decompose_bloq(self) -> 'CompositeBloq':
...
                 `build_composite_bloq` returns `NotImplemented`.
         """
         from qualtran.cirq_interop._cirq_to_bloq import decompose_from_cirq_op
 
-        return decompose_from_cirq_op(self, decompose_once=True)
+        try:
+            return Bloq.decompose_bloq(self)
+        except DecomposeNotImplementedError:
+            return decompose_from_cirq_op(self, decompose_once=True)
```
mpharrigan, line 168: can you please update this methods docstring to describe the fallback behavior there (and why I would want it)

qualtran/_infra/gate_with_registers.py
```diff
@@ -172,17 +175,21 @@ def add_my_tensors(
...
-            self.signature,
-            self.short_name(),
-            tn=tn,
-            tag=tag,
-            incoming=incoming,
-            outgoing=outgoing,
-        )
+        try:
+            # First try to add tensors via decomposition.
+            Bloq.add_my_tensors(self, tn, tag, incoming=incoming, outgoing=outgoing)
```
mpharrigan, line 180: super().add_my_tensors(...)? or maybe we should factor out the body of the default `Bloq.add_my_tensors` into a descriptively named, private free function

qualtran/_infra/gate_with_registers.py
```diff
@@ -172,17 +175,21 @@ def add_my_tensors(
...
+        except DecomposeNotImplementedError:
+            from qualtran.cirq_interop._cirq_to_bloq import _add_my_tensors_from_gate
+
+            _add_my_tensors_from_gate(
```
mpharrigan, line 184: I'm just ... can you document the use case of when we'd rely on this fallback vs the other? You know I don't like multiple default fallbacks. 

The alternative would be for bloq(/gate) authors to be explicit about which fallback they're relying on.

qualtran/simulation/quimb_sim.py
```diff
@@ -120,6 +124,18 @@ def _assign_outgoing(cxn: Connection) -> Soquet:
         if pos is not None:
             fix[tuple([binst])] = pos[binst]
 
+    inds = get_right_and_left_inds(cbloq.signature)
+    if len(inds) != 2:
+        return tn, fix
+
+    l, r = inds
```
fdmalone, line 131: Some comments here about what is going on here would be helpful (since this was a bug, it'd be good make a note).

qualtran/simulation/quimb_sim.py
```diff
@@ -120,6 +124,18 @@ def _assign_outgoing(cxn: Connection) -> Soquet:
         if pos is not None:
             fix[tuple([binst])] = pos[binst]
 
+    inds = get_right_and_left_inds(cbloq.signature)
+    if len(inds) != 2:
+        return tn, fix
```
mpharrigan, line 129: add a quick comment that this is for the case where the tensor network is a vector (bra or ket) instead of an operation (unitary)

qualtran/simulation/quimb_sim.py
```diff
@@ -120,6 +124,18 @@ def _assign_outgoing(cxn: Connection) -> Soquet:
...
+
+    l, r = inds
+    for in_soq in l:
+        if in_soq not in tn.ind_map:
```
mpharrigan, line 133: my defensive programming instincts would probably prefer to use `cbloq.connections` and just look for any connection between `LeftDangle` and `RightDangle`. 

I think this would remove the need for the nested for loop as well

qualtran/bloqs/swap_network.py
```diff
@@ -116,20 +138,44 @@ def bloq_counts(
 
 
 @frozen
-class SwapWithZero(Bloq):
+class SwapWithZero(GateWithRegisters):
```
fdmalone, line 141: What's going on here? Am I wrong that a GateWithRegisters should implement a decompose_from_registers method and not a build_composite bloq method?

tanujkhattar: It can implement any, and both will work for both the use cases. You'll get the correct expected behavior for both `cirq.Circuit(cirq.decmopose_once(op))` and `show_bloq(bloq.decompose_bloq())`, i.e. it doesn't matter whether the bloq author implements `build_composite_bloq` or `decompose_from_registers`, a Bloq user would be able to do

1. For `cirq.Circuit(cirq.decmopose_once(op))`: And get a decomposed circuit with cirq style native operations, except only for those bloqs which are not defined as `GateWithRegisters`. These bloqs will get wrapped in a `BloqAsCirqGate(bloq)`
2. For `bloq.decompose_bloq()`: You'll get bloq native objects, except for cirq gates which are not defined as `GateWithRegisters`. These gates will get wrapped in a `CirqGateAsBloq(gate)` wrapper. 


So, from our perspective, it is always strictly better to define every new bloq as a `GateWithRegisters` and we can override any of `build_composite_bloq` or `decompose_from_registers` as per our choice and the other side of the interop we get for free.

qualtran/simulation/quimb_sim.py
tanujkhattar: added comments

qualtran/_infra/gate_with_registers.py
tanujkhattar: updated

qualtran/_infra/gate_with_registers.py
tanujkhattar: added a detailed docstring

qualtran/_infra/gate_with_registers.py
tanujkhattar: removed the `add_my_tensors` override. The only benefit now was if `GateWithRegisters` doesn't define a decomposition but defines a `_unitary_` then the fallback would take care of it, but it's not a common use case so I've removed the override.

qualtran/_infra/gate_with_registers.py
tanujkhattar: not relevant anymore. See comment above.

qualtran/simulation/quimb_sim.py
tanujkhattar: done

qualtran/simulation/quimb_sim.py
tanujkhattar: done

tanujkhattar:
@fdmalone @mpharrigan This is ready for a re-review. I've addressed all the comments.

qualtran/_infra/gate_with_registers.py
```diff
@@ -149,16 +148,37 @@ class GateWithRegisters(Bloq, cirq.Gate, metaclass=abc.ABCMeta):
...
         """Decompose this Bloq into its constituent parts contained in a CompositeBloq.
 
-        Bloq users can call this function to delve into the definition of a Bloq. If you're
-        trying to define a bloq's decomposition, consider overriding `build_composite_bloq`
-        which provides helpful arguments for implementers.
+        Bloq users can call this function to delve into the definition of a Bloq. The function
+        returns the decomposition of this Bloq represented as an explicit compute graph wrapped
+        in a `CompositeBloq` object.
+
+        Bloq authors can the bloq's decomposition by overriding any of the two convenience methods:
```
mpharrigan, line 155: "Bloq authors can the"

mpharrigan: I'd personally remove "convenience" and just call them methods. A convenience method in my book is a method that removes a few keystrokes from calling it but under the hood just delegates to the full-featured thing maybe with some more aggressive defaults

qualtran/_infra/gate_with_registers.py
```diff
@@ -149,16 +148,37 @@ class GateWithRegisters(Bloq, cirq.Gate, metaclass=abc.ABCMeta):
...
+
+        - `build_composite_bloq`: Override this method to define a bloq-style decomposition using a
+            `BloqBuilder` builder class to construct the `CompositeBloq` directly.
+        - `decompose_from_registers`: Override this method to define a cirq-style decomposition by
+            yielding cirq style operations applied on qubits.
+
+        Irrespective of the choice of backend chosen by the Bloq authors to implement the
+        decomposition, Bloq users will be able to access both the Bloq-style and Cirq-style
+        interfaces, i.e. for a bloq that inherits from `GateWithRegisters`, usrs can do:
```
mpharrigan, line 164:
```suggestion
        Irrespective of the bloq author's choice of backend to implement the
        decomposition, bloq users will be able to access both the bloq-style and Cirq-style
        interfaces. For example, users can call:
```

qualtran/_infra/gate_with_registers.py
```diff
@@ -149,16 +148,37 @@ class GateWithRegisters(Bloq, cirq.Gate, metaclass=abc.ABCMeta):
...
+
+        - `cirq.decompose_once(bloq.on_registers(**cirq_quregs))`: This will yield a cirq-style
+            `OPTREE` where only bloqs which do not inherit from `GateWithRegisters` will be wrapped
+            in a `BloqAsCirqGate` wrapper.
```
mpharrigan, line 168:
```suggestion
        - `cirq.decompose_once(bloq.on_registers(**cirq_quregs))`: This will yield a `cirq.OPTREE`.
            Bloqs will be wrapped in `BloqAsCirqGate` as needed.
```

qualtran/_infra/gate_with_registers.py
```diff
@@ -149,16 +148,37 @@ class GateWithRegisters(Bloq, cirq.Gate, metaclass=abc.ABCMeta):
...
+        - `bloq.decompose_bloq()`: This will return a `CompositeBloq` where only cirq gates which do
+            not inherit from `GateWithRegisters` will be be wrapped in a `CirqGateAsBloq` wrapper.
```
mpharrigan, line 170:
```suggestion
        - `bloq.decompose_bloq()`: This will return a `CompositeBloq`.
           Cirq gates will be be wrapped in `CirqGateAsBloq` as needed.
```

qualtran/_infra/gate_with_registers.py
mpharrigan: The "bloqs which do not inherit from `GateWithRegisters`" verbiage is clunky, and we should avoid it if possible. One of the perils of multiple inheritance is that the question "is `obj` a bloq" now has tri-state logic. 

"Wrapped in a wrapper" is also a little redundant :)

tanujkhattar: sure. We can communicate this detail in a notebook / tutorial at a later point I guess.

mpharrigan (approved):
Some documentation suggestions. 

LGTM for this PR. 

In general going forward, I'd like to continue to be vigilant about our API surface; enabling conversions because we want to and not because it's information-theoretically possible :)

qualtran/_infra/gate_with_registers.py
tanujkhattar: done

PR #419, opened 2023-10-17
`CirqGateAsBloq` should raise `DecomposeNotImplementedError` if underlying gate does not have a decomposition

tanujkhattar:
Fixes https://github.com/quantumlib/Qualtran/pull/350#discussion_r1362789784

A recent PR updated the error raising and handling logic where `bloq.decompose_bloq()` should raise a `DecomposeNotImplementedError` if there doesn't exist a valid decomposition. `CirqGateAsBloq` was still raising a `ValueError` in some of these cases. This PR updates the logic and thus fixes the bug mentioned in above conversation.


fdmalone (approved):

mpharrigan (commented):
thanks!

PR #420, opened 2023-10-18
Move all `Qubitization` infrastructure from Cirq-FT to Qualtran

tanujkhattar:
Part of [Qualtran & Cirq-FT Integration](https://docs.google.com/document/d/1miypjcCNt6JV_qAypUVPy5eH3EmKadLqeZomA5QoWK0/edit?usp=sharing&resourcekey=0-ciQ8jPfQmK8gtey2lNKesw)

Specifically, it moves the following Cirq-FT gates as qualtran Bloqs:
- `SelectOracle`
- `GenericSelect` -> `SelectPauliLCU`
- `PrepareOracle`
- `StatePreparationAliasSampling`
- `ReflectionUsingPrepare`
- `QubitizationWalkOperator`

Most of this PR is just copy / paste. The only change worth highlighting is that I've added a condition to `Signature.build()` so that we don't create registers with bitsize `0`.


fdmalone (commented):
This seems fine, I'd just wonder where this leaves #300?

tanujkhattar:
There are parts of #300 which we still need to add to qualtran, for example the ability to construct `Controlled` and `Inverse` bloqs. Once these are added, we can very well replace `cirq.inverse(bloq)` with `bloq.dagger()` and update the implementation of `.controlled()` methods on Bloq's added as part of this PR.

I also support the idea of using `Partition` to have a single `select` register coming in and then splitting it into multiple named registers as part of the decomposition. But these changes can be done incrementally once this is merged. 

The high level interfaces of #300 are essentially the same as this PR, so I don't think there's anything blocking that we should be concerned about.

qualtran/bloqs/reflection_using_prepare.py
```diff
@@ -0,0 +1,133 @@
...
+        if self.control_val is None:
+            qm.qfree([phase_target])
+
+    def _circuit_diagram_info_(self, args: cirq.CircuitDiagramInfoArgs) -> cirq.CircuitDiagramInfo:
+        wire_symbols = ['@' if self.control_val else '@(0)'] * total_bits(self.control_registers)
+        wire_symbols += ['R_L'] * total_bits(self.selection_registers)
+        return cirq.CircuitDiagramInfo(wire_symbols=wire_symbols)
+
+    def __repr__(self):
+        return f'cirq_ft.ReflectionUsingPrepare({self.prepare_gate}, {self.control_val})'
```
fdmalone, line 110: qualtran?

qualtran/linalg/lcu_util_test.py
```diff
@@ -0,0 +1,163 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+"""Tests for lcu_util.py."""
+
+import random
+import unittest
+
+from cirq_ft.linalg.lcu_util import (
```
fdmalone, line 20: qualtran?

fdmalone (changes requested):

qualtran/bloqs/reflection_using_prepare.py
tanujkhattar: done

qualtran/linalg/lcu_util_test.py
tanujkhattar: done

tanujkhattar:
@fdmalone Removed the stale references to `cirq_ft`. Ready for a re-review.

mpharrigan:
> There are parts of https://github.com/quantumlib/Qualtran/pull/300 which we still need to add to qualtran, for example the ability to construct Controlled and Inverse bloqs. Once these are added, we can very well replace cirq.inverse(bloq) with bloq.dagger() and update the implementation of .controlled() methods on Bloq's added as part of this PR.
>
> I also support the idea of using Partition to have a single select register coming in and then splitting it into multiple named registers as part of the decomposition. But these changes can be done incrementally once this is merged.
>
> The high level interfaces of https://github.com/quantumlib/Qualtran/pull/300 are essentially the same as this PR, so I don't think there's anything blocking that we should be concerned about.

Thanks, this SGTM

mpharrigan (approved):
I just skimmed it, but approve of the move.

I reserve the right to organize these into submodules at a later date

tanujkhattar:
@mpharrigan Don't worry, even I want to reorganize this into submodules later  :)

Issue #422, opened 2023-10-18
`CompositeBloq` cannot store isolated Bloqs, like a `GlobalPhaseBloq`.


tanujkhattar:
`CompositeBloq` represents a graph using a list of edges stored as `connections`. This has a fundamental limitation that we cannot store isolated nodes in the graph, for example a `GlobalPhaseBloq` (equivalent to a cirq.GlobalPhaseGate). 

It is often important to store Global phase as part of the compute graph and I think we need a way to get around this limitation. Some possible options are as follows:

1. One way would be to store both nodes (as `FrozenSet[BloqInstance]`) and connections (the way we do now, `Tuple[Connection, ...]`) 
2. A hacky solution would be to store a self loop (i.e. a connection with both the end points on the same node) to represent an isolated node. 

I think way-1 is strictly better and we should do that. cc @mpharrigan

mpharrigan:
What's the use case for storing global phase? That should be an emergent property from the bloqs.

tanujkhattar:
Applying a `CompositeBloq.controlled()` will turn global phase into local phase and give you wrong results the global phase of your composite bloq is not correct.

PR #423, opened 2023-10-18
Add support for storing isolated bloqs, like `GlobalPhaseGate`, in a `CompositeBloq`

tanujkhattar:
Fixes https://github.com/quantumlib/Qualtran/issues/422

Does the following:
1. Updates `CompositeBloq` to store both nodes and edges of the compute graph instead of only storing the nodes
2. Updates `cbloq_to_quimb` to add logic for adding isolated bloqs to the tensor networks
3. Updates `CirqGateAsBloq` to have an empty signature when `cirq.num_qubits(self.gate) == 0`. 
4. Updates `cirq_optree_to_cbloq` to correctly operations acting on 0 qubits. 
5. Tests the entire end to end workflow by wrapping `cirq.GlobalPhaseGate` in a bloq and using tensor contraction to verify that the resulting unitary is identical to the expected cirq unitary. 

This is needed for porting over hubbard model implementation from Cirq-FT to Qualtran.


mpharrigan (changes requested):
this is a pretty fundamental change to our primary datastructure. I don't think we should support scalars as bloqs

tanujkhattar:
What are your concerns? Why do you think we should not support scalars as Bloqs? 


Also, we can not just drop the global phase, so we need to come up with an alternative solution if you think this is not the right way. Do you have alternative proposals in mind?

mpharrigan:
Some historical discussion:

https://github.com/quantumlib/Cirq/pull/4697

https://github.com/quantumlib/Cirq/pull/1743

----

I don't know what you mean by "drop the global phase". Every bloq and composite bloq currently has a well-defined global phase.

tanujkhattar:
> I don't know what you mean by "drop the global phase". Every bloq and composite bloq currently has a well-defined global phase.

Okay, let me try to explain this differently. The use case is that often times it is possible that I want to implement a unitary `U` but, using the existing primitives, I have a way of implementing a unitary `Q` where `Q` and `U` are equal up to global phase. Having access to a global phase gate let's me define a decomposition of `U` in terms of simpler primitives + `GlobalPhaseGate`; which is convenient. It is important to ensure that the decomposition has the same global phase as `U`, or else the auto generated decomposition for a controlled invocation of `U` will result in a wrong unitary. 


Not having the ability to add `GlobalPhaseGate` would make it annoying for us to find ways to shoehorn this global phase into one of the other top level bloqs used in the decomposition; or use an additional qubit in `|1>` state and apply a `Z` gate to it -- both of these are wasteful / inconvenient.

qualtran/cirq_interop/_cirq_to_bloq.py
```diff
@@ -80,7 +80,8 @@ def signature(self) -> 'Signature':
                     for reg in self.gate.signature
                 ]
             )
-        return Signature([Register('qubits', shape=cirq.num_qubits(self.gate), bitsize=1)])
+        shape = cirq.num_qubits(self.gate)
+        return Signature([Register('qubits', shape=shape, bitsize=1)]) if shape else Signature([])
```
mpharrigan, line 84: should `shape` be called `n_qubits`? is it always an integer? If it is: the condition should be `if shape > 0`

qualtran/simulation/quimb_sim.py
```diff
@@ -125,11 +127,13 @@ def _assign_outgoing(cxn: Connection) -> Soquet:
         if pos is not None:
             fix[tuple([binst])] = pos[binst]
 
-    inds = get_right_and_left_inds(cbloq.signature)
-    if len(inds) != 2:
-        # In this case, the tensor network is a vector (bra/ket) instead of a unitary operation.
-        return tn, fix
+    # Special case-1: Add all isolated Bloqs with no connections (eg: Global phase gates)
+    for binst in cbloq.bloq_instances:
+        print(f'DEBUG QUIMB: {binst=}')
```
mpharrigan, line 132: debug print

qualtran/simulation/quimb_sim.py
```diff
@@ -104,8 +104,10 @@ def _assign_outgoing(cxn: Connection) -> Soquet:
             return cxn.right
         return cxn.left
 
+    visited_bloqs: Set[BloqInstance] = set()
     for binst, incoming, outgoing in cbloq.iter_bloqnections():
         bloq = binst.bloq
+        visited_bloqs.add(binst)
```
mpharrigan, line 110: I think the correct fix would be to modify `iter_bloqnections` to yield all the bloqs with empty `incoming` and `outgoing`.

mpharrigan:
> I want to implement a unitary U but, using the existing primitives, I have a way of implementing a unitary Q where Q and U are equal up to global phase.

Ok, I'm open to this. We have other bookkeeping bloqs. 

I suspect this is still going to cause breakages elsewhere, but I guess we'll cross that bridge when we get there.

Does the drawing code work?

qualtran/simulation/quimb_sim.py
tanujkhattar: done

qualtran/simulation/quimb_sim.py
tanujkhattar: removed

qualtran/cirq_interop/_cirq_to_bloq.py
tanujkhattar: Yes, it's always an integer. Changed.

tanujkhattar:
>I suspect this is still going to cause breakages elsewhere, but I guess we'll cross that bridge when we get there.

Yeah, the good thing is that all the tests pass and drawing code works (see below) and the quimb simulation works (as shown in the test). For whatever is remaining, we'll fix it as we discover it.

> Does the drawing code work? -

Yep! 

image

mpharrigan (approved):
Thanks for your patience. LGTM

Issue #400, opened 2023-10-13
Bloq count graphs can be quite wide


fdmalone:
see trotter.ipynb #359  and #392, should they be more vertical?

mpharrigan:
What about making the node labels more configurable? and/or changing the default to be more sensitive to long reprs

mpharrigan:
In terms of making them "more vertical": currently I think it tries to have each level correspond to a topological generation. We use graphviz under-the-hood so be on the lookout for layout options there

mpharrigan:
With the new call graphs with costs, they're at least a bit more vertical; and the `__str__` reprs are much shorter so they won't be as wide. Closing for now unless there are feasible suggestions on how to improve the layout further.

PR #428, opened 2023-10-19
Better node labels for bloq counts graph

mpharrigan:
Towards #400 

 - use more sensible defaults for the "details" in a bloq counts node. Aggressively abbreviate the field list
 - Refactor the methods so you can override them easily.
 - The arithmetic bloqs ... the little math expression should be `short_name` because it's quite ambiguous (but also short). We use `pretty_name` for the bloq titles. @fdmalone 
 - Fixed `pretty_name` for `CirqGateAsBloq` bloqs.


tanujkhattar (approved):

fdmalone (approved):

fdmalone:
@mpharrigan is this good to go?

mpharrigan:
qrom has changed; I need to (re-)fix potentially

mpharrigan:
false alarm; I had a stale jupyter kernel. I removed the QROM repr hack to show that the default is no longer crashing graphviz; but it's still a little wide

PR #437, opened 2023-10-23
Allow a and b bitsize for greater than bloq.

fdmalone:


tanujkhattar (approved):

mpharrigan:
worth having a default where -- if not specified -- b_bitsize defaults to a_bitsize?

PR #438, opened 2023-10-23
First quantized prepare

fdmalone:
Adds the majority of the prepare bloqs for T and UV for the first quantized chemistry Hamiltonians.

Missing is Sec 2.A which prepares the state for selecting between T, U and V which I need to understand a bit better. 

I think I found an error in the reference in the nu state preparation in eq 90. see test_prepare_nu_bloq_counts which compares the bloq counts, my addition of the expected costs and eq 90. Unless I've made a mistake in parsing the text I think it's off by 5 Toffolis. In particular:

the cost for the arithmetic is

$$
3 n_p^2 + n_p + 1 + 4 n_\mathcal{M}(n_p + 1) \hspace{1em} (90) 
$$

We also need to add 3 Toffolis which can be inverted at zero Toffoli cost for flagging success.

The other costs are 4(np-1) controlled hadamards (not inverted at zero cost) and 6np + 2 Toffolis (free inversion).
So focusing on the n_p terms and constants
$n_p + 2 . 4 n_p + 6 n_p = 15 n_p$ (consistent) 
and the constants
$ 4 - 2.4 + 2 = -2$ (not -7).

Several somewhat self contained primitives are good candidates for implementation without much chemistry context.

PR #440, opened 2023-10-23
First quantized select

fdmalone:
Adds select bloqs for first quantized hamiltonian and notebook. Bloq counts (when added up) match those in the paper up to some small issues (see #438). Requires #438 to be merged first.

This is somewhat incomplete as SELECT requires swaps from the system to ancilla before and after the select operation is applied, I'll follow up with musical score type diagrams which may refactor how bloqs are grouped. E.g. the ij preparation should be exterior to T prepare and in the paper the swaps are considered (but should probably be contained in a larger SELECT bloq).

Fixes some errors in latex in #438.
 
Screenshot 2023-10-23 at 11 42 44 AM


fdmalone:
cc @ncrubin

mpharrigan:
does this actually depend on prepare?

fdmalone:
Eh I fixed some latex errors from #438 so I think it'd be better to go in order.

fdmalone:
@mpharrigan is it ok if I come back and add the new autodoc / test things. I can do the existing chemistry algorithms in one go.

mpharrigan:
> is it ok if I come back and add the new autodoc / test things. I can do the existing chemistry algorithms in one go.

yes

qualtran/bloqs/chemistry/pbc/first_quantization/select_t.py
```diff
@@ -0,0 +1,65 @@
...
+            https://arxiv.org/abs/2105.12767) page 20, section B
+    """
+    num_bits_p: int
+    eta: int
+
+    @cached_property
+    def signature(self) -> Signature:
+        return Signature(
+            [
+                Register("sys", bitsize=self.num_bits_p, shape=(self.eta, 3)),
```
mpharrigan, line 51: this can result in a lot of soquets, right? I think the ultimate strategy will be to do something like #300 where we have a high-level interface bloq that has two registers "select" and "system" and then implementation bloqs that has the parts all factored into meaningful components but there's a lot of soquets and you use a `Partition` bloq to go betwixt the two. Out of the scope for this PR but I don't have any actionable comments, so I figured I'd write this comment here.

qualtran/bloqs/chemistry/pbc/first_quantization/select_t.py
```diff
@@ -0,0 +1,65 @@
...
+        )
+
+    def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
+        # Cost is $5(n_{p} - 1) + 2$ which comes from copying each $w$ component of $p$
+        # into an ancilla register ($3(n_{p}-1)$), copying the $r$ and $s$ bit of into an
+        # ancilla ($2(n_{p}-1)$), controlling on both those bit perform phase flip on an
+        # ancilla $|+\rangle$ state. This requires $1$ Toffoli, Then erase which costs
+        # only Cliffords. There is an additional control bit controlling the application
+        # of $T$ thus we come to our total.
+        # Eq 73. page
```
mpharrigan, line 64: Another sortof note-to-self: We should be capturing this knowledge in a way that we can put it in the docs. I'll keep that in mind

mpharrigan (approved):

qualtran/bloqs/chemistry/pbc/first_quantization/select_t.py
fdmalone: right, I've been sort of running into this pain in https://github.com/quantumlib/Qualtran/pull/493/files#diff-c6f700b3dd40e8b12b6388c73fc9700b8331972cd0a508392cda5687a07acef4R338

fdmalone: It would be helpful to consolidate some of these things with the existing SelectOracle and PrepareOracle.

PR #441, opened 2023-10-24
Jupyter autogen v2

mpharrigan:
This uses bloq examples from #397 to form the basis of the example instances we document.

Each bloq class has a number of examples. They're tied together with an instance of new `BloqDocSpec` colocated with the bloq definition and example definition. 

An important motivating change is now the example instances have unique variable names in the rendered notebooks. I.e. everything isn't named `bloq`. It's named according to the `BloqExample.name`. 

The actual sphinx / jupyter / doc functionality is kept separate in dev_tools. To facilitate the transition, I've factored out the changes here into a `v2` module. After everything is ported, I'll remove the existing "v1" code and there will be one version of doc generation. 

Future PRs will add additional "get cells" functions for drawing decompositions, etc.


qualtran/bloqs/basic_gates/swap.py
```diff
@@ -147,14 +159,17 @@ def t_complexity(self) -> 'TComplexity':
...
+
 
 @frozen
 class CSwap(GateWithRegisters):
     """Swap two registers controlled on a control bit.
 
     Implements a multi-target controlled swap unitary $CSWAP_n = |0><0| I + |1><1| SWAP_n$.
 
-    This decomposes into a qubitwise SWAP on the two target registers, and takes 14*n T-gates.
+    This decomposes into a qubitwise SWAP on the two target registers, and takes $14*n$ T-gates.
```
fdmalone, line 172: Maybe remove the * while you're at it.

qualtran/_infra/bloq_example.py
```diff
@@ -69,6 +70,18 @@ def _bloq_cls_from_func_annotation(func: Callable[[], Bloq]) -> Type[Bloq]:
     return cls
 
 
+@typing.overload
```
fdmalone, line 73: Question: What are these new overloads for?

mpharrigan: this is a type annotation thing. It lets you be more precise about the annotations in two different regimes rather than just `Union`ing all the types

mpharrigan: https://docs.python.org/3/library/typing.html#overload

fdmalone:
Is the idea that the notebooks would (eventually) automatically generate cells for show the decomposed bloq / musical score etc. One thing I've noticed is that the current default of show_bloq'ing the bloq can result in sort of unwieldy diagrams if the bloq has many registers. Would we be able to pick and choose between what is visualized in the notebook?

fdmalone (approved):
Looks like a positive development.

mpharrigan:
> Is the idea that the notebooks would (eventually) automatically generate cells for show the decomposed bloq / musical score etc. One thing I've noticed is that the current default of show_bloq'ing the bloq can result in sort of unwieldy diagrams if the bloq has many registers. Would we be able to pick and choose between what is visualized in the notebook?

Yes, the idea is to add more demo cells (i.e. show decomp) and add more options (i.e. whether or not to include each of the type of cells; e.g. disable the "graphical signature")

PR #454, opened 2023-10-26
Decompose first quantization prepare bloqs

fdmalone:
Requires #438 and #440 first. I just added some high level decompositions to enable musical score diagrams, e.g. 
Screenshot 2023-10-26 at 8 53 28 AM


mpharrigan:
cool diagam.

how are lines originating from THRU (i.e. rounded-rectangle) registers like flag_mu? Does that pass `assert_valid_cbloq`?

fdmalone:
oh lol, I didn't like the arrows so locally modified the drawing code to make some slides. They are RIGHT and LEFT registers in practice (I didn't commit the modifications of the drawing).

mpharrigan:
Ah excellent

fdmalone:
Arrows are fine (I guess some configuration might be ideal), but I think the size of the arrow was a little off-putting.

mpharrigan (approved):

PR #455, opened 2023-10-26
Migrate THC files to cleaner path

### LINKED ISSUE #429: Add THC SELECT to THC notebook.

fdmalone:
Also fixes #429


fdmalone:
This just does the renaming, is the one bloq per file a requirement?

mpharrigan (approved):

mpharrigan:
> s the one bloq per file a requirement?

no; but let's use it as a guideline and make a note of places where it feels ridiculous so we can evaluate

PR #457, opened 2023-10-26
Migrate trotter files

fdmalone:
Move trotter bloqs to more sensible layout. I think these should probably be reorganized again as the inverse square root is independent of Trotter, it should probably be called grid_ham/xyz.py.


fdmalone:
Depends on #455

mpharrigan (approved):

PR #458, opened 2023-10-27
Fix some pytest warnings

mpharrigan:


qualtran/cirq_interop/cirq_interop.ipynb
```diff
@@ -466,6 +466,8 @@
   {
    "cell_type": "code",
```
fdmalone, line 467: What's going on with this file being added?

mpharrigan: what? this adds the "id" field to this cell which the jupyter notebook tests were warning about it being missing. I also cleaned up the metadata field

fdmalone: I should have said modified. It wasn't clear what the jupyter warning was just lookin at the wildcard expression

fdmalone: wait, the ignore for deprecation warnings doesn't silence the missing id warning right?

mpharrigan: the ignores are pretty specific. There's a place where quimb uses a deprecated scipy function which I've scoped to that particular quimb module; and I get a warning that something about the configuration platformdirs are changing in a seemingly inconsequential way from jupyter. The ignore spec includes a regex against that particular message. The id thing wasn't a deprecation warning either

fdmalone (approved):

PR #459, opened 2023-10-27
[cirq-ft vestige] can't assert two things are consistent if you don't have two things

mpharrigan:


NoureldinYosri:
The reason it was setup in this way is to be able to add the test even when we don't have a decomposition yet. The idea was that we could have a list of consistancy checks like cirq which run on every gate/block and any of these checks should pass if it's no-op. so tests that need a decomposition should pass in case no decomposition exits 

it's a different way of thinking and it follows test driven development principles.

mpharrigan:
that's not how it's being used (?) and I found bugs by changing the behavior

NoureldinYosri (approved):

Issue #370, opened 2023-09-26
Alternative rotation compilations/costs


fdmalone:
Currently our T costs derive from  https://arxiv.org/pdf/1404.5320.pdf, which gives a mean cost of 1.149 log2
(1/ε) + 9.2 for a target precision of ε.

Many papers use a phase gradient gate instead for rotations (appendix A of https://doi.org/10.1103/PRXQuantum.1.020312) which requires b_r - 2 Toffolis for b_r bits of precision (it uses the first method to prepare a phase gradient register first and then adds into this). This is often significantly cheaper.

Need to think about rotations.

mpharrigan:
cc @NoureldinYosri who has been thinking about this

NoureldinYosri:
we currenly have a unified approach to the vanilla rotation models [RotationLogarithmicModel](https://github.com/quantumlib/Qualtran/blob/57de77ef6fd2172330e99e121f603045197fd884/qualtran/surface_code/rotation_cost_model.py#L37) and for constant cost models [ConstantWithOverheadRotationCost](https://github.com/quantumlib/Qualtran/blob/57de77ef6fd2172330e99e121f603045197fd884/qualtran/surface_code/rotation_cost_model.py#L69C7-L69C39) both of which implement the same interface of [RotationCostModel](https://github.com/quantumlib/Qualtran/blob/57de77ef6fd2172330e99e121f603045197fd884/qualtran/surface_code/rotation_cost_model.py#L24)

Do we need to do more for this?

mpharrigan:
Related: #1250

mpharrigan:
Going to close this in favor of #1250 or open new issues to track specific asks

Issue #269, opened 2023-06-23
Physical costing cleanup


mpharrigan:
There's code in the physical costing module that hasn't been looked at in a while. 

- [x] take a look at it
- [ ] commit the azure cost model

PR #461, opened 2023-10-27
Add rotation cost model

NoureldinYosri:
This abstraction covers both standarad models which have a cost of $A \log_2(1/\epsilon) + B$ Where $A$ 
and $B$ depend on the gateset, approximation protocol and hardware parameters and are obtained by fitting regression models. It also covers models with constant per rotation of cost which require a circuit that gets prepared once which gives preparation overhead.

I'm reporting the cost as an `AlgorithmSummary` object rather than a `TComplexity` object because it's more natural to report the cost of the second model in terms of Toffoli count rather than T count.

Related to https://github.com/quantumlib/Qualtran/issues/370 and is part of https://github.com/quantumlib/Qualtran/issues/269

### DIFF
--- a/qualtran/surface_code/algorithm_summary.py
+++ b/qualtran/surface_code/algorithm_summary.py
@@ -22,7 +22,7 @@ _PRETTY_FLOAT = field(default=0.0, converter=float, repr=lambda x: f'{x:g}')
 class AlgorithmSummary:
     """Properties of a quantum algorithm that impact its physical cost
 
-    Counts of different properities that affect the physical cost of
+    Counts of different properties that affect the physical cost of
     running an algorithm (e.g. number of T gates).
     All counts default to zero.
 
@@ -41,3 +41,21 @@ class AlgorithmSummary:
     toffoli_gates: float = _PRETTY_FLOAT
     rotation_gates: float = _PRETTY_FLOAT
     rotation_circuit_depth: float = _PRETTY_FLOAT
+
+    def __mul__(self, other: int) -> 'AlgorithmSummary':
+        if not isinstance(other, int):
+            raise TypeError(
+                f'Multiplication isn\'t supported between AlgorithmSummary and not integer type {type(other)}'
+            )
+
+        return AlgorithmSummary(
+            algorithm_qubits=self.algorithm_qubits * other,
+            measurements=self.measurements * other,
+            t_gates=self.t_gates * other,
+            toffoli_gates=self.toffoli_gates * other,
+            rotation_gates=self.rotation_gates * other,
+            rotation_circuit_depth=self.rotation_circuit_depth * other,
+        )
+
+    def __rmul__(self, other: int) -> 'AlgorithmSummary':
+        return self.__mul__(other)
--- a/qualtran/surface_code/algorithm_summary.py
+++ b/qualtran/surface_code/algorithm_summary.py
@@ -45,7 +45,7 @@ class AlgorithmSummary:
     def __mul__(self, other: int) -> 'AlgorithmSummary':
         if not isinstance(other, int):
             raise TypeError(
-                f'Multiplication isn\'t supported between AlgorithmSummary and non integer type {type(other)}'
+                f"Multiplication isn't supported between AlgorithmSummary and non integer type {type(other)}"
             )
 
         return AlgorithmSummary(
@@ -59,3 +59,31 @@ class AlgorithmSummary:
 
     def __rmul__(self, other: int) -> 'AlgorithmSummary':
         return self.__mul__(other)
+
+    def __add__(self, other: 'AlgorithmSummary') -> 'AlgorithmSummary':
+        if not isinstance(other, AlgorithmSummary):
+            raise TypeError(
+                f"Addition isn't supported between AlgorithmSummary and type {type(other)}"
+            )
+        return AlgorithmSummary(
+            algorithm_qubits=self.algorithm_qubits + other.algorithm_qubits,
+            measurements=self.measurements + other.measurements,
+            t_gates=self.t_gates + other.t_gates,
+            toffoli_gates=self.toffoli_gates + other.toffoli_gates,
+            rotation_gates=self.rotation_gates + other.rotation_gates,
+            rotation_circuit_depth=self.rotation_circuit_depth + other.rotation_circuit_depth,
+        )
+
+    def __sub__(self, other: 'AlgorithmSummary') -> 'AlgorithmSummary':
+        if not isinstance(other, AlgorithmSummary):
+            raise TypeError(
+                f"Subtraction isn't supported between AlgorithmSummary and type {type(other)}"
+            )
+        return AlgorithmSummary(
+            algorithm_qubits=self.algorithm_qubits - other.algorithm_qubits,
+            measurements=self.measurements - other.measurements,
+            t_gates=self.t_gates - other.t_gates,
+            toffoli_gates=self.toffoli_gates - other.toffoli_gates,
+            rotation_gates=self.rotation_gates - other.rotation_gates,
+            rotation_circuit_depth=self.rotation_circuit_depth - other.rotation_circuit_depth,
+        )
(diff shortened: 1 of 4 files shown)


qualtran/surface_code/algorithm_summary.py
```diff
@@ -41,3 +41,21 @@ class AlgorithmSummary:
     toffoli_gates: float = _PRETTY_FLOAT
     rotation_gates: float = _PRETTY_FLOAT
     rotation_circuit_depth: float = _PRETTY_FLOAT
+
+    def __mul__(self, other: int) -> 'AlgorithmSummary':
+        if not isinstance(other, int):
+            raise TypeError(
+                f'Multiplication isn\'t supported between AlgorithmSummary and non integer type {type(other)}'
```
mpharrigan, line 48: extremely minor nit: Usually we'd just switch to `"`-strings instead of escaping the apostrophe

qualtran/surface_code/algorithm_summary.py
```diff
@@ -41,3 +41,21 @@ class AlgorithmSummary:
     toffoli_gates: float = _PRETTY_FLOAT
     rotation_gates: float = _PRETTY_FLOAT
     rotation_circuit_depth: float = _PRETTY_FLOAT
+
+    def __mul__(self, other: int) -> 'AlgorithmSummary':
```
mpharrigan, line 45: is it weird that this supports multiplication but not other arithmetic?

qualtran/surface_code/rotation_cost_model.py
```diff
@@ -0,0 +1,109 @@
...
+
+    $$
+    \#T = -\textit{slope} \log_2{\textit{budget}} + \textit{overhead}
+    $$
+
+    Attributes:
+        slope: The coefficient of $log_2{budget}$.
+        overhead: The overhead.
+        gateset: A human-readable description of the gate set (e.g. 'Clifford+T').
+        approximation_protocol: A description or reference to the approximation protocol (e.g. `Diagonal:https://arxiv.org/abs/1403.2975`)
```
mpharrigan, line 47: can you wrap to 100 lines (and there's a missing space after the colon)

qualtran/surface_code/rotation_cost_model.py
```diff
@@ -0,0 +1,109 @@
...
+    def rotation_cost(self, error_budget: float) -> AlgorithmSummary:
+        """Cost of a single rotation."""
+
+    @abc.abstractmethod
+    def prepartion_overhead(self, error_budget) -> AlgorithmSummary:
+        """Cost of preparation circuit."""
+
+
+@frozen
+class LogarithmicModel(RotationCostModel):
```
mpharrigan, line 36: This name makes sense here, where you can clearly see that it's an instance of `RotationCostModel` but I suspect if it shows up outside of the class definition it will lose this critical context. Should probably have rotation in the name somewhere

qualtran/surface_code/rotation_cost_model.py
```diff
@@ -0,0 +1,109 @@
...
+    r"""A linear model in the log of the error budget with no preparation cost.
+
+    $$
+    \#T = -\textit{slope} \log_2{\textit{budget}} + \textit{overhead}
```
mpharrigan, line 40: why `textit`? Usually we use `textrm` to have text that doesn't look like a bunch of variables multiplied together

qualtran/surface_code/rotation_cost_model.py
```diff
@@ -0,0 +1,109 @@
...
+        return AlgorithmSummary(
+            t_gates=math.ceil(-self.slope * math.log2(error_budget) + self.overhead)
+        )
+
+    def prepartion_overhead(self, error_budget) -> AlgorithmSummary:
+        return AlgorithmSummary()
+
+
+@frozen
+class ConstantCostWithPreparation(RotationCostModel):
```
mpharrigan, line 66: again, in actual use the fact that this is for rotations is lost; but we don't want it to be _too_ long.

`ConstOverheadRotationCost` or something like that

qualtran/surface_code/rotation_cost_model.py
```diff
@@ -0,0 +1,109 @@
...
+    r"""A rotation cost of bitsize-2 toffoli per rotation independent of the error budget.
```
mpharrigan, line 67: `bitsize - 2`; to make it look like subtraction instead of hyphenation

qualtran/surface_code/rotation_cost_model.py
```diff
@@ -0,0 +1,109 @@
...
+
+    This model assumes a state $\ket{\phi}$ has been prepared using a standard technique, then
+    each rotation is applied with bitsize digits of accuracy using bitsize-2 Toffoli gates.
```
mpharrigan, line 70: bitsize - 2

qualtran/surface_code/rotation_cost_model.py
```diff
@@ -0,0 +1,109 @@
...
+    $$
+    \ket{\phi} = \frac{1}{\sqrt{2^{b}}} \sum_{k=0}^{2^b-1} e^{-2\pi i k/2^b} \ket{k}
+    $$
+    Where $b$ is the bitsize/number of digits of accuracy.
+
+    reference: https://doi.org/10.1103/PRXQuantum.1.020312
+
+    Attributes:
+        bitsize: Number of digits of accuracy for approximating a rotation.
+        slope: The coefficient of $log_2{budget}$.
```
mpharrigan, line 80: why is there a slope? isn't this supposed to be constant?

qualtran/surface_code/rotation_cost_model.py
```diff
@@ -0,0 +1,109 @@
...
+    def rotation_cost(self, error_budget: float) -> AlgorithmSummary:
+        return AlgorithmSummary(toffoli_gates=max(self.bitsize - 2, 0))
+
+    def prepartion_overhead(self, error_budget) -> AlgorithmSummary:
+        return self.bitsize * AlgorithmSummary(
+            t_gates=math.ceil(-self.slope * math.log2(error_budget / self.bitsize) + self.overhead)
+        )
+
+
+BeverlandEtAl = LogarithmicModel(
```
mpharrigan, line 103: Outside of this definition, this variable will appear without the context that it's a rotation cost model

qualtran/surface_code/algorithm_summary.py
NoureldinYosri: I implemented addition and subtraction as well. I don't think division make sense here

qualtran/surface_code/rotation_cost_model.py
NoureldinYosri: ptal

qualtran/surface_code/rotation_cost_model.py
NoureldinYosri: done

qualtran/surface_code/rotation_cost_model.py
NoureldinYosri: ptal

qualtran/surface_code/rotation_cost_model.py
NoureldinYosri: for computing the cost of the overhead rotation

NoureldinYosri: I changed the signature to make this clearer

qualtran/surface_code/rotation_cost_model.py
NoureldinYosri: updated

mpharrigan (approved):
👍

mpharrigan:
> I'm reporting the cost as an AlgorithmSummary object rather than a TComplexity object because it's more natural to report the cost of the second model in terms of Toffoli count rather than T count.

Can you speak more about this? `TComplexity` lets you report toffoli complexity, no?

NoureldinYosri:
@mpharrigan to use TComplexity I will have to convert the toffolis to Ts `TComplexity(t=4*num_toffoli)`. This is why we have a lot 4*toffoli statements in the codebase which you mentioned you didn't like.

The rotations are implemented in Toffolis so I think the runtime estimate will defer between an architectures that synthesize magic T state vs magic CCZ states.

mpharrigan:
oh for some reason I thought we had a toffoli/ccz field in the `TComplexity` dataclass. Disregard!

PR #462, opened 2023-10-27
Bloq report card

mpharrigan:
Some dev-tools scripts for 

 - running the `qualtran.testing.check_xxx` checks against all `BloqExample`s
 - showing the results in a "report card"

image


fdmalone (approved):
More tools are good.

Issue #114, opened 2022-11-22
Keeping an eye on test durations


mpharrigan:
on my cloudtop:

 - `test_fci_util` takes 1:37
 - `test_select_swap_qrom` takes 0:49

for a total execution time of 4 minutes. We should trim down these 1% of tests taking 50% of runtime @tanujkhattar @ncrubin

mpharrigan:
The tests are slow again @tanujkhattar 

```50.08s call     qualtran/bloqs/qrom_test.py::test_qrom_multi_dim[2-data0]
31.11s call     qualtran/bloqs/select_swap_qrom_test.py::test_select_swap_qrom[3-data1]
22.92s call     qualtran/bloqs/qrom_test.py::test_qrom_1d[2-data2]
14.55s call     qualtran/bloqs/hubbard_model_test.py::test_hubbard_model_notebook
13.98s call     qualtran/bloqs/qrom_test.py::test_qrom_multi_dim[1-data0]
12.75s call     qualtran/bloqs/arithmetic_test.py::test_comparison_gates_notebook
12.15s call     qualtran/cirq_interop/_bloq_to_cirq_test.py::test_notebook
12.11s call     qualtran/bloqs/unary_iteration_bloq_test.py::test_multi_dimensional_unary_iteration_gate[target_shape0]
9.03s call     qualtran/bloqs/unary_iteration_bloq_test.py::test_notebook
8.79s call     qualtran/bloqs/chemistry/trotter/trotter_test.py::test_notebook
8.35s call     qualtran/bloqs/apply_gate_to_lth_target_test.py::test_notebook
8.33s call     qualtran/bloqs/hubbard_model_test.py::test_phase_estimation_of_qubitized_hubbard_model
8.19s call     qualtran/_infra/bloq_test.py::test_notebook
8.08s call     qualtran/bloqs/select_pauli_lcu_test.py::test_notebook
7.98s call     qualtran/bloqs/state_preparation_test.py::test_notebook
7.81s call     qualtran/bloqs/qrom_test.py::test_notebook
7.30s call     qualtran/bloqs/qubitization_walk_operator_test.py::test_notebook
7.30s call     qualtran/bloqs/basic_gates_test.py::test_notebook
7.28s call     qualtran/bloqs/chemistry/thc/prepare_test.py::test_notebook
7.23s call     qualtran/_infra/composite_bloq_test.py::test_notebook
[...]

PR #463, opened 2023-10-27
Skip slow test configurations for slow tests

mpharrigan:
xref #114 

This is about 47 seconds worth of tests


tanujkhattar:
To be sure, these tests can still be run if we provide a flag like `--enable-slow` ? 

There's a parallel effort happening in Cirq to solve this problem where we can mark tests as "slow" and they'll be run only as part of daily / weekly CI instead of running them on every PR. 

See https://github.com/quantumlib/Cirq/issues/6211 and https://github.com/quantumlib/Cirq/pull/6328

mpharrigan:
This PR doesn't respond to flags; it just skips one of the configurations. You can run it by commenting out the skip which honestly is as good as a command line flag because I would bet most of the slow tests in projects with an `--enable-slow` flag _rarely_ get executed

tanujkhattar:
> I would bet most of the slow tests in projects with an --enable-slow flag rarely get executed

The way it's handled in Cirq is that the slow tests get executed once as part of the daily CI while the rest of the tests get executed as part of every PR. I think this model is more robust and should be adopted. The reason slow tests exist is because we have found implementation bugs in the past which were not caught by small test cases, and having the daily CI running these slow tests will help us catch such bugs in case we change the implementation. The "run it by commenting out the skip" is not nearly as robust. Also, the infrastructure to add the daily CI and mark tests as slow is not terribly complicated and already exists in Cirq now; thanks to @pavoljuhas; so we can leverage the knowledge we've gained by implementing it in Cirq to simply adopt the same infrastructure here as well.

mpharrigan:
Ok, let's merge this until the nightly CI is set up

tanujkhattar (approved):

pavoljuhas:
@mpharrigan - I think it would be better to apply the skip mark at the test function definition rather than in its body.
If it is OK, I can add a commit to this PR to do so.

mpharrigan:
These are parameterized tests using the "outer product" over the two parameter sets. Most of the parameter settings are not slow; just the cases identified by the `if ... skip` are slow.

mpharrigan:
Got it, cool!

pavoljuhas:
> Got it, cool!

Done in 66af330e5a91e517bbf7ea920027ebc74c3e0baf...d783c5186d0a90d70eb195918571e0064efb56f9

qualtran/bloqs/qrom_test.py
```diff
@@ -27,10 +27,23 @@
 
 
 @pytest.mark.parametrize(
-    "data", [[[1, 2, 3, 4, 5]], [[1, 2, 3], [4, 5, 10]], [[1], [2], [3], [4], [5], [6]]]
+    "data,num_controls",
+    [
+        pytest.param(
```
mpharrigan, line 32: this is a really nice thing to know about, thanks @pavoljuhas !

qualtran/bloqs/qrom_test.py
```diff
@@ -27,10 +27,23 @@
...
+        for idx, data in enumerate(
+            [[[1, 2, 3, 4, 5]], [[1, 2, 3], [4, 5, 10]], [[1], [2], [3], [4], [5], [6]]]
+        )
+        for num_controls in [0, 1, 2]
+    ],
 )
-@pytest.mark.parametrize("num_controls", [0, 1, 2])
 def test_qrom_1d(data, num_controls):
+    if num_controls == 2 and len(data) == 6:
+        pytest.skip("slow")
```
mpharrigan, line 46: I think this should be removed now, right?

mpharrigan:
The problem is we don't actually use or respect the slow marker

qualtran/bloqs/qrom_test.py
mpharrigan: oh I see two of them you skip in the decorator and this one you mark as slow and skip in the body. Is this to show how to do it when we set up a "slow" marker?

pavoljuhas:
> The problem is we don't actually use or respect the slow marker

Yes, thank you for catching that.  I overlooked the mark after a copy-paste from Cirq.

We can setup a `slow` marker with its own test skipping logic like in Cirq, but that is a thing for a next PR.

mpharrigan:
I opened https://github.com/quantumlib/Qualtran/issues/491 to track

PR #464, opened 2023-10-27
[autodoc/autotest] PrepareUniformSuperposition

mpharrigan:
Add `PrepareUniformSuperposition` examples

### DIFF
--- a/dev_tools/autogenerate-bloqs-notebooks-v2.py
+++ b/dev_tools/autogenerate-bloqs-notebooks-v2.py
@@ -49,6 +49,7 @@ from qualtran_dev_tools.jupyter_autogen_v2 import NotebookSpecV2, render_noteboo
 
 import qualtran.bloqs.basic_gates.swap
 import qualtran.bloqs.factoring.mod_exp
+import qualtran.bloqs.prepare_uniform_superposition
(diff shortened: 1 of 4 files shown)


dev_tools/autogenerate-bloqs-notebooks-v2.py
```diff
@@ -69,17 +70,21 @@
...
         title='Modular Exponentiation',
         module=qualtran.bloqs.factoring.mod_exp,
-        path_stem='mod_exp',
         bloq_specs=[qualtran.bloqs.factoring.mod_exp._MODEXP_DOC],
         directory=f'{SOURCE_DIR}/bloqs/factoring',
     ),
     NotebookSpecV2(
         title='Modular Multiplication',
         module=qualtran.bloqs.factoring.mod_mul,
-        path_stem='mod_mul',
```
fdmalone, line 79: ?

fdmalone (approved):

dev_tools/autogenerate-bloqs-notebooks-v2.py
mpharrigan: The default is to match the module name. I must have had a revision where I was moving things around or doing something non-standard; but it's back to standard.

PR #469, opened 2023-10-31
Add `XPow/YPow/ZPow` rotations gate

tanujkhattar:
This PR does a bunch of changes/improvements in cirq interop
- Introduces a new `CirqGateAsBloqBase` abstract class which has all the functionality to auto populate the bloq methods using underlying Cirq gates. There is one abstract method called `cirq_gate` which can be over-riden by derived classes. 
- The wrapper class `CirqGateAsBloq` simply derives from `CirqGateAsBloqBase` and uses the user specified `self.gate` as the corresponding `cirq_gate`.
- However, users can now also derive from `CirqGateAsBloqBase` class and specify a custom cirq gate as the `cirq_gate` property. For commonly used gates in Cirq-Core, this inheritance based method for wrapping cirq gates solves the discoverability issue and reduces boilerplace code. 
- This is now used in the `rotations.py` file to wrap `Rx / Ry / Rz / XPow / YPow / ZPow` gates.  As you can see, the wrapper code is now much simpler. 
- Another drive-thru change is to change the default signature of wrapped gates from `Register(qubits, shape=(nqubits), bitsize=1)` to `Register('q', shape=(nqubits), bitsize=1)` if `nqubits > 1` or `Register('q', shape=(), bitsize=1)` if nqbits = 1. Changing `qubits` to `q` is nicer because we don't need to worry about singular / plural when number of qubits are > 1 or = 1. 

The `XPow/YPow/ZPow` gates have an option to configure the global phase and follow a pattern similar to Cirq. These gates are useful when implementing phase gradient states.  Docstrings for these gates are copied over from Cirq. 

As a follow up, we can also consider making these new gates implementation more robust so we can get rid of `XGate` / `YGate` / `ZGate` that also currently exist.

### DIFF
--- a/qualtran/bloqs/basic_gates/rotation.py
+++ b/qualtran/bloqs/basic_gates/rotation.py
@@ -14,12 +14,13 @@
 
 import abc
 from functools import cached_property
-from typing import Dict, Set, Tuple, TYPE_CHECKING
+from typing import Dict, Protocol, Set, Tuple, TYPE_CHECKING
 
+import cirq
 import numpy as np
 from attrs import frozen
 
-from qualtran import Bloq, Signature
+from qualtran import GateWithRegisters, Signature
 from qualtran.bloqs.basic_gates.t_gate import TGate
 from qualtran.cirq_interop.t_complexity_protocol import TComplexity
 
@@ -30,16 +31,18 @@ if TYPE_CHECKING:
     from qualtran.resource_counting import BloqCountT, SympySymbolAllocator
 
 
-@frozen
-class RotationBloq(Bloq, metaclass=abc.ABCMeta):
-    angle: float
-    eps: float = 1e-11
+class _HasEps(Protocol):
+    """Protocol for typing `RotationBloq` base class mixin that has accuracy specified as eps."""
 
+    eps: float
+
+
+class RotationBloq(GateWithRegisters, metaclass=abc.ABCMeta):
     @cached_property
     def signature(self) -> 'Signature':
         return Signature.build(q=1)
 
-    def t_complexity(self):
+    def t_complexity(self: _HasEps):
         # TODO Determine precise clifford count and/or ignore.
         # This is an improvement over Ref. 2 from the docstring which provides
         # a bound of 3 log(1/eps).
@@ -48,17 +51,40 @@ class RotationBloq(Bloq, metaclass=abc.ABCMeta):
         num_t = int(np.ceil(1.149 * np.log2(1.0 / self.eps) + 9.2))
         return TComplexity(t=num_t)
 
-    def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
-        num_t = int(np.ceil(1.149 * np.log2(1.0 / self.eps) + 9.2))
+    def build_call_graph(self: _HasEps, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
+        num_t = RotationBloq.t_complexity(self).t
         return {(TGate(), num_t)}
 
 
 @frozen
-class Rz(RotationBloq):
-    """Single-qubit Rz gate.
+class ZPowGate(RotationBloq):
+    r"""A gate that rotates around the Z axis of the Bloch sphere.
+
+    The unitary matrix of `ZPowGate(exponent=t, global_shift=s)` is:
+    $$
+        e^{i \pi s t}
+        \begin{bmatrix}
+            1 & 0 \\
+            0 & e^{i \pi t}
+        \end{bmatrix}
+    $$
+
+    Note in particular that this gate has a global phase factor of
+    $e^{i\pi t/2}$ vs the traditionally defined rotation matrices
+    about the Pauli Z axis. See `Rz` for rotations without the global
+    phase. The global phase factor can be adjusted by using the `global_shift`
+    parameter when initializing.
 
     Args:
-        angle: Rotation angle.
+        exponent: The t in gate**t. Determines how much the eigenvalues of
+            the gate are phased by. For example, eigenvectors phased by -1
+            when `gate**1` is applied will gain a relative phase of
+            e^{i pi exponent} when `gate**exponent` is applied (relative to
+            eigenvectors unaffected by `gate**1`).
+        global_shift: Offsets the eigenvalues of the gate at exponent=1.
+            In effect, this controls a global phase factor on the gate's
+            unitary matrix. The factor for global_shift=s is:
+                exp(i * pi * s * t)
         eps: precision for implementation of rotation.
 
     Registers:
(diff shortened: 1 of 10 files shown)


qualtran/bloqs/basic_gates/rotation.py
```diff
@@ -48,17 +51,179 @@ def t_complexity(self):
...
+    Args:
+        exponent: The t in gate**t. Determines how much the eigenvalues of
+            the gate are phased by. For example, eigenvectors phased by -1
+            when `gate**1` is applied will gain a relative phase of
+            e^{i pi exponent} when `gate**exponent` is applied (relative to
+            eigenvectors unaffected by `gate**1`).
+        global_shift: Offsets the eigenvalues of the gate at exponent=1.
+            In effect, this controls a global phase factor on the gate's
+            unitary matrix. The factor for global_shift=s is:
+
```
fdmalone, line 141: nit: extra blank line?

qualtran/bloqs/basic_gates/rotation_test.py
```diff
@@ -61,3 +61,24 @@ def test_as_cirq_op():
     op, _ = bloq.as_cirq_op(SimpleQubitManager(), **quregs)
     circuit = cirq.Circuit(op)
     assert circuit == cirq.Circuit(cirq.Rz(rads=bloq.angle).on(cirq.NamedQubit("q")))
+    bloq = XPowGate(exponent=1 / 5, global_shift=-0.5)
+    quregs = bloq.signature.get_cirq_quregs()
+    op, _ = bloq.as_cirq_op(SimpleQubitManager(), **quregs)
+    circuit = cirq.Circuit(op)
+    assert circuit == cirq.Circuit(
+        cirq.XPowGate(exponent=1 / 5, global_shift=-0.5).on(cirq.NamedQubit("q"))
+    )
```
fdmalone, line 70: Is it worth testing the tensor contraction route to a unitary, or is this covered enough already?

qualtran/bloqs/basic_gates/rotation.py
```diff
@@ -48,17 +51,179 @@ def t_complexity(self):
...
+    phase. The global phase factor can be adjusted by using the `global_shift`
+    parameter when initializing.
+
+    Args:
+        exponent: The t in gate**t. Determines how much the eigenvalues of
+            the gate are phased by. For example, eigenvectors phased by -1
+            when `gate**1` is applied will gain a relative phase of
+            e^{i pi exponent} when `gate**exponent` is applied (relative to
+            eigenvectors unaffected by `gate**1`).
+
```
fdmalone, line 191: Nit: extra line.

qualtran/bloqs/basic_gates/rotation.py
```diff
@@ -48,17 +51,179 @@ def t_complexity(self):
...
+        global_shift: Offsets the eigenvalues of the gate at exponent=1.
+            In effect, this controls a global phase factor on the gate's
+            unitary matrix. The factor for global_shift=s is:
+
```
fdmalone, line 195: nit: extra line.

fdmalone:
This LGTM % nits. The more standard rotation bloqs are preserved.

qualtran/bloqs/basic_gates/rotation.py
```diff
@@ -30,16 +31,18 @@
...
-class RotationBloq(Bloq, metaclass=abc.ABCMeta):
-    angle: float
-    eps: float = 1e-11
+class _HasEps(Protocol):
+    """Protocol for typing `RotationBloq` base class mixin that has accuracy specified as eps."""
 
+    eps: float
+
+
+class RotationBloq(GateWithRegisters, metaclass=abc.ABCMeta):
```
mpharrigan, line 40: should this be private, i.e. `_RotationBloq`? Or should this be non-abstract to count arbitrary rotations?

tanujkhattar: I think it should be private for now; I'll make it private

mpharrigan:
should we remove/replace the existing Rx,Ry,Rz? I'm not a huge fan of the "pow" nomenclature since I don't think this has really caught on outside of Cirq; but this matches cirq so I can see why we'd want to keep the name

fdmalone:
Why would we remove them? It's convenient having the textbook rotation gates around no?

tanujkhattar:
> Why would we remove them? It's convenient having the textbook rotation gates around no?

Yeah, I don't think we should remove them. We can probably make them derive from `XPow` / `YPow` etc. like Cirq does but then it becomes harder to continue to keep them frozen dataclasses.

qualtran/bloqs/basic_gates/rotation.py
tanujkhattar: done

qualtran/bloqs/basic_gates/rotation.py
tanujkhattar: done

qualtran/bloqs/basic_gates/rotation.py
tanujkhattar: done

qualtran/bloqs/basic_gates/rotation_test.py
tanujkhattar: I've refactored the code so that these gates now derive from `CirqGateAsBloqBase`  abstract class. This has the added benefit that all the convenience methods (like `add_my_tensors` etc.) get auto populated and it gives a nicer way to wrap Cirq-Core gates that we need frequently in Qualtran as Bloqs.

tanujkhattar:
@mpharrigan @fdmalone This is ready for a re-review

qualtran/cirq_interop/_cirq_to_bloq.py
```diff
@@ -53,34 +54,36 @@ def _get_cirq_quregs(signature: Signature, qm: InteropQubitManager):
...
     def signature(self) -> 'Signature':
-        if isinstance(self.gate, Bloq):
-            return self.gate.signature
-        nqubits = cirq.num_qubits(self.gate)
+        if isinstance(self.cirq_gate, Bloq):
+            return self.cirq_gate.signature
+        nqubits = cirq.num_qubits(self.cirq_gate)
         return (
-            Signature([Register('qubits', shape=nqubits, bitsize=1)])
-            if nqubits > 0
```
fdmalone, line 73: Is this conditional (> 0) no longer necessary?

tanujkhattar: Correct, because `Signature.build(x=0)` now returns `Signature([])`

fdmalone (approved):
I think this LGTM. It took me a second to get the inheritance straight in my head but I think this is nice. Presumably we may want to replace various other single-qubit bloqs this way?

tanujkhattar:
> Presumably we may want to replace various other single-qubit bloqs this way?

That's right, and this approach can be used for multi qubit cirq-core gates as well (like XX, YY, ZZ, etc.); the exact same code will work.

qualtran/cirq_interop/_cirq_to_bloq.py
```diff
@@ -53,34 +54,36 @@ def _get_cirq_quregs(signature: Signature, qm: InteropQubitManager):
     return ret
 
 
-@frozen
-class CirqGateAsBloq(GateWithRegisters):
+class CirqGateAsBloqBase(GateWithRegisters):
```
mpharrigan, line 57: don't you need to mark this as abstract? even if the metaclass is inherited, it would probably help for documenting

qualtran/cirq_interop/_cirq_to_bloq.py
```diff
@@ -53,34 +54,36 @@ def _get_cirq_quregs(signature: Signature, qm: InteropQubitManager):
...
     """A Bloq wrapper around a `cirq.Gate`"""
 
-    gate: cirq.Gate
+    @property
+    @abc.abstractmethod
+    def cirq_gate(self) -> cirq.Gate:
```
mpharrigan, line 62: update class docstring .. this is no longer a wrapper, right? it's something else

qualtran/cirq_interop/_cirq_to_bloq.py
```diff
@@ -106,15 +109,24 @@ def add_my_tensors(
...
-        return self.gate.on(*qubits), in_quregs
+        if isinstance(self.cirq_gate, GateWithRegisters):
+            return self.cirq_gate.as_cirq_op(qubit_manager, **in_quregs)
+        qubits = in_quregs.get('q', ()).flatten()
+        return self.cirq_gate.on(*qubits), in_quregs
+
+
+@frozen
+class CirqGateAsBloq(CirqGateAsBloqBase):
+    gate: cirq.Gate
```
mpharrigan, line 125: docstring!

mpharrigan:
post-hoc review; would appreciate if you could open a small follow-up pr

mpharrigan:
@tanujkhattar ^

Issue #476, opened 2023-11-01
Uncontrolled unary iteration sometimes gives incorrect toffoli cost


fdmalone:
``` python
def _z_to_odd(n: int):
    if n % 2 == 1:
        return cirq.Z
    return cirq.I

iteration_length = 4
for num_controls in [0, 1, 2]:
    apply_z_to_odd = ApplyGateToLthQubit(
        SelectionRegister('selection', 3, iteration_length),
        nth_gate=_z_to_odd,
        control_regs=Signature.build(control=num_controls),
    )
    print(apply_z_to_odd.t_complexity().t, 4 * (iteration_length - 2 + num_controls))
```

prints

```
12 8
12 12
16 16
```

i.e. uncontrolled gives an incorrect T complexity of 12 rather than 8. The solution is to not use a bitsize (3) that is greater than is needed for iteration_length (bitsize=2 works just fine). Opening this to track in case we run into it again.

fdmalone:
cc @tanujkhattar

PR #477, opened 2023-11-01
Fix T-complexity for UnaryIteration with 0 controls and larger selection bitsize

tanujkhattar:
Fixes https://github.com/quantumlib/Qualtran/issues/476


NoureldinYosri:
nit: I think this PR should be titled`Optimize t-complexity of unary iteration contruction for the case with zero conrols`

fdmalone (approved):
LGTM

PR #478, opened 2023-11-01
Add Multiplexed Swaps

fdmalone:
Adds "unary iterated cswaps" (not sure what a good name is) required for first quantized select. In the first quantized algorithm we need to swap system registers (there are $\eta$, registers each of size $\(3, n_p\)$) into an ancilla register. The swaps are peformed for each electron register ($i$) by performing unary iteration with $i$ as the selection register. The cost should be $\eta 3 n_p$ for the swaps and $\eta - 2$ for the unary iteration. In practice this cost is multiplied by 4 because we want to iterate over two selection registers (separately) $i$ and $j$ (and then swap back) so it becomes $12 \eta n_p + 4 \eta - 8$.
 
Screenshot 2023-11-01 at 2 31 31 PM

This bloq is more primitive and just assumes we have iteration_length system registers, one ancilla register and builds the circuit necessary for performing unary iteration + cswaps. The cost is iteration_length * bitsize + iteration_length - 2 Toffolis which is tested.


qualtran/bloqs/cswap_lth_reg.py
```diff
@@ -0,0 +1,122 @@
...
+    @cached_property
+    def selection_registers(self) -> Tuple[SelectionRegister, ...]:
+        return self.selection_regs
+
+    @cached_property
+    def target_registers(self) -> Tuple[Register, ...]:
+        iteration_length = self.selection_registers[0].iteration_length
+        regs = [Register(f'x{i}', bitsize=self.bitsize) for i in range(iteration_length)]
+        regs += [Register('y', bitsize=self.bitsize)]
+        return tuple(regs)
```
tanujkhattar: This assumes there's a single selection register and also does not use the fact that we can support multi-dimensional target registers. Support the more generic case using
```suggestion
        target_shape = tuple(sreg.iteration_length for sreg in self.selection_registers)
        return Signature([Register('targets', bitsize=self.target_bitsize, shape=target_shape), Register('output', bitsize=self.target_bitsize)])
```

qualtran/bloqs/cswap_lth_reg.py
```diff
@@ -0,0 +1,122 @@
...
+        return cirq.CircuitDiagramInfo(wire_symbols=wire_symbols)
+
+    def nth_operation(
+        self, context: cirq.DecompositionContext, control: cirq.Qid, **kwargs
+    ) -> cirq.OP_TREE:
+        selection_idx = tuple(kwargs[reg.name] for reg in self.selection_registers)[0]
+        target_regs = {reg.name: kwargs[reg.name] for reg in self.target_registers}
+        return CSwap(self.bitsize).make_on(
+            ctrl=control, x=target_regs[f'x{selection_idx}'], y=target_regs['y']
+        )
```
tanujkhattar: Again, support the more general case using 
```suggestion
        selection_idx = tuple(kwargs[reg.name] for reg in self.selection_registers)
        target_regs = kwargs['targets']
        return CSwap(self.bitsize).make_on(
            ctrl=control, x=target_regs[selection_idx], y=target_regs['y']
        )
```

tanujkhattar (changes requested):

qualtran/bloqs/cswap_lth_reg.py
```diff
@@ -0,0 +1,122 @@
...
+        [Fault-Tolerant Quantum Simulations of Chemistry in First Quantization](
+            https://arxiv.org/abs/2105.12767) page 20 paragraph 2.
+    """
+    selection_regs: Tuple[SelectionRegister, ...] = attr.field(
+        converter=lambda v: (v,) if isinstance(v, SelectionRegister) else tuple(v)
+    )
+    bitsize: int
+    control_regs: Tuple[Register, ...] = attr.field(
+        converter=lambda v: (v,) if isinstance(v, Register) else tuple(v),
+        default=(Register('ctrl', 0),),
```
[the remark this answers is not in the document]
fdmalone: Done.

qualtran/bloqs/cswap_lth_reg.py
```diff
@@ -0,0 +1,122 @@
...
+            labelled x`i`, where i is an integer and TO a single register called y
+
+    References:
+        [Fault-Tolerant Quantum Simulations of Chemistry in First Quantization](
+            https://arxiv.org/abs/2105.12767) page 20 paragraph 2.
+    """
+    selection_regs: Tuple[SelectionRegister, ...] = attr.field(
+        converter=lambda v: (v,) if isinstance(v, SelectionRegister) else tuple(v)
+    )
+    bitsize: int
```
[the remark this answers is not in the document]
fdmalone: Done.

qualtran/bloqs/cswap_lth_reg.py
fdmalone: Done % some changes (shouldn't return a Signature)

qualtran/bloqs/cswap_lth_reg.py
fdmalone: Done.

qualtran/bloqs/cswap_lth_reg.py
```diff
@@ -0,0 +1,122 @@
...
+from cirq._compat import cached_property
+
+from qualtran import bloq_example, BloqDocSpec, Register, SelectionRegister
+from qualtran._infra.gate_with_registers import total_bits
+from qualtran.bloqs.basic_gates import CSwap
+from qualtran.bloqs.unary_iteration_bloq import UnaryIterationGate
+
+
+@attr.frozen
+class ApplyCSwapToLthReg(UnaryIterationGate):
```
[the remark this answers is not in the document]
fdmalone: Done.

qualtran/bloqs/swap_network.py
```diff
@@ -255,3 +257,105 @@ def _swz_small() -> SwapWithZero:
...
+
+    The cost should be $L n_b + L - 2 + n_c$, where $L$ is the
+    iteration length, $n_b$ is the bitsize of
+    the registers to swap, and $n_c$ is the number of controls.
+
+    Args:
+        selection_regs: Indexing `select` signature of type Tuple[`SelectionRegisters`, ...].
+            It also contains information about the iteration length of each selection register.
+        bitsize: The size of the registers we want to swap.
+        nth_gate: A function mapping the composite selection index to a single-qubit gate.
```
tanujkhattar, line 282: nit: fix docstring

fdmalone: Done. One thing I realized (as this was just copied from the ApplyGateToLth.. bloq) is that there (or at least in the tests you pass a Signature into control_regs. Is it recommended to use Signature in place of a Tuple[Registers...]?

fdmalone: (I've just been using Signature as the return value of the Bloq.signature property.

tanujkhattar: > Is it recommended to use Signature in place of a Tuple[Registers...]?

As a function argument, we should use `Signature` only to define the `signature()` property of a Bloq. At call sites, I think it can be a convenient way of initializing a bunch of registers so it's fine. A `Signature` is an `Iterable[Register]`, so any function expecting a `Iterable[Register]` would work fine if we pass it a `Signature`.

qualtran/bloqs/swap_network.py
```diff
@@ -255,3 +257,105 @@ def _swz_small() -> SwapWithZero:
...
+
+    Applies the unitary which peforms
+    $$
+        U |l\rangle|\psi_0\rangle\cdots|\psi_l\rangle\cdots|\psi_n\rangle|\mathrm{junk}\rangle
+        \rightarrow
+        |l\rangle|\psi_0\rangle\cdots|\mathrm{junk}\rangle\cdots|\psi_n\rangle|\psi_l\rangle
+    $$
+    through a combination of unary iteration and CSwaps.
+
+    The cost should be $L n_b + L - 2 + n_c$, where $L$ is the
```
tanujkhattar, line 274:
```suggestion
    The toffoli cost should be $L n_b + L - 2 + n_c$, where $L$ is the
```

tanujkhattar (approved):

Issue #482, opened 2023-11-02
Registers not getting uncomputed for recently added `PairPotential` Bloq under chemistry/trotter/?


tanujkhattar:
The `diff_ij` register stores sum of `system_i` and `system_j`. We compute the register, use it to compute the `SumOfSquares` in register `sos` and then directly free it without uncomputing the result?  A similar thing seems to be happening for a bunch of other registers like `sos` etc. https://github.com/quantumlib/Qualtran/blob/00682cb3f1ba464e31c5bf1e199462ca611a9808/qualtran/bloqs/chemistry/trotter/potential.py#L89-L143

Maybe this is also an issue with some of the other recently added chemistry Bloqs? cc @fdmalone

fdmalone:
I did not account for uncomputing the arithmetic (it's also ignored in the paper). What's the best way to go about this?

fdmalone:
I should say the costs are ignored in the paper (or at least it was sort of ambiguous to me what is actually considered here):

<img width="1098" alt="Screenshot 2023-11-02 at 8 59 31 AM" src="https://github.com/quantumlib/Qualtran/assets/12097876/4f49d411-3836-408c-9cb7-ec2d0ee033b6">

The bloqs were meant to just capture the costs (as a first pass) but we (I) should definitely make it more complete.

fdmalone:
xref #485

PR #483, opened 2023-11-02
Add implementation of Gidney's `OutOfPlaceAdder` and use optimized `TComplexity`

tanujkhattar:
Adds an implementation and tests for `OutOfPlaceAdder` described in Figure 4b of https://arxiv.org/abs/1709.06648

Makes the following drive by changes
1. Changes signature of `OutOfPlaceAdder` such that the output register `c` is a newly allocated RIGHT (or LEFT for adjoint) register of size `bitsize+1`. Uncomputing the `c` register now takes 0 T-gates.
2. Fixes a bug in `PairPotential` where the register `c` was not getting uncomputed. There are potentially still more bugs in this implementation. See https://github.com/quantumlib/Qualtran/issues/482
3. Improves `assert_decompose_is_consistent_with_t_complexity` so that it checks for both `_t_complexity_` and `t_complexity` methods because `GateWithRegisters` is now `Bloq`. We can eventually get rid of `_t_complexity_` completely.
4. Adds additional imports to `qualtran/_infra/gate_with_registers.py` which are required for type checking.

### DIFF
--- a/qualtran/bloqs/arithmetic.py
+++ b/qualtran/bloqs/arithmetic.py
@@ -539,43 +539,83 @@ class Add(GateWithRegisters, cirq.ArithmeticGate):
 
 
 @frozen
-class OutOfPlaceAdder(Bloq):
+class OutOfPlaceAdder(GateWithRegisters, cirq.ArithmeticGate):
     r"""An n-bit addition gate.
 
     Implements $U|a\rangle|b\rangle 0\rangle \rightarrow |a\rangle|b\rangle|a+b\rangle$
-    using $4n - 4 T$ gates.
+    using $4n - 4 T$ gates. Uncomputation requires 0 T-gates.
 
     Args:
-        bitsize: Number of bits used to represent each integer. Must be large
-            enough to hold the result in the output register of a + b.
+        bitsize: Number of bits used to represent each input integer. The allocated output register
+            is of size `bitsize+1` so it has enough space to hold the sum of `a+b`.
 
     Registers:
      - a: A bitsize-sized input register (register a above).
-     - b: A bitsize-sized input/output register (register b above).
+     - b: A bitsize-sized input register (register b above).
+     - c: A bitize+1-sized LEFT/RIGHT register depending on whether the gate adjoint or not.
 
     References:
         [Halving the cost of quantum addition](https://arxiv.org/abs/1709.06648)
     """
 
     bitsize: int
+    adjoint: bool = False
 
     @property
     def signature(self):
-        return Signature.build(a=self.bitsize, b=self.bitsize, c=self.bitsize)
+        side = Side.LEFT if self.adjoint else Side.RIGHT
+        return Signature(
+            [
+                Register('a', self.bitsize),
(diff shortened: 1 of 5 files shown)


fdmalone:
I think this also may be a step towards resolving part of #390 (pg 13, above Fig 3): 
Screenshot 2023-11-02 at 10 31 36 AM

qualtran/bloqs/arithmetic.py
```diff
@@ -539,43 +539,83 @@ def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
 
 
 @frozen
-class OutOfPlaceAdder(Bloq):
+class OutOfPlaceAdder(GateWithRegisters, cirq.ArithmeticGate):
```
fdmalone, line 542: Question: Is there an advantage to keeping some of the arithmetic bloqs as cirq.ArithmeticGates?

tanujkhattar: Yes, they help with faster state vector cirq style simulations because `ArithmeticGate` has a `_apply_unitary_` overloaded.

qualtran/bloqs/arithmetic_test.py
```diff
@@ -404,14 +405,30 @@ def test_add_mod_n_protocols():
...
-    q2 = bb.add_register('c', bitsize)
-    a, b, c = bb.add(OutOfPlaceAdder(bitsize), a=q0, b=q1, c=q2)
-    cbloq = bb.finalize(a=a, b=b, c=c)
-    cbloq.t_complexity()
+    basis_map = {}
+    for x in range(2**3):
+        for y in range(2**3):
+            basis_map[int(f'0b_{x:03b}_{y:03b}_0000', 2)] = int(f'0b_{x:03b}_{y:03b}_{x+y:04b}', 2)
+    gate = OutOfPlaceAdder(bitsize=3)
+    # ops = GateHelper(gate).operation
```
fdmalone, line 413: nit: remove commented out code.

qualtran/bloqs/arithmetic_test.py
```diff
@@ -404,14 +405,30 @@ def test_add_mod_n_protocols():
...
+    )
+    # Check that inverse un-computes correctly
+    qubit_order = op.qubits
+    circuit = cirq.Circuit(cirq.decompose_once(op), cirq.decompose_once(op_inv))
+    for x in range(2**6):
+        bits = [*iter_bits(x, 10)][::-1]
+        assert_circuit_inp_out_cirqsim(circuit, qubit_order, bits, bits)
+    assert gate.t_complexity().t == 3 * 4
+    assert (gate**-1).t_complexity().t == 0
+    # gate.t_complexity() == cirq.Circuit(cirq.decompse_once(op))
```
fdmalone, line 428: removed commented out code

fdmalone (approved):
LGTM % some nits and a question about ArithmeticGates vs Bloqs with on_classical_vals? Does the ArithmeticGate have some more features?

PR #486, opened 2023-11-03
Call graph bugfix and overhaul

### LINKED ISSUE #367: Bloq counts graph fails to merge edges correctly when using `generalizer`
From the notebook combining Ands:
Screenshot 2023-09-25 at 2 04 56 PM
yields:
Screenshot 2023-09-25 at 2 04 11 PM

The first graph looks fine T count = (2x4 + 3x4) = 20, but the second merged example looks like 5x8 = 40.

mpharrigan:
Fixes #367 and more

The issue was that there was nothing stopping us from visiting a node more than once. The particular bug report causes the generalized `And` bloq to be visited twice; but you could reproduce it in any DAG with two nodes (see new tests; they used to fail).

Of course, the `sigma` value was always correct as noted in that issue. By visiting each node once, it would undercount bloqs for `sigma`. So I pulled that out into a separate step.

There was also a bug/edge case where if you explicitly returned no children/callees, it wouldn’t register that bloq as a leaf node and `sigma` would be empty.

I also took the opportunity to clean up the code. I’ve consolidated the base cases into one code path by throwing and/or handling `DecomposeTypeError`. If you look at the commits, this unified the code path for doing the sigma-computation base-case before ripping that out into its own step. But I left this refactoring in place because it could be useful if we wanted to do additional leaf node logic, hint hint.

I changed the variable names to refer to `blog` and `callee`; cleaned up the docstrings; and used consistent order-of-arguments for all the helper functions.


mpharrigan:
@tanujkhattar can you ptal

qualtran/resource_counting/bloq_counts.py
```diff
@@ -69,66 +69,121 @@ def _build_cbloq_counts_graph(cbloq: CompositeBloq) -> Set[BloqCountT]:
...
+    if bloq in g:
+        # We already visited this node.
+        return
+
+    # Make sure this node is present in the graph. You could annotate
+    # additional node properties here, too.
+    g.add_node(bloq)
 
+    try:
+        return _build_call_graph_inner(bloq, generalizer, ssa, keep, max_depth, g, depth)
```
tanujkhattar, line 161: nit: If I were you, I'd keep the logic of ` _build_call_graph_inner` directly here instead of abstracting away in a separate function. The reason is that I'm not a fan of `mutual recursion` (i.e. a->b->a->b) unless it's absolutely necessary. 


I see the extra level of indentation in the `try: / except:` bloq can be annoying, but to get rid of that I'd rather have a helper method `_get_callee_counts(*args)` and return an empty dict when `keep(bloq)` or `(max_depth is not None and depth >= max_depth)` is satisfied. So, 

```python
def _get_callee_counts(*args):
     if keep(bloq) or (max_depth is not None and depth >= max_depth):
        # Base cases 1 & 2 when `bloq` is a leaf.
        return {}
     return _generalize_callees(bloq.build_call_graph(ssa), generalizer)

def _build_call_graph(*args):
    callee_counts = _get_callee_counts(*args)
    if not callee_counts:
        # We don't do anything bespoke for leaf nodes, but we could...
        return 
    for callee, n in callee_counts:
        # continue with normal logic here
```

You could also raise either of `(DecomposeNotImplementedError, DecomposeTypeError)` in `_get_callee_counts` and catch it in `_build_call_graph` but instead of checking for `if not callee_counts` if you want. But loop that iterates on the children and recursively calls `_build_call_graph` would still be part of the `_build_call_graph` implementation.

qualtran/resource_counting/bloq_counts.py
```diff
@@ -69,66 +69,121 @@ def _build_cbloq_counts_graph(cbloq: CompositeBloq) -> Set[BloqCountT]:
...
+    The fact that this is a separate step will likely be made explicit in future generalizations
+    of 'resource counting'.
+    """
+    callees = list(g.successors(bloq))
+    if not callees:
+        return {bloq: 1}
+
+    sigma: Dict[Bloq, Union[int, sympy.Expr]] = defaultdict(lambda: 0)
+    for callee in callees:
+        callee_sigma = _compute_sigma(callee, g)
```
tanujkhattar, line 182: The current algorithm is `O(number of paths)` which can be exponential and is therefore _very_ inefficient. Note that `sigma` for each `bloq` in the same graph `g` remains the same. So we should compute it only when we visit the `bloq` for the first time and then reuse the previously computed values when we visit the `bloq` again. This will reduce the complexity to `O(n + m)` (no. of nodes + no. of edges). 


Here's an alternate iterative implementation which is fairy straightforward to generalize by generalizing (1) logic to get a default for leaves and (2) logic to combine the values of parent from value of it's children. 
```python
def _compute_sigma(bloq, g: nx.DiGraph) -> Dict[Bloq, Union[int, sympy.Expr]]:
    reverse_topo_sort = list(reversed(list(nx.topological_sort(g))))
    all_sigmas = Dict[Bloq, Dict[Bloq, Union[int, sympy.Expr]]] = defaultdict(lambda: {})
    for node in reverse_topo_sort:
        callees = list(g.successors(bloq))
        sigma = all_sigmas[node]
        if not callees:
            # 1. Default value for leaf nodes
            sigma[node] = 1
            continue
        for callee in callees:
            callee_sigma = all_sigmas[callee]
            # 2. Combination logic to compute sigma of the parent using sigma of the child.
            n = g.edges[bloq, callee]['n']
            for k in callee_sigma.keys():
                sigma[k] += callee_sigma[k] * n
    return all_sigmas[bloq]
```

Note that the recursive implementation with caching has an advantage in the scenario when you wish to compute `sigma` of a node in the graph that can reach only a small subset of nodes instead of reaching _all_ nodes of the graph. Given that for our use case, we are often interested in sigma of the root node, that's not the case and therefore the iterative implementation would almost always be faster than recursive.

tanujkhattar (changes requested):
Looking good overall. Two suggested changes.

qualtran/resource_counting/bloq_counts.py
mpharrigan: done

qualtran/resource_counting/bloq_counts.py
mpharrigan: done

fdmalone (approved):

PR #490, opened 2023-11-03
[docs] Deferred maintenance on sphinx docs

mpharrigan:
- include new notebooks in toc so they actually show up
 - change major section titles
 - minimal organization of sections within a realm
 - reference doc autogen includes new modules

Requires #467


fdmalone (approved):
LGTM

dev_tools/qualtran_dev_tools/reference_docs.py
```diff
@@ -232,11 +232,29 @@ def generate_ref_docs(reporoot: Path):
...
+        linalg,
+        resource_counting,
+        serialization,
+        surface_code,
+        testing,
+    )
+    from qualtran.simulation import classical_sim, tensor
 
     # prevent unused warnings:
-    assert [drawing, resource_counting, cirq_interop, quimb_sim, classical_sim]
```
fdmalone, line 239: Is this better than just a pylint ignore?

mpharrigan: hmm this works in pycharm as well but maybe there's some secret comment that works for all systems. but the assert also works

Issue #485, opened 2023-11-02
List of Arithmetic Bloqs which need decompositions


fdmalone:
## Square integers
 - code https://github.com/quantumlib/Qualtran/blob/main/qualtran/bloqs/arithmetic.py#L648
 - ref: https://arxiv.org/abs/2105.12767. pg 76 for Toffoli complexity (text not circuit)
 - adjoint: No

## Sum of Squares (integers)
- code: https://github.com/quantumlib/Qualtran/blob/main/qualtran/bloqs/arithmetic.py#L690
- ref: https://arxiv.org/abs/2105.12767 pg 80 (text no circuit)
- adjoint: No

## Product (integers)
- code: https://github.com/quantumlib/Qualtran/blob/main/qualtran/bloqs/arithmetic.py#L747
- ref: https://arxiv.org/abs/2105.12767 pg 81 (text no circuit)
- adjoint: No

## Scale integer by real number
- code: https://github.com/quantumlib/Qualtran/blob/main/qualtran/bloqs/arithmetic.py#L797
- ref: https://arxiv.org/pdf/2007.07391.pdf pg 70 (text)
- adjoint: No

## Multiply two real numbers
- code: https://github.com/quantumlib/Qualtran/blob/main/qualtran/bloqs/arithmetic.py#L853
- ref: https://arxiv.org/pdf/2007.07391.pdf pg 71 (text)
- adjoint: No

## Square real number
- code: https://github.com/quantumlib/Qualtran/blob/main/qualtran/bloqs/arithmetic.py#L904
- ref: https://arxiv.org/pdf/2007.07391.pdf) pg 74 (text)
- adjoint: No

## Greater Than
- code: https://github.com/quantumlib/Qualtran/blob/main/qualtran/bloqs/arithmetic.py#L958
- ref: ? (modification of <=)
- adjoint: No

## Greater Than Constant
- code: https://github.com/quantumlib/Qualtran/blob/main/qualtran/bloqs/arithmetic.py#L1000
- ref: ? (modification of <= c)
- adjoint: No

## Equals A constant
- code: https://github.com/quantumlib/Qualtran/blob/main/qualtran/bloqs/arithmetic.py#L1038
- ref: ? (modification of <= c)
- adjoint: No

## ToContiguousIndex
- code: https://github.com/quantumlib/Qualtran/blob/main/qualtran/bloqs/arithmetic.py#L1070
[...]

Issue #237, opened 2023-05-23
Implement catalytic controlled-Hadamard


ncrubin:
https://journals.aps.org/prxquantum/pdf/10.1103/PRXQuantum.2.030305

Fig 17.

fdmalone:
![Screenshot 2023-10-31 at 2 30 44 PM](https://github.com/quantumlib/Qualtran/assets/12097876/2e75a945-5c1a-4c51-9225-4fad42db45a9)

fdmalone:
@mpharrigan / @tanujkhattar what are your thoughts on adding a ControlledHadamard Bloq, which possibly delays the above implementation and just tags the cost? It would be nice to have this bloq to add more colour to the first quantized (and second quantized) decompositions. I know there is some discussion in #134 on how to do this generally, so wanted to check first.

An alternative would be to decompose via cirq and generalize appropriately.

mpharrigan:
I'd be fine with that. I think since this _is_ a special decomposition for the controlled version of an ordinary bloq it doesn't conflict with #134 which deigns to automatically decompose ControlledBloq(bloq) as [ControlledBloq(sb) for sb in bloq.decompose_bloq()]

mpharrigan:
Did this ever get contributed?

tanujkhattar:
no

mpharrigan:
I'll add a specific `CHadamard` bloq but leave it as a leaf bloq #873 and the existing cirq-based magic circuit and the explicit construction can be captured in methods, and the desired resource-counting behavior can be configured via the `QECGatesCost` cost-key

PR #493, opened 2023-11-03
First quantized select and prepare

fdmalone:
Adds final "outer" select and prepare bloqs which piece together the necessary subroutines from the previous first quantization PRs (contains part of #454 which should be merged first). Fixes some errors in previous incarnations of Select (signatures were either lacking some registers, which only really became apparent at a higher level).

I wrapped the multiplexed swaps #478  in to make splits / joins less obvious when moving from (3,) bitsize-sized registers to a (3*bitsize) big register. When combined together the costs match those quoted % some small things I've noted before. The remaining tasks would be to implement several lower level primitives (e.g. several arithmetic primitives need implementations and correct adjoints #485), and some of the state preparation routines are relatively staightforward but need controlled hadamards #237. 

I uses the existing SelectOracle and PrepareOracle as a model but maybe we want to include elements of #300 aswell (splitting big registers e.g.)


qualtran/bloqs/chemistry/pbc/first_quantization/first_quanization_test.py
```diff
@@ -1,19 +0,0 @@
...
-#  Unless required by applicable law or agreed to in writing, software
-#  distributed under the License is distributed on an "AS IS" BASIS,
-#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-#  See the License for the specific language governing permissions and
-#  limitations under the License.
-
-from qualtran.testing import execute_notebook
-
-
-def test_notebook():
```
mpharrigan, line 18: ?

qualtran/bloqs/chemistry/pbc/first_quantization/__init__.py
```diff
@@ -79,7 +79,4 @@
 ignored.
 """
 
-from .prepare_t import PrepareTFirstQuantization
```
mpharrigan, line 82: What's the strategy here of what is re-exported vs not

mpharrigan (approved):

qualtran/bloqs/chemistry/pbc/first_quantization/first_quanization_test.py
fdmalone: I've moved the first_quantization.py to select_and_prepare.py (the notebook is tested there.)

qualtran/bloqs/chemistry/pbc/first_quantization/__init__.py
fdmalone: I think just exporting the big bloqs is sufficient as they're most likely to be interacted with.

PR #495, opened 2023-11-06
[autodoc] qualtran.bloqs.sorting

mpharrigan:
Use v2 doc infrastructure for sorting module `Comparitor` and `BitonicSort`.


qualtran/bloqs/sorting.py
```diff
@@ -92,7 +108,7 @@ class BitonicSort(Bloq):
 
     @property
     def signature(self):
-        return Signature([Register("input", bitsize=self.bitsize, shape=(self.bitsize,))])
+        return Signature([Register("input", bitsize=self.bitsize, shape=(self.k,))])
```
mpharrigan, line 111: @fdmalone can you confirm that this is the correct signature

fdmalone: yikes

fdmalone (approved):

PR #496, opened 2023-11-07
[autotest] Check equivalent bloq counts

mpharrigan:
As part of the autotesting infrastructure, add an automatic check (over all bloq_examples) that checks whether `bloq.bloq_counts() == bloq.decompose_bloq().bloq_counts()`. In addition to PASS or MISSING; this can be UNVERIFIED if only one of (build_call_graph; decomposition) is provided.

This also includes a suite of generalizers to help reconcile any negligible differences between build_call_graph and decomposition.

### DIFF
--- a/dev_tools/bloq-report-card.ipynb
+++ b/dev_tools/bloq-report-card.ipynb
@@ -19,18 +19,31 @@
    ]
   },
   {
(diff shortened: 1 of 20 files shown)


mpharrigan:
replaced by #696

Issue #491, opened 2023-11-03
Set up a "slow" marker for pytest


mpharrigan:
xref #114 #463

 - Add pytest support for a custom marker named "slow"
 - github action for pull requests will not run slow tests
 - set up nightly(?) CI that runs all tests including "slow".
 - What should the default be for running tests on a local workstation? probably skip them, right?

mpharrigan:
This is done

PR #499, opened 2023-11-07
Define custom pytest mark slow

pavoljuhas:
* Skip the slow-marked tests by default, but provide an extra
  pytest option `--enable-slow-tests` to run them anyway.
  Also run slow tests if selected with `pytest -m slow`.

* Support a quick check of test selection with `pytest --co -m skip`
  and `pytest --co -m "not skip"`.

* Use the new `slow` mark for the slow tests instead of unconditional skip

Partially implements #491


mpharrigan (approved):
niiiiiice. Thank you!

PR #500, opened 2023-11-07
delete `source dev_tools/pypath` in check scripts

anurudhp:
As file `dev_tools/pypath` does not exist.


tanujkhattar (approved):
Simply delete the line instead of commenting it.

anurudhp:
deleted the lines

PR #501, opened 2023-11-08
Refactor `arithmetic.py` into `arithmetic/` module

tanujkhattar:
Mostly just moving code around for better organization.


NoureldinYosri:
it looks like you are creating new files and copying the relevant parts, right? if so then we are losing the file histroy. Instead lets copy the files and then remove the unnecessary parts

mpharrigan:
I'm pretty sure git can't distinguish between these ^^ i.e. the diff is computed on-the-fly. From command line you can `git diff --color-moved`.

mpharrigan (approved):

NoureldinYosri:
@mpharrigan @tanujkhattar  it's not about the diff, it's about the file history. now if you check the files in arithmetic/ it will look like all of them were created at the same time by tanuj
after
compare this with the history before the change
before
so if a new contributor is looking for more information about more information about a block by looking who added it to contacct them, they wouldn't be able to since that information is lost now.

PR #502, opened 2023-11-08
Add partition bloq

fdmalone:
Stripping out from #300 as a useful utility bloq for a follow up of #392. Writte by @mpharrigan I just added a test and an example to the notebook.

### DIFF
--- a/qualtran/bloqs/util_bloqs.py
+++ b/qualtran/bloqs/util_bloqs.py
@@ -17,6 +17,7 @@
 from functools import cached_property
 from typing import Dict, Tuple, TYPE_CHECKING, Union
 
+import attrs
 import numpy as np
 import quimb.tensor as qtn
 from attrs import frozen
@@ -136,6 +137,41 @@ class Join(Bloq):
             text = f'[{", ".join(str(i) for i in soq.idx)}]'
             return directional_text_box(text, side=soq.reg.side)
         return directional_text_box(' ', side=soq.reg.side)
+
+
+@frozen
+class Partition(Bloq):
+    """Partition a generic index into multiple registers.
+
+    Args:
+        n: The total bitsize of the un-partitioned register
+        regs: Registers to partition into. The `side` attribute is ignored.
+        partition: `False` means un-partition instead.
+
+    Registers:
+        x: the un-partitioned register. LEFT by default.
+        [user spec]: The registers provided by the `regs` argument. RIGHT by default.
+    """
+
+    n: int
+    regs: Tuple[Register, ...]
+    partition: bool = True
+
+    @cached_property
+    def signature(self) -> 'Signature':
+        lumped = Side.LEFT if self.partition else Side.RIGHT
+        partitioned = Side.RIGHT if self.partition else Side.LEFT
+
+        return Signature(
+            [Register('x', bitsize=self.n, side=lumped)]
+            + [attrs.evolve(reg, side=partitioned) for reg in self.regs]
+        )
+
+    def dagger(self):
+        return attrs.evolve(self, partition=not self.partition)
+
+    def __str__(self):
+        return 'Partition(...)'
 
 
 @frozen
--- a/qualtran/bloqs/util_bloqs.py
+++ b/qualtran/bloqs/util_bloqs.py
@@ -14,7 +14,6 @@
 
 """Bloqs for virtual operations and register reshaping."""
 
-from collections import defaultdict
 from functools import cached_property
 from typing import Dict, Tuple, TYPE_CHECKING, Union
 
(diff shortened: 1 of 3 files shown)


fdmalone:
I removed the add_controlled method. Not sure what the status of control is.

qualtran/bloqs/util_bloqs.py
```diff
@@ -138,6 +139,41 @@ def wire_symbol(self, soq: 'Soquet') -> 'WireSymbol':
...
+        return Signature(
+            [Register('x', bitsize=self.n, side=lumped)]
+            + [attrs.evolve(reg, side=partitioned) for reg in self.regs]
+        )
+
+    def dagger(self):
+        return attrs.evolve(self, partition=not self.partition)
+
+    def __str__(self):
+        return 'Partition(...)'
```
mpharrigan, line 174: is there a reason why you're doing `__str__` and not (or instead of) `pretty_name` and/or `short_name`?

fdmalone: because I just copy pasted the code lol. I will fix

qualtran/bloqs/util_bloqs_test.py
```diff
@@ -55,6 +57,32 @@ def test_util_bloqs():
...
+class TestPartition(Bloq):
+    test_bloq: Bloq
+
+    @cached_property
+    def bitsize(self):
+        return sum(reg.total_bits() for reg in self.test_bloq.signature)
+
+    @cached_property
+    def signature(self) -> Signature:
+        return Signature.build(regs=self.bitsize)
```
mpharrigan, line 70: can we use a different name than `regs`? i.e. there's a distinction between the register name and the Partition class attribute that we should clearly delineate in the test

qualtran/bloqs/util_bloqs.py
fdmalone: I just deleted, the default Partition is fine I think (...) not adding much.

qualtran/bloqs/util_bloqs.py
```diff
@@ -138,6 +139,38 @@ def wire_symbol(self, soq: 'Soquet') -> 'WireSymbol':
...
+        partitioned = Side.RIGHT if self.partition else Side.LEFT
+
+        return Signature(
+            [Register('x', bitsize=self.n, side=lumped)]
+            + [attrs.evolve(reg, side=partitioned) for reg in self.regs]
+        )
+
+    def dagger(self):
+        return attrs.evolve(self, partition=not self.partition)
+
```
tanujkhattar, line 172: Can we implement `add_my_tensors` so that it works in simulations?

qualtran/bloqs/util_bloqs.py
```diff
@@ -138,6 +139,38 @@ def wire_symbol(self, soq: 'Soquet') -> 'WireSymbol':
...
+        x: the un-partitioned register. LEFT by default.
+        [user spec]: The registers provided by the `regs` argument. RIGHT by default.
+    """
+
+    n: int
+    regs: Tuple[Register, ...]
+    partition: bool = True
+
+    @cached_property
+    def signature(self) -> 'Signature':
```
tanujkhattar, line 161: Can we implement `as_cirq_op` so that it does the correct reshaping of cirq quregs, similar to `Split` and `Join` ?

qualtran/bloqs/util_bloqs.py
```diff
@@ -138,6 +139,38 @@ def wire_symbol(self, soq: 'Soquet') -> 'WireSymbol':
...
+
+    Registers:
+        x: the un-partitioned register. LEFT by default.
+        [user spec]: The registers provided by the `regs` argument. RIGHT by default.
+    """
+
+    n: int
+    regs: Tuple[Register, ...]
+    partition: bool = True
+
```
tanujkhattar, line 159: Similarly, let's implement `t_complexity`, `on_classical_vals` and `wire_symbol` protocols

qualtran/bloqs/util_bloqs_test.py
```diff
@@ -55,6 +57,34 @@ def test_util_bloqs():
...
+        out_regs = bb.add(self.test_bloq, **{reg.name: sp for reg, sp in zip(bloq_regs, out_regs)})
+        test_regs = bb.add(
+            partition.dagger(), **{reg.name: sp for reg, sp in zip(bloq_regs, out_regs)}
+        )
+        return {'test_regs': test_regs}
+
+
+def test_partition():
+    bloq = TestPartition(test_bloq=TestCNOT())
+    assert_valid_bloq_decomposition(bloq)
```
tanujkhattar, line 85: Can we add a test that verifies that `CompositeBloq` using `Partition` works fine when converted to a Cirq circuit?

tanujkhattar (changes requested):

qualtran/bloqs/util_bloqs_test.py
fdmalone: I tested through conversion to a unitary, is this enough?

fdmalone:
@tanujkhattar PTAL

mpharrigan (approved):

Issue #503, opened 2023-11-08
Document Protocols


mpharrigan:
In qualtran, we have "protocols" for annotating, deriving, and querying properties of bloqs. Practically, a protocol consists of

 - (a name)
 - A method on `Bloq` that takes no arguments and returns the property. This is the **interface**
 - A method on `Bloq` that takes system-supplied arguments which bloq authors can override to provide information for the protocol. This is the **implementation**.
 - A Python module containing functions for plumbing the implementation into the interface and other **additional functionality**.
 - Helper functions for implementing one protocol in terms of another called **fallbacks**. Some protocols have a **default fallback** that requires no special action from the bloq author.

Each protocol should have a documentation page where each bolded item is a heading. 

- [ ] decompose
- [ ] call graph
- [ ] classical sim
- [ ] tensor sim
- [ ] drawing
- [ ] cirq interop

And then there should be a docs page that describes what a protocol is and gives a summary of the bolded elements for each one as a quick reference.

 - [ ] overview

There's a suite of "information-theoretically-possible" fallbacks that we don't provide functions for but maybe we should. And then document them. We could even make a graph of how protocols can be derived from one another and whether those fallbacks happen by default.

PR #504, opened 2023-11-08
Initial overview documentation of protocols

mpharrigan:
This is a step towards #503.

As part of writing full docs for decompose, classical sim, cirq interop, drawing, and other protocols: their corresponding section should be fleshed out in this overview page.


qualtran/Protocols.ipynb
```diff
@@ -0,0 +1,72 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "a3bea287",
+   "metadata": {},
+   "source": [
+    "# Protocols\n",
```
NoureldinYosri, line 8: If I didn't know what a protocol is, I wouldn't have understood this section. I think a simple paragraph that doesn't mention bloqs and references some good explanation should make this section more understandable.

mpharrigan: well I don't think we're using the word "protocol" accurately in an established way:

https://en.wikipedia.org/wiki/Protocol

but rather leaning in to its ambiguous meaning (https://en.wiktionary.org/wiki/protocol) to define it exactly how we want; i.e. below.

fdmalone (approved):

qualtran/Protocols.ipynb
fdmalone: To be clear is this the idiomatic meaning of protocol and not any formal CS protocol type thing

mpharrigan: how about this updated wording:

> We write bloqs so we can query properties, draw diagrams, and check the correctness of quantum algorithms. Basically: we write bloqs so we can do things with them! Each 'thing'--or genre of thing--we want to do has an associated "protocol". In our usage of the term, a protocol is a set of methods, functions, or procedures to structure a relevant piece of functionality.

PR #508, opened 2023-11-10
Block encoding bloq

fdmalone:
Hopefully addressing the comment from #392 (which I will close in favour of this). Also rips out BlackBoxSelect and BlackBoxPrepare from #300, and uses #502.

This is a first step towards implementing some of the higher level block encodings like this
Screenshot 2023-11-09 at 4 34 36 PM

I'll follow up with bloqs for the above picture but here I just added a BlockEncodingBloq. Similar to #300 I added a BlackBoxBlockEncoding(Bloq) which expects a SelectOracle and PrepareOracle Bloq. Unlike the picture above we define 3 registers (two ancilla registers and a system register). These are then partitioned according to the details of Select and Prepare. Ideally I think I'd like just two in keeping with the _a and _s convention often used, I guess another Parition could achieve this in Prepare. I should also move Select and Prepare to their own files (or the existing select_and_prepare).


fdmalone:
Screenshot 2023-11-09 at 10 53 44 PM

qualtran/bloqs/block_encoding.py
```diff
@@ -0,0 +1,272 @@
...
+    def selection_register(self) -> Register:
+        return Register('selection', bitsize=self.prepare.selection_bitsize)
+
+    @property
+    def junk_register(self) -> Register:
+        return Register('junk', bitsize=self.prepare.junk_bitsize)
+
+    @property
+    def system_register(self) -> Register:
+        return Register('system', bitsize=self.select.system_bitsize)
```
mpharrigan, line 223: Can you explain and motivate the design decision for having the abstract base construction for `BlockEncoding`. You could imagine un-abstracting these properties and make them always be defined like this. Instead your "api" would be that select and prepare need to expose a selection, junk, and system bitsize.

mpharrigan (approved):
I have a fundamental question; but if you have a good answer happy to merge and iterate

qualtran/bloqs/block_encoding.py
fdmalone: My idea here was that if we want more complicated block encodings (i.e. add, chebychev, multiply etc which are given in Eq. 12-15 and Eq 33 above) these are also block encodings but should define the same interface as the "arguments" right?

I was also considering the case were say we have some complicated block encoding of the hamiltonian and maybe we don't want to worry about all the details but we want B[T_10[H]], then something like

``` python
@attrs.frozen
class ChebyshevPolynomial(BlockEncoding):
    r"""Block encoding of $T_j[H]$ where $T_j$ is the $j$-th Chebyshev polynomial.

    Here H is a Hamiltonian with spectral norm $|H| \le 1$ and we assume we have
    an $n_L$ qubit ancilla register and assume that $j > 0$ to avoid block
    encoding the identity operator.

    Recall:

    \begin{align*}
        T_0[H] &= \mathbb{1} \\
        T_1[H] &= H \\
        T_2[H] &= 2 H^2 - \mathbb{1} \\
        T_3[H] &= 4 H^3 - 3 H \\
        &\dots
    \end{align*}

    Args:
        block_encoding: Block encoding of a Hamiltonian $H$, $\mathcal{B}[H]$.
        order: order of Chebychev polynomial.

    References:
        [Quantum computing enhanced computational catalysis](
            https://arxiv.org/abs/2007.14460). Page 45; Theorem 1.
    """

    block_encoding: BlockEncoding
    order: int

        if self.order < 1:
            raise ValueError(f"order must be greater >= 1. Found {self.order}.")

    def short_name(self) -> str:
        return f"T_{self.order}[{self.block_encoding.short_name()}]"

    @property
    def selection_register(self) -> Register:
        return Register('selection', bitsize=self.block_encoding.selection_bitsize)

    @property
    def junk_register(self) -> Register:
        return Register('junk', bitsize=self.block_encoding.junk_bitsize)

    @property
    def system_register(self) -> Register:
        return Register('system', bitsize=self.block_encoding.system_bitsize)

    def build_composite_bloq(
        self, bb: 'BloqBuilder', selection: 'SoquetT', junk: 'SoquetT', system: 'SoquetT'
    ) -> Dict[str, 'Soquet']:
        # includes selection registers and any selection registers used by PREPARE
        selection, junk, system = bb.add(
            self.block_encoding, selection=selection, junk=junk, system=system
        )
        for iorder in range(1, self.order):
            selection = bb.add(Reflection(bitsize=self.selection_bitsize), refl=selection)
            be = attrs.evolve(self.block_encoding, adjoint=((iorder % 2) == 1))
            selection, junk, system = bb.add(be, selection=selection, junk=junk, system=system)
        return {'selection': selection, 'junk': junk, 'system': system}

        n = self.order
        num_refl = self.block_encoding.selection_bitsize
        return {(Reflection(bitsize=num_refl), n - 1), (self.block_encoding, n)}


```

fdmalone:
Actually there's an issue here which is select may expect control registers (which may be prepared in junk during prepare). I need to think about how to wire these up properly.

fdmalone:
> Actually there's an issue here which is select may expect control registers (which may be prepared in junk during prepare). I need to think about how to wire these up properly.

Actually, my thinking was wrong here. The control registers internal to select are separate from the control_registers abstract property which are for controlling the whole bloq.

qualtran/bloqs/block_encoding.py
mpharrigan: but in your answer, you still have the same register names; albeit with a different source of the bitsizes

mpharrigan (approved):
this still lgtm

PR #505, opened 2023-11-09
Add `QFTPhaseGradient` which synthesizes rotations via addition into gradient register

tanujkhattar:
This PR is a first proof of concept (with simulations to verify) for synthesizing rotations via addition into a phase gradient register. As a concrete example, the PR adds 2 different versions of `QFT` to the library
- `QFTSchoolBook` : Uses $\frac{n * (n + 1)}{2}$ arbitrary rotations + cliffords. Note that naively, each of these rotations would require `O(log(1/eps)` T-gates so the overall T-complexity is $O(n^2 \log_2{\frac{1}{\epsilon}})$
- `QFTPhaseGradient`: Uses the trick described in https://algassert.com/post/1620 to synthesize gradient rotations via addition into the phase gradient register. This trick is very commonly used in other papers as well (eg: https://arxiv.org/abs/2007.07391) and this PR adds the necessary infrastructure and tests to support this trick. The resulting QFT gate requires $< 4 * n^2$ T-gates. 

Note that I currently use a naive version of `PlusEqualProduct` gate to compute the product using `O(n ** 2)` T-gates. We can probably use a more optimal multiplication Bloq (for eg: Karatsuba, or it's more recent optimized versions) but that will be TODO for future.

PR #509, opened 2023-11-10
Add `HammingWeightPhasingViaPhaseGradient` Bloq

tanujkhattar:
This is currently rebased on top of https://github.com/quantumlib/Qualtran/pull/505

This is a more advanced example of using phase gradient states to synthesize rotations and combines it with the `HammingWeightPhasing` technique introduced earlier.

Note that combining the a) hamming weight phasing trick with b) synthesizing rotations using phase gradient trick is not commonly used in the literature and I have added a detailed docstring to the `HammingWeightPhasingViaPhaseGradient` bloq explaining why that is the case. 

Nevertheless, this is a good educational example and maybe also useful for specific cases of `(bitsize, eps)` pairs (see the tests). 

cc @ncrubin @fdmalone


qualtran/bloqs/rotations/hamming_weight_phasing.py
```diff
@@ -86,3 +89,71 @@ def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
...
+    @cached_property
+    def b_phase(self) -> int:
+        return int(np.log2(1 / self.eps))
+
+    @cached_property
+    def b_grad(self) -> int:
+        return max(self.bitsize.bit_length(), self.b_phase + int(np.ceil(np.log2(self.b_phase))))
+
+    def build_composite_bloq(self, bb: 'BloqBuilder', **soqs: 'SoquetT') -> Dict[str, 'SoquetT']:
+        x, phase_grad = soqs['x'], soqs['phase_grad']
```
fdmalone, line 146: Just have x and phase_grad as arguments rather than **soqs?

qualtran/bloqs/rotations/hamming_weight_phasing.py
```diff
@@ -86,3 +89,71 @@ def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
...
+    bitsize: int
+    exponent: float = 1
+    eps: int = 1e-10
+
+    @cached_property
+    def signature(self) -> 'Signature':
+        return Signature.build(x=self.bitsize, phase_grad=self.b_grad)
+
+    @cached_property
+    def b_phase(self) -> int:
```
fdmalone, line 138: Add a one line  docstring OR rename to bitsize_phase?

fdmalone: Ehh, sorry I guess you define these in the docstring.

qualtran/bloqs/rotations/hamming_weight_phasing.py
```diff
@@ -86,3 +89,71 @@ def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
...
+    phase gradient register. See reference [1] for more details on this technique.
+
+    Note: For most reasonable values of `bitsize` and `eps`, the naive `HammingWeightPhasing` would
+    have better constant factors than `HammingWeightPhasingViaPhaseGradient`. This is because, in
+    general, the primary advantage of using phase gradient is to reduce the complexity from
+    $O(n * log(1/eps))$ to O(log(1/eps) ** 2) (the phase gradient register is of size
+    $O(log(1/eps))$ and a scaled addition into the target takes $(b_{grad} - 2)(log(1/eps) + 2)$).
+    Therefore, to apply $n$ individual rotations on a target register of size $n$, the complexity is
+    independent of $n$ and is essentially a constant (scales only with log(1/eps)).
+    However, for the actual constant values to be better, value of $n$ needs to be $> log(1/eps)$.
```
fdmalone, line 111: nit: ", the value"

qualtran/bloqs/rotations/hamming_weight_phasing.py
```diff
@@ -86,3 +89,71 @@ def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
...
+    In the case of hamming weight phasing, $n$ corresponds to the hamming weight register which itself
+    is $log(\text{bitsize})$. Thus, as `eps` becomes smaller, the required value of $\text{bitsize}$,
+    for the phase gradient version to become more performant, becomes larger.
```
fdmalone, line 114: Question: Is this specific to hamming weight phasing? In many of the chemistry algorithms rotations are always costed assuming a phase gradient gate (so like  (n - 2) Toffolis).

qualtran/bloqs/rotations/hamming_weight_phasing_test.py
```diff
@@ -39,3 +48,58 @@ def test_hamming_weight_phasing(n: int, theta: float):
     hw_phasing = cirq.Circuit(state_prep, HammingWeightPhasing(n, theta).on(*gh.quregs['x']))
     hw_final_state = sim.simulate(hw_phasing).final_state_vector
     assert np.allclose(expected_final_state, hw_final_state, atol=1e-7)
+
+
+@attrs.frozen
```
fdmalone, line 53: Wait, what's the point of this test class, why not just use the implemented bloq?

qualtran/bloqs/rotations/hamming_weight_phasing.py
tanujkhattar: So far, I have been prefering this style because it keeps mypy happy and I don't like the yellow pycharm squiggles in all my classes. I'll change for now but I think this is worth discussing in our weekly sync so everyone follows the same style

qualtran/bloqs/rotations/hamming_weight_phasing_test.py
tanujkhattar: This `HammingWeightPhasingViaPhaseGradient` class assumes that a phase gradient register is correctly prepared. The test class encapsulates the circuit to prepare a phase gradient register, apply rotations via hamming weight phasing, and then clean up the phase gradient register. 

The test class exists mainly for convenience.

qualtran/bloqs/rotations/hamming_weight_phasing.py
tanujkhattar: The statement regarding tradeoff is valid in general but it becomes relevant mainly in the case of hamming weight phasing. Let's look at some numbers for more details -

- Let $eps \approx 10^{-12}$, $\log_2(\frac{1}{eps}) \approx 40$. 
- Let $n$ be the size of your system register s.t. $|\Psi_{n}\rangle$ is an n-qubit state we care about. 
$$|\Psi_{n}\rangle = \sum_{x=1}^{2^n}\alpha_{x}|x\rangle$$ 
- Suppose you want to apply a phase $e^{i \gamma x}$ to each state $|x\rangle$. 
- Naively, this would involve performing $n$ rotations, one on each qubit in the system register. Thus, the cost of the phasing operation would be $O(n \log(1/eps))$
- Using phase gradient trick, we do a scaled addition to a phase gradient register of size $b_\text{grad}$. One scaled addition reduces to $b_\text{gamma}$ normal additions, each of cost $b_\text{grad}$. Generally, both these variables are set $\approx O(\log(1/eps))$ and thus the total cost is $O(\log(1/eps)^ 2)$
- So, whenever the system register $n$ is greater than $b_\text{gamma}$, using phase gradient trick is better than naive approach. Assuming a system register of size > 40 for any beyond classical application is a reasonable assumption so for general chemistry algorithms it makes sense to ignore this tradeoff and always use phase gradient trick.
- However, for HammingWeightPhasing, for a system register of size $n$ the first reduction reduces the problem to applying only $\log(n)$ rotations on the hamming weight register. For $\log(n)$ to be greater than $\log(1/eps)$, $n$ should be $>1/eps$ -- this is a much more constrained scenario and one needs to be careful of the tradeoff when doing the costing.

tanujkhattar:
@fdmalone Addressed all your comments, this is ready for a re-review!

qualtran/bloqs/rotations/hamming_weight_phasing.py
fdmalone: makes sense, thanks for the explanation

fdmalone (approved):

Issue #240, opened 2023-05-26
Multi-Control Pauli Decomposition is probably incorrect


tanujkhattar:
Right now, we don't have a test for correctness and as part of writing a gate locally,  I discovered that the gate is probably not doing what it's promised to do (I tested for `cirq.Z(q).controlled_by(*control_qubits, control_vals = [0] * len(control_qubits))`) 

This issue is to verify the correctness and add a validation test:

https://github.com/quantumlib/cirq-qubitization/blob/a2f283e0269d92d89aba2ce19e5b973019db5aca/cirq_qubitization/cirq_algos/multi_control_multi_target_pauli.py#L44

fdmalone:
@tanujkhattar I hit this recently and remembered I was supposed to look into it. Any idea what's going on before I dig in?

mpharrigan:
@tanujkhattar did you fix this?

anurudhp:
#790 added tensor tests to verify the action, closing.

PR #510, opened 2023-11-10
Chebyshev polynomial block encoding

fdmalone:
Builds on #508, high level polynomial bloq which builds the construction 

Screenshot 2023-11-09 at 10 51 33 PM

Screenshot 2023-11-09 at 10 53 03 PM


fdmalone:
I added a wrapper Reflection bloq around a multi-controlled pauli Z. I think there are some issues with this implementation #240 which I should revisit, but nevertheless it will make for nicer diagrams for chemistry algos too.

qualtran/bloqs/block_encoding.py
```diff
@@ -46,13 +48,62 @@
...
+    cvs: Tuple[int]
+
+    def __attrs_post_init__(self):
+        if len(self.bitsizes) != len(self.cvs):
+            raise ValueError(
+                f"cvs must be same length as bitsizes: {len(self.cvs)} vs {len(self.bitsizes)}"
+            )
+
+    def pretty_name(self) -> str:
+        return 'Refl'
```
mpharrigan, line 73: should this be `short_name`? I know we don't really have guidance on this

mpharrigan (approved):

qualtran/bloqs/block_encoding.py
fdmalone: Done.

Issue #517, opened 2023-11-16
protobuf not getting installed


mstechly:
Hey!

I have followed the regular installation instruction and when I tried to use serialization module I got the following error:
```
from google.protobuf import descriptor as _descriptor
ModuleNotFoundError: No module named 'google'
```

Installing `protobuf` manually solved the issue, but I just wanted to point out that perhaps something is missing from the installation instructions?

mpharrigan:
fixed by #521 and shouldn't happen again with fixed CI

mstechly:
Awesome, thanks @mpharrigan !

PR #521, opened 2023-11-17
Use pytest env for pytest CI check

mpharrigan:


fdmalone:
our pytest env is insufficient / out of date it seems

mpharrigan:
sorry this is in progress to fix #517 among other issues.

This first commit gives the expected failure; I'll push a fix -- and we'll be left with the _other_ failures which then I'll think about how to fix :p

mpharrigan:
ok. This is now ready. I changed the pytest job to run the runtime tests with runtime dependencies and added a new, separate job to test the dev dependencies. Previously: the dev environment was used for testing so badly configured factorized dependencies could sneak by causing #517

check/pytest-dev-tools
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+################################################################################
+# Runs pytest on the repository.
```
fdmalone, line 4: runs pytest on dev_tools right?

check/pytest-dev-tools
```diff
@@ -0,0 +1,26 @@
...
+#
+# Usage:
+#     check/pytest [--flags for pytest] [file-paths-relative-to-repo-root]
```
fdmalone, line 7: fix docstring

fdmalone (changes requested):
LGTM, just a nit about the docstring.

mpharrigan:
ahh I had that commit locally but didn't push it because I didn't want to have to start the CI over :D

fdmalone:
a [skip ci] might slide.

mpharrigan:
is this good now @fdmalone ?

fdmalone (approved):

PR #525, opened 2023-11-20
Cross-checking classical simulation

mpharrigan:
This is a sketch of a function that should exist -- i.e. use `bloq.on_classical_vals` to set up a truth-table tensor network which you can test.


qualtran/bloqs/basic_gates/toffoli.py
```diff
@@ -59,6 +61,40 @@ def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
...
+            incoming['ctrl'][1],
+            outgoing['ctrl'][0],
+            outgoing['ctrl'][1],
+            internal,
+        )
+        CTRL = np.zeros((2,) * 5, dtype=np.complex128)
+        for a, b in itertools.product([0, 1], repeat=2):
+            CTRL[a, b, a, b, int(a == 1 and b == 1)] = 1.0
+
+        # Wire up the CTRL tensor to XOR to flip `target` when active.
```
fdmalone, line 90: Is there an obvious reason why this is called CTRL here and the analogue is COPY for CNOT add_my_tensors?

qualtran/simulation/xcheck_classical_quimb_test.py
```diff
@@ -0,0 +1,49 @@
...
+def test_xgate():
+    x = XGate()
+    x_tt = flank_with_classical_vectors(x, {'q': 0})
+    assert x_tt.tensor_contract() == 1.0
+
+    result = flank_with_classical_vectors(x, in_vals={'q': 0}, out_vals={'q': 0}).tensor_contract()
+    assert result == 0.0
+
+
+def test_cnot():
```
fdmalone, line 30: Is it worth parametrizing these tests over inputs / outputs.

fdmalone (approved):
LGTM. Consider parametrizing the tests.

qualtran/simulation/xcheck_classical_quimb_test.py
mpharrigan: these tests are in the test file associated with the flank function rather than for e.g. cnot. I don't think it would give us enough added coverage to try all the values vs the extra runtime.

qualtran/bloqs/basic_gates/toffoli.py
mpharrigan: You could use the (generalized) COPY tensor with three n-bit indices to implement e.g. unary iteration. 

```
COPY[a, a, a] = 1 for a in range(1,10) else 0
target[a, ...] = list_of_unitaries[a] for a in range(1, 10)
```

fdmalone: interesting

PR #529, opened 2023-11-21
Add resource estimation chem tutorial

fdmalone:
Adds a high level overview / tutorial of the chemistry Hamiltonians and how one might obtain / analyze costs from them. Depends on #520 and #508.

I'm thinking of adding another tutorial on how to add an algorithm (like the naive 2nd quantized chem hamiltonian I allude to in this), as a follow up.


fdmalone:
> while I'm generally a proponent of less formal diction, I think most of the "Ok"s can be removed from the notebook

OK!

mpharrigan (approved):

PR #531, opened 2023-11-21
[autotest] Switch to using exceptions in check_bloq_example implementations 

mpharrigan:
So `qualtran.testing` has these new `check_bloq_example_xxx` functions. They return `BloqCheckResult` enums that have pass and fail; but also na, missing, unverified, and error. This is used for the "bloq report card" where we want more info than a straight pass/fail.

For unit tests, we need a straight pass fail. Previously: the check results were wrapped in `conftest.py` and any bad status would raise an `AssertionError` or `pytest.skip`. Unfortunately, this means the tracebacks in pytest failures would be very unhelpful: they would only trace back to the wrapper function in conftest.py.

This PR refactors the check functions. The implementations now raise an exception for anything other than `PASS`. Unknown errors are passed through. The `check` functions catch these special exceptions to turn them into `BloqCheckResult`. `BloqCheckException` is a special exception type to make this translation easy peasy.

I was motivated to make this change when adding a new check and needing to debug individual instances via the unit tests. And the traceback thing was hindering me.


fdmalone:
Awesome, I was going to raise this issue.

qualtran/testing.py
```diff
@@ -242,17 +242,71 @@ class BloqCheckResult(Enum):
...
+        msg: A message providing details for the exception.
+    """
+
+    def __init__(self, cr: BloqCheckResult, msg: str):
+        super().__init__(cr, msg)
+        self._cr = cr
+        self._msg = msg
+
+    @property
+    def cr(self) -> BloqCheckResult:
```
fdmalone, line 265: cr is a little terse as method / property name no?

fdmalone (approved):
Not a fan of "cr"

mpharrigan:
changed cr -> `check_result`

PR #542, opened 2023-11-28
Fix bug in auto generated `_decompose_with_context_` from `decompose_bloq` in GateWithRegisters

tanujkhattar:
This PR fixes a bug in `GateWithRegisters` class for the case where we auto generate a cirq-style `_decompose_with_context_` implementation when a user provides a Bloq style `decompose_bloq` implementation. Specifically, the bug occurs because the auto generated `_decompose_with_context_`  first computes a composite bloq and then delegates to `cbloq.to_cirq_circuit(qubit_manager=qm, **in_quregs)` to get the corresponding cirq decomposition. This function assumes that all input qubits specified in `in_quregs` are managed by the qubit manager `qm` and thus can get deallocated as part of the decomposition. This was currently not the case. 

Thus, the fix is to mark all the input qubits as "managed" using an `InteropQubitManager` as part of the auto generated `_decompose_with_context_`. See the added tests for more details, which fail on master but pass after changes in the  current PR.


qualtran/_infra/gate_with_registers.py
```diff
@@ -308,17 +308,22 @@ def decompose_from_registers(
...
             return self.decompose_from_registers(context=context, **quregs)
-        except DecomposeNotImplementedError as e:
+        except DecomposeNotImplementedError:
             pass
         try:
-            qm = context.qubit_manager
             cbloq = self.decompose_bloq()
-            circuit, out_quregs = cbloq.to_cirq_circuit(qubit_manager=qm, **quregs)
+            in_quregs = {reg.name: quregs[reg.name] for reg in self.signature.lefts()}
+            # Input qubits can get de-allocated by cbloq.to_cirq_circuit, thus mark them as managed.
```
tanujkhattar, line 323: @mpharrigan this was the issue

mpharrigan: oh, so it was just an accounting issue rather than the accounting finding an actual error?

tanujkhattar: Yes, that's correct. I've added tests so the PR is ready for review now.

mpharrigan (approved):

PR #544, opened 2023-11-29
Nits

mpharrigan:
found during my travels


qualtran/_infra/composite_bloq.py
```diff
@@ -771,7 +771,9 @@ def add_register(
         return None
 
     @classmethod
-    def from_signature(cls, signature: Signature, add_registers_allowed=False):
+    def from_signature(
```
fdmalone, line 774: does mypy / pylint not raise an error without a type hint?

qualtran/_infra/composite_bloq.py
```diff
@@ -771,7 +771,9 @@ def add_register(
...
+        cls, signature: Signature, add_registers_allowed=False
```
fdmalone, line 775: nit on your nits: add_registers_allowed is missing a type hint

fdmalone (changes requested):

qualtran/_infra/composite_bloq.py
mpharrigan: it does not

qualtran/_infra/composite_bloq.py
mpharrigan: done

fdmalone (approved):

PR #545, opened 2023-11-30
Use select and prepare oracles for sparse hamiltonian and add decompositions

fdmalone:
1. Uses the SelectOracle and PrepareOracle for chem/sparse
2. Adds decompositions for both.
3. Adds factory method to build all the alt / keep values from the chemistry integrals.


fdmalone:
Split up into #546  #547 #548

PR #554, opened 2023-12-01
An assortment of generalizers

mpharrigan:
When we're computing call graphs, we can "generalize" bloqs: ignore some or lump others together. This PR provides a library of off-the-shelf generator functions and now supports passing in multiple such functions to the public api.

 - Add a type annotation for the generalizer callable interface
 - helper function to turn many generalizers into one generalizer that calls each of them
 - which is used by the new public api: can provide a sequence of generalizers
 - generalizers module contains common generalizer functions
 - use these common functions elsewhere in the library instead of always re-implementing them

### DIFF
--- a/qualtran/resource_counting/generalizers.py
+++ b/qualtran/resource_counting/generalizers.py
@@ -1,0 +1,110 @@
+#  Copyright 2023 Google LLC
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


qualtran/resource_counting/generalizers.py
```diff
@@ -0,0 +1,114 @@
...
+
+    from qualtran.bloqs.basic_gates import CNOT, Hadamard, TGate, Toffoli, TwoBitSwap, XGate, ZGate
+    from qualtran.bloqs.util_bloqs import ArbitraryClifford
+    from qualtran.cirq_interop import CirqGateAsBloq
+
+    if not isinstance(b, CirqGateAsBloq):
+        return b
+
+    gate = b.gate
+    if gate == cirq.T:
```
fdmalone, line 94: Is there an issue open for the TODOs? I also was wondering about if we need to add an S gate. Will the new controlled protocol remove the need for having CX/Y/Z too?

mpharrigan: Changed to only handle cirq gates for which a perfect match exists in qualtran

fdmalone (approved):
still LGTM

Issue #553, opened 2023-12-01
Reproducing physical costings from literaure with CCZ2T factories


StefanoPolla:
I am trying to reproduce physical costings from literature, e.g. (THC)[https://arxiv.org/abs/2011.03494].

To do this, there are two features that I'm missing in the physical_costing module:
1. the possibility of using multiple magic state factories. In the THC paper, 4 CCZ2T factories are used.
2. an optimization over the factory parameters (code distance in level 1 and 2 distillation), similar to what is done in (`openfermion.resource_estimates.surface_code_compilation.physical_costing.cost_estimator`)[https://github.com/quantumlib/OpenFermion/blob/master/src/openfermion/resource_estimates/surface_code_compilation/physical_costing.py#L171] through the `iter_known_factories` generator.

This issue connects to issue #535 and PR #475, as an extension of `qualtran.surface_code.ccz2T_cost_model`

StefanoPolla:
I have written some code to solve this, I can try to fork and integrate this code. Let me know if I should go on with this.

To solve 1, I wrote an overlay class `MultiCCZ2TFactories(CCZ2TFactory)`, which naively multiplies the footprint and divides the n_cycles by the number of factories used. I could generalize it an overlay of any MagicStateFactory.

fdmalone:
cc @NoureldinYosri who has been developing these things.

mpharrigan:
Ideally it would be `MultiFactory` that works with any sub-factory

mpharrigan:
This is done

PR #555, opened 2023-12-01
[physical costing] Grid search best parameters

mpharrigan:
Addresses #553 

Opening this as a draft as it's not adequately tested or documented. If someone wants to take this up and validate this model against some published references and/or document it I would appreciate it

@NoureldinYosri @StefanoPolla

### DIFF
--- a/qualtran/surface_code/ccz2t_cost_model.py
+++ b/qualtran/surface_code/ccz2t_cost_model.py
@@ -13,7 +13,7 @@
 #  limitations under the License.
 
 import math
-from typing import Optional
+from typing import Callable, Iterator, Optional, Tuple
 
 from attrs import frozen
 
@@ -155,6 +155,27 @@ def get_ccz2t_costs(
     *,
     n_magic: AlgorithmSummary,
     n_algo_qubits: int,
+    phys_err: float,
+    cycle_time_us: float,
+    factory: MagicStateFactory,
+    data_block: DataBlock,
+) -> PhysicalCost:
+    distillation_error = factory.distillation_error(n_magic=n_magic, phys_err=phys_err)
+    n_cycles = factory.n_cycles(n_magic=n_magic)
+    data_error = data_block.data_error(
+        n_algo_qubits=n_algo_qubits, n_cycles=n_cycles, phys_err=phys_err
+    )
+    failure_prob = distillation_error + data_error
+    footprint = factory.footprint() + data_block.footprint(n_algo_qubits=n_algo_qubits)
+    duration_hr = (cycle_time_us * n_cycles) / (1_000_000 * 60 * 60)
+
+    return PhysicalCost(failure_prob=failure_prob, footprint=footprint, duration_hr=duration_hr)
+
+
+def get_ccz2t_costs_from_error_budget(
+    *,
+    n_magic: AlgorithmSummary,
+    n_algo_qubits: int,
     phys_err: float = 1e-3,
     error_budget: Optional[float] = 1e-2,
     cycle_time_us: float = 1.0,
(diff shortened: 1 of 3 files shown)


qualtran/surface_code/ccz2t_cost_model.py
```diff
@@ -204,11 +225,60 @@ def get_ccz2t_costs(
...
+    n_magic: AlgorithmSummary,
+    n_algo_qubits: int,
+    phys_err: float = 1e-3,
+    cycle_time_us: float = 1.0,
+    factory_iter: Callable[[], Iterator] = iter_ccz2t_factories,
+    data_block_iter: Callable[[], Iterator] = iter_simple_data_blocks,
+) -> Tuple[PhysicalCost, CCZ2TFactory, SimpleDataBlock]:
+    """Grid search over parameters to minimize space time volume.
+
+    This is from openfermion. What paper was it introduced or used in?
```
StefanoPolla, line 262: As far as I understand, this search was performed by hand using a tweaked version of the CCZ2T spreadsheet in the [costing of THC (page 25)](https://link.aps.org/doi/10.1103/PRXQuantum.2.030305).

qualtran/surface_code/ccz2t_cost_model_test.py
```diff
@@ -30,3 +34,16 @@ def test_vs_spreadsheet():
     np.testing.assert_allclose(re.failure_prob, 0.0084, rtol=1e-3)
     np.testing.assert_allclose(re.footprint, 4.00e5, rtol=1e-3)
     np.testing.assert_allclose(re.duration_hr, 7.53, rtol=1e-3)
+
+
+def test_grid_search_runs():
+    # TODO: a better test vs some published reference
```
StefanoPolla, line 40: I can add the MultiFactory and attempt to test this against THC. In my previous tries, the numbers are close but not quite match the paper exactly. I'm not sure about other published sources to check against.

qualtran/surface_code/ccz2t_cost_model.py
```diff
@@ -152,6 +152,27 @@ def n_cycles(self, n_magic: AlgorithmSummary) -> int:
 
 
 def get_ccz2t_costs(
```
StefanoPolla, line 154: This seems more general than CCZ2T, should we rename `get_physical_cost` and refactor into the PhysicalCost file?

qualtran/surface_code/ccz2t_cost_model.py
```diff
@@ -204,11 +225,60 @@ def get_ccz2t_costs(
...
+        yield SimpleDataBlock(data_d=logical_data_qubit_distance)
+
+
+def get_ccz2t_costs_from_grid_search(
+    *,
+    n_magic: AlgorithmSummary,
+    n_algo_qubits: int,
+    phys_err: float = 1e-3,
+    cycle_time_us: float = 1.0,
+    factory_iter: Callable[[], Iterator] = iter_ccz2t_factories,
```
StefanoPolla, line 257: Why do we request a generator factory_iter: Callable[[], Iterator] = iter_ccz2t_factories rather than an iterator factory_iter: Iterator = iter_ccz2t_factories()?

qualtran/surface_code/ccz2t_cost_model.py
mpharrigan: Ah great. Kindly add that to the docstring

qualtran/surface_code/ccz2t_cost_model.py
mpharrigan: @NoureldinYosri has been working to generalize the infrastructure to support additional models (including the "Azure" model). What do you think, Nour?

qualtran/surface_code/ccz2t_cost_model.py
mpharrigan: that could work. You'd have to change the way we set the default value since I _think_ calling the generator in the default argument would exhibit some really tricky and unwanted behavior

StefanoPolla: What about allowing both an iterator and a generator? 
i.e. 
```
def get_ccz2t_costs_from_grid_search(
    [...],
    factory_iter: Union[Callable[[], Iterator], Iterator] = iter_ccz2t_factories,
    [...]):
    if isinstance(iter_ccz2t_factories, Callable):
        iter_ccz2t_factories = iter_ccz2t_factories()
```

StefanoPolla: If we like this (or another) solution, I can add it in #561 before finalizing

mpharrigan: Feel free to change it to (just) an iterator. Just make sure the defaults work!

NoureldinYosri: I think `Iterable` is most suitable here since it covers lists, tuples and iterators. The default value should be be `tuple(iter_ccz2t_factories())` however since the default value gets initialized once so the iterator will get consumed during the first call

qualtran/surface_code/ccz2t_cost_model.py
NoureldinYosri: I think we should do it but not now. we currently have two cost models: CCZ2T and the azure model. but they do certain things differently. I don't think having one big method is a good idea.

mpharrigan:
finished and merged in #561

PR #560, opened 2023-12-02
Add quantum-classical addition bloq

fpapa250:
As a part of adding PsiQuantum ECC-256 Primitives from https://arxiv.org/pdf/2306.08585.pdf we must first add a bloq for quantum-classical addition. We do so by simply converting a classical integer k into a quantum parameter using XGates corresponding to the binary representation of the integer. This approach is shown in fig 2a at https://arxiv.org/pdf/2001.09580.pdf.


qualtran/bloqs/arithmetic/addition.py
```diff
@@ -266,6 +268,120 @@ def _add_oop_large() -> OutOfPlaceAdder:
...
+        target_val = args[-1]
+        new_target_val = target_val + self.k
+
+        if self.cvs and args[0] != int(''.join(str(x) for x in self.cvs), 2):
+            new_target_val = target_val
+        ret = (args[0], new_target_val) if self.cvs else (new_target_val,)
+        return ret
+
+    def on_classical_vals(self, *args) -> Dict[str, 'ClassicalValT']:
+        return dict(zip([reg.name for reg in self.signature], self.apply(*args)))
```
mpharrigan, line 320: what is happening here? This should look something like

```python
def on_classical_vals(self, x, **vars):
  if self.cvs:
    ctrl = vars['ctrl']
  else:
     return {'x': x+self.k}

   if ctrl_active:
     return {'ctrl': ctrl, 'x': x+self.k}
  else:
    return {'ctrl': ctrl, 'x': x}
```


i.e. this should be a classical expression of the concept of conditionally adding an integer

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -266,6 +268,120 @@ def _add_oop_large() -> OutOfPlaceAdder:
...
+
+    def build_composite_bloq(self, bb: 'BloqBuilder', **regs: SoquetT) -> Dict[str, 'SoquetT']:
+
+        # Assign registers to variables and allocate ancilla bits for classical integer k.
+        x = regs['x']
```
mpharrigan, line 325: i'd personally stil put `x` in the signature and let `**regs` capture the optional control soquet

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -266,6 +268,120 @@ def _add_oop_large() -> OutOfPlaceAdder:
...
+                    ctrls, k_split[i] = bb.add(
+                        MultiControlPauli(cvs=self.cvs, target_gate=XGate()),
+                        controls=ctrls,
+                        target=k_split[i],
+                    )
+                else:
+                    k_split[i] = bb.add(XGate(), q=k_split[i])
+
+        # Rejoin the qubits representing k for in-place addition.
+        k = bb.join(k_split)
```
mpharrigan, line 354: can you open an issue to track turning this into `IntState`? It's blocked on being able to control it, right?

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -266,6 +268,120 @@ def _add_oop_large() -> OutOfPlaceAdder:
...
+                else:
+                    k_split[i] = bb.add(XGate(), q=k_split[i])
+
+        # Free the ancilla qubits.
+        k = bb.join(k_split)
+        bb.free(k)
+
+        # Return the output registers.
+        if len(self.cvs) > 0:
+            return {'x': x, 'ctrl': ctrls}
```
mpharrigan, line 377: best practice to always list registers in order, so `ctrl` first

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -266,6 +268,120 @@ def _add_oop_large() -> OutOfPlaceAdder:
...
+        else:
+            return {'x': x}
+
+    def short_name(self) -> str:
+        return f'x = x + {self.k}'
```
mpharrigan, line 382: `x += {self.k}`

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -266,6 +268,120 @@ def _add_oop_large() -> OutOfPlaceAdder:
 )
 
 
+@frozen
+class SimpleAddConstant(Bloq):
+    r"""Applies U_{k}|x> = |x + k>.
```
mpharrigan, line 273: I personally don't find it useful to have this `U` in all these different expressions. Suggest something like:

```
"""Takes |x> to |x + k> for a classical integer `k`
```

qualtran/bloqs/arithmetic/addition.py
fpapa250: Will do, but for my own knowledge what is IntState? It has something to do with the controlled bloq protocol, right?

mpharrigan: `qualtran.bloqs.basic_gates.z_basis.IntState` is a bloq that represents allocating a state initialized to an unsigned integer

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -266,6 +269,122 @@ def _add_oop_large() -> OutOfPlaceAdder:
...
+            ctrls = None
+        k = bb.allocate(n=self.bitsize)
+
+        # Get binary representation of k and split k into separate wires.
+        k_split = bb.split(k)
+        if self.signed:
+            binary_rep = list(iter_bits_twos_complement(self.k, self.bitsize))
+        else:
+            binary_rep = list(iter_bits(self.k, self.bitsize))
+        binary_rep.reverse()
```
mpharrigan, line 341: can you speak on this? This is an endianness thing, right? What endianness do you assume for the `x` register? We should make sure this is consistent throughout the library and we should document it somewhere

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -266,6 +269,122 @@ def _add_oop_large() -> OutOfPlaceAdder:
...
+    @cached_property
+    def signature(self) -> 'Signature':
+        if len(self.cvs) > 0:
+            return Signature(
+                [Register('ctrl', bitsize=len(self.cvs)), Register('x', bitsize=self.bitsize)]
+            )
+        else:
+            return Signature([Register('x', bitsize=self.bitsize)])
+
+    def on_classical_vals(
```
mpharrigan, line 311: since this is all classical logic, you should be able to write a test.

```python
for a variety of values of k, x:
  bloq = SimpleAddConstant(k=k)
  ret1 = bloq.call_classically(x=x)
  ret2 = bloq.decompose_bloq().call_classically(x=x)
  assert ret1 == ret2
```

This will help you catch the issues I suspect exist in this method

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -266,6 +269,122 @@ def _add_oop_large() -> OutOfPlaceAdder:
...
+        self, x: 'ClassicalValT', **vars: 'ClassicalValT'
+    ) -> Dict[str, 'ClassicalValT']:
+        if self.cvs:
+            ctrl = vars['ctrl']
+        else:
+            return {'x': x + self.k}
+
+        if np.equal(ctrl, self.cvs):
```
mpharrigan, line 319: In this line, there is an object you wish to be a value but is an array; and another one which is a value but you want it to be an array

`np.equal` returns an array

`ctrl` will be an integer. `Register('x', bitsize=32)` represents a 32-bit integer; `Register('x', bitsize=1, shape=(32,))` represents an array of one-bit bits

qualtran/bloqs/arithmetic/addition.py
fpapa250: Yeah, it's not really ideal, but it is because the functions return big-endian representations, but iterating through the qubits is indexed little-endian so I chose to reverse the output of the bit representation functions. I'll add a comment about this.

mpharrigan:
From communications offline: There's apparently a bug where `np.uint` + python `int` gives you a float. And split/join give uints. https://github.com/numpy/numpy/issues/7126

Do you have a minimal example of where this comes into play? I'm not seeing the issues with the following thing I threw together

```python
@frozen
class SetInt(Bloq):
    bitsize: int
    k: int

    @cached_property
        return Signature([Register('x', bitsize=self.bitsize, side=Side.RIGHT)])

    def on_classical_vals(self) -> Dict[str, 'ClassicalValT']:
        return {'x': 0 + self.k}

    def build_composite_bloq(self, bb: 'BloqBuilder') -> Dict[str, 'SoquetT']:
        k = bb.allocate(n=self.bitsize)

        # Get binary representation of k and split k into separate wires.
        k_split = bb.split(k)
        binary_rep = list(iter_bits(self.k, self.bitsize))

        # Apply XGates to qubits in k where the bitstring has value 1. Apply CNOTs when the gate is
        # controlled.
        for i in range(self.bitsize):
            if binary_rep[i] == 1:
                k_split[i] = bb.add(XGate(), q=k_split[i])

        # Rejoin the qubits representing k for in-place addition.
        k = bb.join(k_split)
        return {'x': k}

def test_set_int():
    si = SetInt(k=5, bitsize=4)
    ret1 = si.call_classically()
    ret2 = si.decompose_bloq().call_classically()
    print()
    print(ret1)
    print(ret2)
```

fpapa250:
> From communications offline: There's apparently a bug where `np.uint` + python `int` gives you a float. And split/join give uints. [numpy/numpy#7126](https://github.com/numpy/numpy/issues/7126)
> 
> Do you have a minimal example of where this comes into play? I'm not seeing the issues with the following thing I threw together
> 
> ```python
> @frozen
> class SetInt(Bloq):
>     bitsize: int
>     k: int
> 
>     @cached_property
>     def signature(self) -> 'Signature':
>         return Signature([Register('x', bitsize=self.bitsize, side=Side.RIGHT)])
> 
>     def on_classical_vals(self) -> Dict[str, 'ClassicalValT']:
>         return {'x': 0 + self.k}
> 
>     def build_composite_bloq(self, bb: 'BloqBuilder') -> Dict[str, 'SoquetT']:
>         k = bb.allocate(n=self.bitsize)
> 
>         # Get binary representation of k and split k into separate wires.
>         k_split = bb.split(k)
>         binary_rep = list(iter_bits(self.k, self.bitsize))
> 
>         # Apply XGates to qubits in k where the bitstring has value 1. Apply CNOTs when the gate is
>         # controlled.
>         for i in range(self.bitsize):
>             if binary_rep[i] == 1:
>                 k_split[i] = bb.add(XGate(), q=k_split[i])
> 
>         # Rejoin the qubits representing k for in-place addition.
>         k = bb.join(k_split)
>         return {'x': k}
> 
> def test_set_int():
>     si = SetInt(k=5, bitsize=4)
>     ret1 = si.call_classically()
>     ret2 = si.decompose_bloq().call_classically()
>     print()
>     print(ret1)
>     print(ret2)
> ```

This specifically occurs when I do a join before an Add() bloq. If you add an Add() bloq after the join the output of the ClassicalValT b (b = a + b) will give you an error as it is a float not an int. I just committed the test which is giving this error and you can try it out yourself.

mpharrigan:
here's the latest, open issue on numpy: https://github.com/numpy/numpy/issues/20905

In the classical simulation protocol, we _do indeed_ want to support both python integers (which are bigint, which is relevant for the numbers used in factoring algorithms) and numpy integers (which can be put into an ndarray).

Reading the linked numpy issues, I guess there is an ambiguity in what you want when you add a unit64 to a bigint. For better or for worse, I think the fix should be to cast `self.k` to a `np.uint` inside the `on_classical_vals` function. We want to match the behavior of the bloq decomposition which itself has to care about overflows and stuff. Honestly: if we wanted to be 100% accurate we'd have to support overflows for arbitrary `bitsize`-bit integers

mpharrigan:
```diff
diff --git a/qualtran/bloqs/arithmetic/addition.py b/qualtran/bloqs/arithmetic/addition.py
index ebcd587a..d5a1edc8 100644
--- a/qualtran/bloqs/arithmetic/addition.py
+++ b/qualtran/bloqs/arithmetic/addition.py
@@ -82,7 +82,8 @@ class Add(GateWithRegisters, cirq.ArithmeticGate):
     def on_classical_vals(
         self, a: 'ClassicalValT', b: 'ClassicalValT'
     ) -> Dict[str, 'ClassicalValT']:
-        return {'a': a, 'b': a + b}
+        assert self.bitsize <= 64 # TODO: be smarter
+        return {'a': a, 'b': np.uint64(a) + np.uint64(b)}
 
     def short_name(self) -> str:
         return "a+b"
```

fpapa250:
Sorry this is so ugly. I rebased it using the commits adding MultiControlX so I could implement classical simulation of the bloq before I go snowboarding this weekend. The commits/merge got a little out of hand.

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -69,7 +82,11 @@ def apply(self, *register_values: int) -> Union[int, Iterable[int]]:
     def on_classical_vals(
         self, a: 'ClassicalValT', b: 'ClassicalValT'
     ) -> Dict[str, 'ClassicalValT']:
-        return {'a': a, 'b': a + b}
+        assert self.bitsize <= 64  # TODO: be smarter
+        return {
+            'a': a,
+            'b': np.uint64(a) + np.uint64(b),
+        }  # TODO: account for signed integer addition
```
mpharrigan, line 89: for todo's please open an issue describing the problem and what a solution would look like and link to it in the todo comment

qualtran/bloqs/arithmetic/addition_test.py
```diff
@@ -233,3 +239,56 @@ def test_out_of_place_adder():
...
+    cbloq_classical = cbloq.call_classically(ctrls=ctrls, x=x)
+
+    assert len(bloq_classical) == len(cbloq_classical)
+    for i in range(len(bloq_classical)):
+        np.testing.assert_array_equal(bloq_classical[i], cbloq_classical[i])
+
+    assert bloq_classical[-1] == result
+
+
+# TODO: write tests for signed integer addition (subtraction)
```
mpharrigan, line 282: open an issue and link it

mpharrigan (approved):
Looks good, but let's track the todos in issues

PR #561, opened 2023-12-04
[physical costing] Grid search best parameters with multiple CCZ factories

StefanoPolla:
This is a continuation of #555. Please let me know if I should rather PR in [`2023-11-grid-search`](https://github.com/quantumlib/Qualtran/tree/2023-11-grid-search) or organize differently.

Here I add:
- `MultiFactory`, an overlay of `MagicStateFactory` that allows to use multiple factories
- Some documentation and additional functionality on `get_ccz2t_costs_from_grid_search`
- A notebook providing a comparison between the grid search results applied to the THC FeMoco costing and the original published results. The plan is to convert this into a test.


mpharrigan:
Thanks for taking this on! Yes, please work on it in this PR and we can close #555

qualtran/surface_code/multi_factory.py
```diff
@@ -0,0 +1,41 @@
...
+from attrs import frozen
+
+
+@frozen
+class MultiFactory(MagicStateFactory):
+    """Overlay of MagicStateFactory representing multiple factories of the same kind.
+
+    Args:
+        base_factory (MagicStateFactory): the base factory to be replicated.
+        n_factories (int): number of factories to construct.
```
mpharrigan, line 27: we typically don't repeat the type in the docstring since we're using type annotations.

qualtran/surface_code/multi_factory.py
```diff
@@ -0,0 +1,41 @@
...
+
+from qualtran.surface_code.algorithm_summary import AlgorithmSummary
+from qualtran.surface_code.magic_state_factory import MagicStateFactory
+
+from attrs import frozen
+
+
+@frozen
+class MultiFactory(MagicStateFactory):
+    """Overlay of MagicStateFactory representing multiple factories of the same kind.
```
mpharrigan, line 23: I'd add a one-sentence paragraph that details that all quantities are just proportional to `base_factory` with a factor of `n_factories`

qualtran/surface_code/multi_factory.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+
+    base_factory: MagicStateFactory
+    n_factories: int
+
+    def footprint(self) -> int:
+        return self.base_factory.footprint() * self.n_factories
+
+    def n_cycles(self, n_magic: AlgorithmSummary) -> int:
+        return self.base_factory.n_cycles(n_magic) // self.n_factories
```
mpharrigan, line 37: hmm, `//` does integer division which is similar to `floor`. I think technically we want `ceil`

qualtran/surface_code/multi_factory.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def distillation_error(self, n_magic: AlgorithmSummary, phys_err: float) -> float:
+        # the distillation error per magic state does not change for different number of factories
```
mpharrigan, line 40: include this note in the class docstring

qualtran/surface_code/thc_compliation.ipynb
```diff
@@ -0,0 +1,189 @@
...
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "distillation error: 0.069%\n",
+      "data error: 0.373%\n",
+      "failure probability: 0.442%\n",
+      "wall time: 4.1369453125 days\n",
+      "footprint: 2.81 million qubits\n"
+     ]
```
mpharrigan, line 71: before we merge, make sure you do `All Cells -> Clear Output` from juypter notebook. We don't commit the notebooks to reduce churn in the diffs

mpharrigan (commented):
looking forward to this, thanks again

mpharrigan:
cc @fdmalone 

I'd say do both things:
- you can use those figures to write a unit test in the `_test.py` file
- I suspect we don't want to completely delete the notebook ... but if you can replace all the code in it with a much simpler call to your new functionality that would be great!

mpharrigan (approved):
Nice!

PR #564, opened 2023-12-06
Factor out generic bloqs for testing

mpharrigan:
There's been a bunch of ad hoc bloqs in various `*_test.py` files. This PR tries to consolodate some of that into `qualtran.bloqs.for_testing`. Whereas other bloqs in the standard library can and will evolve and have their features change -- theses should be used to test qualtran infra and protocols where a unit test wants to make sure things work on a minimal, generic, unchanging bloq.

This includes one- and two- bit atomic (leaf/non-decomposable) bloqs; two bloqs with decompositions in serial and parallel configuration; and the many-register ("complicatedbloq") from the block encoding tests. 

This PR changes the tests and notebooks to use these centralized generic bloqs.

When a specific bloq property is required for a unit test, we leave those bloqs as defined inside the `_test.py` file. Rule of thumb: if you want to use it in two different test modules consider factoring out.

### DIFF
--- a/qualtran/bloqs/util_bloqs_test.py
+++ b/qualtran/bloqs/util_bloqs_test.py
@@ -21,9 +21,9 @@ import pytest
 from attrs import frozen
 
 from qualtran import Bloq, BloqBuilder, Register, Side, Signature, Soquet, SoquetT
-from qualtran._infra.bloq_test import TestCNOT
 from qualtran._infra.gate_with_registers import get_named_qubits
-from qualtran.bloqs.basic_gates import CNOT, Hadamard, XGate
+from qualtran.bloqs.basic_gates import CNOT, XGate
+from qualtran.bloqs.for_testing import TestMultiRegister
 from qualtran.bloqs.util_bloqs import Allocate, Free, Join, Partition, Split
 from qualtran.simulation.classical_sim import _cbloq_call_classically
 from qualtran.simulation.tensor import bloq_to_dense, cbloq_to_quimb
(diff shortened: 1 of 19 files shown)


qualtran/bloqs/util_bloqs_test.py
```diff
@@ -107,30 +84,39 @@ def build_composite_bloq(self, bb: 'BloqBuilder', test_regs: 'SoquetT') -> Dict[
...
+    bloq = TestPartition(test_bloq=CNOT())
     quregs = get_named_qubits(bloq.signature.lefts())
     op, quregs = bloq.as_cirq_op(cirq.ops.SimpleQubitManager(), **quregs)
     unitary = cirq.unitary(cirq.Circuit(op))
     assert np.allclose(unitary, bloq_to_dense(CNOT()))
-    bloq = TestPartition(test_bloq=ComplicatedBloq())
-    quregs = get_named_qubits(bloq.signature.lefts())
-    op, quregs = bloq.as_cirq_op(cirq.ops.SimpleQubitManager(), **quregs)
-    unitary = cirq.unitary(cirq.Circuit(op))
-    assert np.allclose(unitary, bloq_to_dense(bloq))
```
fdmalone, line 133: Where did the unitary test go?

fdmalone (approved):

qualtran/bloqs/util_bloqs_test.py
mpharrigan: the factored out `TestMultiRegister` uses the testing bloqs which don't have cirq equivalents. The first half of this test does a unitary comparison; I added a diagram check, which should validate the `Partition` behavior

PR #565, opened 2023-12-06
Adjoint protocol

mpharrigan:
Given a bloq, we often want an adjoint (often equal to the inverse) version of it.

### Requirements
 - A method that a bloq user can call to get the adjoint version of a bloq
 - Some encapsulation of the default behavior: i.e. by default, we get the adjoint by reversing the order of subbloqs and taking the adjoint of each subbloq.
 - A way for bloq authors to provide specialized adjoint or controlled versions of a bloq. 

### Design
Introduce a new method on Bloq: `Bloq.adjoint()`. By default, it will return `Adjoint(bloq=self)` where the latter is a new special-purpose “metabloq” that takes a bloq as its main attribute.

The Adjoint metabloq delegates all its protocols with modification: The decomposition partially delegates to its member-bloq’s decomposition but will reverse the order of operations and call subbloq.adjoint() on each of the subbloqs.

Bloq authors can override the adjoint() method to provide special constructions of an adjoint operation.

For example: the two-bit And operation has a special decomposition for its adjoint that saves toffolis. This is currently flagged by an attribute on And called “adjoint” that is a boolean. This sort of bloq dispatch will continue to be supported and is appropriate here.

[Note: the existing attribute will have to be renamed to not conflict with the method (follow-on PR)]

 
For the second, continued example: The MultiAnd (multi-bit and) operation should lose its adjoint attribute since it uses the default adjoint behavior. Make a very important note that even though the subbloqs (i.e. And) have a special adjoint, the way they are composed in Adjoint(MultiAnd()) is standard. 

The signature of an adjoint bloq must follow the rules: the same set of registers in the same order except LEFT and RIGHT are switched. This has implications for e.g. Split and Join which need to have their registers renamed to be a mutually agreed-upon name. 

----

this requires #564


qualtran/_infra/adjoint.py
```diff
@@ -0,0 +1,168 @@
...
+    old_i_soqs = [_reg_to_soq(RightDangle, reg) for reg in old_signature.rights()]
+    new_i_soqs = [_reg_to_soq(LeftDangle, reg) for reg in new_signature.lefts()]
+    soq_map: List[Tuple[SoquetT, SoquetT]] = list(zip(old_i_soqs, new_i_soqs))
+
+    # Then we reverse the order of subbloqs
+    bloqnections = reversed(list(cbloq.iter_bloqnections()))
+
+    # And add subbloq.adjoint() back in for each subbloq.
+    bb, _ = BloqBuilder.from_signature(new_signature)
+    for binst, preds, succs in bloqnections:
```
fdmalone, line 67: preds is unused

qualtran/resource_counting/bloq_counts.py
```diff
@@ -164,7 +164,7 @@ def _compute_sigma(root_bloq: Bloq, g: nx.DiGraph) -> Dict[Bloq, Union[int, symp
             for k in callee_sigma.keys():
                 sigma[k] += callee_sigma[k] * n
 
-    return bloq_sigmas[root_bloq]
+    return dict(bloq_sigmas[root_bloq])
```
fdmalone, line 167: Why?

qualtran/_infra/adjoint_test.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def test_build_call_graph():
+    n = sympy.Symbol('_n0')
+    bc = Adjoint(TestBloqWithCallGraph()).bloq_counts()
+    assert bc == {
+        Adjoint(subbloq=TestAtom()): n,
+        Adjoint(subbloq=TestSerialCombo()): 1,
+        Adjoint(subbloq=TestParallelCombo()): 1,
+    }
```
fdmalone, line 122: Do we have a test bloq that requires more than one level of decomposition? E.g. I might be worried about the adjoint propagating all the way down the call graph.

qualtran/cirq_interop/_cirq_to_bloq.py
```diff
@@ -119,6 +119,9 @@ def as_cirq_op(
         qubits = in_quregs.get('q', ()).flatten()
         return self.cirq_gate.on(*qubits), in_quregs
 
+    def adjoint(self) -> 'Bloq':
```
fdmalone, line 122: Do we have a test for cirq-bloq interop with adjoint? If not, I think there should be a unit test tests the unitary / simulated result is correct.

qualtran/_infra/adjoint.py
```diff
@@ -0,0 +1,168 @@
...
+       with its adjoint.
+     - **Cirq Interop**: The default `Bloq` implementation is used, which goes via `BloqAsCirqGate`
+       as usual.
+     - **Wire Symbol**: The wire symbols are the adjoint of `subbloq`'s wire symbols. Namely,
+       left- and right-oriented symbols are flipped.
+     - **Names**: The string names / labels are that of the `subbloq` with a dagger symbol appended.
+
+    Some protocols are impossible to delegate specialized implementations. The `Adjoint` bloq
+    supports the following protocols with "decompose-only" implementations. This means we always
+    go via the bloq's decomposition instead of preferring specialized implementations provided by
```
fdmalone, line 120: To clarify, we go by the Adjoint bloq's decomposition right? Can we add some tests for the adjoint tensor simulations?

qualtran/serialization/bloq.py
```diff
@@ -163,27 +163,26 @@ def bloqs_to_proto(
         _add_bloq_to_dict(bloq, bloq_to_idx)
         _populate_bloq_to_idx(bloq, bloq_to_idx, pred, max_depth)
 
+    # Decompose[..]Error is raised if `bloq` does not have a decomposition.
```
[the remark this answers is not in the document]
mpharrigan: this is from #564 

https://github.com/quantumlib/Qualtran/pull/564/files#r1417984550

qualtran/resource_counting/bloq_counts.py
mpharrigan: I must've ran into this during testing. Otherwise `sigma` is a defaultdict. I think the main consequence of that is the repr is really cluttered

qualtran/cirq_interop/_cirq_to_bloq.py
mpharrigan: The adjoint protocol will break on any bloq that has an attribute named `adjoint`. For reviewing convenience, I split those changes into #566

qualtran/_infra/adjoint.py
mpharrigan: `Adjoint` doesn't override any of the tensor methods, so as long as the decomposition is correct and the tensors-from-decomposition code is correct there's nothing new to test.

I think tensor tests will be merited and valuable when we have some bloqs that support the adjoint protocol #566 . Right now: you'd keep decomposing and wrapping in `Adjoint(...)` until you get to atomic bloqs and an exception will be raised. You need `MyBloq.adjoint()` to sometimes return concrete leaf bloqs

mpharrigan:
Happy to combine this with #566 if you want one big but complete PR

qualtran/cirq_interop/_cirq_to_bloq.py
fdmalone: OH, there's an And here, got you

fdmalone:
> Happy to combine this with #566 if you want one big but complete PR
Eh, no, this is good.

PR #566, opened 2023-12-06
Idiomatic adjoints for bloqs

mpharrigan:
Following #565 

This PR updates bloq to use the new adjoint protocol

 - basic_gates learn their adjoints
 - `adjoint: bool` flag removed from bloqs that do not have special adjoint behavior.
 - flag renamed for those that do


qualtran/Adjoint.ipynb
```diff
@@ -26,7 +26,9 @@
     "from qualtran.drawing import show_bloq\n",
     "from qualtran.bloqs.basic_gates import ZeroState\n",
     "\n",
+    "print(ZeroState())\n",
```
fdmalone, line 29: remove?

qualtran/_infra/gate_with_registers.py
```diff
@@ -309,17 +309,21 @@ def decompose_from_registers(
     def _decompose_with_context_(
         self, qubits: Sequence[cirq.Qid], context: Optional[cirq.DecompositionContext] = None
     ) -> cirq.OP_TREE:
+        from qualtran.cirq_interop._cirq_to_bloq import InteropQubitManager
```
fdmalone, line 312: Why not top level import?

fdmalone (approved):
LGTM. Mostly using the new adjoing + sneaking in a T gate.

qualtran/_infra/gate_with_registers.py
mpharrigan: this is from #542

mpharrigan:
I reverted the changes to some of the comparison gates, tracking in https://github.com/quantumlib/Qualtran/issues/580

qualtran/Adjoint.ipynb
mpharrigan: this is on purpose to show that it turns into "ZeroEffect"

mpharrigan:
Why was this able to be merged even though the tests were failing?

tanujkhattar:
That's because we didn't update the branch protection rules after adding new github actions (for notebooks & pytest-dev-tools) so these tests were not marked as "required". I've updated the branch protections rules so that all the 6 CI actions are now required and this should not happen again.

PR #575, opened 2023-12-07
`on_classical_vals` consistency

mpharrigan:
- add a pylint checker (manual use only, for now)
 - use keyword arguments everywhere
 - some style commits
 - Fix the `on_classical_vals` methods in arithmetic bloqs

Then there's a more controversial commit that lets us use the classical sim protocol on cirq-ft-style bloqs. 

Take a look; The commits should be pretty atomic and I can pull out the less controversial ones; let me know

### DIFF
--- a/qualtran/bloqs/arithmetic/conversions.py
+++ b/qualtran/bloqs/arithmetic/conversions.py
@@ -65,7 +65,7 @@ class ToContiguousIndex(Bloq):
         )
 
     def on_classical_vals(
-        self, mu: 'ClassicalValT', nu: 'ClassicalValT'
+        self, *, mu: 'ClassicalValT', nu: 'ClassicalValT'
     ) -> Dict[str, 'ClassicalValT']:
         return {'mu': mu, 'nu': nu, 's': nu * (nu + 1) // 2 + mu}
 
--- a/qualtran/cirq_interop/_cirq_to_bloq.py
+++ b/qualtran/cirq_interop/_cirq_to_bloq.py
@@ -264,6 +264,25 @@ def _gather_input_soqs(
     return qvars_in
 
 
+def _bloq_from_op(op: cirq.Operation) -> Bloq:
+    """Get the `Bloq` corresponding to `op.gate`."""
+    from qualtran.bloqs.basic_gates import CNOT, XGate, ZGate
+
+    if op.gate is None:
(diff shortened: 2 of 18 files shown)


qualtran/bloqs/arithmetic/conversions.py
```diff
@@ -65,7 +65,7 @@ def signature(self) -> Signature:
         )
 
     def on_classical_vals(
-        self, mu: 'ClassicalValT', nu: 'ClassicalValT'
+        self, *, mu: 'ClassicalValT', nu: 'ClassicalValT'
```
fdmalone, line 68: why are these all key-word only now?

fdmalone:
where's the controversial commit?

qualtran/cirq_interop/_cirq_to_bloq.py
```diff
@@ -264,6 +264,25 @@ def _gather_input_soqs(
     return qvars_in
 
 
+def _bloq_from_op(op: cirq.Operation) -> Bloq:
+    """Get the `Bloq` corresponding to `op.gate`."""
+    from qualtran.bloqs.basic_gates import CNOT, XGate, ZGate
```
fdmalone, line 269: Is there a reason to limiting this to just this subset of basic gates?

fdmalone:
> where's the controversial commit?

nvm, got it

qualtran/cirq_interop/_cirq_to_bloq.py
```diff
@@ -264,6 +264,25 @@ def _gather_input_soqs(
...
+
+    if op.gate is None:
+        raise ValueError(f"Only gate operations are supported, not {op}.")
+
+    if isinstance(op.gate, Bloq):
+        return op.gate
+    if op.gate == cirq.CNOT:
```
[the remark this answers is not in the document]
mpharrigan: `cirq.CNOT` is a particular instance of `CXPowGate` in cirq land

qualtran/bloqs/arithmetic/conversions.py
mpharrigan: Mainly I was having fun with my pylint check.

The system will always call this method with keyword arguments, and users probably shouldn't be calling it directly. The real issue is if you accept position only arguments (including `*args`) it will break

mpharrigan:
pulled out the good stuff into #585

PR #576, opened 2023-12-07
Fixes bug with addition endianness.

fdmalone:
I needed to reverse the qubit order in the cirq decomposition to ensure big-endianness was preserved (the giveaway was the original unit test was reversing the iter_bits list of bits).

### DIFF
--- a/qualtran/bloqs/arithmetic/addition.py
+++ b/qualtran/bloqs/arithmetic/addition.py
@@ -66,8 +66,10 @@ class Add(GateWithRegisters, cirq.ArithmeticGate):
         a, b = register_values
         return a, a + b
 
-    def on_classical_vals(self, *args) -> Dict[str, 'ClassicalValT']:
-        return dict(zip([reg.name for reg in self.signature], self.apply(*args)))
+    def on_classical_vals(
+        self, a: 'ClassicalValT', b: 'ClassicalValT'
+    ) -> Dict[str, 'ClassicalValT']:
+        return {'a': a, 'b': a + b}
 
     def short_name(self) -> str:
         return "a+b"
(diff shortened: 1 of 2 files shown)


fdmalone:
We (I) may want to write this decomposition in bloqs so we can do the helpful decompose on classical vals test too!

fdmalone:
@mpharrigan I think you committed test-revealing-bug in #575  not sure if deliberate. But this PR should fix it.

mpharrigan:
cc @fpapa250 who kicked off this journey of discovery. Should unblock #560 

> deliberate

yes, it was

> helpful decompose on classical vals

You can do this with the cirq-ft style decomposition except for the fact that we're using `cirq.CNOT` instead of the `CNOT` block; so it doesn't support classical simulation

---

In cirq world, it should still be theoretically possible to write a test that apply unitary from `ArithmeticGate.apply` gives the same result as apply_unitary_from_decomposition. Both these code paths were exercised in the existing tests but their answers were never compared.

mpharrigan:
please add this additional test

```patch
diff --git a/qualtran/bloqs/arithmetic/addition_test.py b/qualtran/bloqs/arithmetic/addition_test.py
index 0bfb2120..f69b7927 100644
--- a/qualtran/bloqs/arithmetic/addition_test.py
+++ b/qualtran/bloqs/arithmetic/addition_test.py
@@ -38,6 +38,7 @@ def test_add_decomposition(a: int, b: int, num_bits: int):
     op = gate.on_registers(a=qubits[:num_bits], b=qubits[num_bits:])
     greedy_mm = cirq.GreedyQubitManager(prefix="_a", maximize_reuse=True)
     context = cirq.DecompositionContext(greedy_mm)
+    circuit0 = cirq.Circuit(op)
     circuit = cirq.Circuit(cirq.decompose_once(op, context=context))
     ancillas = sorted(circuit.all_qubits())[-num_anc:]
     initial_state = [0] * (2 * num_bits + num_anc)
@@ -47,6 +48,7 @@ def test_add_decomposition(a: int, b: int, num_bits: int):
     final_state[:num_bits] = list(iter_bits(a, num_bits))[::-1]
     final_state[num_bits : 2 * num_bits] = list(iter_bits(a + b, num_bits))[::-1]
     assert_circuit_inp_out_cirqsim(circuit, qubits + ancillas, initial_state, final_state)
+    assert_circuit_inp_out_cirqsim(circuit0, qubits + ancillas, initial_state, final_state)
     # Test diagrams
     expected_wire_symbols = ("In(x)",) * num_bits + ("In(y)/Out(x+y)",) * num_bits
     assert cirq.circuit_diagram_info(gate).wire_symbols == expected_wire_symbols
```

fdmalone:
this is already there no?

mpharrigan:
let's merge as is and figure out how best to get that working later

Issue #556, opened 2023-12-02
Bug in SelectSwapQROM when the number of datasets >= 10


fdmalone:
If we have more than 10 data sets to load then some of the target registers can overlap with the named qubits (I think this is the problem anyway.)

``` python
from qualtran.bloqs.select_swap_qrom import SelectSwapQROM, find_optimal_log_block_size

target_bitsizes = (3,)*10 + (1,)*2 + (3,)
block_size = 2 ** find_optimal_log_block_size(10, sum(target_bitsizes))
qrom = SelectSwapQROM(
    (1,)*10,
    (1,)*10,
    (1,)*10,
    (1,)*10,
    (1,)*10,
    (1,)*10,
    (1,)*10,
    (1,)*10,
    (1,)*10,
    (1,)*10,
    (0,)*10,
    (1,)*10,
    (3,)*10,
    target_bitsizes=target_bitsizes,
    block_size=block_size,
)
qrom.call_graph()
```

```
[...]

PR #577, opened 2023-12-15
Rename qro(a)m target registers to avoid name collision with named qubits

fdmalone:
If the number of data sets is >= 10 collisions could occur (see #556).


fdmalone:
also a hot fix for a flaky THC test.

qualtran/bloqs/select_swap_qrom.py
```diff
@@ -151,7 +151,7 @@ def selection_registers(self) -> Tuple[SelectionRegister, ...]:
     @cached_property
     def target_registers(self) -> Tuple[Register, ...]:
         return tuple(
-            Register(f'target{sequence_id}', self._target_bitsizes[sequence_id])
+            Register(f'target{sequence_id}_', self._target_bitsizes[sequence_id])
```
mpharrigan, line 154: add a comment explaining why the underscore is necessary and link to issue

mpharrigan (approved):
let's leave the issue open so we don't forget about this kludge

qualtran/bloqs/select_swap_qrom.py
fdmalone: Done.

PR #578, opened 2023-12-16
[autodoc] Chemistry algorithms

fdmalone:
Adds autodoc / autotest for missing chemistry algorithms (first quantization and trotter). Not sure what to do with all the trotter docs given that the original trotter.ipynb is more of an exposition.


mpharrigan:
sweeeeeeeet. 

Can you please split the trotter exposition vs the trotter bloqdocs into separate notebooks?

fdmalone:
@mpharrigan should be done now along with THC autodoc added to -v2.

mpharrigan (approved):

PR #581, opened 2024-01-02
GQSP: Implement complementary polynomial computation

anurudhp:
theorem 4 of https://arxiv.org/abs/2308.01501

Still some debugging left to do:
- [x] Computed Q satisfies the definitional constraints, but still produces incorrect GQSP circuits
- [x] computing a polynomial from roots is inaccurate even for degree 30 or so. TODO check if this causes a problem.

### DIFF
--- a/qualtran/bloqs/generalized_qsp.py
+++ b/qualtran/bloqs/generalized_qsp.py
@@ -92,9 +92,9 @@ def qsp_complementary_polynomial(
 
     The exact method for computing $Q$ is described in the proof of Theorem 4.
     The method computes an auxillary polynomial R, whose roots are computed
-    and re-interpolated the obtain the required polynomial Q.
+    and re-interpolated to obtain the required polynomial Q.
 
-    TODO Also implement the more efficient optimization-based method described in Eq. 52
+    TODO: Also implement the more efficient optimization-based method described in Eq. 52
 
     Args:
         P: Co-efficients of a complex polynomial.
@@ -151,10 +151,6 @@ def qsp_complementary_polynomial(
 
         assert is_permutation(smaller_roots, 1 / np.array(larger_roots).conj())
 
-    # Leading co-efficient of R described in Eq. 37.
-    c = R.coef[-1]
-    scaling_factor = np.sqrt(np.abs(c * np.prod(larger_roots)))
-
     # pair up roots in `units`, claimed in Eq. 40 and the explanation preceding it.
     # all unit roots must have even multiplicity.
     paired_units: list[complex] = []
@@ -179,6 +175,13 @@ def qsp_complementary_polynomial(
     # - \hat{G}^2 is the monomials which are unit roots of R, which occur in pairs.
     # - G*(z) G(z) is the interpolation of the conjugate paired non-unit roots of R,
     #   described in Eq. 37 - Eq. 38
+
+    # Leading co-efficient of R described in Eq. 37.
+    # Note: In the paper, the polynomial is interpolated from larger_roots,
+    #       but this is swapped in our implementation to reduce the error in Q.
+    c = R.coef[-1]
+    scaling_factor = np.sqrt(np.abs(c * np.prod(larger_roots)))
+
     Q = scaling_factor * Polynomial.fromroots(paired_units + smaller_roots)
 
     return np.around(Q.coef, decimals=10)
--- a/qualtran/bloqs/generalized_qsp.py
+++ b/qualtran/bloqs/generalized_qsp.py
@@ -106,20 +106,19 @@ def qsp_complementary_polynomial(
     """
     d = len(P) - 1  # degree
 
-    r"""R(z) = z^d (1 - P^*(z) P(z))
-    obtained on simplifying Eq. 34, Eq. 35 by substituting H, T.
-    The definition of $R$ is given in the text from Eq.34 - Eq. 35,
-    following the chain of definitions below:
-
-        $$
-        T(\theta) = \abs{P(e^{i\theta}),
-        H = I - T,
-        H = e^{-id\theta} R(e^{i\theta})
-        $$
-        
-    Here H and T are defined on reals, so the initial definition of R is only on the unit circle.
-    We analytically continue this definition to the entire complex plane by replacing $e^{i\theta}$ by $z$.
-    """
+    # R(z) = z^d (1 - P^*(z) P(z))
+    # obtained on simplifying Eq. 34, Eq. 35 by substituting H, T.
+    # The definition of $R$ is given in the text from Eq.34 - Eq. 35,
+    # following the chain of definitions below:
+    #
+    #     $$
+    #     T(\theta) = \abs{P(e^{i\theta}),
+    #     H = I - T,
+    #     H = e^{-id\theta} R(e^{i\theta})
+    #     $$
+    #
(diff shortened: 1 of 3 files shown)


anurudhp:
After some debugging, I realized why complex polynomials were producing incorrect circuits:
https://github.com/quantumlib/Qualtran/blob/7c9f3a0af0705e8bf72e0297c7cd2ce2c9b94628/qualtran/bloqs/generalized_qsp.py#L120

the $\phi_j$ are computed as $arg(a / b)$, which is numerically unstable around 0 - really small non-zero values can have arbitrary angles. This isn't an issue for polynomials with real coefficients though, as thetas are computed correctly.

qualtran/bloqs/generalized_qsp.py
```diff
@@ -79,6 +79,88 @@ def decompose_from_registers(
...
+) -> Sequence[complex]:
+    r"""Computes the Q polynomial given P
+
+    Computes polynomial $Q$ of degree at-most that of $P$, satisfying
+
+        $$ \abs{P(e^{i\theta})}^2 + \abs{Q(e^{i\theta})}^2 = 1 $$
+
+    for every $\theta \in \mathbb{R}$.
+
+    The exact method for computing $Q$ is described in the proof of Theorem 4.
```
tanujkhattar, line 93: Maybe add a comment saying that this function computes the complementary polynomial using root finding following the ideas laid out in Theorem 4. But there's also a cheaper optimization based approach described in Eq 52

qualtran/bloqs/generalized_qsp.py
```diff
@@ -115,9 +197,9 @@ def qsp_phase_factors(
     for d in reversed(range(n)):
         assert S.shape == (2, d + 1)
 
-        a, b = np.around(S[:, d], decimals=10)
+        a, b = S[:, d]
         theta[d] = np.arctan2(np.abs(b), np.abs(a))
-        phi[d] = (np.angle(a) - np.angle(b)) % (2 * np.pi)
+        phi[d] = np.angle(a) - np.angle(b)
```
tanujkhattar, line 202: Can you add a comment with a TODO that highlights the numerical instability for complex polynomials and what causes the numerical instability? (or link to a github issue that describes the problem)

qualtran/bloqs/generalized_qsp.py
```diff
@@ -79,6 +79,88 @@ def decompose_from_registers(
...
+
+    # R(z) = z^d (1 - P^*(z) P(z)) (Eq. 34, 35)
+    R = Polynomial.basis(d) - Polynomial(P) * Polynomial(np.conj(P[::-1]))
+    roots = R.roots()
+
+    units: list[complex] = []  # roots r s.t. \abs{r} = 1
+    larger_roots: list[complex] = []  # roots r s.t. \abs{r} > 1
+    smaller_roots: list[complex] = []  # roots r s.t. \abs{r} < 1
+
+    for r in roots:
```
tanujkhattar, line 113: It'd be great if you can add more comments throughout the implementation, mapping it back to the proof of Theorem 4 since I think there are a few differences in the explanation and your implementation? More comments to make it easier for readers to follow

tanujkhattar (approved):
Looking good overall. Request to add more comments and explanations for the implementation.

qualtran/bloqs/generalized_qsp.py
anurudhp: added a comment before `phi[d]`

qualtran/bloqs/generalized_qsp.py
anurudhp: I added comments inline explaining the mapping and respective eqs/paras in the paper. Could you see if something is missing?

qualtran/bloqs/generalized_qsp.py
anurudhp: added a todo note for eq. 52

qualtran/bloqs/generalized_qsp.py
```diff
@@ -79,6 +79,111 @@ def decompose_from_registers(
...
+            unpaired_units.append(z)
+
+    if verify:
+        assert len(unpaired_units) == 0
+
+    # Q = G \hat{G}, where
+    # - \hat{G}^2 is the monomials which are unit roots of R, which occur in pairs.
+    # - G*(z) G(z) is the interpolation of the conjugate paired non-unit roots of R,
+    #   described in Eq. 37 - Eq. 38
+    Q = scaling_factor * Polynomial.fromroots(paired_units + smaller_roots)
```
tanujkhattar, line 182: Maybe a trivial question - Is my understanding correct that in the paper they seem to define `Q` with the opposite usage of `smaller_roots` and `larger_roots` ?  Something like

```python
scaling_factor = np.sqrt(np.abs(c * np.prod(smaller_roots)))
Q = scaling_factor * Polynomial.fromroots(paired_units + larger_roots)
```

This follows from Eq 37
image
where the scaling factor is 

$$
G = \sqrt{\prod_{k=1}^{m}\frac{1}{w_{k}^*}} \prod_{k=1}^{m} ||x - w_{k}||
$$

It would probably not matter but just want to make sure that I'm understanding this correct.

tanujkhattar (approved):
Looks great now! I've added one doubt but feel free to merge after responding to my comment. Thanks!

qualtran/bloqs/generalized_qsp.py
anurudhp: Ah yes, that is true. I was debugging the issue of the polynomial being too large when substituting its zeroes, so I swapped the two around hoping it'll reduce the error.  I just tried swapping the two around, and it fails for degree 4.

qualtran/bloqs/generalized_qsp.py
```diff
@@ -79,6 +79,114 @@ def decompose_from_registers(
...
+
+    # Leading co-efficient of R described in Eq. 37.
+    # Note: In the paper, the polynomial is interpolated from larger_roots,
+    #       but this is swapped in our implementation to reduce the error in Q.
+    c = R.coef[-1]
+    scaling_factor = np.sqrt(np.abs(c * np.prod(larger_roots)))
+
+    Q = scaling_factor * Polynomial.fromroots(paired_units + smaller_roots)
+
+    return np.around(Q.coef, decimals=10)
```
tanujkhattar, line 187: Why the magic number `10` ? Does this have any impact on potential loss of precision?

anurudhp: actually this isn't needed, I removed it now.

But the current tests only work up till degree 4, the errors from degree 5 are quite large already. I plan to look into this next.

tanujkhattar: >  the errors from degree 5 are quite large already. I plan to look into this next.

So swapping smaller_roots and larger_roots doesn't help? Or it works till degree 4 because of reduced error due to swapping the use of two ?

anurudhp: Btw, the complementary polynomial satisfies the required constraint $|P(z)|^2 + |Q(z)|^2 = 1$ even for degree 100, so the error seems to blow-up in the QSP phase computation.

anurudhp: @tanujkhattar thoughts?

tanujkhattar: Looks good, happy to merge now and debug the precision issue in follow up PRs.

PR #585, opened 2024-01-05
More support for classical simulation

mpharrigan:
- use idiomatic `on_classical_val` methods in arithmetic
 - support some more uncomputations


mpharrigan:
This contains the non-superfluous changes that originated from #575

fdmalone (dismissed):

mpharrigan:
ah, right. Now I know why I expanded the scope in #575. If `cirq.CX` is used, we can't support classical simulation.

fdmalone:
What's the status with this one? I have some vested interested in the interception of the cirq basic gates in favor of bloqs :)

Screenshot 2024-01-26 at 2 20 48 PM

mpharrigan:
I think the quote-unquote right way forward is to:

 1. make sure we have bloqs in this library for everything including CZ
 2. following #591 , we can replace instances of `cirq.CNOT` with our CNOT bloq 

We'd do step 2 for all cases where we need to have control over the protocols: e.g. in this PR I want to add classical simulation to CNOT which I can't do to `cirq.CNOT` and you want to add wire symbols to `cirq.CZ`, which honestly should work via the cirq interop and I don't know why it's not translating the `@` but nonetheless we should turn it into a CZ bloq

mpharrigan:
I changed `Add` to a `Bloq` and used the `CNOT` bloq in place of `cirq.CNOT`, which unblocked the classical test. I did light modifications to the other `Add` tests to use bloq idioms.

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -69,15 +73,26 @@ class Add(GateWithRegisters, cirq.ArithmeticGate):
...
+        incoming: Dict[str, 'SoquetT'],
+        outgoing: Dict[str, 'SoquetT'],
+    ):
+        import quimb.tensor as qtn
+
+        N = 2**self.bitsize
+        inds = (incoming['a'], incoming['b'], outgoing['a'], outgoing['b'])
+        unitary = np.zeros((N,) * len(inds), dtype=np.complex128)
+        for a, b in itertools.product(range(N), range(N)):
+            unitary[a, b, a, (a + b) % N] = 1
```
fdmalone, line 90: Why is this % N

mpharrigan: otherwise it would be out of range

mpharrigan:
Tests failing because `Add(bitsize=1)` used to work ...

qualtran/bloqs/arithmetic/addition.py
fdmalone: Do you know what happens with overflow? I'm sort of curious.

mpharrigan: The tests are written to never trigger overflow, which is something to say. 

looking at `cirq.ArithmeticGate` line 219 seems to imply that `% N` is used, which was the previous behavior.

mpharrigan:
I factored out #623

Issue #523, opened 2023-11-18
Use of `eval` in proto deserialization


mpharrigan:
`_ndarray_from_proto` uses `eval` on wire input. This seems like a bad idea.

mpharrigan:
@tanujkhattar can you take a look at this

tanujkhattar:
Looking at it now

PR #588, opened 2024-01-09
Update NDArray proto and serialization to avoid running eval on serialized code

tanujkhattar:
Fixes https://github.com/quantumlib/Qualtran/issues/523

Note that this is backwards incompatible change but it's fine because we currently don't make any backwards compatibility guarantees.


mpharrigan:
Much nicer. Thanks. 

Yes, worth re-iterating that at some point, backwards compatibility in the serialization format will be very important but that is not the case yet

mpharrigan (approved):

PR #591, opened 2024-01-17
Cirq interface opt-in

mpharrigan:
The Cirq-interop code is high quality and highly refactorable. It’s currently configured as a [kitchen sink](https://www.merriam-webster.com/dictionary/everything%20but%20the%20kitchen%20sink). This change reconfigures some of the defaults.

This PR adds support for writing "cirq-style" decompositions with qubits and bloq.on(*qubits) for all bloqs. This is an informationally-complete way of defining decompositions that may be 1) more familiar to cirq people 2) more convenient if you’re wiring up a bunch of thru-registers and don’t want to redundantly write down inputs and outputs everywhere. `Bloq` learns `on` and `on_registers` method so that all bloqs can be wired up this way.

`BloqAsCirqGate` loses support for the bloq api. It’s just a Cirq gate and doesn’t implement the Bloq API. It’s confusing for it to be both. Note: `CirqGateAsBloq` is still both since it didn't confuse me when debugging this refactor but could be disentangled in a follow-on. 

GateWithRegisters that don’t implement any special cirq protocols can migrate to Bloq. They can still be wired up using Bloq.on(). As an example, this PR changes `MultiAnd` to a `Bloq` with no other modifications to the library required.

--------

Specific changes:
 - `Bloq.on` and `Bloq.on_registers` wrap in `BloqAsCirqGate` which now gets unwrapped by `cirq_optree_to_cbloq`. 
 - `GateWithRegisters` is unchanged apart from factoring out code into helper functions.
 - `BloqAsCirqGate` uses those helper functions

### DIFF
--- a/qualtran/_infra/bloq.py
+++ b/qualtran/_infra/bloq.py
@@ -369,6 +369,55 @@ class Bloq(metaclass=abc.ABCMeta):
             bloq=self, cirq_quregs=cirq_quregs, qubit_manager=qubit_manager
         )
 
+    def on(self, *qubits: 'cirq.Qid') -> 'cirq.Operation':
+        """A `cirq.Operation` of this bloq operating on the given qubits.
+
+        This method supports an alternative decomposition backend that follows a 'Cirq-style'
+        association of gates with qubits to form operations. Instead of wiring up `Soquet`s,
+        each gate operates on qubit addresses (`cirq.Qid`s), which are reused by multiple
+        gates. This method lets you operate this bloq on qubits and returns a `cirq.Operation`.
+
+        The primary, bloq-native way of writing decompositions is to override
+        `build_composite_bloq`. If this is what you're doing, do not use this method.
+
+        To provide a Cirq-style decomposition for this bloq, implement a method (typically named
+        `decompose_from_registers` for historical reasons) that yields a list of `cirq.Operation`s
+        using `cirq.Gate.on(...)`, `Bloq.on(...)`, `GateWithRegisters.on_registers(...)`, or
+        `Bloq.on_registers(...)`.
+
+        See Also:
+            `Bloq.on_registers`: Provides the same functionality, but with named registers
+                instead of a flat list of qubits.
+            `decompose_from_decompose_from_registers`: More details on how to write a cirq-style
+                decomposition.
+        """
+        import cirq
+
+        from qualtran.cirq_interop import BloqAsCirqGate
+
+        return cirq.Gate.on(BloqAsCirqGate(bloq=self), *qubits)
+
+    def on_registers(
+        self, **qubit_regs: Union['cirq.Qid', Sequence['cirq.Qid'], 'NDArray[cirq.Qid]']
+    ) -> 'cirq.Operation':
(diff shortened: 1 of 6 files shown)


qualtran/_infra/bloq.py
```diff
@@ -369,6 +369,55 @@ def as_cirq_op(
...
+        This method supports an alternative decomposition backend that follows a 'Cirq-style'
+        association of gates with qubits to form operations. See `Bloq.on()` for more details.
+
+        Args:
+            **qubit_regs: A mapping of register name to the qubits comprising that register.
+
+        See Also:
+            `Bloq.on`: Provides the same functionality, but with a flat list of qubits.
+                instead of named registers.
+            `decompose_from_decompose_from_registers`: More details on how to write a cirq-style
```
tanujkhattar, line 414: `decompose_from_decompose_from_registers` - looks like a typo? Or is `decompose_from_decompose_from` expected and doesn't currently exist?

mpharrigan: ah this was an old name for what I have named `decompose_from_cirq_style_method`

qualtran/bloqs/and_bloq.py
```diff
@@ -291,6 +294,9 @@ def decompose_from_registers(
         )
         yield self._decompose_via_tree(control, self.cvs, ancilla, *target)
 
+    def decompose_bloq(self) -> 'CompositeBloq':
+        return decompose_from_cirq_style_method(self)
```
mpharrigan, line 298: @tanujkhattar the idea is to explicitly flag that we're using a cirq style `decompose_from_registers` method by overriding `decompose_bloq` with this non-default way of doing a decomposition. You can actually name the decomposition method whatever you want and specify its name as the second argument. There's a variety of alternatives centering around a) plumbing through the method b) defaults.

I picked this way to keep the `Bloq` abstract interface streamlined, with the happy path clearly highlighted. The explicit delegation with `decompose_from_cirq_style_method` removes a category of headaches when debugging a decomposition.

qualtran/cirq_interop/_cirq_to_bloq.py
```diff
@@ -280,6 +285,29 @@ def _gather_input_soqs(
...
+    true cirq gates with `CirqGateAsBloq`.
+    """
+    from qualtran.cirq_interop._bloq_to_cirq import BloqAsCirqGate
+
+    if op.gate is None:
+        raise ValueError(f"Only gate operations are supported, not {op}.")
+
+    gate = op.gate
+    if isinstance(gate, BloqAsCirqGate):
+        # Perhaps this was operation was constructed from `Bloq.on()`.
```
tanujkhattar, line 301:
```suggestion
        # Perhaps this operation was constructed from `Bloq.on()`.
```

tanujkhattar (approved):

qualtran/cirq_interop/_cirq_to_bloq.py
mpharrigan: fixed offline, thanks

qualtran/_infra/bloq.py
mpharrigan: fixed, thanks!

PR #592, opened 2024-01-18
Add MultiControlX() bloq with functional classical simulation

fpapa250:
Adds MultiControlX() bloq which is similar to MultiControlPauli(), but it is a bloq, has a classical simulation, and does not use any And() gates. It allows for 0-controlled operations. It also decomposes into the XGate(), CNOT(), and Toffoli() when the number of controls matches those bloqs.


qualtran/bloqs/multi_control_multi_target_pauli.py
```diff
@@ -137,6 +140,160 @@ def _has_unitary_(self) -> bool:
...
+    References:
+        [Constructing Large Controlled Nots]
+        (https://algassert.com/circuits/2015/06/05/Constructing-Large-Controlled-Nots.html)
+    """
+
+    cvs: Tuple[int, ...] = field(converter=lambda v: (v,) if isinstance(v, int) else tuple(v))
+
+    @cached_property
+    def signature(self) -> 'Signature':
+        if len(self.cvs) > 0:
```
mpharrigan, line 156: just philosophically, I think we should mandate that `cvs` must be greater than 0

qualtran/bloqs/multi_control_multi_target_pauli.py
```diff
@@ -137,6 +140,160 @@ def _has_unitary_(self) -> bool:
         return True
 
 
+@frozen
+class MultiControlX(Bloq):
+    r"""Implements multi-control, single-target X gate as a bloq using $n-1$ clean ancillas.
+
+    References:
+        [Constructing Large Controlled Nots]
+        (https://algassert.com/circuits/2015/06/05/Constructing-Large-Controlled-Nots.html)
```
mpharrigan, line 149: what's the relevant section of this? The final construction looks like it uses 1 ancilla but above you say n-1 ancilla

qualtran/bloqs/multi_control_multi_target_pauli.py
```diff
@@ -137,6 +140,160 @@ def _has_unitary_(self) -> bool:
...
+
+    def on_classical_vals(
+        self, x: 'ClassicalValT', **vars: 'ClassicalValT'
+    ) -> Dict[str, 'ClassicalValT']:
+        if len(self.cvs) > 0:
+            ctrl = vars['ctrl']
+        else:
+            return {'x': (x + 1) % 2}
+
+        ctrls = bits_to_ints(self.cvs)
```
mpharrigan, line 169: Document the `cvs` attribute. They're supposed to be 0s and 1s, right? i.e. bits

qualtran/bloqs/multi_control_multi_target_pauli.py
```diff
@@ -137,6 +140,160 @@ def _has_unitary_(self) -> bool:
...
+        if ctrls == ctrl:
```
mpharrigan, line 170: here's another philosophical thing: we support both

```python
Register('ctrl', bitsize=5)
Register('ctrl', bitsize=1, shape=(5,)
```

the former is one 5-bit-wide wire whereas the latter is a 1d array of 1-bit wires. If you want `cvs` to be a sequence of bits and you want this to be called `MultiControlledX` then you should probably use the second construction (and remove the `bits_to_ints` call inside this method). If you want `cvs` to be one control integer, use the first one, call it `IntControlledX`, and remove the `bits_to_ints` call here.

qualtran/bloqs/multi_control_multi_target_pauli.py
```diff
@@ -137,6 +140,160 @@ def _has_unitary_(self) -> bool:
...
+                ctrl = bb.add(XGate(), q=ctrl)
+            ctrl, x = bb.add(CNOT(), ctrl=ctrl, target=x)
+            if self.cvs[0] == 0:
+                ctrl = bb.add(XGate(), q=ctrl)
+            return {'ctrl': ctrl, 'x': x}
+
+        # Base case 2: Toffoli()
+        if n == 2:
+            ctrls = regs['ctrl']
+            ctrls_split = bb.split(ctrls)
```
mpharrigan, line 201: the soquets will come 'pre-split' if you use a `bitsize=1, shape=(n,)` register instead

qualtran/bloqs/multi_control_multi_target_pauli.py
```diff
@@ -137,6 +140,160 @@ def _has_unitary_(self) -> bool:
...
+
+            # Allows for 0-controlled implementations.
+            for i in range(len(self.cvs)):
+                if self.cvs[i] == 0:
+                    ctrls_split[i] = bb.add(XGate(), q=ctrls_split[i])
+
+            ctrl = [ctrls_split[0], ctrls_split[1]]
+            ctrl, x = bb.add(Toffoli(), ctrl=ctrl, target=x)
```
mpharrigan, line 209: `ctrls_split` is already a length-2 1d array, right? you can just pass that in to `bb.add(... ctrl=ctrls_split)`

qualtran/bloqs/multi_control_multi_target_pauli.py
```diff
@@ -137,6 +140,160 @@ def _has_unitary_(self) -> bool:
...
+            ctrls_split[0] = ctrl[0]
+            ctrls_split[1] = ctrl[1]
```
mpharrigan, line 211: likewise you can just name the first return value `ctrls_split`

qualtran/bloqs/multi_control_multi_target_pauli.py
```diff
@@ -137,6 +140,160 @@ def _has_unitary_(self) -> bool:
...
+        bb.free(ancillas)
+
+        # Return the output registers.
+        if n > 0:
+            return {'ctrl': ctrls, 'x': x}
+        else:
+            return {'x': x}
+
+    def short_name(self) -> str:
+        return f'C^{len(self.cvs)}-NOT(x)'
```
mpharrigan, line 295: probably don't need the `'(x)'` in the short name to make it shorter

qualtran/bloqs/multi_control_multi_target_pauli_test.py
```diff
@@ -41,3 +45,39 @@ def test_t_complexity_mcp(num_controls: int, pauli: cirq.Pauli, cv: int):
...
+
+
+@pytest.mark.parametrize("cvs", [(), (0), (1, 0), (1, 1, 1), (1, 0, 1, 0)])
+def test_multi_control_x(cvs):
+    bloq = MultiControlX(cvs=cvs)
+    assert_valid_bloq_decomposition(bloq=bloq)
+
+
+@pytest.mark.parametrize(
+    "cvs,x,ctrl",
```
mpharrigan, line 57: this doesn't match the order of the names you pass in to the actual function (??)

qualtran/bloqs/multi_control_multi_target_pauli_test.py
```diff
@@ -41,3 +45,39 @@ def test_t_complexity_mcp(num_controls: int, pauli: cirq.Pauli, cv: int):
...
+    [((), 0, 0), ((0), 1, 0), ((1, 0), 0, 2), ((1, 1, 1), 1, 7), ((1, 0, 1, 0), 1, 10)],
```
mpharrigan, line 58: `(0)` in python is just the integer 0. If you want a tuple you need to `(0,)`

qualtran/bloqs/multi_control_multi_target_pauli_test.py
```diff
@@ -41,3 +45,39 @@ def test_t_complexity_mcp(num_controls: int, pauli: cirq.Pauli, cv: int):
...
+def test_classical_multi_control_x(x, cvs, ctrl):
+    bloq = MultiControlX(cvs=cvs)
+    cbloq = bloq.decompose_bloq()
+    ret1 = bloq.call_classically(x=x, ctrl=ctrl)
+    ret2 = cbloq.call_classically(x=x, ctrl=ctrl)
+    assert ret1 == ret2
+
+
+@pytest.mark.parametrize(
+    "cvs,x,ctrl,result",
```
mpharrigan, line 69: this doesn't match either

qualtran/bloqs/multi_control_multi_target_pauli_test.py
```diff
@@ -41,3 +45,39 @@ def test_t_complexity_mcp(num_controls: int, pauli: cirq.Pauli, cv: int):
...
+        ((1, 1, 1), 1, 7, 0),
+        ((1, 0, 1, 0), 1, 10, 0),
+        ((1), 0, 0, 0),
+    ],
+)
+def test_classical_multi_control_x_correctness(x, cvs, ctrl, result):
+    bloq = MultiControlX(cvs=cvs)
+    cbloq = bloq.decompose_bloq()
+    ret = bloq.call_classically(x=x, ctrl=ctrl)
+    assert ret[-1] == result
```
mpharrigan, line 83: what is this testing? `cbloq` is unused

mpharrigan (changes requested):

qualtran/bloqs/multi_control_multi_target_pauli_test.py
fpapa250: I merged the two tests, but basically one will test that the output of the classical composite_bloq matches the classical implementation. The other test asserts that the value that matches both is actually correct.

qualtran/bloqs/multi_control_multi_target_pauli.py
```diff
@@ -148,3 +160,141 @@ def _ccpauli() -> MultiControlPauli:
...
     examples=(_ccpauli,),
 )
+
+
+@frozen
+class MultiControlX(Bloq):
+    r"""Implements multi-control, single-target X gate as a bloq using $n-2$ clean ancillas.
+
+    Args:
+        cvs: A tuple of control variable settings. Each entry specifies whether that
```
mpharrigan, line 170: I think we call these "control values". They're not really variables

qualtran/bloqs/multi_control_multi_target_pauli.py
```diff
@@ -148,3 +160,141 @@ def _ccpauli() -> MultiControlPauli:
...
+            control line is a "positive" control (`cv[i]=1`) or a "negative" control (`cv[i]=0`).
+
+    Registers:
+        ctrls: An input register with n 1-bit controls corresponding to the size of the control
+            variable settings above.
+        x: A 1-bit input register bit-flipped based on the values in the ctrls register.
+
+    References:
+        [$C^n$NOT from n-2 Zeroed bits from Constructing Large Controlled Nots]
+        (https://algassert.com/circuits/2015/06/05/Constructing-Large-Controlled-Nots.html)
```
mpharrigan, line 180: Sorry we don't have a style guide for these references, but please use the actual title from the post as the link and then a period and then a note on what part of the reference we're referencing.

e.g.

[Constructing Large CNOTS](https:/...). Section title "n−2 Ancilla Bits", Figure titled CnNOT from n-2 zeroed bits.

qualtran/bloqs/multi_control_multi_target_pauli.py
```diff
@@ -148,3 +160,141 @@ def _ccpauli() -> MultiControlPauli:
...
+
+    @cached_property
+    def signature(self) -> 'Signature':
+        assert len(self.cvs) > 0
+        return Signature(
+            [Register('ctrls', bitsize=1, shape=(len(self.cvs),)), Register('x', bitsize=1)]
+        )
+
+    def on_classical_vals(
+        self, x: 'ClassicalValT', ctrls: 'ClassicalValT'
```
mpharrigan, line 193: match the order of `signature`

mpharrigan (commented):
some nits, then lgtm

mpharrigan:
run `check/format-incremental --apply` to format the code and `check/pylint` to check pylint

mpharrigan (approved):
awesome lgtm

PR #597, opened 2024-01-22
Quantum data types

fdmalone:


qualtran/_infra/data_types.py
```diff
@@ -0,0 +1,147 @@
...
+            raise ValueError("bitsize must be greater than 1. Use QBit() instead.")
+
+    @property
+    def num_qubits(self):
+        return self.bitsize
+
+
+@attrs.frozen
+class QInt(QDType):
+    """Integer of a given width bitsize.
```
mpharrigan, line 58: signed? twos compliment? I think we need to be specific.

qualtran/_infra/data_types.py
```diff
@@ -0,0 +1,147 @@
...
+        if self.bitsize == 1:
+            raise ValueError("bitsize must be greater than 1. Use QBit() instead.")
+
+    @property
+    def num_qubits(self):
+        return self.bitsize
+
+
+@attrs.frozen
+class QUnsignedInt(QDType):
```
mpharrigan, line 76: do we need a type heirarchy? do we need an abstract QInt where is_dtype(x, QAbstractInt) returns True for QInt, QUnsignedInt, BoundedQInt?

qualtran/_infra/data_types.py
```diff
@@ -0,0 +1,147 @@
...
+class BoundedQInt(QDType):
+    """Integer whose values are bounded within a range.
+
+    Attributes:
+        bitsize: The number of qubits used to represent the integer.
+        iteration_range: The
+    """
+
+    bitsize: Union[int, sympy.Expr]
+    iteration_range: range
```
mpharrigan, line 106: this (`range`) isn't really a type, is it?

qualtran/_infra/registers.py
```diff
@@ -40,7 +43,7 @@ class Side(enum.Flag):
 
 
 @frozen
-class Register:
+class Register(Generic[QDTypeT]):
```
mpharrigan, line 46: I have strong questions about whether this actually needs to be generic.

From a quick inspection, it looks like it's useful in the `registers_test` where you're making registers and then using them and want to access specific properties of the datatype. I fear this has limited utility outside of this niche unless we make the generic-ness pervasive -- i.e. Signature would need to be generic w.r.t. a tuple of QDTypeT or something wild. 

These types are for the quantum program we're reasoning about at python run time. Remember: everything is sortof kicked up a level when we talk about programming concepts. Python class attributes (i.e. runtime values) are quantum program static/template parameters (i.e. concrete values in the quantum "binary"). We want qualtran to do the quantum type checking. We use mypy to do python type checking that we built the library correctly. 

Maybe the python type system can be hijacked to do more static validation of our quantum programs, but I'd strongly suggest removing the generic, using `QDType` (_not_ `QDTypeT`) everywhere, and considering the addition of these generic python types in follow-on work.

qualtran/_infra/registers.py
```diff
@@ -181,14 +189,14 @@ def __init__(self, registers: Iterable[Register]):
         self._rights = _dedupe((reg.name, reg) for reg in self._registers if reg.side & Side.RIGHT)
 
     @classmethod
-    def build(cls, **registers: int) -> 'Signature':
+    def build(cls, **registers: QDTypeT) -> 'Signature':
```
mpharrigan, line 192: this should be `QDType`, right? Signature isn't generic

qualtran/_infra/composite_bloq_test.py
```diff
@@ -39,6 +42,7 @@
     SoquetT,
 )
 from qualtran._infra.composite_bloq import _create_binst_graph, _get_dangling_soquets
+from qualtran._infra.data_types import QUnsignedInt
```
mpharrigan, line 45: import from `qualtran`

qualtran/_infra/registers.py
```diff
@@ -49,15 +52,16 @@ class Register:
...
+            array of 32-bit fixed-point floats.
         shape: A tuple of integer dimensions to declare a multidimensional register. The
             total number of bits is the product of entries in this tuple times `bitsize`.
         side: Whether this is a left, right, or thru register. See the documentation for `Side`
             for more information.
     """
 
     name: str
-    bitsize: int
+    dtype: QDTypeT
```
mpharrigan, line 64: we should write an attrs validator that raises a helpful error message if an integer is passed here.

qualtran/_infra/registers.py
mpharrigan, line 1: `SelectionRegister` is broken

qualtran/_infra/registers.py
mpharrigan: Some philosophy:

I'd like to keep `Signature.build` really dead simple where conciseness is key and you don't want to import anything and direct any cases that don't work with `Signature.build` to the structured list-of-registers thing. We've gotten pretty far in the library with the number "1" meaning qubit and any number greater than 1 meaning "unsigned integer with this many bits". I'd like to keep this quaint set of assumptions in `Signature.build`. 

```python
def build(cls, **registers: int):
  if reg == 1: return Register(regname, QBit(), THRU)
  return Register(regname, QInt(bitsize=reg), THRU)
```

mpharrigan (commented):
initial comments

qualtran/_infra/registers.py
fdmalone: It's supposed to be right now :)

qualtran/_infra/data_types.py
```diff
@@ -0,0 +1,147 @@
...
+    """A single qubit. The smallest addressable unit of quantum data."""
+
+    @property
+    def num_qubits(self):
+        return 1
+
+
+@attrs.frozen
+class QAny(QDType):
+    """Opaque bag-of-qbits type. Should be used sparingly"""
```
mpharrigan, line 43: It will be essential to document or codify the consequences of these data types. I.e. what's the different between `QInt` and `QAny`? is it just a different name to aid the programmer? I see you use `QAny` in the definition of SWAP. Was the intention that you can hook up any dtype to a QAny register?

qualtran/_infra/data_types.py
fdmalone: This is in the doc but hasn't made it into code yet. I'm thinking it would just an attribute.

qualtran/_infra/data_types.py
fdmalone: I think we might down the line, for example, the adder circuits should work for integer types (I think it works for fixed point too). For the moment I think it's best to see where things land given that circuit bloqs are implemented which may secretly be assuming unsigned integers etc, ditto for classical simulation.

qualtran/_infra/registers.py
fdmalone: yeah I didn't find much use for it so far, so it can go.

qualtran/_infra/registers.py
fdmalone: I've been thinking of QAny as the default here, but, sure this makes sense. We're taking QInt as a signed integer so I was a little hesitant to make it the default.

qualtran/_infra/data_types.py
fdmalone: Yes, I think we should (although my type check currently doesn't allow) for wiring any register into a QAny register. QInt etc should be used when we know for sure the register is representing some sort of numeric type (like an int) and not just a chunk of bits we are swapping.

qualtran/_infra/registers.py
fdmalone: SelectionRegister is going to be deprecated in favor of a Register with a BoundedQInt dtype.

qualtran/_infra/data_types.py
mpharrigan: how do you decide what's an attribute (complimentiness?) vs a new type (signedness)

qualtran/_infra/registers.py
mpharrigan: QAny would work too. I'll leave that up to you. There's a bit of fineesse  where the default should be chosen to simplify the cases we think should be simple rather than a super principled argument.

qualtran/_infra/data_types.py
mpharrigan: Be forewarned that you're sortof implementing your own generic type system. We'll have QAny which is compatible with any QDType. Then you'll get a request for QAnyInt which is compatible with any integer variant. Then you'll get a request for arbitrary "quantum generics"

qualtran/_infra/data_types.py
fdmalone: I was thinking that complimentiness is more of a hardware detail (sort of like endianness) but it's a fine line given that we're implementing algorithms at such a low level anyway and don't have ALUs. It's probably not a clear boundary. Adding another signed integer type would be fine too and maybe more explicit given that certain chemistry algorithms assume one complement quite explicitly. Maybe this is the correct thing to do since we're actually using the sign bit explicitly in places. WDYT?

Issue #449, opened 2023-10-25
Graph re-writing


mpharrigan:
Add functions and rules for re-writing ("transforming") parts of CompositeBloq graphs. 

One example optimization is to annihilate adjacent Split and Join bloqs, where applicable.

Other related transformations would involve swapping out different implementations of the same quantum interface.

anurudhp:
Hey @mpharrigan, I am interested in contributing to this. 

Is there already some infra like [cirq transformers](https://github.com/quantumlib/Cirq/tree/main/cirq-core/cirq/transformers) being worked on for qualtran? If not, I could implement a prototype graph-rewriter, with declarative patterns.

PR #600, opened 2024-01-24
Controlled protocol

mpharrigan:
This expands the "controlled" protocol. 

A key feature of any program is the ability to do branching. That is, choose what operations to do based on input. The primitive branching feature in quantum algorithms is the idea of a “controlled” gate. 

This PR introduces

 - `CtrlSpec`: an encapsulation of control behavior. I have the abstract interface factored out into `AbstractCtrlSpec`; which I may remove for now until more general behavior is actually needed. The concrete implementation in this PR supports expressing boolean activation functions whose inputs can be specified by one register, but the rest of the protocol would support an arbitrary number of registers.
 - `Controlled(Bloq)` can now do any sort of (abstract) control spec. 
 - `Bloq.controlled()` to get the (potentially bespoke) controlled version of a bloq.
 - `Bloq.get_ctrl_system()`: override-able method to provide custom controlled versions. For the protocol to be fully generic to any possible future ctrl spec, this returns an 'adder' function that knows how to actually wire up the returned controlled version.

Limits of this PR

 - I overrode the controlled versions of X, CNOT, Split, and Join in this PR but more to follow.
 - quimb simulation will go via decomposition. Follow-on work will get the subbloq's tensor and attach a control tensor. the abstract ctrlspec interface will be extended to support this.

image


qualtran/_infra/controlled.py
```diff
@@ -11,59 +11,292 @@
...
+            # dictionary, so we can correspond it to `old_out_soqs`.
+            ctrl_soqs = tuple(new_out_d.pop(creg_name) for creg_name in ctrl_reg_names)
+            return ctrl_soqs, new_out_d.values()
+
+        return cb, add_controlled
 
     @cached_property
-    def signature(self) -> Signature:
-        return Signature((Register(name="control", bitsize=1),) + tuple(self.subbloq.signature))
+    def ctrl_reg_names(self) -> Sequence[str]:
```
tanujkhattar, line 208: Does it make sense to make this an independent function outside of `Controlled` bloq?

mpharrigan: why?

mpharrigan: I'll factor it into a private helper function

mpharrigan:
ptal

qualtran/bloqs/basic_gates/cnot.py
```diff
@@ -96,6 +99,27 @@ def add_my_tensors(
...
+            def add_controlled(
+                bb: 'BloqBuilder', ctrl_soqs: Sequence['SoquetT'], in_soqs: Dict[str, 'SoquetT']
+            ) -> Tuple[Iterable['SoquetT'], Iterable['SoquetT']]:
+                (c2,) = ctrl_soqs
+                (c1, c2), target = bb.add(
+                    bloq, ctrl=np.array([in_soqs['ctrl'], c2]), target=in_soqs['target']
+                )
+                return (c1,), (c2, target)
+
+            return bloq, add_controlled
```
tanujkhattar, line 119: It's worth noting that this introduces an implicit ordering between the wires that get plugged into the two soquets of the newly added Toffoli gate. We've faced this issue in the past where two circuits are equivalent but they won't compare equal if, for example, one has `Toffoli(c1, c2, target)` and other has `Toffoli(c2, c1, target)`. 

Special purpose code was added in Cirq to make this equality return `True`. I don't think we have any similar plans for Qualtran right now, but this is one of the issue that arises with controlled bloqs so good to keep an eye on.

qualtran/serialization/bloq_test.py
```diff
@@ -124,13 +124,6 @@ def test_cbloq_to_proto_test_mod_exp():
     num_binst = len(set(binst.bloq for binst in cbloq.bloq_instances))
     assert len(proto_lib.table) == 1 + num_binst
 
-    cbloq = ControlledBloq(mod_exp)
-    proto_lib = bloq_serialization.bloqs_to_proto(cbloq, max_depth=1)
-    # 2x that of ModExp.make_for_shor(17 * 19).decompose_bloq() because each bloq in the
-    # decomposition is now controlled and each Controlled(subbloq) requires 2 entries in the
-    # table - one for ControlledBloq and second for subbloq.
-    assert len(proto_lib.table) == 2 * (1 + num_binst)
-
```
tanujkhattar, line 133: TODO(Tanuj, Arren): Add this test back once we update serialization code to support newly added `CtrlSpec` and corresponding `ControlledBloq`.

tanujkhattar (approved):
LGTM % nits

qualtran/_infra/controlled.py
```diff
@@ -11,59 +11,313 @@
...
+
+        # Otherwise, it's part of the control register.
+        i = self.ctrl_reg_names.index(soq.reg.name)
+        return self.ctrl_spec.wire_symbol(i, soq)
+
+    def pretty_name(self) -> str:
+        return f'C[{self.subbloq.pretty_name()}]'
+
+    def short_name(self) -> str:
+        return f'C[{self.subbloq.short_name()}]'
```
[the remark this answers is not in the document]
mpharrigan: We usually don't include class attributes in these methods, which are primarily used for the text on diagrams. This is worth revisiting when we overhaul the way we label diagram https://github.com/quantumlib/Qualtran/pull/602#issuecomment-1912789161

qualtran/bloqs/basic_gates/cnot.py
mpharrigan: We don't really have a notion of circuit equality in Qualtran. Treating the two control lines of e.g. toffoli as identical is already a problem with or without this PR. This might be a part of #449 roadmap

qualtran/serialization/bloq_test.py
mpharrigan: https://github.com/quantumlib/Qualtran/issues/648

qualtran/_infra/controlled.py
```diff
@@ -11,59 +11,313 @@
...
+     2. integer-equality controls where a `bitsize`-sized input must match an integer control value.
+     3. ndarrays of control values, where the bloq is active if **all** inputs are active.
+
+    For example: `CtrlSpec(cvs=[0, 1, 0, 1])` is active if the four input bits match the pattern.
+
+    A generalized control spec could support any number of "activation functions". The methods
+    `activation_function_dtypes` and `is_active` are defined for future extensibility.
+
+    Args:
+        bitsize: The bitsize of the control input.
```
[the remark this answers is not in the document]
mpharrigan: fixed this here and some other docs/variable names

mpharrigan:
I feel like the test failure is unrelated ...

PR #601, opened 2024-01-25
Fix _cirq_to_bloq wire symbols.

fdmalone:
The automatic wire symbol function was broken when drawing musical score diagrams. This fixes that.

Before:
Screenshot 2024-01-25 at 1 14 23 PM

After:
Screenshot 2024-01-25 at 1 07 16 PM

### DIFF
--- a/qualtran/bloqs/chemistry/thc/prepare_test.py
+++ b/qualtran/bloqs/chemistry/thc/prepare_test.py
@@ -23,6 +23,7 @@ from qualtran.bloqs.chemistry.thc.prepare import (
     PrepareTHC,
     UniformSuperpositionTHC,
 )
+from qualtran.drawing.musical_score import get_musical_score_data, MusicalScoreData
 from qualtran.linalg.lcu_util import preprocess_lcu_coefficients_for_reversible_sampling
 from qualtran.testing import execute_notebook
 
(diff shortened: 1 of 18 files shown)


mpharrigan:
Was this related to #591 ? I don't think so, right? Or was this always a bug?

Would it be possible to add a small test

fdmalone:
It was always a bug, see #530. I can add a test.

fdmalone:
Or rather, I imagine it became a bug once the cirq-ft adopted qualtran registers.

fdmalone:
I added a test for qrom which would have failed before.

qualtran/bloqs/chemistry/thc/prepare_test.py
```diff
@@ -81,5 +82,14 @@ def test_prepare_graph():
     assert isinstance(sigma, dict)
 
 
+def test_musical_score():
+    uni = _thc_uni()
+    msd = get_musical_score_data(uni)
+    assert isinstance(msd, MusicalScoreData)
+    prep = _thc_prep()
+    msd = get_musical_score_data(prep)
+    assert isinstance(msd, MusicalScoreData)
```
mpharrigan, line 91: so this used to just literally error out?

mpharrigan (approved):
thanks for the fix!

qualtran/bloqs/chemistry/thc/prepare_test.py
fdmalone: No, it would just print the the same register name for each wire as the (like mu, mu, mu, mu, ... ). BUT my first fix would then cause this to error out because it didn't handle shape > () bitsize > 1 properly (the prep.decompose() would work)

PR #602, opened 2024-01-25
Nicer musical score diagrams

fdmalone:
Builds on #601 trying to make nice circuit diagrams:

Before:
Screenshot 2024-01-25 at 3 08 29 PM

After:
Screenshot 2024-01-25 at 3 06 43 PM

I wonder if short_name should be optional in musical score diagrams and if there was some way to make a big bloq wire symbol diagram similar (for e.g. the $\sum_{\mu\nu} |\mu \nu\rangle$) bloq above, similar to to subprepare here:
Screenshot 2024-01-25 at 3 05 55 PM

in the THC prepapare bloq above


fdmalone:
There is one more PR to fix these diagrams which affects the circuits themselves so kept separate. Will open shortly

qualtran/bloqs/arithmetic/comparison.py
```diff
@@ -489,6 +495,14 @@ def short_name(self) -> str:
     def t_complexity(self) -> 'TComplexity':
         return t_complexity(LessThanEqual(self.a_bitsize, self.b_bitsize))
 
+    def wire_symbol(self, soq: Soquet) -> WireSymbol:
+        if soq.name == 'a':
```
mpharrigan, line 499: soquet doesn't have a name field (?)

mpharrigan:
I think it's highly non-trivial to create big-block rectangles. Happy to be proved wrong

mpharrigan:
> I wonder if short_name should be optional in musical score diagrams

`short_name` and `pretty_name` are badly designed methods (I can say this because I put them there). It's a kludge because we needed _something_ and giving it a silly method name like `pretty_name()` would spur people on to change it :D.  I'd love a cohesive design for a "labels" protocol that somehow elegantly lets a bloq express how it should be diagramed in various context (musical score, call graph, data flow graph); and maybe has some hooks or configuration for the style to be modified at diagram creation time rather than bloq authorship time. Note that these two things are slightly in opposition to each other.

qualtran/bloqs/arithmetic/comparison.py
fdmalone: yikes, should be soq.reg.name. I'll add a test.

mpharrigan (approved):
good to merge after syntax fix

fdmalone:
@mpharrigan I pulled out the wire symbol test from qrom into testing.

qualtran/testing.py
```diff
@@ -370,3 +371,18 @@ def check_bloq_example_decompose(bloq_ex: BloqExample) -> Tuple[BloqCheckResult,
...
+    """Assert a bloq's wire symbols match the expected ones.
+
+    Args:
+        bloq: the bloq whose wire symbols we want to check.
+        expected_ws: A list of the expected wire symbols.
+    """
+    ws = []
+    regs = bloq.signature
+    for i, r in enumerate(regs):
+        ws.append(bloq.wire_symbol(Soquet(i, r)).text)
```
mpharrigan, line 386: this will only work for `shape=()` situations 

And passing an integer in place of the `binst` is sortof a type error; but it's sortof an error for the wire_symbol method to actually use the `binst` field (whoops). Maybe make a note of that.

mpharrigan:
yeah it's hard to test the wire symbols. happy to merge and expand the utility of the new testing helper method later

qualtran/testing.py
fdmalone: Opened #608 to track, and a note.

PR #604, opened 2024-01-26
Use reflection in chem diagrams

fdmalone:
Depends on #602. Uses the Reflection bloq (reflect abound zero) in the chemistry diagrams. The final step is to use the new controlled protocol which will bring the musical score diagrams into line with the published circuits (remove the unsightly splits and joins).

Before:
Screenshot 2024-01-26 at 1 57 02 PM

After:
Screenshot 2024-01-26 at 1 09 57 PM


fdmalone:
Reference:

Screenshot 2024-01-26 at 1 57 42 PM

mpharrigan:
cool. I'll withhold approval until #602 is merged so I can have an honest look at the diff, but the proof is in the diagram

mpharrigan (approved):

PR #605, opened 2024-01-27
General costs

mpharrigan:
One of qualtran's main objective is to aid in the accounting of resource requirements ("costs") for quantum algorithms. The call graph protocol will happily count up bloqs for you, but we want to account for other costs (qubit 'width', precision, ...).

**This is a draft PR that sketches out my current prototype, described a bit below.**

This PR builds on the call graph. Recall that nodes in the call graph are bloqs and edges indicate that the predecessor bloq "calls" the successor bloq as a subroutine. The edges are annotated with a quantity `'n'`, which is how many times the callee is invoked. 

Now, nodes in the call graph (i.e. bloqs) will be annotated with a `"costs"` field. This is a table of costs (namely a list of `(CostKey, CostVal)` tuples). `CostKey` can be things like `BloqCount(bloq=TGate())` or `MaxQubits()` and `CostVal` can be an additive, multiplicative, or extrememizing general 'value'. The costs are derived from the call graph.

   cost(caller, is_leaf=False) = caller.static_costs + sum(n * cost(callee) for n, callee in caller.callees)
   cost(bloq, is_leaf=True) = bloq.static_costs + bloq.leaf_costs

That is: we use _either_ the sum of callee costs _or_ specifically annotated 'leaf' costs depending on whether the bloq is a leaf (or not) in the call graph. You can use the existing call graph machinery to treat bloqs as leafs (or not) with `generalizer`, `keep`, and `max_depth` arguments. 

Consider if we wanted to query either Total T Count or T+Tof count. With one "costs" method; you'd either never get the total T count or always get the total T count. If you did something really naive, you'd double-count the Ts from the toffoli decomposition. We need to be able to say that it costs *either* the sum of its callees *or* this fixed value of 1 toffoli.

The pseudocode above uses addition and multiplication for convenience, but these operations depend on the `CostVal` type. `+` should be considered a generic reduction step and `* n` should mean "do that reduction `n` times". For something like success probability, you'd use multiplication and power. For something like max qubit width, you'd use max and identity. Each type of cost also has to report its "identity" aka 0 value to make the reduction work.

The pseudocode above implies one cost per bloq; but all methods handle a list of `(CostKey, CostVal)`. Reductions are done by joining on the `CostKey` column.

### DIFF
--- a/qualtran/Protocols.ipynb
+++ b/qualtran/Protocols.ipynb
@@ -24,26 +24,66 @@
    "id": "1e356672",
    "metadata": {},
    "source": [
-    "## Summary of Qualtran protocols\n",
+    "## Decomposition\n",
     "\n",
-    "### Decomposition\n",
+    "### Interface\n",
+    "You can get a decomposition of a bloq in terms of its component bloqs by calling `bloq.decompose_bloq()`. Additional functionality is provided by the `qualtran.BloqBuilder` class. \n",
     "\n",
-    "You can get a decomposition of a bloq in terms of its component bloqs by calling `bloq.decompose_bloq()`. Additional functionality is provided by the `qualtran.BloqBuilder` class. To implement this protocol, bloq authors can override `Bloq.build_composite_bloq(...)`. A fallback to use a Cirq decomposition is provided. There is no default fallback -- `qualtran.DecomposeNotImplementedError` will be raised.\n",
+    "### Implementation\n",
+    "\n",
+    "To implement this protocol, bloq authors can override `Bloq.build_composite_bloq(...)`.\n",
+    "\n",
+    "### Fallbacks\n",
+    "\n",
+    "There is no default implementation; `qualtran.DecomposeNotImplementedError` will be raised.\n",
+    "`qualtran.cirq_interop.decompose_from_cirq_style_method` can be used as a fallback."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "26313128",
+   "metadata": {},
+   "source": [
+    "## Call Graph\n",
+    "\n",
+    "### Interface\n",
+    "\n",
+    "You can get a directed-acyclic graph representing the hierarchical decomposition of bloqs by calling `bloq.call_graph()` or direct callees with `bloq.bloq_counts()`. Additional functionality is contained in the `qualtran.resource_counting` module. \n",
+    "\n",
+    "### Implementation\n",
+    "To implement this protocol, bloq authors can override `Bloq.build_call_graph(...)`.\n",
     "\n",
     "\n",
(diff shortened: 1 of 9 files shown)


tanujkhattar:
Tracking `t-depth` can also be a good metric for generalized costs and would probably be a bit tricky to get right by looking only at the call graph (following up on our in-person discussions). Before approximation error, I think we should demonstrate computing
1) Bloq counts (trivial)
2) Qubit width (non-trivial)
3) T-depth / Bloq depth (probably non-trivial) ?

My understanding is that for (2) and (3) you'd propose to use `static_costs` and I'm curious to see if that works, compared to directly processing the composite bloq instead of call graph (as per my proposal).

mpharrigan:
What do you mean by T-depth?  One way would be to ignore all clifford gates, left-align, and count the number of topological generations.

In the surface code, there's a more nuanced definition of T depth https://arxiv.org/abs/1210.4626 perhaps "T reaction depth"  that I don't know a priori how to code up.

If you want to compute quantities that necessitate fully unrolling the circuit into Clifford+T ... that's not the point of the hierarchical costing framework. You can use `bloq.as_composite_bloq.flatten(lambda :True)` to fully unroll stuff and then write functions that operate on a single `CompositeBloq`. 

The conceit is that you can get cost estimates by looking at each level of decomposition at a time. That's why the qubit width is the maximum qubit width you'd get from serializing all operations (with #618) which you can do one level at a time without fully compiling down.

tanujkhattar:
> What do you mean by T-depth? 

I don't mean reaction depth, I mean literally just the number of topological generations in the unrolled circuit (or an approximate estimate thereof). Reaction depth is upper bounded by this number.

> If you want to compute quantities that necessitate fully unrolling the circuit into Clifford+T

There's another way which doesn't require unrolling everything in the circuit to get a reasonable upper bound. For example, we know QROM has linear depth, different arithmetic gate decompositions can have linear vs log depths etc. If each bloq in a composite bloq knows it's max_depth (or can be computed recursively), then we can compute (approx) max depth of the composite bloq without unrolling it by simply propagating 

```
max_depth(x_bloq)  = depth(x_bloq) + max(max_depth(y_bloq) for y_bloq in callees(x_bloq))
```

Here `callees(x_bloq)` yields all opposite end points for outgoing connections of `x_bloq`, similar to `max_width` analysis. 

-------

The point that I have been trying to make for a while now, which is apparent in both the examples of qubit width and bloq depth, is that propagating costs on the composite bloq is an important workflow and general costs design completely ignores that use case by saying you should do it as part of `static_costs`. We've seen two examples now where the way to compute the cost for a composite bloq is to propagate costs along the DAG instead of propagating it on the corresponding (one-level) of the Tree corresponding to the call graph.

PR #609, opened 2024-01-27
Add build_call_graph for qrom.

### LINKED ISSUE #559: Bloq counts / t_complexity for QRO(A)M is pretty slow
[add evidence]

fdmalone:
Fixes half of #559 by adding a build_call_graph method. The QROM Circuit is unary iteration + a bunch of cnots to write the data. The call graph works out the cnot count through the bit pattern of the data and the Toffoli count through the usual unary iteration formula.

Currently this doesn't work for the variable space qrom which requires a specific pattern for the input data. There are two options: 1) add this checking to the method, or 2) Make a new VariableSpacedQROM bloq which does some checks on the input data and uses the correct cost. I prefer option 2) as it makes things more explicit and makes this pretty useful feature more visible. What do you think @tanujkhattar

### DIFF
--- a/qualtran/bloqs/qrom.py
+++ b/qualtran/bloqs/qrom.py
@@ -23,7 +23,7 @@ from numpy.typing import ArrayLike, NDArray
 from qualtran import Register, SelectionRegister, Soquet
 from qualtran._infra.gate_with_registers import merge_qubits, total_bits
 from qualtran.bloqs.and_bloq import And, MultiAnd
-from qualtran.bloqs.basic_gates import CNOT, XGate
+from qualtran.bloqs.basic_gates import CNOT
 from qualtran.bloqs.unary_iteration_bloq import UnaryIterationGate
 from qualtran.drawing import Circle, TextBox, WireSymbol
 from qualtran.resource_counting import BloqCountT, SympySymbolAllocator
--- a/qualtran/bloqs/qrom.py
+++ b/qualtran/bloqs/qrom.py
@@ -12,7 +12,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from typing import Callable, Sequence, Tuple
+from typing import Callable, Sequence, Set, Tuple
 
 import attrs
 import cirq
@@ -23,8 +23,10 @@ from numpy.typing import ArrayLike, NDArray
 from qualtran import Register, SelectionRegister, Soquet
 from qualtran._infra.gate_with_registers import merge_qubits, total_bits
 from qualtran.bloqs.and_bloq import And, MultiAnd
+from qualtran.bloqs.basic_gates import CNOT, XGate
 from qualtran.bloqs.unary_iteration_bloq import UnaryIterationGate
 from qualtran.drawing import Circle, TextBox, WireSymbol
+from qualtran.resource_counting import BloqCountT, SympySymbolAllocator
 
 
 @cirq.value_equality()
@@ -197,3 +199,15 @@ class QROM(UnaryIterationGate):
     def _value_equality_values_(self):
         data_tuple = tuple(tuple(d.flatten()) for d in self.data)
         return (self.selection_registers, self.target_registers, self.control_registers, data_tuple)
+
+    def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
+        # Note this does not give the correct T complexity if we expect variable spaced qrom.
+        # QROM is unary iteration + a bunch of CX gates to write the data.
+        data_len = self.data[0].size
+        num_ands = data_len - 2 + self.num_controls
+        # number of cnots is just given by how many set bits each data entry has.
+        num_data_cnots = sum([int(x).bit_count() for x in np.array(self.data).ravel()])
+        extra_cnot = int(self.num_controls > 0)
+        num_data_cnots += data_len - 2 + extra_cnot
+        # ignoring the additional potential 2 X gates if num_controls == 0
+        return {(And(), num_ands), (And(uncompute=True), num_ands), (CNOT(), num_data_cnots)}
--- a/qualtran/bloqs/qrom_test.py
+++ b/qualtran/bloqs/qrom_test.py
@@ -19,10 +19,12 @@ import numpy as np
 import pytest
 
 from qualtran._infra.gate_with_registers import split_qubits, total_bits
+from qualtran.bloqs.basic_gates import CNOT, TGate
 from qualtran.bloqs.qrom import QROM
 from qualtran.cirq_interop.bit_tools import iter_bits
 from qualtran.cirq_interop.t_complexity_protocol import t_complexity
 from qualtran.cirq_interop.testing import assert_circuit_inp_out_cirqsim, GateHelper
+from qualtran.resource_counting.generalizers import cirq_to_bloqs
 from qualtran.testing import (
     assert_valid_bloq_decomposition,
     assert_wire_symbols_match_expected,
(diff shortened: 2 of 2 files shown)


fdmalone:
New timings (see #559):

10 time build = 4.363059997558594e-05, t_compl = 0.0039501190185546875, call_graph = 0.0003993511199951172, decomp = 0.014151573181152344
257 time build = 4.124641418457031e-05, t_compl = 0.03728079795837402, call_graph = 0.00031375885009765625, decomp = 0.31090688705444336
505 time build = 5.9604644775390625e-05, t_compl = 0.06827211380004883, call_graph = 0.00037169456481933594, decomp = 0.6993441581726074
752 time build = 6.341934204101562e-05, t_compl = 0.09933924674987793, call_graph = 0.0004017353057861328, decomp = 0.9915351867675781
1000 time build = 6.151199340820312e-05, t_compl = 0.25572872161865234, call_graph = 0.00044083595275878906, decomp = 1.3253419399261475

Old timings:
10 time build = 4.7206878662109375e-05, t_compl = 0.003383636474609375, call_graph = 0.015896081924438477, decomp = 0.013347148895263672
257 time build = 3.719329833984375e-05, t_compl = 0.03769230842590332, call_graph = 0.3068056106567383, decomp = 0.41684818267822266
505 time build = 6.270408630371094e-05, t_compl = 0.07230806350708008, call_graph = 0.725104570388794, decomp = 0.5794270038604736
752 time build = 6.198883056640625e-05, t_compl = 0.10921454429626465, call_graph = 1.1440081596374512, decomp = 1.0013768672943115
1000 time build = 5.91278076171875e-05, t_compl = 0.13857412338256836, call_graph = 1.3271982669830322, decomp = 1.4297850131988525

tanujkhattar:
Ideally, we should add a `build_call_graph` on `UnaryIteration` class so that all derived classes get the speedup for `build_callgraph` and the `break_early` logic is handled for all cases and not just variable spaced QROM. Here is a quick draft that shows how it can be done - https://github.com/quantumlib/Qualtran/pull/610

I've done some profiling based on running your example code (given as follows)
```python
for dim in np.linspace(10, 10000, 5, dtype=int):
    data = np.random.randint(0, 100, size=dim)
    start = time.time()
    qrom = QROM.build(data)
    time_build = time.time() - start
    start = time.time()
    t_comp = qrom.t_complexity()
    time_t_compl = time.time() - start
    start = time.time()
    graph, sigma = qrom.call_graph()
    time_call_graph = time.time() - start
    print(
        f"{dim} time build = {time_build}, t_compl = {time_t_compl}, call_graph = {time_call_graph}"
    )
```

On main
```
10 time build = 8.487701416015625e-05, t_compl = 0.006656169891357422, call_graph = 0.01882004737854004
2507 time build = 8.177757263183594e-05, t_compl = 0.46207594871520996, call_graph = 3.4360601902008057
5005 time build = 0.00021004676818847656, t_compl = 1.3898670673370361, call_graph = 8.454498052597046
7502 time build = 6.103515625e-05, t_compl = 1.344743013381958, call_graph = 12.302637100219727
10000 time build = 7.390975952148438e-05, t_compl = 1.6425161361694336, call_graph = 17.790980100631714
```

On your current PR that adds a restrictive `build_callgraph` only for `QROM`
```
10 time build = 8.916854858398438e-05, t_compl = 0.00453495979309082, call_graph = 0.0006778240203857422
2507 time build = 0.00032210350036621094, t_compl = 0.5399672985076904, call_graph = 0.0007531642913818359
5005 time build = 0.00014090538024902344, t_compl = 1.0233731269836426, call_graph = 0.0008797645568847656
7502 time build = 6.175041198730469e-05, t_compl = 1.413398027420044, call_graph = 0.0014181137084960938
10000 time build = 0.00013589859008789062, t_compl = 1.9211628437042236, call_graph = 0.0015759468078613281
```

On my PR that adds a generalized `build_callgraph` on `UnaryIterationGate` and thus also handles the variable qrom case. 
```
10 time build = 0.00018405914306640625, t_compl = 0.007563114166259766, call_graph = 0.0019731521606445312
2507 time build = 7.224082946777344e-05, t_compl = 0.48326563835144043, call_graph = 0.05669093132019043
5005 time build = 0.0001201629638671875, t_compl = 0.9212100505828857, call_graph = 0.08840084075927734
7502 time build = 0.00015306472778320312, t_compl = 1.2891829013824463, call_graph = 0.1206820011138916
10000 time build = 6.985664367675781e-05, t_compl = 1.9097709655761719, call_graph = 0.18745183944702148
```


In general, the current approach in your PR is O(1) whereas the one in my PR is O(N). My PR can be further optimized to bring down the cost (eg: iterative vs recursive O(N), constructing the bloqs `And` / `CNOT` only once instead of O(N) times etc.) but it would still pretty much be O(N). 

To support the break-early case in general (i.e. variable spaced qrom) I think we'd have to pay O(N) cost in the worse case since for each internal node of the segment tree we need to check whether we can break early or not. The same holds true for QROM because we need to at least look at the data (in O(N)) to figure out the optimal bit pattern. But I'm optimistic that after optimizations, the constant factors behind O(N) cost should come down significantly. 


Let me know what you think. I can go ahead and add tests / make further optimizations to my PR.

fdmalone:
Well I guess my question would be (maybe philosophical), couldn't we just formulaically return the number of And/And^ gates in unary iteration (+ any cliffords) rather than doing any O(N) operation? My concern would be (maybe you could convince me otherwise) that if we have say 10^6 entries (so some qrom with (10^2)^3 entries which is not uncommon in chemistry how fast is your approach. So this would be 100-1000 times slower than your benchmark which would mean 10s-100s of seconds which is probably fine.

fdmalone:
But yeah, your solution is nicer as it would presumably help speeding up SelectSwapQROM too.

tanujkhattar:
Okay, I made some improvements and did some more profiling. 

The majority of the time spent in my approach was in the `_break_early(l, r)` method for QROM. Right now, the method is O(NlogN) because it uses `np.unique`, that iterates over the entire `r - l` elements in the range `[l, r]` and returns a list of unique elements. And this method is being called from every node of the segment tree, so we are spending O(NlogN) time at each of the O(logN) levels of the segment tree for a total of `O(Nlog^2(N))` time complexity. 

I've refactored the method to now be `O(N)` in the worst case and break early if at least 2 different elements are found. The overall complexity should now be `O(N)` instead of `O(Nlog^2(N)`. I've also written an iterative version of segment tree traversal instead of a recursive one. 


Here's some profiling data due to all the improvements

I now construct a call graph for 1e6 elements (the range you mentioned in the previous comment)
```
for dim in np.linspace(10_000, 1000_000, 5, dtype=int):
    data = np.random.randint(0, 100, size=dim)
    start = time.time()
    qrom = QROM.build(data)
    time_build = time.time() - start
    start = time.time()
    graph, sigma = qrom.call_graph()
    time_call_graph = time.time() - start
    print(
        f"{dim} time build = {time_build}, call_graph = {time_call_graph}"
    )
```

On my first draft

```
10000 time build = 0.0004858970642089844, call_graph = 0.2071681022644043
257500 time build = 0.0003261566162109375, call_graph = 4.352848768234253
505000 time build = 0.0007529258728027344, call_graph = 8.924990892410278
752500 time build = 0.004110097885131836, call_graph = 15.377625703811646
1000000 time build = 0.005636692047119141, call_graph = 18.472869157791138
```

On the improved version ([post this commit](https://github.com/quantumlib/Qualtran/pull/610/commits/189477107efad30133f798741728fe316240ed52)) - ~3x faster

```
10000 time build = 0.0005826950073242188, call_graph = 0.11071014404296875
257500 time build = 0.0016162395477294922, call_graph = 1.7726051807403564
505000 time build = 0.0021660327911376953, call_graph = 3.72778582572937
752500 time build = 0.0046787261962890625, call_graph = 5.437635183334351
1000000 time build = 0.002295970916748047, call_graph = 7.241891145706177
```

So it takes ~7 seconds to build a call graph for 1e6 elements and it supports the variable QROM case.

tanujkhattar:
> couldn't we just formulaically return the number of And/And^ gates in unary iteration (+ any cliffords) rather than doing any O(N) operation?

If we want to support `_break_early` in general, then NO since the structure of the constructed segment tree circuit depends upon the behavior of `break_early`, so at the very least we need to iterate over the 2 * N nodes of the segment tree and ask if we can break early at this point or not (which is what we do now). 

For the vanilla case (i.e. no support for break early), we can definitely do it in O(1). But a fast implementation of `O(N)` shouldn't be bad since we need to spend `O(N)` just to construct the `QROM` bloq (specify the input data). 

The performance improvements mentioned in my previous comment are a step in this direction but I think there's still scope to shave off constant factors hiding in my `O(N)` implementation.

tanujkhattar:
> But yeah, your solution is nicer as it would presumably help speeding up SelectSwapQROM too.

`SelectSwapQROM` doesn't derive from `UnaryIteration`. It just decomposes into `QROM` and `SwapWithZero`, so my version wouldn't help there. But it would help in other unary iteration gates like `SelectMajoranaFermion` etc.

mpharrigan:
posting here for posterity: https://jiffyclub.github.io/snakeviz/ 

snakeviz is really useful for figuring out what's slow

fdmalone:
Ok, I'll close this in favor of your PR. We'll need to revisit SelectSwapQROM too, but I imagine just overwriting the call graph into the appropriate number of qroms and swaps will be now much faster than before.

tanujkhattar:
@mpharrigan Yeah,  I used snakeviz to do the profiling and figure out `_break_early` was a bottleneck. +1 to it being super useful for profiling

fdmalone:
@tanujkhattar on the _break_early method could we precompute this or is it more complicated. From memory the variable spacing patterns I've seen are so specific presumably you could just check a handful of elements of the input to check if it's necessary and then switch?

tanujkhattar:
We could precompute it, but it's not really a bottleneck now since for the general case when all elements are distinct, we look at 2 elements and return False. So it's pretty much `O(1)`. 

I actually did some more profiling and looks like allocating `2 * n` memory for `l` and `r` arrays in my segment tree implementation was also an unnecessary bottleneck that gets worse as `n` increases. So I switched to computing the `l` and `r` ranges on the fly using arithmetic instead of caching them in `l` and `r` arrays. That leads to another ~30% improvement s.t. it now takes ~4.5 seconds to construct callgraph for 1e6 elements (output of my previous profiling code given below)

```
10000 time build = 0.0016388893127441406, call_graph = 0.05440402030944824
257500 time build = 0.0017690658569335938, call_graph = 1.1562600135803223
505000 time build = 0.003152132034301758, call_graph = 2.1690218448638916
752500 time build = 0.004446983337402344, call_graph = 3.4806017875671387
1000000 time build = 0.0055999755859375, call_graph = 4.495238780975342
```

I'll stop being nerd sniped now and merge my PR. I'm sure we can further push down the number if needed but it looks pretty good now.

Issue #596, opened 2024-01-22
Add matrix product state (MPS) preparation algorithm


StefanoPolla:
The sequential MPS preparation algorithm of Schön et al. ([arXiv:quant-ph/0501096](https://arxiv.org/abs/quant-ph/0501096)) would be a valuable addition to Qualtran, providing a leading state preparation algorithm.
The Toffoli compilation of this algorithm is costed in appendix E of the recent Fomichev et al. paper ([arXiv:2310.18410](https://arxiv.org/abs/2310.18410)), and it used building blocks like PREPARE-based reflections.

The goal is to construct a composite bloq which will have as input the MPS (or possibly only the shape of the MPS tensors and the required bit precision) and implements the MPS preparation. 

As an element of this algorithm, we need a bloq combining PREPARE-based reflections with an ancilla to synthesize $k$-qubit unitary operations where only the first $m \leq 2^{k}$ columns are defined. See eq. E1 to E8 of [arXiv:2310.18410](https://arxiv.org/abs/2310.18410). This bloq can be implemented by using existing primitives.

Vicara12:
I'm going to look into it

StefanoPolla:
To construct this algorithm, we need a building bloq $W$ preparing a state $|w\rangle = W|0\rangle$.

This state has a specific structure: the output register can be split into an auxiliary qubit $R$ (reflection ancilla) and a target register T
$$|w\rangle = \frac{1}{\sqrt{2}} ( |1\rangle_R |\alpha\rangle_T - |0\rangle |u_\alpha\rangle_T)$$
where $|j\rangle_T$ is a computational basis state which can be prepared only using CX gates (controlled on $R$), and $|u_j\rangle_T$ is a generic state which should be constructed by controlled-PREPARE.
[See Eq. E5 and E8 of [arXiv:2310.18410](https://arxiv.org/abs/2310.18410))

To do this, we should first implement a controlled version of PREPARE.

fdmalone:
Copying this over from the closed PR (#615) for visibility:

[...]

Issue #513, opened 2023-11-13
ZeroState vs Allocate


mpharrigan:
These two things result in confusion. Especially since there's an unused/unsupported `n` attribute in ZeroState. I suggest removing this and making ZeroState always a one-bit thing for use in smaller circuits and bit twiddling. Maybe rename to `ZeroBitState`? That's pretty long though. If you just need a bunch of workspace use `Allocate` or its convenience method `bb.alloc()`. If you need a particular integer there's `IntState`.

mpharrigan:
Un-closing. We should remove `n` from `ZeroState`.

PR #615, opened 2024-01-31
Adding build_composite_bloq to PhaseGradientState

Vicara12:
Adding `build_composite_bloq` to `PhaseGradientState` so that it is possible to perform tensor contraction on circuits that use it.


tanujkhattar:
Thanks for @Vicara12 for opening your first PR! We have a weekly sync where we meet and discuss issues, and you can get an invite by joining https://groups.google.com/g/qualtran-dev

Regarding your PR, `PhaseGradientState` already has a decomposition defined using the cirq style `decompose_from_registers` and thus the `decompose_bloq` is derived automatically. You can verify that by calling `show_bloq(PhaseGradientState(6).decompose_bloq())` which prints the correct decomposition as follows: 

image

The reason tensor decomposition fails is because `Allocate` and `Free` bloqs currently don't have a tensor decomposition specified. This was first raised in https://github.com/quantumlib/Qualtran/issues/513 and I've opened https://github.com/quantumlib/Qualtran/issues/616 to track it's impact on phase gradient state bloq. 

Adding tensor decomposition to `Allocate` fixes the problem for `PhaseGradientState` as I've done in my PR here - https://github.com/quantumlib/Qualtran/pull/617 

Doing a similar thing for `Free` should fix the problem for the adjoint of phase gradient state. I'll open a separate PR to do that and then the issue should be resolved. 

We can close your PR now, but thanks for highlighting this. I hope to see you in our weekly sync's to discuss more about your usage for Qualtran. Thanks!

Vicara12:
Thank you for your detailed answer, now the tensor contract works.

Also, I plan to implement an MPS preparation algoritm (issue #596
<https://github.com/quantumlib/Qualtran/issues/596>) and I would like to
tell you what I think I need to implement to avoid duplicating code.
In order to do so I need to compile a n qubit quantum gate of which only k
columns are specified using Householder reflections (refer to
https://arxiv.org/abs/2310.18410 sections IIIB page 7 and apendix E on page
24 for the details) . The blocks that I need are:

   - A bloq for the reflection I-2|w><w|, for which I plan to use
   *ReflectUsingPrepare* with a custom prepare oracle.
   - A prepare oracle to prepare the state |w> (this I plan to implement
   myself, as the state |w> is easily decomposed in other gates and using a
   generic state prepare oracle would be overkill)
   - For the prepare oracle W I also need a controlled state preparation
   bloq. I saw that qualtran has the class *StatePreparationAliasSampling*,
   but I see two problems with it for this specific use case. The first one is
   that it produces entangled junk qubits, which need special handling and
   could interfere with the rest of the algorithm. The second one is that, as
   far as I can tell, there is not a controlled version of it. For this
   reason, I implemented the state preparation algorithm from
   https://arxiv.org/abs/1812.00954 page 3, which produces no entangled
   garbage qubits and can be easily adapted to be controlled.
   - Finally, for the previous step I need a programable rotation gate
   array. I found the class *ProgrammableRotationGateArray*, which does
   something similar, but I think that for this algorithm it would be much
   more efficient to use a gradient state and perform the rotations using
   phase kickback (which is the technique used in the paper), because there
   are multiple rotation gate arrays and *ProgrammableRotationGateArray*
   would lead to redundancies.

Please let me know if those would be valuable contributions to the project
or if there are already valid alternatives to them which I can use. Thanks!

Regards,
Víctor

PR #618, opened 2024-01-31
Max width of composite bloq

mpharrigan:
A helper function for getting the width of a composite bloq. This was raised as a problem in the review of #605 . This is just the helper function. There's a "TODO" where one could plumb in the max-widths of the callees, which would be available if this was part of the call graph summation. 

image


mpharrigan:
@tanujkhattar I think this can go in as a private helper method, ptal

qualtran/resource_counting/_qubit_counting.py
```diff
@@ -0,0 +1,60 @@
...
+    `_bloq_max_width` callable) and the bystander connections that are "in play". The max
+    width is the maximum over all the time points.
+    """
+    max_width: Union[int, sympy.Expr] = 0
+    series_i: int = 0
+    in_play: Set[Connection] = set()
+
+    for gen_i, binsts in enumerate(nx.topological_generations(binst_graph)):
+        for binst in binsts:
+            pred_cxns, succ_cxns = _binst_to_cxns(binst, binst_graph=binst_graph)
```
tanujkhattar, line 41: I'd suggest to add an outer loop for analyzing each connected component independently, so we have (pseudocode) as follows

```
for cc_i, cc_binst_graph in enumerate_connected_components(binst_graph):
    for gen_i, binsts in enumerate(nx.topological_generations(cc_binst_graph)): 
        for binst in binsts:
            ....
```

This should result in output being 2 * n vs 3 * n for the two examples discussed in the issue

mpharrigan: I didn't understand that in your example. If you drew that circuit diagram with a random alloc and free happening alongside your calculation you'd say the width included that line. I guess if your bloq is composed of disconnected components, you could (theoretically) do each component sequentially so the max widths should be considered separately?

mpharrigan:
image

@tanujkhattar ptal

tanujkhattar (approved):

Issue #446, opened 2023-10-25
Quantum Data Types


mpharrigan:
At present, we have `Register` which contains all the attributes to define part of a function signature: the name and the side attributes go towards the function signature. The bitsize (and shape) argument describes the quantum data type. This singular family of dependent quantum data types is pretty coarse grained. We can extend our library of available data types to include

 - Bounded domain registers; i.e. with a min and max value. This is a generalization of cirq-ft “selection register”
 - LittleEndian, BigEndian integer registers. Signed and Unsigned integer registers
 - FixedPoint; FloatingPoint registers
 - ControlRegister that carries with it one or more “control values”; i.e. whether something is positive control or negative control.

Things that are not in scope are 
 - Sum and product types.
 - Strings/lists/jagged arrays

These elements would comprise a full type system, but would add a lot of complexity and are likely not relevant for the type of resource estimates we aim to do.

Support for qudit registers (i.e. arbitrary dimension d units and registers thereof) and/or quopit (prime dimension) are not out of scope but they shall only be added if there’s a clear resource-estimation use-case for them that cannot be accomplished with Register+bitsize or bounded domain registers.

fdmalone:
Rough PR plan:

- [ ] Simple gates (basic_gates, and, ...) #628 
- [ ] Arithmetic and sorting
- [ ] factoring
- [ ] rotations / qft
- [ ] deprecate selection registers #626 (I might split this up if possible)
- [ ] chemistry (> n PRs)
- [ ] util_bloqs (splits / joins)
- [ ] Probably one on interop.
- [ ] enforce type checking.

fpapa250:
This is in reference to how on_classical_vals() treats the classical version of the data type, correct? I'm finding that as I build new arithmetic bloqs out of previous ones I've built (`ModAdd()` requires a `GreaterThan()` bloq) each sub-bloq is taking issue with the type. For example a simple `bb.join()` operation on the sign bit in the following commit required a change to the classical sim of the `XGate()`. Not sure on the best way to proceed with this.

https://github.com/quantumlib/Qualtran/commit/78938148e8fdc6dc0f7b24291f60e842cb1df9c8

fdmalone:
classical simulation is part of it for sure. I'm not sure what the best way forward is. If an arithmetic operation has a different decomposition dependent on the type of the inputs then it probably makes sense to dispatch based on the register type, rather than trying reinterpret the bitsize.

fdmalone:
[...]

PR #624, opened 2024-02-02
Add quantum data types

fdmalone:
Factored out from #597.

* Adds the basic data types under _infra/data_types, with some documentation.
* Adds a dtype property to registers. I couldn't find a non-invasive way of shimming in the register to use the dtype: adding a converter for bitsize is problematic because then reg.bitsize would return a QDType. changing the name to reg.dtype would require a huge diff (replacing Register(.., bitsize=n) everywhere).
* No type checking or any other features of data types are used. Essentially nothing has changed.

Part 2: deprecate bitsize in register for dtype and add bitsize as a parameter. Add some interception code for Signature.build as discussed in #597. This will be a big PR. passing an int will work but will raise a warning potentially (although this is probably too annoying in practice). This will lead to some temporary weirdness like reg = Register(dtype=1) which is unfortunate but I don't see a way around it.

Part3: deprecate SelectionRegister in favour of Register(dtype=BoundedQInt(...))

Part 4: A handful of PRs for the registers which actually use numeric types (arithmetic / chemistry, ...). Splits / Joins. Type checking won't be enforced yet.

Part 5:  Totally deprecate integer values for Register.

Part 6: Enforce type checking of registers during _process_soquets. Splits/Joins expect dtypes rather than register sizes.

### DIFF
--- a/qualtran/_infra/data_types.py
+++ b/qualtran/_infra/data_types.py
@@ -73,12 +73,12 @@ class QAny(QDType):
 
     bitsize: Union[int, sympy.Expr]
 
+    def __attrs_post_init__(self):
+        assert self.bitsize > 1, "bitsize must be > 1."
+
     @property
     def num_qubits(self):
         return self.bitsize
-
-    def __eq__(self, other) -> bool:
-        return other.num_qubits == self.num_qubits
 
 
 @attrs.frozen
@@ -92,6 +92,9 @@ class QInt(QDType):
     """
 
     bitsize: Union[int, sympy.Expr]
+
+    def __attrs_post_init__(self):
+        assert self.bitsize > 1, "bitsize must be > 1."
 
     @property
     def num_qubits(self):
@@ -109,6 +112,9 @@ class QIntOnesComp(QDType):
     """
 
     bitsize: Union[int, sympy.Expr]
+
+    def __attrs_post_init__(self):
+        assert self.bitsize > 1, "bitsize must be > 1."
 
     @property
     def num_qubits(self):
@@ -128,6 +134,9 @@ class QUnsignedInt(QDType):
 
     bitsize: int
 
+    def __attrs_post_init__(self):
+        assert self.bitsize > 1, "bitsize must be > 1."
+
     @property
     def num_qubits(self):
         return self.bitsize
@@ -146,12 +155,10 @@ class BoundedQInt(QDType):
     iteration_length: int
 
     def __attrs_post_init__(self):
-        if self.iteration_range.start > self.iteration_range.stop:
-            raise ValueError("iteration_range limits should be increasing in value.")
-
-        if len(self.iteration_range) > 2**self.bitsize:
+        assert self.bitsize > 1, "bitsize must be > 1."
+        if self.iteration_length > 2**self.bitsize:
             raise ValueError(
-                f"BoundedQInt iteration length is too large for given bitsize. {len(self.iteration_range)} vs {2**self.bitsize}"
+                f"BoundedQInt iteration length is too large for given bitsize. {self.iteration_length} vs {2**self.bitsize}"
             )
 
     @property
@@ -177,3 +184,6 @@ class QFixedPoint(QDType):
     @property
     def num_qubits(self):
         return self.int_bitsize + self.frac_bitsize + 1
+
+    def __attrs_post_init__(self):
+        assert self.num_qubits > 1, "Number of qubits must be > 1."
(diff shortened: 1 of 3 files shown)


mpharrigan:
> I couldn't find a non-invasive way of shimming dtype into Register

Is this because of `attrs` which ties ones hands when trying to do some of this funny business? I was thinking


```python
def __init__(self, name: str, bitsize: Union[int, QDType]):
  if isinstance(bitsize, int):
     self._dtype = QBit() or QAny(bitsize)
  else:
    self._dtype = bitsize

@property
def bitsize(self):
  return self._dtype.num_qubits

@property
def dtype(self):
  return self._dtype
```

maybe this is wonky to get into attrs. Maybe you can use the fact that attrs will strip leading underscores from attribute names in the constructor

```python
class Register:
  name: str
  _bitsize: Union[...] = field(converter=turn_everything_secretly_to_qdtype)

  @property
  def bitsize(self): ... # ok now

  @property
  def dtype(self): return self._bitsize # not the most readable, but works in a pinch.
```

qualtran/_infra/data_types.py
```diff
@@ -0,0 +1,189 @@
...
+For example, a bloq multiplying two fixed point reals should be built with an
+explicit QFixedPoint dtype.
+3. The smallest addressable unit is a QBit. Other types are interpretations of
+collections of QBits. A QUnsigedInt(32) is intended to represent a register
+encoding positive integers.
+4. To avoid too much overhead we have a QAny type, which is meant to represent
+an opaque bag of bits with no particular significance associated with them. A
+bloq defined with a QAny register (e.g. a n-bit CSwap) will accept any other type assuming the
+bitsizes match. QInt(32) == QAny(32), QInt(32) != QFixedPoint(16, 16). QInt(32) != QUnsignedInt(32).
+5. We assume a big endian convention for addressing QBits in registers throughout qualtran.
```
mpharrigan, line 41: might be worth reminding people what big endian actually is (I never can remember). i.e. "most significant bit at index 0" or whatever it actually is

qualtran/_infra/data_types.py
```diff
@@ -0,0 +1,189 @@
...
+"""
+
+import abc
+from typing import Union
+
+import attrs
+import sympy
+
+
+class QDType:
```
mpharrigan, line 54: `(metaclass=abc.ABCMeta)`

and a little docstring saying this is the abstract interface for quantum data types.

qualtran/_infra/data_types.py
```diff
@@ -0,0 +1,189 @@
...
+encoding positive integers.
+4. To avoid too much overhead we have a QAny type, which is meant to represent
+an opaque bag of bits with no particular significance associated with them. A
+bloq defined with a QAny register (e.g. a n-bit CSwap) will accept any other type assuming the
+bitsizes match. QInt(32) == QAny(32), QInt(32) != QFixedPoint(16, 16). QInt(32) != QUnsignedInt(32).
+5. We assume a big endian convention for addressing QBits in registers throughout qualtran.
+6. Ones' complement integers are used extensively in quantum algorithms. We have
+two types QInt and QIntOnesComp for integers using two's and ones' complement
+respectively.
+"""
```
mpharrigan, line 45: at a future point, I will ask you to take this docstring out and put it in a notebook that documents the whole qdatatype thing so we can actually include it in the docs.

qualtran/_infra/data_types.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+@attrs.frozen
+class QAny(QDType):
+    """Opaque bag-of-qbits type."""
+
+    bitsize: Union[int, sympy.Expr]
+
+    def __attrs_post_init__(self):
+        assert self.bitsize > 1, "bitsize must be > 1."
```
mpharrigan, line 77: this will complain if bitsize is a sympy.Expr

mpharrigan (approved):
lgtm modulo nits

fdmalone:
right I couldn't figure out how to do it without an init + private. Your solution should work looking through attrs docs again.

qualtran/_infra/data_types.py
fdmalone: Done.

qualtran/_infra/data_types.py
fdmalone: Done.

qualtran/_infra/data_types.py
fdmalone: Fixed.

fdmalone:
I'll skip part 2 as bitsize can stay until the migration is complete.

fdmalone:
xref #446

Issue #235, opened 2023-05-23
Make comparison gates use 4 * N T-gates to compare and 0 T-gates to uncompute using MBUC


tanujkhattar:
We can use Craig's out of place adder to compute using 4 * N T-gates and uncompute using 0 T-gates as described in https://arxiv.org/pdf/1709.06648.pdf.

mpharrigan:
Is this done?

NoureldinYosri:
not completly, some comparison gates use 4T others use 8T. I will send a couple of PR to address this

mpharrigan:
Is this done now?

NoureldinYosri:
not all of them .. some comparison bloqs use the n log n decomposition

NoureldinYosri:
now there are versions for the linear depth 4T comparators ... there are also some comparators that use log depth (8T)  constructors ... so both options are supported

PR #625, opened 2024-02-04
Implements GreaterThan() w/ Classical Sim

fpapa250:


fpapa250:
@mpharrigan how can I update the notebook since the bloq construction has changed?

fdmalone:
you can rerun dev_tools/autogenerate-bloqs-notebooks-v2.py. A lot of these tests are failing because it looks like the t-complexity changed, was that expected?

fpapa250:
> you can rerun dev_tools/autogenerate-bloqs-notebooks-v2.py. A lot of these tests are failing because it looks like the t-complexity changed, was that expected?

Yes, the GreaterThan() construction I did uses logical-and bloqs to reduce the T-Complexity from 8n to 4n (that of Add() which is what this construction is based on). I also changed the args to only require one bitsize.

mpharrigan:
@NoureldinYosri can you take a quick look at this construction

qualtran/bloqs/arithmetic/comparison.py
```diff
@@ -465,50 +467,172 @@ def _has_unitary_(self):
...
+        true_bitsize = self.bitsize if self.signed else (self.bitsize + 1)
+
+        # Flip all the bits in the x register.
+        a_split = bb.split(a)
+
+        for i in range(true_bitsize):
+            a_split[i] = bb.add(XGate(), q=a_split[i])
+        b_split = bb.split(b)
+
+        # Iteratively implements the left adder circuit building block of the Gidney Adder. On
```
NoureldinYosri, line 558: you don't need to implement the decompisition of the adder gate. We already have the Gidney adder in qualtran `qualtran.bloqs.arithmetic.addition.Add` which you can use in your decomposition.

https://github.com/quantumlib/Qualtran/blob/642dec21cb6844bbe2f5f33e4312a4951f73e8c9/qualtran/bloqs/arithmetic/addition.py#L48

qualtran/bloqs/arithmetic/comparison.py
```diff
@@ -465,50 +467,172 @@ def _has_unitary_(self):
 class GreaterThan(Bloq):
     r"""Compare two integers.
 
-    Implements $U|a\rangle|b\rangle|0\rangle \rightarrow
-    |a\rangle|b\rangle|a > b\rangle$ using $8n T$  gates.
+    Implements |a>|b>|t> => |a>|b>|t ⨁ (b > a)> using $4n$ T gates.
```
NoureldinYosri, line 470: This flips the gate. This gate is supposed to do `a > b` not `b > a` as is written in the deleted docstring

$$U|a\rangle|b\rangle|0\rangle \rightarrow
    |a\rangle|b\rangle|a > b\rangle$$

The change is why it is failing the tests. if that is intended then I think you should leave the `GreaterThan` bloq as is and implement a `LessThan` bloq

NoureldinYosri (changes requested):
We don't need to implement the docomposition here since we already did that in [Add](https://github.com/quantumlib/Qualtran/blob/642dec21cb6844bbe2f5f33e4312a4951f73e8c9/qualtran/bloqs/arithmetic/addition.py#L48) bloq which we can use in the decomposition. 

The bigger issue here is that this PR flips the `GreaterThan` bloq into a `LessThan` bloq

qualtran/bloqs/arithmetic/comparison.py
fpapa250: I implemented the decomposition because we only need to see the most significant bit of the b register after the add and then use that to modify the target qubit. After this we need to uncompute the add so that the b register is untouched so the right side is the uncomputation of the left which is different than the Add() bloq.

NoureldinYosri: the `Add` block splits the decomposition in half `_left_building_block` and `_right_building_block`. where `_right_building_block` is the uncomputation part. you can create an add bloq with the appropiate parameters and do `yield from b._left_building_block()` and `yield from b._right_building_block()`  instead of reimplementing the decomposition.

fpapa250: Two problems with this:
1. b._right_building_block() isn't the same as the uncomputation of b._left_building_block() because it is in place addition (one of the CX gates is on the wrong qubit pair for full uncomputation
2. cirq.CX() does not support classical simulation and I think we want this for testing purposes

tanujkhattar: > cirq.CX() does not support classical simulation and I think we want this for testing purposes

The solution to this should be to replace the usages of `cirq.CX` with `CNOT().on_registers()` because the `Bloq` API now supports `on_registers` to yield a cirq operation. Cirq style decompositions should continue to be supported as first class citizens in the library.

fdmalone: The CX will be removed in #https://github.com/quantumlib/Qualtran/pull/585 fyi

fpapa250: After https://github.com/quantumlib/Qualtran/pull/585 is merged I will modify GreaterThan to use add.left_building_block(). Is it okay to keep the right side as-is because the uncomputation operation is different or do you have any suggestions on shortening the right uncomputation?

fdmalone: I need to look more carefully, but from the comments, is the issue that there is a bug in the Add right_building_block or just that it is necessarily different?

fpapa250: It is necessarily different. At the end of Add(x, y): y=x+y. For GreaterThan(x,y,target) y needs to be the same so the uncomputation differs by the control/target of one CNOT.

fdmalone: Right, that makes sense then.

NoureldinYosri: I don't want to block this PR but I think we should revisit this and do some refactoring to avoid having two slightly different variations of the same decomposition, but that can be a follow up.

a larger question is about the different decomposition of the same bloq. this decomposition uses 4n Ts but is linear in depth. the older estimate of 8n is based on what we did for the LessThanEqual bloq and at that time we prefered the logarithmic depth decomposition (8n Ts, log(n) depth) over the (4n Ts, n depth). I keep trying to put a proposal to track these but nothing feels good enough. anyway I will propose something asap

qualtran/bloqs/arithmetic/comparison.py
```diff
@@ -465,50 +467,172 @@ def _has_unitary_(self):
...
+
+        # Return the output registers.
+        return {'a': a, 'b': b, 'target': target}
 
-    def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
-        # TODO Determine precise clifford count and/or ignore.
-        # See: https://github.com/quantumlib/Qualtran/issues/219
-        # See: https://github.com/quantumlib/Qualtran/issues/217
-        t_complexity = self.t_complexity()
-        return {(TGate(), t_complexity.t)}
```
NoureldinYosri, line 511: @mpharrigan I remember in an older comment you said the call graph should report the complexity in the most natural way possible or something like that. are we still following that?

here -and in all arithmatic operations- the natural way would be to report it in Toffoli count rather than T count. since that will affect the resource estimation result if CZ magic factories are used instead of T factories.

qualtran/bloqs/arithmetic/comparison.py
fpapa250: Do the chemistry bloqs dependent on `GreaterThan()` prefer the logarithmic depth decomposition? If that's the case then it would make fixing the tests simpler - maintain a way to add `GreaterThan()` with logarithmic depth decomposition.

I think having both implementations is preferable anyway since it seems in the paper we want to show that Qualtran can be used to compare differing implementations of the same algorithm to compare various resource-optimized metrics.

Not sure on the best way to organize them, but if that's what we want I can just create a second `GreaterThan()` bloq with the optimal T-count decomposition and file a tracking issue to go back and combine these bloqs when an idea for bloqs with differently optimized decompositions is proposed. LMK what the preferred resolution is.

fdmalone: I think we should have both implementations to maintain the current order until we find a neat way to switch between decomposition strategies (or not). Actually what you've done is a step towards making the chemistry costs more consistent with the quoted literature values (#235), currently the tests have several "corrections" to use 4n instead of 8n (the current cost), but to avoid some accounting pain it's probably best to maintain the current costs until we can replace all the comparators (rather than 2 less thans having a certain cost and a greater than having a different cost).

fdmalone: I'm also happy to make the changes to the chemistry unit tests to reflect this though, if we don't want to maintain both.

mpharrigan: > I don't want to block this PR

yeah, I'd prefer to merge-then-organize than vice versa.

> I keep trying to put a proposal to track these

do we have an issue tracking this? I think we might. Can you update it with your latest thoughts?

> I think we should have both implementations to maintain the current order until

agree

> prefered the logarithmic depth decomposition

If you forced me to choose, I think for the regime we most care about when we have like 1 T-factory, the T _count_ is more important than the T _depth_. There's no consensus in the literature on this cost model, so we should have flags or whatever to let the user decide what they want.

qualtran/bloqs/arithmetic/comparison.py
mpharrigan: The call graph "callees" should be as high-level as possible. The base arithmetic operations should report toffoli (probably) and the arithmetic operations that are built on top of other arithmetic operations should use those.

fpapa250: Closing this since it is in reference to Log Depth GreaterThan()

qualtran/bloqs/arithmetic/comparison.py
fpapa250: For now I created `LinearDepthGreaterThan()` separate from `GreaterThan()`.

NoureldinYosri (approved):

PR #626, opened 2024-02-04
[data types] Deprecate selection register

fdmalone:
Deprecates SelectionRegister in favour of BoundedQUInt Registers. It's a rather big diff but apart from data_types, registers and maybe unary_iteration it's pretty much just search and replace. I only modified SelectionRegisters in this PR deliberately (e.g. chemistry bloqs other regs still  need updating). 

xref #446


tanujkhattar (approved):
Awesome! Glad to see this happen finally!

Issue #549, opened 2023-11-30
Select/Prepare Oracles


fdmalone:
This might be a misunderstanding on my part, but currently we expect that the SelectOracle and PrepareOracle selection registers should match up (in terms of bitsizes) (this is assumed in #508), AND that the selection registers (from prepare) are those registers that we want to reflect about for constructing the walk operator. However, this leads to some awkward cases:

1. We typically perform a state preparation over a register (sigma) for alias sampling which needs to be reflected on, so it should be a selection register. Now Select would require this register as part of its signature (but won't act on it) and require knowledge of the bitsize.
2. For sparse state preparation we use a flat register to index non-zero matrix elements and the store the indices in another register (called $|ind_j\rangle$), this ind register doesn't need to be reflected on BUT the ind_j register is used a selection register for SELECT, so again there's some inconsistency.

A possible workaround would be to define black boxes which partition the incoming registers and only select the ones absolutely necessary for the gate, and then prepare should probably define a registers_for_reflection?

fdmalone:
cc @tanujkhattar @mpharrigan

mpharrigan:
What's the essence of the problem? We use PREPARE to create a state but certain parts of that state are reflect-only or use-as-index-only? And the current system assumes the prepared state is both the thing we want to reflect and the indices used for select?

mpharrigan:
is my understanding of the issue (close to) correct?

fdmalone:
yes that's right

PR #627, opened 2024-02-05
Adding controlled state preparation using rotations

Vicara12:
This code implements the last two points in [this comment](https://github.com/quantumlib/Qualtran/issues/596#issuecomment-1922366568) from issue [#596 ](https://github.com/quantumlib/Qualtran/issues/596). It adds a prepare oracle with a control qubit that prepares an arbitrary state given a list of its coefficients using the technique described in page 3 of the paper [Trading T-gates for dirty qubits in state preparation and unitary synthesis](https://arxiv.org/abs/1812.00954).

Despite Qualtran already having a prepare oracle that uses alias sampling, that method produces entangled junk qubits and also does not include a control. These two factors are an inconvenience for some applications and this technique solves both problems.

This code is a necessary part for the goal of implementing a MPS preparation algorithm, even though it is useful on its own for other applications.

Regarding the tests, they basically tests both states that can be prepared exactly and approximately (because the rotation angles have limited resolution, in many cases the result is an approximation of the desired state). It also tests that the control works as expected both for the states 0, 1 and +.


qualtran/bloqs/controlled_state_preparation.ipynb
```diff
@@ -0,0 +1,240 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#  Copyright 2023 Google LLC\n",
```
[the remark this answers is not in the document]
tanujkhattar: Good point, we do that in Cirq. Should probably do it in notebooks as well

qualtran/bloqs/controlled_state_preparation.ipynb
```diff
@@ -0,0 +1,240 @@
...
+    "Assume one wants to implement an arbitrary state whose coefficients are stored in `state_coefs` with `n_qubits` sites using a resolution for the rotations of `rot_reg_size` qubits."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def genRandomState (sites):\n",
```
fdmalone, line 70: nit: use snake case for function names, also add type hints

qualtran/bloqs/controlled_state_preparation.ipynb
```diff
@@ -0,0 +1,240 @@
...
+    "  state = [random.uniform(-1,1) + random.uniform(-1,1)*1j for _ in range(2**sites)]\n",
+    "  norm = np.sqrt(sum([abs(s)**2 for s in state]))\n",
+    "  return [s/norm for s in state]\n",
```
fdmalone, line 73: nit: just np.array(state) / norm?

qualtran/bloqs/controlled_state_preparation.ipynb
```diff
@@ -0,0 +1,240 @@
...
+    "\n",
+    "show_bloq(circuit)\n",
+    "coefficients = circuit.tensor_contract()\n",
+    "\n",
+    "accuracy = coefficients[0] # <coefficients|0> = coefficients[0]\n",
+    "print(f\"accuracy: {abs(accuracy)}\\n\")\n",
+    "\n",
+    "print(\"Coefficients in polar form:\")\n",
+    "for i, c in enumerate(coefficients):\n",
+    "  print(f\"  |{i:0{n_qubits}b}> result: {round(abs(c),4)} ∠{round(np.angle(c, deg=True),2)}º\")"
```
fdmalone, line 215: Why are the "angle" values here so odd? I get

Coefficients in polar form:
  |000> result: 1.0 ∠0.0º
  |001> result: 0.0 ∠-115.57º
  |010> result: 0.0 ∠-64.32º
  |011> result: 0.0 ∠57.47º
  |100> result: 0.0 ∠-53.75º
  |101> result: 0.0 ∠17.61º
  |110> result: 0.0 ∠37.75º
  |111> result: 0.0 ∠-150.2º

qualtran/bloqs/controlled_state_preparation.py
```diff
@@ -0,0 +1,349 @@
...
+from qualtran.bloqs.arithmetic import Add
+from qualtran.bloqs.basic_gates import OneEffect, OneState, ZeroEffect, ZeroState
+from qualtran.bloqs.basic_gates.rotation import Rx
+from qualtran.bloqs.qrom import QROM
+from qualtran.bloqs.select_and_prepare import PrepareOracle
+
+
+@attrs.frozen
+class ControlledStatePreparationUsingRotations(PrepareOracle):
+    r"""Class that implements controlled state preparation using Ry and Rz rotations from [1]. It
```
fdmalone, line 31: nit: The class summary should be on line long (no line breaks) You can break it in two (separated by a blank line). "Class" is redundant here. Maybe just, Controlled state preparation using ....

qualtran/bloqs/controlled_state_preparation.py
```diff
@@ -0,0 +1,349 @@
...
+    References:
+        [Trading T-gates for dirty qubits in state preparation and unitary synthesis]
+        (https://arxiv.org/abs/1812.00954).
+            Low, Kliuchnikov, Schaeffer. 2018.
+    """
+
+    n_qubits: int
+    rot_reg_size: int
+    state: Tuple
+    adjoint: bool = False
```
fdmalone, line 58: I think this has to be called something else so as to not clash with the bloqs adjoint method (see and_bloq which has a special uncompute flag.)

qualtran/bloqs/controlled_state_preparation.py
```diff
@@ -0,0 +1,349 @@
...
+            control, target_state, rot_reg, phase_gradient = self.__preparePhases(
+                rom_vals, bb, control, target_state, rot_reg, phase_gradient
+            )
+        # deallocate rotation register's qubits
+        qs = bb.split(rot_reg)
+        for q in qs:
+            bb.add(ZeroEffect(), q=q)
+        return {"control": control, "target_state": target_state, "phase_gradient": phase_gradient}
+
+    def __prepareAmplitudes(
```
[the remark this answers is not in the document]
fdmalone: Also snake_case not camelCase

qualtran/bloqs/controlled_state_preparation.py
```diff
@@ -0,0 +1,349 @@
...
+            selection=target_state,
+            qubit=rot_ancilla,
+            rot_reg=rot_reg,
+            phase_gradient=phase_gradient,
+        )
+        bb.add(OneEffect(), q=rot_ancilla)
+        return control, target_state, rot_reg, phase_gradient
+
+    def __getPhaseROMValues(self, amplitude_rom_vals):
+        """As we are using the equivalent to controlled Z to do the rotations instead of Rz, there
```
fdmalone, line 174: Single summary line followed by more descriptive docstring.

qualtran/bloqs/controlled_state_preparation.py
```diff
@@ -0,0 +1,349 @@
...
+
+    def __preparePhases(
+        self,
+        amplitude_rom_vals: ArrayLike,
+        bb: BloqBuilder,
+        control: SoquetT,
+        target_state: SoquetT,
+        rot_reg: SoquetT,
+        phase_gradient: SoquetT,
+    ):
```
fdmalone, line 158: Add a docstring.

qualtran/bloqs/controlled_state_preparation.py
```diff
@@ -0,0 +1,349 @@
...
+            angles = [offset - np.angle(c) for c, offset in zip(self.state, offset_angles)]
+        else:
+            angles = [np.angle(c) - offset for c, offset in zip(self.state, offset_angles)]
+        rom_values = [RotationTree.angle2RomValue(a, self.rot_reg_size) for a in angles]
+        return rom_values
+
+
+@attrs.frozen
+class ControlledQROMRotateQubit(Bloq):
+    r"""Class that performs an array of controlled rotations $Z^{\theta_i/2}$ for a list of angles
```
fdmalone, line 200: ditto about single line followed by more detailed summary.

qualtran/bloqs/controlled_state_preparation.py
```diff
@@ -0,0 +1,349 @@
...
+        return Signature.build(
+            prepare_control=1,
+            selection=self.n_selections,
+            qubit=1,
+            rot_reg=self.rot_reg_size,
+            phase_gradient=self.rot_reg_size,
+        )
+
+    def build_composite_bloq(self, bb: BloqBuilder, **soqs: SoquetT) -> Dict[str, SoquetT]:
+        """Parameters:
```
fdmalone, line 236: I think it's clearer to not use **soqs and just spell out the args.

qualtran/bloqs/controlled_state_preparation.py
```diff
@@ -0,0 +1,349 @@
...
+            )
+        else:
+            qrom_control, soqs["rot_reg"] = bb.add(
+                qrom, control=qrom_control, target0_=soqs["rot_reg"]
+            )
+        return qrom_control, soqs
+
+
+class RotationTree:
+    r"""Class used by ControlledStatePreparationUsingRotations to get the corresponding rotation
```
fdmalone, line 277: same comment about docstring style.

qualtran/bloqs/controlled_state_preparation.py
```diff
@@ -0,0 +1,349 @@
...
+
+    References:
+        [Trading T-gates for dirty qubits in state preparation and unitary synthesis]
+        (https://arxiv.org/abs/1812.00954).
+            Low, Kliuchnikov, Schaeffer. 2018.
+    """
+
+    @staticmethod
+    def extractRomValuesFromState(state: ArrayLike, rot_reg_size: int, adjoint: bool = False):
+        r"""Gives list in which the ith element is a list of the rom values to be loaded when
```
fdmalone, line 292: ditto.

qualtran/bloqs/controlled_state_preparation.py
```diff
@@ -0,0 +1,349 @@
...
+
+    @staticmethod
+    def angle2RomValue(angle, rot_reg_size):
+        r"""Returns the value to be loaded to a QROM to encode the given angle with a certain value
+        of rot_reg_size.
+        """
+        rom_value_decimal = 2**rot_reg_size * angle / (2 * np.pi)
+        return round(rom_value_decimal) % (2**rot_reg_size)
+
+    # do not call, use rotationTreeFromState
```
fdmalone, line 338: Could this be a frozen class?

qualtran/bloqs/controlled_state_preparation.py
```diff
@@ -0,0 +1,349 @@
...
+                rom_vals_this_layer.append(rom_val)
+                if tree.branch0 is not None:
+                    next_layer.append(tree.branch0)
+                if tree.branch1 is not None:
+                    next_layer.append(tree.branch1)
+            rom_vals.append(rom_vals_this_layer)
+        return rom_vals
+
+    @staticmethod
+    def rotationTreeFromState(state):
```
fdmalone, line 316: this is a factory method, mabye build_rotation_tree_from_state?

qualtran/bloqs/controlled_state_preparation.py
```diff
@@ -0,0 +1,349 @@
...
+        self.sum0 = sum0
+        self.sum1 = sum1
+        self.sum_total = sum0 + sum1
+        self.branch0 = branch0
+        self.branch1 = branch1
+
+    def __getP0(self):
+        if self.sum_total == 0:
+            return 0
+        return self.sum0 / self.sum_total
```
fdmalone, line 349: have a look in say and_bloq for building examples test cases which are good for testing the bloq protocols (`@bloq_example`).

You can also add to our autodocumentation tool: dev_tools/autogenerate-bloqs-notebooks-v2.py which helps document the bloqs separate from your nice tutorial. This is helpful to check your docs are rendering properly. This shouldn't be a blocker but something to think about.

fdmalone (changes requested):
Great contribution! I left some initial comments which are mostly style based. I need to read through the paper but you have tests so this is very close to getting merged.

qualtran/bloqs/controlled_state_preparation.ipynb
Vicara12: It is normal, because the amplitude of the complex value is zero (or a value of the order of 1e-16) I think `np.angle` just outputs some random value.

qualtran/bloqs/controlled_state_preparation.py
Vicara12: I've seen that some classes, such as `ReflectionUsingPrepare`, need the adjoint of a gate. Aren't they using this parameter? How to I change the behavior of the gate then?

qualtran/bloqs/controlled_state_preparation.py
Vicara12: I used `**soqs` because in some cases the `selection` soquet is not used (as when preparing the amplitude of the first qubit). I tried calling the function without it and it complained that a selection soquet must be provided. Is there a cleaner alternative to this?

qualtran/bloqs/controlled_state_preparation.py
```diff
@@ -0,0 +1,349 @@
...
+import numpy as np
+from numpy.typing import ArrayLike
+
+from qualtran import Bloq, BloqBuilder, SelectionRegister, Signature, SoquetT
+from qualtran.bloqs.arithmetic import Add
+from qualtran.bloqs.basic_gates import OneEffect, OneState, ZeroEffect, ZeroState
+from qualtran.bloqs.basic_gates.rotation import Rx
+from qualtran.bloqs.qrom import QROM
+from qualtran.bloqs.select_and_prepare import PrepareOracle
+
```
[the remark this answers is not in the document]
Vicara12: The problem with the paper is that it is not really about state preparation, but mentions it and gives the main idea behind the algorithm (the best I could find). Many of the things that are used in the code are not really explained there but referenced to other huge papers where it is nowhere to be found. So I had to figure them out myself or with the help of @StefanoPolla. I will add an outline of what it does though, because as it is it might be a bit hard to understand.

qualtran/bloqs/controlled_state_preparation.py
Vicara12: Also, `PhaseGradientState` has the attribute `adjoint`, in case that is a bug or something.

qualtran/bloqs/controlled_state_preparation.py
Vicara12: I tried adding my bloq to the autogen tool but it did nothing. Actually, I think that most of the bloqs there are not being generated, or at least it does not seem to be the case from the script's output. The commits are [aa82ece](https://github.com/quantumlib/Qualtran/pull/627/commits/aa82ece16af2e38ec512a03af25991960c836ba4) and [35e45c8](https://github.com/quantumlib/Qualtran/pull/627/commits/35e45c89877c6af45544b65fba755f5bc1b88fc7).

qualtran/bloqs/controlled_state_preparation.py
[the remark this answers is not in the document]
StefanoPolla: To get a short explanation we could connect the QROM call to the first line of Eq.8 (loading the binary representation of $\theta$ depending on $|y\rangle$), the controlled-rotation to the second line, and the QROM computation to the third.

The additional control needs a mention but no special explanation.

qualtran/bloqs/controlled_state_preparation.py
fdmalone: Yeah I think this is something that will go away in the future: see [here for example](https://github.com/quantumlib/Qualtran/pull/583/files#diff-70447a4c3b0e7c3f6d090278b95bddb69e98e009b70a54f62d8732dcc13736f3). If you look at the [And bloq](https://github.com/quantumlib/Qualtran/blob/main/qualtran/bloqs/and_bloq.py#L78) as an example, it defines the flag as uncompute (other places it's been updated to is_adjoint). And then you would have to update the bloq method [adjoint ](https://github.com/quantumlib/Qualtran/blob/main/qualtran/bloqs/and_bloq.py#L89). For the moment because it's not throwing an error you can probably ignore this, and we can clear it up later. Indeed this hasn't been uniformly updated throughout qualtran as you point out.

qualtran/bloqs/controlled_state_preparation.py
fdmalone: Could you try the pattern from [here](https://github.com/quantumlib/Qualtran/blob/main/qualtran/bloqs/and_bloq.py#L213), i.e., I think something like

``` python

@bloq_example
def _controlled_state_prep():
    state_coefs = (
        (-0.42677669529663675 - 0.1767766952966366j),
        (0.17677669529663664 - 0.4267766952966367j),
        (0.17677669529663675 - 0.1767766952966368j),
        (0.07322330470336305 - 0.07322330470336309j),
        (0.4267766952966366 - 0.17677669529663692j),
        (0.42677669529663664 + 0.17677669529663675j),
        (0.0732233047033631 + 0.17677669529663678j),
        (-0.07322330470336308 - 0.17677669529663678j),
    )
    controlled_state_prep = ControlledStatePreparationUsingRotations(n_qubits=2, rot_reg_size=4, state=state_coefs)
    return controlled_state_prep
```
(the names have to match up to the _ which is stripped later on). Note to self: this should be documented better.

fdmalone: As to why it's helpful, you can add [unit tests](https://github.com/quantumlib/Qualtran/blob/main/qualtran/bloqs/and_bloq_test.py#L30), which will run through several checks on your bloq which can be very helpful.

qualtran/bloqs/controlled_state_preparation.py
```diff
@@ -0,0 +1,469 @@
...
+        """
+        Encodes the phase of each coefficient, taking into account both the phase of the original
+        coefficient and offsets caused by the amplitude preparation.
+
+        It applies a rotation to the target_state register through phase kickback. By using
+        target_register as the selection register for rotating an ancilla qubit that is then
+        erased, leaving the desired phase in target_register.
+
+        Args:
+            - amplitude_rom_vals: list of ROM values that correspond to the angles of the phases to
```
fdmalone, line 237: nit: remove the - here, it won't play nice with the autogenerated docs.

qualtran/bloqs/controlled_state_preparation.py
```diff
@@ -0,0 +1,469 @@
...
+@attrs.frozen
+class ControlledQROMRotateQubit(Bloq):
+    r"""Array of controlled rotations $Z^{\theta_i/2}$ for a list of angles $\theta$.
+
+    It uses phase kickback and thus needs a phase gradient state in order to work. This
+    state must be provided externally for efficiency, as it is unaffected and can thus be reused.
+    Refer to [1], section on arbitrary quantum state preparation on page 3.
+
+    Args:
+        n_selections: number of qubits used for encoding the selection register of the QROM, it
```
fdmalone, line 319: nit: this is called selection_bitsizes elsewhere in the library. It's helpful to keep it consistent.

fdmalone: If you're eager you could make this change to other bitsize variables.

qualtran/bloqs/controlled_state_preparation.py
```diff
@@ -0,0 +1,469 @@
...
+        return RotationTree(dn_l.sum_total, dn_r.sum_total, dn_l, dn_r)
+
+    @property
+    def angle_0(self):
+        r"""Angle that corresponds to p_0."""
+        return 2 * np.arccos(np.sqrt(self._p0))
+
+    @staticmethod
+    def angle_2_ROM_value(angle, rot_reg_size):
+        r"""Returns the value to be loaded to a QROM to encode the given angle with a certain value
```
fdmalone, line 451: nit: single line summaries.

qualtran/bloqs/controlled_state_preparation.ipynb
fdmalone: right, you have a bunch of zeros with different relative signs. I think adding

``` python
# zero out small coefficients
coefficients[np.where(abs(coefficients) < 1e-12)] = 0.0
```
right before your print is probably a good idea. Also it looks like you're using 2 spaces instead of 4 in the notebook, I guess black doesn't check the notebooks.

qualtran/bloqs/controlled_state_preparation.py
fdmalone: also sorry for the nits but I think we aim for all lowercase variable names / function names. I would also replace _2_ with _to_.

qualtran/bloqs/controlled_state_preparation.py
Vicara12: Now it works! But the result looks a bit weird, because there is first the tutorial and then a giant block of documentation without any context. Maybe it would be a good idea to rename the tutorial as `controlled_state_preparation_tutorial.ipynb` and leave the autogen documentation as a separate file? This I think looks much cleaner.

Vicara12: Okay so I figured out why the autogen doc wasn't working. It turns out that the command in the file that indicates how to run the script points to an older version. https://github.com/quantumlib/Qualtran/blob/061f17269ea8f06e46bf65bd2e81ab86e0e2bf15/dev_tools/autogenerate-bloqs-notebooks-v2.py#L41-L42
A `-v2` is missing from it. I copy-pasted it without checking and what was running was the old version. I fixed it in the most recent commit.

fdmalone: right, we normally separate tutorials out from the bloq documentation if there is a detailed tutorial.

fdmalone:
@tanujkhattar could you take a glance as I know you have more context with rotations and phase gradients.

tanujkhattar:
@fdmalone Thanks for the detailed reviews. I'll take a look at this PR by EOD today.

qualtran/bloqs/controlled_state_preparation.py
```diff
@@ -0,0 +1,483 @@
...
+    References:
+        [Trading T-gates for dirty qubits in state preparation and unitary synthesis]
+        (https://arxiv.org/abs/1812.00954).
+            Low, Kliuchnikov, Schaeffer. 2018.
+    """
+
+    state_bitsizes: int
+    rot_reg_bitsizes: int
+    state: Tuple
+    uncompute: bool = False
```
tanujkhattar: Do we imagine a scenario where `len(state) != 2^{state_bitsize}` ? If not, then we can only ask the user to specify the `state: Tuple[complex, ...]` and we can infer the required bitsize for the register using `(len(state) - 1).bit_length()`

Another nit would be to rename the attribtutes as follows:
```suggestion
    state_bitsize: int
    state_coefficients: Tuple[complex, ...]
    phase_bitsize: int
    uncompute: bool = False
```

qualtran/bloqs/controlled_state_preparation.py
```diff
@@ -0,0 +1,483 @@
...
+
+    @property
+    def selection_registers(self) -> Tuple[SelectionRegister, ...]:
+        return (
+            SelectionRegister(
+                "target_state", bitsize=self.state_bitsizes, iteration_length=self.state_bitsizes
```
tanujkhattar: The iteration length should be `len(self.state)` (or `2 ** self.state_bitsize`)

fdmalone (approved):

fdmalone:
I just committed a patch for the migration #661  and the deprecation of selection registers #626. I think it's probably good to merge this now and follow up with anything later given all the churn from data types etc.

tanujkhattar (changes requested):

qualtran/bloqs/state_preparation/controlled_state_preparation.py
```diff
@@ -0,0 +1,484 @@
...
+)
+from qualtran.bloqs.arithmetic import Add
+from qualtran.bloqs.basic_gates import OneEffect, OneState, ZeroEffect, ZeroState
+from qualtran.bloqs.basic_gates.rotation import Rx
+from qualtran.bloqs.qrom import QROM
+from qualtran.bloqs.select_and_prepare import PrepareOracle
+
+
+@attrs.frozen
+class ControlledStatePreparationUsingRotations(PrepareOracle):
```
tanujkhattar, line 100:
```suggestion
class ControlledStatePreparationUsingRotations(GateWithRegisters):
```

qualtran/bloqs/state_preparation/controlled_state_preparation.py
```diff
@@ -0,0 +1,484 @@
...
+                "target_state",
+                BoundedQUInt(bitsize=self.state_bitsizes, iteration_length=self.state_bitsizes),
+            ),
+        )
+
+    @property
+    def signature(self):
+        return Signature.build(
+            control=1, target_state=self.state_bitsizes, phase_gradient=self.rot_reg_bitsizes
+        )
```
tanujkhattar, line 142: We'd have to update the interface of `PrepareOracle` to account for `control_registers` and `extra_registers` cached properties so that the call sites that expect a `PrepareOracle` know how to wire it up properly. For the sake of this PR, I'd suggest we don't derive from `PrepareOracle` and tackle the change in a follow up PR

fdmalone: I would be interested in revisiting these oracles: see #549

qualtran/bloqs/state_preparation/controlled_state_preparation.py
```diff
@@ -0,0 +1,484 @@
...
+            num_controls=2,
+        )
+        # both prepare_control and qubit will control the QROM so that this acts as a control Z
+        qrom_control = bb.join(np.array([soqs["prepare_control"], soqs["qubit"]]))
+        # load angles in rot_reg (line 1 of eq (8) in [1])
+        qrom_control, soqs = self._apply_qrom(qrom, bb, qrom_control, soqs)
+        # phase kickback via phase_grad += rot_reg (line 2 of eq (8) in [1])
+        soqs["rot_reg"], soqs["phase_gradient"] = bb.add(
+            Add(bitsize=self.rot_reg_bitsizes), a=soqs["rot_reg"], b=soqs["phase_gradient"]
+        )
```
tanujkhattar, line 375: We also have a `AddIntoPhaseGrad` gate which has a slightly better Toffoli complexity and can be used here after slight modifications. 

Add the following function to `AddIntoPhaseGrad` 

```python
    def add_my_tensors(
        self,
        tn: 'qtn.TensorNetwork',
        tag: Any,
        *,
        incoming: Dict[str, 'SoquetT'],
        outgoing: Dict[str, 'SoquetT'],
    ):
        import quimb.tensor as qtn

        N, M = 2**self.inp_bitsize, 2**self.phase_bitsize
        inds = (incoming['x'], incoming['phase_grad'], outgoing['x'], outgoing['phase_grad'])
        unitary = np.zeros((N, M) * 2, dtype=np.complex128)
        for a, b in itertools.product(range(N), range(M)):
            unitary[a, b, a, int(math.fmod(a + b, M))] = 1

        tn.add(qtn.Tensor(data=unitary, inds=inds, tags=[self.short_name(), tag]))
```

and modify this code as 

```python
        soqs["rot_reg"], soqs["phase_gradient"] = bb.add(
            AddIntoPhaseGrad(self.rot_reg_bitsizes, self.rot_reg_bitsizes),
            x=soqs["rot_reg"],
            phase_grad=soqs["phase_gradient"],
        )
```

qualtran/bloqs/state_preparation/controlled_state_preparation.py
```diff
@@ -0,0 +1,484 @@
...
+                control, sel, state_qubits[qi], rot_reg, phase_gradient = bb.add(
+                    ctrl_rot_q,
+                    prepare_control=control,
+                    selection=sel,
+                    qubit=state_qubits[qi],
+                    rot_reg=rot_reg,
+                    phase_gradient=phase_gradient,
+                )
+                state_qubits[:qi] = bb.split(sel)
+            state_qubits[qi] = bb.add(Rx(angle=-np.pi / 2), q=state_qubits[qi])
```
tanujkhattar, line 215: I see a bunch of code is duplicated mainly to pass the right set of inputs to the bloq. You can rewrite this function in a more succinct way by using the helper `bb.add_d` instead of `bb.add` and using a dictionary to keep track of arguments. For example:

```python
    def _prepare_amplitudes(
        self,
        rom_vals: ArrayLike,
        bb: BloqBuilder,
        control: SoquetT,
        target_state: SoquetT,
        rot_reg: SoquetT,
        phase_gradient: SoquetT,
    ):
        # if it is the adjoint gate, load the modular negative values to undo the rotations that
        # loaded the amplitudes
        if self.uncompute:
            rom_vals = RotationTree(self.state).get_rom_values(
                self.rot_reg_bitsizes, uncompute=True
            )
        state_qubits = bb.split(target_state)
        soqs = {"prepare_control": control, "rot_reg": rot_reg, "phase_gradient": phase_gradient}
        for i in range(self.state_bitsizes):
            # for the normal gate loop from qubit 0 to state_bitsizes-1, if it is the adjoint
            # then the process is run backwards with the opposite turn angles
            if self.uncompute:
                qi = self.state_bitsizes - i - 1
            else:
                qi = i
            ctrl_rot_q = ControlledQROMRotateQubit(qi, self.rot_reg_bitsizes, tuple(rom_vals[qi]))
            state_qubits[qi] = bb.add(Rx(angle=np.pi / 2), q=state_qubits[qi])
            if qi:
                # first qubit does not have selection registers, only controls
                soqs["selection"] = bb.join(state_qubits[:qi])
            soqs = bb.add_d(ctrl_rot_q, **soqs, qubit=state_qubits[qi])
            if qi:
                state_qubits[:qi] = bb.split(soqs.pop("selection"))
            state_qubits[qi] = bb.add(Rx(angle=-np.pi / 2), q=soqs.pop("qubit"))

        target_state = bb.join(state_qubits)
        return soqs["prepare_control"], target_state, soqs["rot_reg"], soqs["phase_gradient"]
```

qualtran/bloqs/state_preparation/controlled_state_preparation.py
```diff
@@ -0,0 +1,484 @@
...
+            phase_gradient=self.rot_reg_bitsizes,
+        )
+
+    def build_composite_bloq(self, bb: BloqBuilder, **soqs: SoquetT) -> Dict[str, SoquetT]:
+        """Parameters:
+        * prepare_control
+        * selection (not necessary if selection_bitsizes == 0)
+        * qubit
+        * rot_reg
+        * phase_gradient
```
tanujkhattar, line 360: Similarly, we can get rid of the `apply_qrom` helper and rewrite this in a more succinct way using `bb.add_d` method. 

```python
    def build_composite_bloq(self, bb: BloqBuilder, **soqs: SoquetT) -> Dict[str, SoquetT]:
        """Parameters:
        * prepare_control
        * selection (not necessary if selection_bitsizes == 0)
        * qubit
        * rot_reg
        * phase_gradient
        """
        qrom = QROM(
            [np.array(self.rom_values)],
            selection_bitsizes=(self.selection_bitsizes,),
            target_bitsizes=(self.rot_reg_bitsizes,),
            num_controls=2,
        )
        # both prepare_control and qubit will control the QROM so that this acts as a control Z
        qrom_control = bb.join(np.array([soqs.pop("prepare_control"), soqs.pop("qubit")]))
        phase_grad = soqs.pop("phase_gradient")
        # load angles in rot_reg (line 1 of eq (8) in [1])
        soqs = bb.add_d(qrom, control=qrom_control, target0_=soqs.pop("rot_reg"), **soqs)
        # phase kickback via phase_grad += rot_reg (line 2 of eq (8) in [1])
        soqs["target0_"], phase_grad = bb.add(
            AddIntoPhaseGrad(self.rot_reg_bitsizes, self.rot_reg_bitsizes),
            x=soqs["target0_"],
            phase_grad=phase_grad,
        )
        # uncompute angle load in rot_reg to disentangle it from selection register
        # (line 1 of eq (8) in [1])
        soqs = bb.add_d(qrom, **soqs)
        separated = bb.split(soqs.pop('control'))
        return {
            "prepare_control": separated[0],
            "qubit": separated[1],
            "phase_gradient": phase_grad,
            "rot_reg": soqs.pop('target0_'),
        } | soqs
```

qualtran/bloqs/state_preparation/controlled_state_preparation.py
```diff
@@ -0,0 +1,484 @@
...
+
+_CONTROLLED_STATE_PREP_DOC = BloqDocSpec(
+    bloq_cls=ControlledStatePreparationUsingRotations,
+    import_line='from qualtran.bloqs.state_preparation.controlled_state_preparation import ControlledStatePreparationUsingRotations',
+    examples=(_controlled_state_prep,),
+)
+
+
+@attrs.frozen
+class ControlledQROMRotateQubit(Bloq):
```
tanujkhattar, line 317: I think it would be nice to add a more general version of it in the same family as `ProgrammableRotationGateArray` (for example: support arbitrary number of controls, including 0) but we can do that as a follow up PR. For now, it'd be sufficient to open an issue and track improvements to this important primitive.

tanujkhattar (commented):
Thanks a lot once again Victor for this PR - state preparation via rotations is a super useful primitive and it's great to have it implemented in Qualtran.

I've left another round of comments for improving the code quality. I know this PR has been stalling for a while and apologies for the delayed review. However, I've also committed all my suggestions to my local branch. If you want, you can simply apply my commit to your PR and it should resolve most of my comments 

https://github.com/tanujkhattar/Qualtran/commit/9a182cfa4db8542b45901546480a6be82e8a5ae8

Once this is done, we can go ahead and merge the PR.

qualtran/bloqs/controlled_state_preparation.py
```diff
@@ -0,0 +1,483 @@
...
+)
+from qualtran.bloqs.arithmetic import Add
+from qualtran.bloqs.basic_gates import OneEffect, OneState, ZeroEffect, ZeroState
+from qualtran.bloqs.basic_gates.rotation import Rx
+from qualtran.bloqs.qrom import QROM
+from qualtran.bloqs.select_and_prepare import PrepareOracle
+
+
+@attrs.frozen
+class ControlledStatePreparationUsingRotations(PrepareOracle):
```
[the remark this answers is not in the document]
Vicara12: Maybe it is a good idea to do not have `is_controlled` and document that it is possible to have an uncontrolled version if `control_bitsize = 0`? It feels a bit redundant and would require an extra check to ensure that if `is_controlled` is true then `control_bitsize` must be greater than zero.

qualtran/bloqs/controlled_state_preparation.py
Vicara12: Silly question: as far as I've seen, no bloq has a constructor. Where should I place code that must be called when the object is initialized? I also want to run some checks on the parameters to ensure the user utilizes the bloq correctly, but so far I have no idea where to place it.

qualtran/bloqs/controlled_state_preparation.py
Vicara12: What does iteration length do exactly? Maybe I understood it wrong

qualtran/bloqs/state_preparation/controlled_state_preparation.py
```diff
@@ -0,0 +1,484 @@
...
+                qrom, control=qrom_control, selection=soqs["selection"], target0_=soqs["rot_reg"]
+            )
+        else:
+            qrom_control, soqs["rot_reg"] = bb.add(
+                qrom, control=qrom_control, target0_=soqs["rot_reg"]
+            )
+        return qrom_control, soqs
+
+
+# @attrs.frozen
```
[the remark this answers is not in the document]
Vicara12: Great idea! I suppose that accessing array items is way faster than the old recursive version. Also, is there any reason to use bit logic instead of `2*i` and `2*i + 1`? I would assume that the assembly produced ends up being the same.

qualtran/bloqs/state_preparation/controlled_state_preparation.py
```diff
@@ -0,0 +1,484 @@
...
+        return Signature.build(
+            control=1, target_state=self.state_bitsizes, phase_gradient=self.rot_reg_bitsizes
+        )
+
+    def build_composite_bloq(
+        self, bb: BloqBuilder, *, control: SoquetT, target_state: SoquetT, phase_gradient: SoquetT
+    ) -> Dict[str, SoquetT]:
+        rom_vals = RotationTree.extract_rom_values_from_state(self.state, self.rot_reg_bitsizes)
+        # allocate the qubits for the rotation angle register
+        rot_reg = bb.join(np.array([bb.add(ZeroState()) for _ in range(self.rot_reg_bitsizes)]))
```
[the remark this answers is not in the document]
Vicara12: Absolutely, I did that nasty trick because when I programmed it `allocate` and `free` didn't have a block decomposition.

qualtran/bloqs/state_preparation/controlled_state_preparation.py
Vicara12: I created the [issue](https://github.com/quantumlib/Qualtran/issues/676) but can't assign it to myself.

qualtran/bloqs/controlled_state_preparation.py
tanujkhattar: > as far as I've seen, no bloq has a constructor

That's because most Bloqs are annotated with `@attr.frozen` which automatically generates a constructor for the Bloq. And the auto generated constructor is super simple - just expects and assigns values to `self.<attribute_name>` where `<attribute_name>` corresponds to the list of attributes we define on top of the bloq. In this case, `state_bitsize` etc. See https://www.attrs.org/en/stable/examples.html for more details. 

The philosophy is that we almost always want to keep the constructors super simple (no fancy logic) and using the auto generated constructors enforces this along with other nice things like the generated class is hashable etc.

> I also want to run some checks on the parameters to ensure the user utilizes the bloq correctly

You can use validators for this purpose. See https://www.attrs.org/en/stable/examples.html#val and https://github.com/quantumlib/Qualtran/blob/28c0c1d3651120f3bb2f25dd40346ae4e94585e0/qualtran/bloqs/qrom.py#L85 for examples of how it's done.

qualtran/bloqs/state_preparation/state_preparation_via_rotation.py
```diff
@@ -0,0 +1,443 @@
...
+    rot_reg_bitsizes: int
+    rom_values: Tuple[int, ...]
+    control_bitsize: int
+
+    @property
+    def signature(self) -> Signature:
+        return Signature.build(
+            prepare_control=self.control_bitsize,
+            selection=self.selection_bitsizes,
+            qubit=1,
```
tanujkhattar, line 336: I think we should the `qubit` register with the `control_bitsize` argument so that we combine the `prepare_control` and `qubit` into a single register before calling this Bloq. It's not clear to me from the interface of this Bloq that why do we need a separate control qubit ?

The goal of this bloq is to
1) Load N multiplexed rotation angles via QROM in a selection register

$$
|x\rangle|0\rangle \rightarrow |x\rangle|\theta_x\rangle
$$

2) Apply multiplexed rotations on the target register by the loaded angle, i.e. 

$$
|x\rangle|\theta_{x}\rangle \rightarrow e^{i \pi \theta_x / 2} |x\rangle |\theta_{x}\rangle
$$

3) Uncompute step-1 s.t. you get

$$
e^{i \pi \theta_x / 2} |x\rangle |\theta_{x}\rangle \rightarrow e^{i \pi \theta_x / 2} |x\rangle |0\rangle
$$


If users specify a `control_bitsize` then steps 1 & 3 can be made controlled (via controlled `QROM`) and therefore step-2 becomes controlled for free. 

There's no reason for us to put `qubit` as part of the signature of this Bloq since it's a problem specific requirement that we should always control the i'th step of our controlled rotations using `i-1`'th qubit. 

Thus, I propose the following changes:

1) Update the name of the class to something like `ProgrammableRotationGateArrayViaPhaseGradient` or `PRGAViaPhaseGradient` or `MultiplexedRotationsViaPhaseGradient` 
2) Change the signature of the class to 
```python
Signature.build(
    control=self.control_bitsize, 
    selection=self.selection_bitsize,
    phase_grad=self.rot_reg_bitsizes,
)
```
3. Note that my suggestion is to remove the `rot_reg` from the signature of the class. This is because the `rot_reg` is a temporary ancilla register that the class needs to allocate for loading QROM angles in order to achieve the required phasing. The `rot_reg` allocation and deallocation should happen inside the `build_composite_bloq` of this class.  The signature of the class then matches the 3 step description defined above.

tanujkhattar (commented):
This is looking great! A few more nits for improving the interface of  `ControlledQROMRotateQubit` and I think we'll be good to merge. 

Sorry for the repeated iterations, but I really think this is a great addition to the library so it's important to get it right. Really appreciate your patience here!

qualtran/bloqs/controlled_state_preparation.py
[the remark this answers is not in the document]
tanujkhattar: > I will have a look at CtrlSpec on Monday

I think we can merge as is and deal with `CtrlSpec` later. I don't want to block this PR by opening another can of worms.

qualtran/bloqs/state_preparation/controlled_state_preparation.py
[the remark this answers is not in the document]
tanujkhattar: > Also, is there any reason to use bit logic instead of 2*i and 2*i + 1

Not really. Just old habits die hard :)

qualtran/bloqs/state_preparation/state_preparation_via_rotation.py
```diff
@@ -0,0 +1,443 @@
...
+
+
+class RotationTree:
+    r"""Used by `StatePreparationUsingRotations` to get the corresponding rotation angles.
+
+    The rotation angles are used to encode the amplitude of a state using the method described in
+    [1], section on arbitrary quantum state preparation, page 3.
+
+    The only methods to be used externally are extract_rom_values_from_state, angle_to_rom_value and
+    build_rotation_tree_from_state.
```
StefanoPolla, line 391: the names of the methods in the docstrings don't match the new names of the methods

qualtran/bloqs/state_preparation/state_preparation_via_rotation.py
```diff
@@ -0,0 +1,443 @@
...
+    SoquetT,
+)
+from qualtran.bloqs.basic_gates import XGate
+from qualtran.bloqs.basic_gates.rotation import Rx
+from qualtran.bloqs.qrom import QROM
+from qualtran.bloqs.rotations.phase_gradient import AddIntoPhaseGrad
+
+
+@attrs.frozen
+class StatePreparationViaRotations(GateWithRegisters):
```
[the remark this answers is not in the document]
StefanoPolla: We could also deal with this by adding to `RotationTree` a method `get_approximate_state` which combines `get_rom_values` with a new internal method `state_from_rom_values`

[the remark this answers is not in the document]
Vicara12: Yes, I also thought about moving the phase calculation (including offsets) to rotation tree. This way it would be quite easy to write a function that completely decomposes the state and then is able to return an approximation from the computed values as you mentioned

qualtran/bloqs/state_preparation/state_preparation_via_rotation.py
Vicara12: Done. This add some more code that handles some special cases when joining the qubit with the control register, but I think that's unavoidable.

Vicara12:
@tanujkhattar It seems that some tests are not passing from the latest commits in main? This is a bit weird, but some tests in `AddIntoPhaseGradient` fail due to an incorrect attribute name, which should be `x_bitsize` (fixed that in my last commit) and also the test `test_t_complexity_mcp` from `bloqs/multi_control_multi_target_pauli_test.py` fails. This should not be caused by my code (as I haven't touched anything remotely related to it, as far as I know).

qualtran/bloqs/rotations/phase_gradient.py
```diff
@@ -200,6 +202,24 @@ def _t_complexity_(self) -> 'TComplexity':
...
+        import quimb.tensor as qtn
+
+        N, M = 2**self.x_bitsize, 2**self.phase_bitsize
+        inds = (incoming['x'], incoming['phase_grad'], outgoing['x'], outgoing['phase_grad'])
+        unitary = np.zeros((N, M) * 2, dtype=np.complex128)
+        for a, b in itertools.product(range(N), range(M)):
+            unitary[a, b, a, int(math.fmod(a + b, M))] = 1
+
+        tn.add(qtn.Tensor(data=unitary, inds=inds, tags=[self.short_name(), tag]))
+
```
tanujkhattar, line 222: The `add_my_tensorts` would be a bit more complex now given the new parameters `sign` and `right_shift`. Since we already have a unitary implemented, we can derive tensor data via

```suggestion
    def add_my_tensors(
        self,
        tn: 'qtn.TensorNetwork',
        tag: Any,
        *,
        incoming: Dict[str, 'SoquetT'],
        outgoing: Dict[str, 'SoquetT'],
    ):
        from qualtran.cirq_interop._cirq_to_bloq import _add_my_tensors_from_gate

        _add_my_tensors_from_gate(
            self, self.signature, self.short_name(), tn, tag, incoming=incoming, outgoing=outgoing
        )
```

qualtran/bloqs/rotations/phase_gradient.py
```diff
@@ -12,6 +12,8 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import itertools
+import math
```
tanujkhattar, line 16: These imports won't be required anymore
```suggestion
```

qualtran/bloqs/state_preparation/state_preparation_via_rotation.py
```diff
@@ -0,0 +1,450 @@
...
+    def selection_registers(self) -> Tuple[Register, ...]:
+        return (
+            Register(
+                "target_state",
+                BoundedQUInt(
+                    bitsize=self.state_bitsize, iteration_length=len(self.state_coefficients)
+                ),
+            ),
+        )
+
```
tanujkhattar, line 153: This is moot now given we don't derive from `PrepareOracle`. 
```suggestion
```

PR #628, opened 2024-02-05
[data types] Simple data types

fdmalone:
Working on #446 and following on from adding the types in #624 I'll start incrementally replacing registers bitsizes with dtypes. Here I just take care of basic_gates, and and on_each

### DIFF
--- a/qualtran/bloqs/basic_gates/z_basis.py
+++ b/qualtran/bloqs/basic_gates/z_basis.py
@@ -76,9 +76,7 @@ class _ZVector(Bloq):
 
     @cached_property
     def signature(self) -> 'Signature':
-        return Signature(
-            [Register('q', bitsize=QBit(), side=Side.RIGHT if self.state else Side.LEFT)]
-        )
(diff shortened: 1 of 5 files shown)


qualtran/bloqs/basic_gates/z_basis.py
```diff
@@ -74,7 +76,9 @@ def __attrs_post_init__(self):
 
     @cached_property
     def signature(self) -> 'Signature':
-        return Signature([Register('q', bitsize=1, side=Side.RIGHT if self.state else Side.LEFT)])
+        return Signature(
+            [Register('q', bitsize=QBit(), side=Side.RIGHT if self.state else Side.LEFT)]
```
mpharrigan, line 80: this `bitsize=QBit()` uses the attrs shim to make everything work properly, yes? 

Do you want to take the opportunity to remove the `bitsize=` part of it. `bitsize=1` is much more informative than `1` but `QBit()` -- in my mind -- captures the full essence of the argument.

mpharrigan (approved):
lgtm, consider removing `bitsize=QBit()` in favor of `QBit()` which now conveys all the info

Issue #632, opened 2024-02-06
[data types] Better names for data types


fdmalone:
We should probably make the changes:

1. QUnsignedInt -> QUInt
2. QFixedPoint -> QFxP
3. BoundedQInt -> BoundedQUInt (we're never using the signed variation)

I think we should just add types as necessary rather than trying to be exhaustive (QUFxP for example). For 3 I will make the change in the selection register deprecation.

mpharrigan:
QFxP is pushing it lol

fdmalone:
Glass houses

tanujkhattar:
`QFxP` is motivated from similar terminology in https://github.com/francof2a/fxpmath, which we'll probably end up using extensively. See https://github.com/quantumlib/Qualtran/pull/635 for a first usage.

fdmalone:
I'm fine with either, I sort of prefer the capitalization for separate words, but if we eventually adopt their spec then it's most natural.

PR #634, opened 2024-02-06
[data types] Add data types for rotation and QFT bloqs

fdmalone:
* Sneak in: Shorten QFixedPoint -> QFxP, add QUFxP (#632)
* Updated registers for HammingWeightCompute.

### DIFF
--- a/qualtran/bloqs/arithmetic/hamming_weight.py
+++ b/qualtran/bloqs/arithmetic/hamming_weight.py
@@ -20,7 +20,7 @@ import cirq
 from attrs import frozen
 from numpy.typing import NDArray
 
-from qualtran import Bloq, GateWithRegisters, Register, Side, Signature
+from qualtran import Bloq, GateWithRegisters, QBit, QUInt, Register, Side, Signature
 from qualtran.bloqs.and_bloq import And
 from qualtran.bloqs.util_bloqs import ArbitraryClifford
 from qualtran.cirq_interop.t_complexity_protocol import TComplexity
@@ -58,11 +58,13 @@ class HammingWeightCompute(GateWithRegisters):
     @cached_property
     def signature(self):
         side = Side.LEFT if self.uncompute else Side.RIGHT
+        jnk_size = self.bitsize - self.bitsize.bit_count()
+        out_size = self.bitsize.bit_length()
         return Signature(
             [
-                Register('x', self.bitsize),
-                Register('junk', self.bitsize - self.bitsize.bit_count(), side=side),
-                Register('out', self.bitsize.bit_length(), side=side),
+                Register('x', QUInt(self.bitsize)),
+                Register('junk', QBit() if jnk_size == 1 else QUInt(jnk_size), side=side),
+                Register('out', QBit() if out_size == 1 else QUInt(out_size), side=side),
             ]
         )
 
--- a/qualtran/bloqs/arithmetic/hamming_weight.py
+++ b/qualtran/bloqs/arithmetic/hamming_weight.py
@@ -20,7 +20,7 @@ import cirq
 from attrs import frozen
 from numpy.typing import NDArray
 
-from qualtran import Bloq, GateWithRegisters, QBit, QUInt, Register, Side, Signature
+from qualtran import Bloq, GateWithRegisters, QAny, QUInt, Register, Side, Signature
 from qualtran.bloqs.and_bloq import And
 from qualtran.bloqs.util_bloqs import ArbitraryClifford
 from qualtran.cirq_interop.t_complexity_protocol import TComplexity
@@ -63,8 +63,8 @@ class HammingWeightCompute(GateWithRegisters):
         return Signature(
             [
                 Register('x', QUInt(self.bitsize)),
-                Register('junk', QBit() if jnk_size == 1 else QUInt(jnk_size), side=side),
-                Register('out', QBit() if out_size == 1 else QUInt(out_size), side=side),
+                Register('junk', QAny(jnk_size), side=side),
+                Register('out', QUInt(out_size), side=side),
             ]
         )
 
--- a/qualtran/bloqs/qft/qft_phase_gradient.py
+++ b/qualtran/bloqs/qft/qft_phase_gradient.py
@@ -54,8 +54,6 @@ class QFTPhaseGradient(GateWithRegisters):
 
     @cached_property
     def signature(self) -> 'Signature':
-        if self.bitsize == 1:
-            return Signature([Register("q", QBit()), Register("phase_grad", QBit())])
         return Signature(
             [Register("q", QUInt(self.bitsize)), Register("phase_grad", QUInt(self.bitsize))]
         )
--- a/qualtran/bloqs/qft/qft_phase_gradient.py
+++ b/qualtran/bloqs/qft/qft_phase_gradient.py
@@ -17,7 +17,7 @@ import attrs
 import cirq
 from numpy.typing import NDArray
 
-from qualtran import GateWithRegisters, Signature
+from qualtran import GateWithRegisters, QBit, QUInt, Register, Signature
 from qualtran.bloqs.arithmetic.multiplication import PlusEqualProduct
 
 
@@ -54,7 +54,11 @@ class QFTPhaseGradient(GateWithRegisters):
 
     @cached_property
     def signature(self) -> 'Signature':
-        return Signature.build(q=self.bitsize, phase_grad=self.bitsize)
+        if self.bitsize == 1:
+            return Signature([Register("q", QBit()), Register("phase_grad", QBit())])
+        return Signature(
+            [Register("q", QUInt(self.bitsize)), Register("phase_grad", QUInt(self.bitsize))]
+        )
 
     def decompose_from_registers(
         self, *, context: cirq.DecompositionContext, **quregs: NDArray[cirq.Qid]
--- a/qualtran/bloqs/qft/qft_phase_gradient.py
+++ b/qualtran/bloqs/qft/qft_phase_gradient.py
@@ -17,7 +17,7 @@ import attrs
 import cirq
 from numpy.typing import NDArray
 
-from qualtran import GateWithRegisters, QBit, QUInt, Register, Signature
+from qualtran import GateWithRegisters, QFxp, QUInt, Signature
 from qualtran.bloqs.arithmetic.multiplication import PlusEqualProduct
 
 
@@ -54,8 +54,8 @@ class QFTPhaseGradient(GateWithRegisters):
 
     @cached_property
     def signature(self) -> 'Signature':
-        return Signature(
-            [Register("q", QUInt(self.bitsize)), Register("phase_grad", QUInt(self.bitsize))]
+        return Signature.build_from_dtypes(
+            q=QUInt(self.bitsize), phase_grad=QFxp(self.bitsize, self.bitsize)
         )
 
     def decompose_from_registers(
--- a/qualtran/bloqs/qft/qft_text_book.py
+++ b/qualtran/bloqs/qft/qft_text_book.py
@@ -17,7 +17,7 @@ import attrs
 import cirq
 from numpy.typing import NDArray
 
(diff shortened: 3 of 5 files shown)


fdmalone:
@tanujkhattar I wasn't sure about some of the data types for these registers. In particular I have a feeling the register i put as fixed point should actually not be (just a QUInt), but it was a little confusing with the gamma parameter (I guess there is a FP register holding this value omitted from the signature?)

qualtran/bloqs/arithmetic/hamming_weight.py
```diff
@@ -58,11 +58,13 @@ class HammingWeightCompute(GateWithRegisters):
...
         side = Side.LEFT if self.uncompute else Side.RIGHT
+        jnk_size = self.bitsize - self.bitsize.bit_count()
+        out_size = self.bitsize.bit_length()
         return Signature(
             [
-                Register('x', self.bitsize),
-                Register('junk', self.bitsize - self.bitsize.bit_count(), side=side),
-                Register('out', self.bitsize.bit_length(), side=side),
+                Register('x', QUInt(self.bitsize)),
+                Register('junk', QBit() if jnk_size == 1 else QUInt(jnk_size), side=side),
```
tanujkhattar, line 66: Is `QUint(1)` discouraged? This looks very verbose if we have to do this for every bloq that creates a `QUInt`

qualtran/bloqs/qft/qft_phase_gradient.py
```diff
@@ -54,7 +54,11 @@ class QFTPhaseGradient(GateWithRegisters):
 
     @cached_property
     def signature(self) -> 'Signature':
-        return Signature.build(q=self.bitsize, phase_grad=self.bitsize)
+        if self.bitsize == 1:
+            return Signature([Register("q", QBit()), Register("phase_grad", QBit())])
+        return Signature(
+            [Register("q", QUInt(self.bitsize)), Register("phase_grad", QUInt(self.bitsize))]
+        )
```
tanujkhattar, line 61: Ditto: Do we need the `QUInt` vs `QBit` special casing?

qualtran/bloqs/rotations/phase_gradient.py
```diff
@@ -222,7 +227,12 @@ def __attrs_post_init__(self):
 
     @cached_property
     def signature(self):
-        return Signature.build(x=self.inp_bitsize, phase_grad=self.phase_bitsize)
+        return Signature(
+            [
+                Register("x", QUInt(self.inp_bitsize)),
+                Register("phase_grad", QUFxP(0, self.phase_bitsize)),
+            ]
+        )
```
tanujkhattar, line 235: I'm confused how / why the tests pass. This definitely looks like a bug? 

`AddScaledValIntoPhaseReg` exepcts a `phase_grad` register of type `QUFxp` with `phase_bitsize` fractional bitsize and `0` word bitsize. 

This bloq is called from ` HammingWeightPhasingViaPhaseGradient.build_decompose_bloq` which expects `phase_grad` as a `QUInt(self.b_grad)` and does no explicit cast before calling 

```
        out, phase_grad = bb.add(
            AddScaledValIntoPhaseReg(
                self.bitsize.bit_length(), self.b_grad, self.exponent / 2, self.b_phase
            ),
            x=out,
            phase_grad=phase_grad,
        )
```

Do you know what's going on here?

tanujkhattar (changes requested):
A few nits and a question regarding a potential bug

qualtran/_infra/data_types.py
```diff
@@ -184,7 +184,7 @@ def num_qubits(self):
 
 
 @attrs.frozen
-class QFixedPoint(QDType):
+class QFxP(QDType):
```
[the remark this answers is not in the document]
fdmalone: Sure, now means less pain later, but it's only a small amount of pain. Alternatively rather than a string we could just have sign, n_word and n_frac as attributes and could return a Fxp string for construction?

qualtran/_infra/data_types.py
```diff
@@ -209,3 +209,31 @@ def __attrs_post_init__(self):
...
+
+    int_bitsize: Union[int, sympy.Expr]
+    frac_bitsize: Union[int, sympy.Expr]
+
+    @property
+    def num_qubits(self):
+        if isinstance(self.int_bitsize, sympy.Expr):
+            return self.int_bitsize + self.frac_bitsize
+        else:
+            return self.int_bitsize + self.frac_bitsize
```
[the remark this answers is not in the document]
fdmalone: No!

qualtran/bloqs/arithmetic/hamming_weight.py
fdmalone: Yeah this is sort of a decision that needs to be made (see #641). QUInt(1) and QFxp(0, ...) etc can probably be size 1. I don't think QInt(1) makes sense. #641 could move a some of the boilerplate to the constructors.

qualtran/_infra/data_types.py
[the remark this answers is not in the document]
tanujkhattar: I'd rather have a builder method using `n_word` and `n_frac` and keep the string as an attribute, especially if we plan to combine `QFxp` and `QUFxp`

qualtran/bloqs/rotations/phase_gradient.py
fdmalone: Well they pass because there is no type checking on the registers currently and they both have the same bitsize (num_qubits). What should the correct type be?

For migration I'm mainly focussed on removing integer bitsizes for building registers. Once that's complete we'll start enforcing types (and not just bitsizes) are consistent % generic bloqs with QAny registers.

qualtran/_infra/data_types.py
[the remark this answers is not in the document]
mpharrigan: it's really easy to go from the actual attributes to a string and not really easy to go the other way...

qualtran/bloqs/arithmetic/hamming_weight.py
mpharrigan: In this particular instance, I wonder whether QAny would make more sense for a junk register, since we make no claims about what it actually is. 

For places in the library that blindly use a bitsize argument (like Add) I would prefer if we mandated that the registers are larger than 1 bit anyways; since e.g. adding a 1 bit value is a very different operation than adding n>1 bit values from the perspective of compiling and costing and algorithm for our desired architecture.

Long story short: it's a code smell if your datatypes change depending on logic like this and you should think about what your bloq is first

qualtran/bloqs/qft/qft_phase_gradient.py
mpharrigan: first, it would be helpful to have a `Registers:` section in the docstring so I know what the "q" register is supposed to be. 

Does it make sense to do a 1-bit QFT?

qualtran/_infra/data_types.py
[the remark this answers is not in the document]
fdmalone: yeah I think 

``` python
qfxp = QFxp(signed=True, n_word=8, n_frac=7)
fxp = Fxp(float_val, dtype=qfxp.fxp_string())
```

is easier / clearer to type than

``` python
qfxp = QFxp(dtype="fxp-s8/7")
```

qualtran/bloqs/arithmetic/hamming_weight.py
tanujkhattar: > For places in the library that blindly use a bitsize argument (like Add) I would prefer if we mandated that the registers are larger than 1 bit anyways; since e.g. adding a 1 bit value is a very different operation than adding n>1 bit values from the perspective of compiling and costing and algorithm for our desired architecture. 

I think this would result in a lot of if/else blocks and code that's not generic. I agree that junk in this case should probably just be `QAny` but I don't see why it's bad to have a 1 bit integer or a 1 bit fxp number. A 1 bit fxp unsigned integer can mean a different thing compared to a 1 bit fxp number with only 1 fractional bit; and that information can be useful in the algorithm. Even if not, I think there's value in implementing classes that work for arbitrary `bitsize`.

qualtran/bloqs/qft/qft_phase_gradient.py
tanujkhattar: > Does it make sense to do a 1-bit QFT?

Yes, it's a hadamard gate. And when writing generic code that works for arbitrary `bitsize`, it's useful to not have special casing checks at the call site to delegate to a hadamard or a QFT separately; when the decomposition can take care of that. Costings would be correct in both the cases.

qualtran/_infra/data_types.py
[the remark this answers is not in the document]
tanujkhattar: Okay, storing and using explicit attributes SGTM. 

Just FYI - note that `n_word` for `Fxp` means total bitsize and in our previous convention, it used to mean bitsize for the integer part. We should pick a convention, document it and keep it consistent.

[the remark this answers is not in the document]
fdmalone: right, I had to look at their docs before writing it. Currently we have int_bitsize and frac_bitsize and the signed bit is hiding there depending on if it's a QUFxp or a QFxp, but I think the n_word is more natural in some way. Given all this discussion I will pull out the Fxp renaming to a separate PR.

mpharrigan:
@tanujkhattar do you think the blockers are resolved once @fdmalone resolves the merge conflicts?

qualtran/bloqs/qft/qft_phase_gradient.py
```diff
@@ -54,7 +54,9 @@ class QFTPhaseGradient(GateWithRegisters):
 
     @cached_property
     def signature(self) -> 'Signature':
-        return Signature.build(q=self.bitsize, phase_grad=self.bitsize)
+        return Signature(
+            [Register("q", QUInt(self.bitsize)), Register("phase_grad", QUInt(self.bitsize))]
```
tanujkhattar, line 58: As discussed offline,phase gradient registers should be unsigned fixed point numbers with 0 integer bits.

fdmalone: I'm tidying this up now, q and x are still unsigned ints right?

qualtran/bloqs/qft/qft_text_book.py
```diff
@@ -40,7 +40,7 @@ class QFTTextBook(GateWithRegisters):
 
     @cached_property
     def signature(self) -> 'Signature':
-        return Signature.build(q=self.bitsize)
+        return Signature([Register("q", QBit() if self.bitsize == 1 else QUInt(self.bitsize))])
```
tanujkhattar, line 43: I think we should not special case on self.bitsize == 1; do we have a consensus yet?

qualtran/bloqs/rotations/phase_gradient.py
```diff
@@ -222,7 +227,12 @@ def __attrs_post_init__(self):
 
     @cached_property
     def signature(self):
-        return Signature.build(x=self.inp_bitsize, phase_grad=self.phase_bitsize)
+        return Signature(
+            [
+                Register("x", QUInt(self.inp_bitsize)),
+                Register("phase_grad", QFxp(self.phase_bitsize, self.phase_bitsize, signed=True)),
```
tanujkhattar, line 233: This should be unsigned I think

qualtran/bloqs/qft/qft_text_book.py
fdmalone: sorry I prematurely pushed, I'm replacing with build_from_dtypes which will handle this.

fdmalone: Or rather we aren't special casing for QUInt / QInt (1)

qualtran/bloqs/qft/qft_phase_gradient.py
tanujkhattar: I think keeping `q` and `x` as unsigned ints is probably fine for now. We can revisit later if needed.

tanujkhattar (approved):

PR #635, opened 2024-02-06
Error propagation analysis due to fixed point arithmetic for multiplication and squaring

tanujkhattar:
@ncrubin This is a first attempt to reproduce the error analysis for fixed point arithmetic for multiplication and squaring a real number from https://arxiv.org/pdf/2007.07391.pdf

Ideally, the bloqs corresponding to these primitives would expect registers of `QFxp` data type and running classical simulation on the corresponding composite bloq should reproduce this analysis. But the quantum data types are still being implemented so this notebook serves as a reference and a good starting point to reproduce the analysis from the paper. 

@fdmalone  @mpharrigan  - This would add an additional dependency to fxpmath, which is MIT licensed.


dev_tools/requirements/envs/dev.env.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.10
+# This file is autogenerated by pip-compile with Python 3.11
```
mpharrigan, line 2: please use the docker image to compile the requirements

qualtran/bloqs/arithmetic/Error propagation due to fixed point arithmetic.ipynb
```diff
@@ -0,0 +1,203 @@
...
+   "id": "92cdfbaa-1e5f-4aac-9fdd-4d4015fd384a",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from fxpmath import Fxp\n",
+    "import numpy as np\n",
+    "\n",
+    "# Initial setup - Helpers\n",
+    "def fxp(x: float, d: int) -> Fxp:\n",
+    "    dtype = f'fxp-u{d}/{d}'\n",
```
mpharrigan, line 27: docstring is necessary here, esp if one is not familiar with fxpmath

qualtran/bloqs/arithmetic/Error propagation due to fixed point arithmetic.ipynb
```diff
@@ -0,0 +1,203 @@
...
+    "    try:\n",
+    "        for _ in range(100):\n",
+    "            a, b = rng.random(2)\n",
+    "            res = mul_via_repeated_add(a, b, d)\n",
+    "            assert_allclose(res, a * b, eps)\n",
+    "        print(f'Success! {d=}, {eps=}')\n",
+    "    except:\n",
+    "        print(f'Failed! {d=}, {eps=}')\n",
+    "\n",
+    "for eps in [1e-3, 1e-4, 1e-5, 1e-6, 1e-7]:\n",
```
mpharrigan, line 94: these all succeed

qualtran/bloqs/arithmetic/Error propagation due to fixed point arithmetic.ipynb
```diff
@@ -0,0 +1,203 @@
...
+    "\n",
+    "def test_multiplication_for_eps(eps: float, d: int):\n",
+    "    rng = np.random.default_rng(int(eps * 1e9))\n",
+    "    try:\n",
+    "        for _ in range(100):\n",
+    "            a, b = rng.random(2)\n",
+    "            res = mul_via_repeated_add(a, b, d)\n",
+    "            assert_allclose(res, a * b, eps)\n",
+    "        print(f'Success! {d=}, {eps=}')\n",
+    "    except:\n",
```
mpharrigan, line 91: bare except. Is this the best pattern here? asserting and catching?

qualtran/bloqs/arithmetic/Error propagation due to fixed point arithmetic.ipynb
```diff
@@ -0,0 +1,203 @@
...
+      "Success! d=21, eps=1e-05\n",
+      "Success! d=24, eps=1e-06\n",
+      "Success! d=28, eps=1e-07\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Multiplying two real numbers\n",
+    "def get_bitsize_for_fxp_mul(eps: float):\n",
+    "    return int(np.ceil(1 + np.log2(1/eps)) + np.log2(1 + np.log2(1/eps)))\n",
```
mpharrigan, line 72: this is from the paper?

mpharrigan (commented):
this is just numerically testing the error formulas from the appendices, yes? If that is the full scope, I'd suggest changing the title of the notebook.

The code in the notebook needs a lot of documentation

mpharrigan:
Taking these implementations and using them for the `on_classical_vals` overrides would be very cool 👍

qualtran/bloqs/arithmetic/Error propagation due to fixed point arithmetic.ipynb
tanujkhattar: Yes, I'll add a comment to reference equation.

qualtran/bloqs/arithmetic/Error propagation due to fixed point arithmetic.ipynb
tanujkhattar: Yeah, they are all expected to succeed. They succeed for `d-1` as well, which is interesting. Similarly for the next section the tests actually fail for the `d` value proposed in the paper and need a `+1` for small bitsizes. 

The goal here is to provide a framework to replicate the analysis in the paper.

tanujkhattar:
> this is just numerically testing the error formulas from the appendices, yes? 

Yup, but the eventual goal of this notebook is to demonstrate this analysis by instantiating appropriate bloqs. This is a first iteration that gets the analysis in place and follow up PRs would add it to appropriate bloqs and update the notebook with bloq instantiation and classical simulation 

> If that is the full scope, I'd suggest changing the title of the notebook.
Do you have suggestions? Maybe "Numerical analysis of error propagation due to fixed point arithmetic" ?

qualtran/bloqs/arithmetic/Error propagation due to fixed point arithmetic.ipynb
tanujkhattar: changed to catch only `AssertionError` instead of bare except

qualtran/bloqs/arithmetic/Error propagation due to fixed point arithmetic.ipynb
tanujkhattar: added docstrings

dev_tools/requirements/envs/dev.env.txt
tanujkhattar: done

fdmalone:
This is a bit off topic but do we have a plan in place for updating classical simulation to allow for non uint64 types?

More on topic another useful primitive to add would be [scaling an float by an integer ](https://github.com/quantumlib/Qualtran/blob/main/qualtran/bloqs/arithmetic/multiplication.py#L242) (should be renamed) which has some interesting shifting / precision properties I hadn't really thought about before.

qualtran/bloqs/arithmetic/Numerical analysis of error propagation due to fixed point arithmetic.ipynb
```diff
@@ -0,0 +1,189 @@
+{
```
fdmalone, line 1: Are we ok with spaces in filenames?

tanujkhattar: Renamed to remove spaces and match pattern used by other notebooks

tanujkhattar:
> More on topic another useful primitive to add to your notebook would be [scaling a real by an integer ](https://github.com/quantumlib/Qualtran/blob/main/qualtran/bloqs/arithmetic/multiplication.py#L242)(should be renamed) which has some interesting shifting / precision properties I hadn't really thought about before

@fdmalone I've added a section to the notebook that analyzes the scaling a real by an integer primitive following Appendix D4 of the paper. 

I think this is good to go now and we can revisit once we have updated classical simulation code to support `Fxp` along with data types.

fdmalone (approved):
LGTM

Issue #630, opened 2024-02-05
[GQSP] Precision issue in computing phase angles


anurudhp:
Algorithm 1 of https://arxiv.org/abs/2308.01501: the QSP angles are calculated as $\theta_d = \arctan(|b_d / a_d|)$, $\phi_d = \arg(a_d / b_d)$.

1. Computing $\phi$ is numerically unstable when $a_d/b_d$ is zero, as small errors result in a non-zero angle.
2. Another issue is the edge case of $a_d = 0$ or $b_d = 0$, for which I could not figure out what the correct behaviour should be.

https://github.com/quantumlib/Qualtran/blob/642dec21cb6844bbe2f5f33e4312a4951f73e8c9/qualtran/bloqs/generalized_qsp.py#L190-L242

cc. @tanujkhattar

tanujkhattar:
We can modify the current implementation of `phi[d] = np.angle(a) - np.angle(b)` to

```python
phi[d] = 0 if np.isclose(np.abs(b), 0) else np.angle(a * np.conj(b) / np.abs(b))
```

and the behavior should be well defined for the various cases you mentioned above? 

Although I'm not sure if this necessarily the problem following discussion on https://github.com/quantumlib/Qualtran/issues/631

anurudhp:
fixed by #636

Issue #631, opened 2024-02-05
[GQSP] Computing complementary polynomial by root-finding is very inaccurate


anurudhp:
Theorem 4 of https://arxiv.org/abs/2308.01501: Computes roots of an auxiliary polynomial R (from P) which are then re-interpolated to construct Q. 

https://github.com/quantumlib/Qualtran/blob/642dec21cb6844bbe2f5f33e4312a4951f73e8c9/qualtran/bloqs/generalized_qsp.py#L82-L187

Interestingly, the computed Q satisfies the constraint $|P(z)|^2 + |Q(z)|^2 = 1$ for polynomials upto degree 100. But the QSP sequence is only accurate till degree 4 (related to #630).

tanujkhattar:
Can you add a small example of a `P` for which you compute the corresponding valid `Q` but the QSP sequence turns out to be incorrect due to precision issue in computing phase angles?

anurudhp:
```py
P = [0.07691151711056517, 0.02506952291383959, 0.02858956904068647, 0.08523950878413064, 0.09750064936065875, 0.08848532934911055, 0.035950784393690226, 0.05988589458757472, 0.03547956116572998, 0.03401902153706458]
Q = [-0.002674721864062315, -0.00375596522126604, -0.006826640697387547, -0.008793329571296443, -0.01680094202314425, -0.024007293148226038, -0.027211524947234132, -0.029928727434094398, -0.03458906260723875, 0.9782155640881396]
```

code to reproduce test:
```py
def test_custom():
    random_state = np.random.RandomState(100)

    U = RandomGate.create(2, random_state=random_state)
    P = random_qsp_polynomial(degree=10, random_state=random_state, only_real_coeffs=True)
    Q = GeneralizedQSP(U, P).Q
    check_polynomial_pair_on_random_points_on_unit_circle(
        P, Q, random_state=random_state, n_points=10**5
    ) # works

    verify_generalized_qsp(U, P) # fails, error is ~ 0.00166
```

tanujkhattar:
For the example you posted, the problem doesn't seem to be related to precision issues in computing phase angles as highlighted in https://github.com/quantumlib/Qualtran/issues/630

For example, when computing the phase angles for the above example, the values of a & b are as follows:

```
d=4, phi[d]= 0.0 a= (0.1+0j) b= (0.8996+0j) a/b= (0.1112+0j)
[...]

PR #636, opened 2024-02-06
[GQSP] fix bug in phase angle calculation

anurudhp:
Fixes issue in precision for real polynomials. The thetas seem robust to error, but not phis, as adding a small imaginary part can result in non zero angles. More discussion in #630.

### DIFF
--- a/qualtran/bloqs/generalized_qsp.py
+++ b/qualtran/bloqs/generalized_qsp.py
@@ -222,6 +222,9 @@ def qsp_phase_factors(
     phi = np.zeros(n)
     lambd = 0
 
+    def safe_angle(x):
+        return 0 if np.isclose(x, 0) else np.angle(x)
+
     for d in reversed(range(n)):
         assert S.shape == (2, d + 1)
 
@@ -237,12 +240,18 @@ def qsp_phase_factors(
             #      even though we want it to be 0.
             #      There should be a more numerically stable way to compute the relevant phis
             #      so that the final QSP sequence is valid.
-            phi[d] = 0 if np.isclose(np.abs(b), 0) else np.angle(a) - np.angle(b)
+            phi[d] = 0 if np.isclose(np.abs(b), 0) else safe_angle(a) - safe_angle(b)
 
         if d == 0:
-            lambd = np.angle(b)
+            lambd = safe_angle(b)
         else:
             S = SU2RotationGate(theta[d], phi[d], 0).rotation_matrix @ S
+
+            # check if the gate was actually correct
+            aa, bb = S[0][0], S[1][d]
+            np.testing.assert_almost_equal(aa, 0, decimal=6)
+            np.testing.assert_almost_equal(bb, 0, decimal=6)
+
             S = np.array([S[0][1 : d + 1], S[1][0:d]])
 
     return theta, phi, lambd
--- a/qualtran/bloqs/generalized_qsp.py
+++ b/qualtran/bloqs/generalized_qsp.py
@@ -224,7 +224,7 @@ def qsp_phase_factors(
         a, b = S[:, d]
         theta[d] = np.arctan2(np.abs(b), np.abs(a))
         # \phi_d = arg(a / b)
-        phi[d] = 0 if np.isclose(np.abs(b), 0) else safe_angle(a) - safe_angle(b)
+        phi[d] = 0 if np.isclose(np.abs(b), 0) else safe_angle(a * np.conj(b))
 
         if d == 0:
             lambd = safe_angle(b)
--- a/qualtran/bloqs/generalized_qsp.py
+++ b/qualtran/bloqs/generalized_qsp.py
@@ -282,7 +282,7 @@ class GeneralizedQSP(GateWithRegisters):
 
     @cached_property
     def _lambda(self) -> float:
-        return 0  # self._qsp_phases[2]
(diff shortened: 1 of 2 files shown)


qualtran/bloqs/generalized_qsp.py
```diff
@@ -220,23 +222,36 @@ def qsp_phase_factors(
...
         if d == 0:
-            lambd = np.angle(b)
+            lambd = safe_angle(b)
         else:
             S = SU2RotationGate(theta[d], phi[d], 0).rotation_matrix @ S
+
+            # check if the gate was actually correct
+            aa, bb = S[0][0], S[1][d]
+            np.testing.assert_almost_equal(aa, 0, decimal=6)
+            np.testing.assert_almost_equal(bb, 0, decimal=6)
```
anurudhp, line 253: @tanujkhattar I narrowed it down to this. These values must be 0, but are not iff the case fails (according to the unittests). I'll now investigate further why this is not working.

qualtran/bloqs/generalized_qsp.py
```diff
@@ -220,23 +215,21 @@ def qsp_phase_factors(
...
         a, b = S[:, d]
         theta[d] = np.arctan2(np.abs(b), np.abs(a))
         # \phi_d = arg(a / b)
-        # TODO this is numerically unstable for values of a and b around 0,
-        #      np.angle(0) = 0, but for near-zero values it produces the "correct angle",
-        #      even though we want it to be 0.
-        #      There should be a more numerically stable way to compute the relevant phis
-        #      so that the final QSP sequence is valid.
-        phi[d] = np.angle(a) - np.angle(b)
+        phi[d] = 0 if np.isclose(np.abs(b), 0) else safe_angle(a) - safe_angle(b)
```
tanujkhattar, line 227: `np.angle(x)` always returns values in `(-pi, pi]` but this would not be true for `angle(a) - angle(b)`. I guess this is not an issue for us because we use the angles for single qubit rotations and they take care of 2 * pi difference by default. 

Do you know if subtracting the angles is strictly better than directly using `np.angle(a * np.conj(b) / np.abs(b))` ?

anurudhp:
this PR fixes #630

anurudhp:
@tanujkhattar this is now ready to review. The algorithm works as expected after the fix, for fairly large degrees.

qualtran/bloqs/generalized_qsp_test.py
```diff
@@ -144,24 +143,35 @@ def verify_generalized_qsp(U: GateWithRegisters, P: Sequence[complex]):
...
-    for _ in range(20):
+    for _ in range(10):
         U = RandomGate.create(bitsize, random_state=random_state)
         P = random_qsp_polynomial(degree, random_state=random_state, only_real_coeffs=True)
         verify_generalized_qsp(U, P)
 
 
+@pytest.mark.parametrize("bitsize", [1, 2, 3])
+@pytest.mark.parametrize("degree", [2, 3, 4, 5, 50, 100, 120])
+def test_generalized_qsp_on_random_unitaries(bitsize: int, degree: int):
```
tanujkhattar, line 166: This test is for general complex unitaries whereas the one above is for real unitaries only, correct?
```suggestion
def test_generalized_qsp_on_random_complex_unitaries(bitsize: int, degree: int):
```

tanujkhattar (approved):
This is great, and exactly the kind of outcome I was hoping for! 

I believe this also fixes https://github.com/quantumlib/Qualtran/issues/631 ?

qualtran/bloqs/generalized_qsp.py
anurudhp: Not sure if it is strictly better, I just wanted to avoid multiplication/division to reduce precision errors. But I think `np.angle(a * np.conj(b))` should also work equally well (can drop the `abs` as it does not affect angle).

qualtran/bloqs/generalized_qsp_test.py
anurudhp: No no, both use complex unitaries. The first test is for P with only real coeffs, and second is complex coeffs.

tanujkhattar: Yeah sorry, I meant real vs complex polynomials. Maybe change the name of the test appropriately.

anurudhp:
Yep, also fixes #631. I will later open a separate issue for scaling this to larger polynomials (e.g. degrees of the order 1e5 - 1e7)

qualtran/bloqs/generalized_qsp.py
tanujkhattar: I'd vote to change it to more closely match the algorithm description in the paper and keep all the phi's in `(-pi, pi]`

qualtran/bloqs/generalized_qsp_test.py
anurudhp: updated

qualtran/bloqs/generalized_qsp.py
anurudhp: updated

anurudhp: actually even `angle(a / b)` works (on tests), I can update it to that as well if it's more preferred.

tanujkhattar: I think this is fine

Issue #637, opened 2024-02-06
[data types] Bloqs which work for multiple types


fdmalone:
As part of #446 we will need to update the signatures of several arithmetic bloqs to allow for some generic behavour.

One way to do this is

``` python
class Add(bloq):
    dtype: QDType

   @dtype.validator(self):
        assert isinstance(dtype, (QInt, QUInt, ...))
```

This will happen after the migration work.

mpharrigan:
I think I asked during the design review whether we'd need abstract qdtypes. Specifically: a named-but-not-instantiatable type to replace `(QInt, QUint, ...)` in the isinstance check (and appropriate for the attrs type annotation)

fdmalone:
right, that is the plan once we sort of land on what bloqs work on which types.

PR #638, opened 2024-02-06
[data types] Addition bloqs

fdmalone:
Depends on #634


fdmalone:
There's a common pattern emerging where my criteria that only the QBit type would have bitsize = 1 is causing additional boilerplate. Should we remove this restriction?

fdmalone:
no longer depends on #634 and I simplified the bitsize argument to allow for QUInt(1) (although we should probably update the addition tests to forbid single bit adders).  Will follow up on #637.

mpharrigan (approved):

PR #639, opened 2024-02-06
[data types] Migrate comparison gates

fdmalone:
Depends on #638.

@NoureldinYosri I wasn't sure what the data types of the registers should be for the comparators (or maybe they work for all integer types, ones / twos complement, ...). For the moment the types don't enforce anything but would be good to track in #637


NoureldinYosri:
@fdmalone unsigned is correct here. I think most if not all comparator decompositions will fail when at least one of its arguments is signed

fdmalone:
no longer depends on #638

tanujkhattar (approved):

PR #641, opened 2024-02-06
[data types] Add signature factory method build_from_dtypes

fdmalone:
The ci is failing in #639 due to the comparator in the [unit test having zero bitsize](https://github.com/quantumlib/Qualtran/actions/runs/7805172959/job/21288750016#step:6:786). I'm not sure what our desired behavior is for registers of size zero, but the test previously passed  because signature.build would silently drop registers of size zero. I thought adding a similar method for dtypes would be helpful in this regard, which is implemented here, but it raises the question if we really want zero bitsize registers or should this just throw an error when constructing the type. I also thought it could be nice to handle the boilerplate of converting a QInt(1) to a QBit() but currently this will raise an ValueError during construction of QInt(1), so:

1. Should QDType(0) raise an error? I can update the test to use a lower epsilon so the bitsize is non-zero for the alias sampling comparator. I did this already but the notebook needs to be updated. 
2. Should QInt(1) be allowed and move the conversion to signature.build_from_dtypes. This seems fine, but I would need to add some attrs magic to Register too probably.

### DIFF
--- a/qualtran/_infra/registers.py
+++ b/qualtran/_infra/registers.py
@@ -198,7 +198,7 @@ class Signature:
 
     @classmethod
     def build(cls, **registers: int) -> 'Signature':
-        """Construct a Signature comprised of simple thru registers.
+        """Construct a Signature comprised of simple thru registers given the register bitsizes.
 
         Args:
             registers: keyword arguments mapping register name to bitsize. All registers
@@ -208,7 +208,7 @@ class Signature:
 
     @classmethod
     def build_from_dtypes(cls, **registers: QDType) -> 'Signature':
-        """Construct a Signature comprised of simple thru registers.
+        """Construct a Signature comprised of simple thru registers given the register dtypes.
 
         Args:
             registers: keyword arguments mapping register name to QDType. All registers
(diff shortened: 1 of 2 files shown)


fdmalone:
WDYT @mpharrigan @tanujkhattar

qualtran/_infra/registers.py
```diff
@@ -206,6 +206,24 @@ def build(cls, **registers: int) -> 'Signature':
         """
         return cls(Register(name=k, bitsize=v) for k, v in registers.items() if v)
 
+    @classmethod
+    def build_from_dtypes(cls, **registers: QDType) -> 'Signature':
+        """Construct a Signature comprised of simple thru registers.
```
mpharrigan, line 211: distinguish from `build` in the first line. by editing this or the line for build or both

mpharrigan (commented):
I think it's weird to rely on falsey values for registers to drop said registers but I know @tanujkhattar found it helpful in the past, so no objections in principle.

fdmalone:
Currently the library allows for unigned single bit registers which is probably fine. We follow the existing behavior of Signature.build silently stripping zero bitsizes-sized registers (dtypes).

qualtran/_infra/registers.py
fdmalone: Done.

mpharrigan (approved):

PR #642, opened 2024-02-06
[data types] Migrate Multiplication bloqs

fdmalone:
Uses the new QFxp registers where appropriate. This leads to one bloq requiring and updated interface which propagated downstream to a trotter bloq. Essentially scaling a real by an int should allow for a different bitsize output register (we're shifting a lot of the bits to the left, so need more integer bits). How this is handled isn't specified in the paper so it would be good to revisit once we hook up classical simulation.

I realized the input bitsizes were flipped there so I fixed that bug too.


fdmalone:
I adopted the suggestion from #446 to allow for single bit twos complement QInt, which otherwise was causing the tests to fail. Maybe that's not a very good justification for doing it now, but..

@tanujkhattar PTAL, particularly at the PlusEqualProduct types which possibly should be unsigned?

mpharrigan (commented):
looks fine to me, @tanujkhattar can you take a look

qualtran/bloqs/arithmetic/multiplication.py
```diff
@@ -253,6 +260,7 @@ class ScaleIntByReal(Bloq):
     Args:
         r_bitsize: Number of bits used to represent the real number.
         i_bitsize: Number of bits used to represent the integer.
+        out_bitsizes: The output bitsize for the scaled fixed point real number.
```
tanujkhattar, line 263: In the linked reference the output size is assumed to be same as the `r_bitsize`, which is chosen based on the tolerable error `eps`.  

I guess this change in the signature is required to support the usage of this bloq for inverse sqrt where we assume the output bitsize can be specified as an independent parameter?

qualtran/bloqs/chemistry/trotter/inverse_sqrt_test.py
```diff
@@ -38,7 +38,7 @@ def test_poly_eval_inverse_sqrt(bloq_autotester):
 def test_newton_raphson_inverse_sqrt_bloq_counts():
     bloq = NewtonRaphsonApproxInverseSquareRoot(7, 8, 12)
     _, counts = bloq.call_graph()
-    assert counts[TGate()] == 1632
+    assert counts[TGate()] == 1424
```
tanujkhattar, line 41: Why did this number change? Because of the bug where input bitsizes were flipped? What's the expected gate count?

qualtran/bloqs/arithmetic/multiplication.py
```diff
@@ -41,7 +41,9 @@ def short_name(self) -> str:
 
     @property
     def signature(self) -> 'Signature':
-        return Signature.build(a=self.a_bitsize, b=self.b_bitsize, result=self.result_bitsize)
+        return Signature.build_from_dtypes(
+            a=QInt(self.a_bitsize), b=QInt(self.b_bitsize), result=QInt(self.result_bitsize)
+        )
```
tanujkhattar, line 46: `PlusEqualProduct` is used to add into the phase gradient register (the `result` here) so the register types should be `QFxp` (I personally think most of the arithmetic gates should support `QFxp` but we can defer to later).

tanujkhattar (approved):
LGTM % nits

qualtran/bloqs/arithmetic/multiplication.py
fdmalone: Hmm I might be wrong here but this is more that if we have a real number in [0, 1) (QFxp(n, n), say) and scale it by a a b-bit integer then this would shift the fractional part by at most b so we would need a QFxp(n, n-b) register right? The trotter bloqs also assume the output register can be of a larger size but maybe this casting should be handled separately?

qualtran/bloqs/arithmetic/multiplication.py
fdmalone: Done.

qualtran/bloqs/chemistry/trotter/inverse_sqrt_test.py
fdmalone: Right, I had accidentally fed the integer bitsize into the real bitsize. The cost for the scaling a real by an int is like b_r * (2 * b_i - 1) - b_i^2, so the error is whatever this is when the arguments are flipped. I added a better test which should have previously failed.

qualtran/bloqs/arithmetic/multiplication.py
tanujkhattar: > if we have a real number in [0, 1) (QFxp(n, n), say) and scale it by a a b-bit integer then this would shift the fractional part by at most b so we would need a QFxp(n, n-b) register right

Yes, that's correct. And the reference assumes that the output register is always of the size `QFxp(n, n-b)`

> The trotter bloqs also assume the output register can be of a larger size but maybe this casting should be handled separately

I think it's fine to add support for it here - it'd be great if we can update the list of references to include the section in the trotter paper that assumes a different output register size. Do they have corresponding analysis for error propagation? Or is that left as an exercise for the reader?

fdmalone: It's a very back of the envelope calculation.

PR #644, opened 2024-02-07
[data types] Update data type definitions (fixed point and unsigned min bitsize)

fdmalone:
Pulled out from #634 and addressing #632. 

* QFixedPoint -> QFxp, with the specification largely matching that defined by Fxp (i.e. provide n_word (total bitsize) and n_frac). Given that their n_word is our bitsize everywhere else I think we could probably remove the num_qubits property from the dtype spec and just rely on bitsize.
* Relax the constraint on single bit unsigned types.
* Move BoundedQInt to BoundedQUInt, as we really only use the unsigned variant in the code currently.


qualtran/_infra/data_types.py
```diff
@@ -184,28 +172,46 @@ def num_qubits(self):
...
+    to represent the number. E.g. n_sign = 0 if bitsize = 8, num_frac = 6, signed = False,
+    then $\pi$ \approx 3.140625 = 11.001001, where the . represents the decimal place.
+
+    We can specify a fixed point real number by the tuple bitsize, num_frac and
+    signed, with num_int determined as (bitsize - num_frac - n_sign).
 
     Attributes:
-        int_bitsize: The number of qubits used to represent the integer part of the float.
-        frac_bitsize: The number of qubits used to represent the fractional part of the float.
+        bitsize: The number of qubits used to represent the fixed point real number.
```
tanujkhattar, line 189: nit: Update docstring. `bitsize` now corresponds to total number of bits for integer and fractional part combined.

qualtran/_infra/data_types.py
```diff
@@ -184,28 +172,46 @@ def num_qubits(self):
...
 
 @attrs.frozen
-class QFixedPoint(QDType):
-    """Fixed point type to represent real numbers.
+class QFxp(QDType):
+    r"""Fixed point type to represent real numbers.
 
-    The integer part of the float and the fractional part (the part after the
-    decimal plase) can be chosen to have different bitsizes. An additional sign bit is assumed.
+    A real number can be approximately represented in fixed point using num_int
```
tanujkhattar, line 178: nit: Escape variable names and equations in the docstring. Eg: `num_int` instead of num_int etc.

qualtran/_infra/data_types.py
```diff
@@ -184,28 +172,46 @@ def num_qubits(self):
...
+    bits for the integer part and num_frac bits for the fractional part. If the
+    real number is signed we require an additional bit to store the sign (0 for
+    +, 1 for -). In total there are bitsize = (n_sign + num_int + num_frac) bits used
+    to represent the number. E.g. n_sign = 0 if bitsize = 8, num_frac = 6, signed = False,
```
tanujkhattar, line 182:
```suggestion
    to represent the number. E.g. Using `(bitsize = 8, num_frac = 6, signed = False)`
```

qualtran/_infra/data_types.py
```diff
@@ -184,28 +172,46 @@ def num_qubits(self):
...
     def num_qubits(self):
-        if isinstance(self.int_bitsize, sympy.Expr):
-            return self.int_bitsize + self.frac_bitsize + sympy.symbols("1")
-        else:
-            return self.int_bitsize + self.frac_bitsize + 1
+        return self.bitsize
+
+    @property
+    def num_int(self) -> Union[int, sympy.Expr]:
+        num_int = self.bitsize - self.num_frac
```
tanujkhattar, line 205:
```suggestion
        return self.bitsize - self.num_frac - int(self.signed)
```

tanujkhattar (approved):
LGTM % nits

qualtran/_infra/data_types.py
fdmalone: Done.

qualtran/_infra/data_types.py
fdmalone: Done.

PR #646, opened 2024-02-08
Add a third installation option

NoureldinYosri:


mpharrigan:
I think the idea here was to have the commands that leaves you with a local, editable copy of the code

NoureldinYosri:
but this applies only to contributors, what about users who just want to use the latest version without waiting for the next release?

mpharrigan (approved):

PR #658, opened 2024-02-14
Notebook illustrating Qualtran -> QIR Compiler

skushnir123:


qualtran/_infra/Bloqs-Tutorial.ipynb
tanujkhattar, line 1: Are the changes here mostly due to running the notebook? We typically clear output of all notebooks before committing to the repository.

qualtran/qir_example.ipynb
tanujkhattar, line 1: Let's create a new folder `qualtran/qir_interop/` and put all the code for QIR interop there.

qualtran/qir_example.ipynb
```diff
@@ -0,0 +1,685 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 42,
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "import pyqir\n",
```
tanujkhattar, line 9: We'll have to add `pyqir` to the requirements file and recompile dependencies using `dev_tools/requirements/re-pip-compile.sh`

qualtran/qir_example.ipynb
```diff
@@ -0,0 +1,685 @@
...
+    "    ) -> Dict[str, SoquetT]:\n",
+    "        xs = bb.split(x)\n",
+    "        ys = bb.split(y)\n",
+    "\n",
+    "        for i in range(self.n):\n",
+    "            xs[i], ys[i] = bb.add(CNOT(), ctrl=xs[i], target=ys[i])\n",
+    "        return {\n",
+    "            'x': bb.join(xs),\n",
+    "            'y': bb.join(ys),\n",
+    "        }\n",
```
tanujkhattar, line 57: I think these three are good examples to use as test cases for the conversion code in the new `bloq_to_qir_test.py` file

qualtran/qir_example.ipynb
```diff
@@ -0,0 +1,685 @@
...
+    "def create_ir_map(bloq: qualtran.Bloq):\n",
+    "    param_counter = 0\n",
+    "    ir_map = {}\n",
+    "    # Loop through all registers in signature\n",
+    "    for register in bloq.signature.lefts():\n",
+    "        shape = register.shape[0] if len(register.shape) != 0 else 1 # get the shape as an int (we are asumming its 1d for simplicity)\n",
+    "        # map the (register_name, index_in_register) to the overall index\n",
+    "        ir_map.update({(register.name, i*register.bitsize + j): param_counter + i*register.bitsize + j for i in range(shape) for j in range(register.bitsize)})\n",
+    "        param_counter += register.bitsize * shape\n",
+    "    return ir_map\n",
```
tanujkhattar, line 166: We can move all the utility functions to a new file `bloq_to_qir.py` and add tests using the example cases shown above.

qualtran/qir_example.ipynb
```diff
@@ -0,0 +1,685 @@
...
+    "\n",
+    "def compile_bloq(bloq: qualtran.Bloq, qubit_type, void_type, module, context, builder, func_dict=dict()):\n",
+    "    func_name = f\"{bloq.pretty_name()}_{get_num_qubits_for_bloq(bloq)}\"\n",
+    "    if func_name in func_dict:\n",
+    "        return func_dict[func_name]\n",
+    "    bloq_func = create_func_for_bloq(bloq, func_name, qubit_type, void_type, module)\n",
+    "    ir_map = create_ir_map(bloq)\n",
+    "    soq_map = {}\n",
+    "\n",
+    "    # It seems like QFTTextBook and PhaseGradientUnitary can be decomposed even though supports_decompose_bloq returns false\n",
```
tanujkhattar, line 205: Seems like a bug in the `GateWithRegisters` base class. We should override this method for the `GateWithRegister` class to check whether the sub-bloq defines a cirq style decomposition and if yes, the return True from `supports_decompose_bloq`. 

Can you open a separate issue to track this?

cc @mpharrigan

qualtran/qir_example.ipynb
```diff
@@ -0,0 +1,685 @@
...
+    "                irs = []\n",
+    "                for i in range(single_soq.reg.bitsize):\n",
+    "                    irs.append(find_ir_for_index(param_indexes[param_index], ir_map))\n",
+    "                    param_index+=1\n",
+    "                soq_map[single_soq] = irs\n",
+    "\n",
+    "    return bloq_func, ir_map\n",
+    "\n",
+    "\n",
+    "def convert_qualtran(bloq: qualtran.Bloq):\n",
```
tanujkhattar, line 261:
```suggestion
    "def bloq_to_qir(bloq: qualtran.Bloq):\n",
```

qualtran/qir_example.ipynb
```diff
@@ -0,0 +1,685 @@
...
+    "    return map_soquet_to_params_helper(soq, soq_map, ir_map)\n",
+    "\n",
+    "def map_soquet_to_params_helper(soquet, soq_map, ir_map):\n",
+    "    if soquet in soq_map:\n",
+    "        irs = soq_map[soquet]\n",
+    "        return [ir_map[ir] for ir in irs]\n",
+    "    irs = generate_irs_from_soquet(soquet)\n",
+    "    return [ir_map[ir] for ir in irs]\n",
+    "\n",
+    "def compile_bloq(bloq: qualtran.Bloq, qubit_type, void_type, module, context, builder, func_dict=dict()):\n",
```
tanujkhattar, line 197: Please add docstrings and return type annotations to all the helper functions.

qualtran/qir_example.ipynb
```diff
@@ -0,0 +1,685 @@
...
+    "\n",
+    "        for i in range(self.n):\n",
+    "            xs[i], ys[i] = bb.add(ExampleBaseBloq(), x=xs[i], y=ys[i])\n",
+    "        return {\n",
+    "            'x': bb.join(xs),\n",
+    "            'y': bb.join(ys),\n",
+    "        }\n",
+    "\n",
+    "@attrs.frozen\n",
+    "class ExampleNonTrivialShapeBloq(Bloq):\n",
```
tanujkhattar, line 87: Another example to add (maybe as a followup) would be `ExampleNonTrivialSideBloq` which has LEFT and RIGHT directional registers instead of all registers being THRU registers. 

I think in terms of QIR, a LEFT register implies de-allocation happens inside that function and a `RIGHT` register corresponds to allocations happening inside that function. If the concepts don't translate directly (eg: because QIR assumes you take care of qubit allocations and deallocations globally and have a sea of globally allocated qubits) then I think both LEFT & RIGHT registers would be treated as inputs to the corresponding the QIR function (we do this when supporting cirq style decomposition where `decompose_from_registers()` method gets both LEFT & RIGHT registers as preallocated ancilla qubits.

Tl;Dr - Converting bloqs with directional registers should be tracked as a future improvement in the issue

tanujkhattar (changes requested):
This is a great start! It'd be pretty cool if we can also hookup the QIR output of QFT with the microsoft resource estimator to produce resource estimation numbers. But we can also do that in a follow up PR

qualtran/_infra/Bloqs-Tutorial.ipynb
skushnir123: Yes, just from running stuff. I can revert these changes.

qualtran/qir_example.ipynb
skushnir123: Ok will do

PR #669, opened 2024-02-15
[GQSP] Negative powers and Hamiltonian simulation

anurudhp:


anurudhp:
@tanujkhattar I've implemented the hamiltonian simulation now, along with the call graph. I added unittests with randomly generated select and prepare oracles. I'm still debugging why the expected block doesn't match e^iHt. In the meantime could you review the rest of the code?

tanujkhattar:
I'll take a look soon, thanks for the update

qualtran/_infra/adjoint.py
```diff
@@ -147,7 +147,7 @@ def decompose_from_registers(
         self, *, context: cirq.DecompositionContext, **quregs: NDArray[cirq.Qid]
     ) -> cirq.OP_TREE:
         if isinstance(self.subbloq, GateWithRegisters):
-            return cirq.inverse(self.subbloq.decompose_from_registers(context=context, **quregs))
+            return cirq.inverse(*self.subbloq.decompose_from_registers(context=context, **quregs))
```
tanujkhattar, line 150: I don't think you should do this.
```suggestion
            return cirq.inverse(self.subbloq.decompose_from_registers(context=context, **quregs))
```

qualtran/bloqs/generalized_qsp.py
```diff
@@ -251,22 +274,28 @@ class GeneralizedQSP(GateWithRegisters):
...
+    negative_power: int = field(default=0, kw_only=True)
 
     @cached_property
     def signature(self) -> Signature:
         return Signature([Register('signal', QBit()), *self.U.signature])
 
-    @cached_property
-    def Q(self):
-        return qsp_complementary_polynomial(self.P)
+    @staticmethod
```
tanujkhattar, line 293: Prefer a class method instead
```suggestion
    @classmethod
```

qualtran/bloqs/generalized_qsp.py
```diff
@@ -239,9 +260,11 @@ def safe_angle(x):
 class GeneralizedQSP(GateWithRegisters):
```
tanujkhattar, line 260: The can be implemented as a frozen dataclass using `@attrs.frozen`

qualtran/bloqs/generalized_qsp.py
```diff
@@ -284,13 +313,110 @@ def _phi(self) -> Sequence[float]:
...
-        for theta, phi in zip(self._theta[1:], self._phi[1:]):
-            yield self.U.on_registers(**quregs).controlled_by(signal_qubit, control_values=[0])
-            yield SU2RotationGate(theta, phi, 0).on(signal_qubit)
+        num_inverse_applications = self.negative_power
+
+        yield self.signal_rotations[0].on(signal_qubit)
+        for signal_rotation in self.signal_rotations[1:]:
+            if num_inverse_applications > 0:
+                # apply C-U^\dagger
+                yield (self.U**-1).on_registers(**quregs).controlled_by(signal_qubit)
```
tanujkhattar, line 336: nit: Use `self.U.adjoint` instead

qualtran/bloqs/generalized_qsp.py
```diff
@@ -284,13 +313,110 @@ def _phi(self) -> Sequence[float]:
...
+        degree = len(self.P)
+        return {
+            (self.U.adjoint().controlled(), min(degree, self.negative_power)),
+            (self.U.controlled(ctrl_spec=CtrlSpec(cvs=0)), max(0, degree - self.negative_power)),
+            (self.U.adjoint(), max(0, self.negative_power - degree)),
+        } | {(rotation, 1) for rotation in self.signal_rotations}
+
+
+@frozen
+class HamiltonianSimulationByGQSP(GateWithRegisters):
```
tanujkhattar, line 359: Add docstrings to the class

qualtran/bloqs/generalized_qsp.py
anurudhp: It is already marked `@frozen`

tanujkhattar: Sorry, I missed that. We shouldn't need an explicit hash function then?

anurudhp: the members `P, Q` could be `list`s or `NDArray`s, which are not hashable. So I had to work-around by hashing the essential data for the bloq (U + signal rotation sub-bloqs)

qualtran/bloqs/generalized_qsp.py
```diff
@@ -284,13 +313,140 @@ def _phi(self) -> Sequence[float]:
...
+            d += 1
+
+        coeff_indices = np.arange(-d, d + 1)
+        approx_cos = 1j**coeff_indices * scipy.special.jv(coeff_indices, self.t * self.alpha)
+        return approx_cos
+
+    @cached_property
+    def gqsp(self) -> GeneralizedQSP:
+        return GeneralizedQSP.from_qsp_polynomial(
+            self.walk_operator, self.approx_cos, negative_power=self.degree
```
anurudhp, line 405: I've narrowed down the issue to this: as the $e^{it\cos\theta}$ approximation is very accurate, it forces $Q = 0$, and therefore all the QSP angles (thetas) are also 0. This gives us the trivial final unitary $U^d$ (d = degree), and not $P(U)$.

Mathematically this makes sense as the only polynomial s.t. $|P(e^{i\theta})| = 1$ is $P(z) = z^k$. But in practice we just have a really good approximation and $Q \approx 0$, which gets treated as $Q = 0$. I'm not sure how to fix this.

@tanujkhattar

qualtran/bloqs/generalized_qsp.py
anurudhp: ah yes, that works! fixed now

qualtran/bloqs/generalized_qsp.py
```diff
@@ -63,15 +71,12 @@ def rotation_matrix(self):
...
-        yield cirq.Rz(rads=np.pi - self.lambd).on(qubit)
-        yield cirq.Ry(rads=2 * self.theta).on(qubit)
-        yield cirq.Rz(rads=-self.phi).on(qubit)
-        yield cirq.GlobalPhaseGate(np.exp(1j * (np.pi + self.lambd + self.phi) / 2)).on()
+    def build_composite_bloq(self, bb: 'BloqBuilder', q: 'SoquetT') -> Dict[str, 'SoquetT']:
+        q = bb.add(ZPowGate(exponent=2, global_shift=0.5), q=q)
+        q = bb.add(ZPowGate(exponent=1 - self.lambd / np.pi, global_shift=-1), q=q)
+        q = bb.add(Ry(angle=2 * self.theta), q=q)
+        q = bb.add(ZPowGate(exponent=-self.phi / np.pi, global_shift=-1), q=q)
+        return {'q': q}
```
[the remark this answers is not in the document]
anurudhp: Yep it works now (inheriting from GWR). But I thought it was qualtran gates/bloqs are preferred to cirq ones?

qualtran/bloqs/generalized_qsp.py
```diff
@@ -12,19 +12,27 @@
...
+from qualtran.bloqs.qubitization_walk_operator import QubitizationWalkOperator
+
+if TYPE_CHECKING:
+    from qualtran import BloqBuilder, SoquetT
+    from qualtran.resource_counting import BloqCountT, SympySymbolAllocator
 
 
 @frozen
-class SU2RotationGate(GateWithRegisters):
+class SU2RotationGate(Bloq):
```
tanujkhattar, line 35: Move `SU2RotationGate` to `basic_gates/` directory, override `as_cirq_op` to return a `PhasedXZGate` and override `CIRQ_GATE_TO_BLOQ_MAP` in `_cirq_gate_to_bloq` to convert a  `cirq.PhasedXZGate` to `SU2RotationGate`.  Add a builder method `.from_matrix`  so you can derive the parameters of the gate given an arbitrary 2-qubit unitary. Override `build_call_graph` and `wire_symbol` and test that `assert_decompose_is_consistent_with_t_complexity` and `bloq_autotester` works. Add a docstring (maybe just move the docstring from `rotation_matrix` method to the class) and add a `BloqExample`, update and run `dev_tools/autogenerate-bloqs-notebooks.py` to generate a doc and make sure `theta`, `phi` and `lambd` can be symbolics.

qualtran/bloqs/generalized_qsp.py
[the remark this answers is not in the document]
tanujkhattar: The good thing about `basic_gates/` is that we now do an automatic conversion from Cirq gates to Qualtran Bloqs wherever applicable. So your earlier implementation (deriving from `GWR`) would (a) Use the Cirq native gates when using a cirq-style API for simulations, which is often faster (see https://github.com/quantumlib/Qualtran/issues/793) and (b) Use the qualtran-native Bloqs when accessed via qualtran API like `bloq.decompose_bloq()` or `bloq.call_graph()`. 

Since this is being moved to `basic_gates/`, having a qualtran-only API (like you have right now) is good and an explicit conversion to/from `PhasedXZGate` in Cirq would be nice.

qualtran/bloqs/generalized_qsp.py
```diff
@@ -284,13 +313,136 @@ def _phi(self) -> Sequence[float]:
...
+        degree = len(self.P)
+        return {
+            (self.U.adjoint().controlled(), min(degree, self.negative_power)),
+            (self.U.controlled(ctrl_spec=CtrlSpec(cvs=0)), max(0, degree - self.negative_power)),
+            (self.U.adjoint(), max(0, self.negative_power - degree)),
+        } | {(rotation, 1) for rotation in self.signal_rotations}
+
+
+@frozen
+class HamiltonianSimulationByGQSP(GateWithRegisters):
```
tanujkhattar, line 356: Let's move out the "support negative powers" part in a separate PR and merge it, since it's working and the tests pass.

qualtran/bloqs/generalized_qsp.py
tanujkhattar, line 260: Let's create a new subpackage for `qualtran/bloqs/qsp/` and move the generalized qsp code inside the new folder. We can have two files right now

1. `qualtran/bloqs/qsp/generalized_qsp.py` - This should have `GeneralizedQSP`, `gqsp_phase_factors`, `gqsp_complementary_polynomial` (rename `qsp` to `gqsp` in the last 2 functions)
2. `qualtran/bloqs/qsp/polynomial_approximations.py` - This file can have functions which return coefficeints for commonly used polynomial approximations. For example: `approx_exp_cos(precision,  scaling_const)`  and `approx_exp_sin(precision,  scaling_const)`, which returns coefficients of the Jacobi-anger polynomial approximation which can be called directly from the HamiltonianSimulationViaGQSP Bloq. 


For `HamiltonianSimulationByGQSP`, we can create a new subpackage for `bloqs/hamiltonian_simulation/` and move `HamiltonianSimulationByGQSP` there. But this can be done once the above refactorings are complete to avoid merge conflicts.

tanujkhattar: For `GeneralizedQSP`, add a `bloq_example`, autogenerate the notebook from docstrings (you can add manually add more cells and they'll be preserve during the autogeneration process), add `bloq_autotester` tests and `assert_decompose_is_consistent_with_t_complexity`. 

Does it make sense to show some kind of symbolic analysis in GQSP?  The current API won't support it since we expect `P` and `Q` to be tuples of complex numbers and we currently don't have a way to specify sequences of variable symbolic lengths but we can file an issue to track this if there's a nice example where users would benefit from seeing the cost of applying GQSP as in terms of symbolic parameters they care about (for example hamiltonian simulation).

qualtran/bloqs/generalized_qsp_test.py
```diff
@@ -251,3 +296,192 @@ def test_generalized_real_qsp_with_symbolic_signal_matrix(degree: int):
...
+        P, Q = gqsp.P, gqsp.Q
+
+        check_polynomial_pair_on_random_points_on_unit_circle(
+            P, Q, random_state=random_state, rtol=2 * precision
+        )
+        verify_generalized_qsp(U, P, Q, negative_power=len(P) // 2)
+
+
+@frozen
+class RandomPrepareOracle(PrepareOracle):
```
tanujkhattar, line 341: Move these newly defined oracles to `bloqs/for_testing/` so they can be reused from elsewhere, if needed.

tanujkhattar (changes requested):
Left a round of comments for code organization and cleanup while I think more about the hamiltonian simulation issue. Thanks for doing this!

qualtran/bloqs/generalized_qsp_test.py
anurudhp: https://github.com/quantumlib/Qualtran/pull/851

anurudhp:
Split into https://github.com/quantumlib/Qualtran/pull/802, https://github.com/quantumlib/Qualtran/pull/810, https://github.com/quantumlib/Qualtran/pull/848, https://github.com/quantumlib/Qualtran/pull/851

PR #671, opened 2024-02-15
[data types] Deprecate register bitsize

fdmalone:
depends on #666. Blocked currently by serialization #672


qualtran/_infra/registers.py
```diff
@@ -61,24 +61,27 @@ class Register:
...
     shape: Tuple[int, ...] = field(
         default=tuple(), converter=lambda v: (v,) if isinstance(v, int) else tuple(v)
     )
     side: Side = Side.THRU
 
     def __attrs_post_init__(self):
-        if isinstance(self._bitsize, BoundedQUInt):
+        if not isinstance(self.dtype, QDType):
+            raise ValueError(f'dtype must be a QDType: found {type(self.dtype)}')
+        if isinstance(self.dtype, BoundedQUInt):
```
mpharrigan, line 76: I missed this from before. Why is this restriction in place? In principle you could have an array of numbers each from e.g. 1 to 5.

If we do want this: maybe make more general by having `QDType.validate_shape(shape)` method if there's some fundamental reason why some types shouldn't be arrayable

mpharrigan:
big if tru

qualtran/_infra/registers.py
fdmalone: This was copied over from a similar check in  SelectionRegister. I can't remember if it was a fundamental limitation there or just in place to prevent some untested functionality.

fdmalone:
big if THRU?

qualtran/_infra/registers.py
fdmalone: @tanujkhattar I remember asking before about why multidimensional selection registers were frowned upon, but can't quite remember the conclusion.

tanujkhattar: `Signature([Register('x', BoundedQUInt(n, N)), Register('y', BoundedQUint(m, M)])` represents signature of a Bloq that expects to do a coherent for-loop over `N * M` values; where variable `x` can take values between `[0, N)` and variable `y` can take values between `[0, M)`. 

`Signature([Register('x', BoundedQUInt(n, N), shape=(2, )])` - Now this represents two variables `x[0]` and `x[1]`, both of which can take values between `[0, N)`. So a total of `N ** 2` values. 

As the size of the multidimensional register increases, the number of values we can iterate upon increases exponentially. Also, the iteration length needs to be same across all variables. And it's a bit nontrivial to reason about the correctness of this case. For all these reasons, we decided to avoid multi-dimensional selection registers and forced users to declare one selection registers for every for-loop variable as part of the signature of the Bloq; even if two of the variables had the same iteration length and could in principle be represented using a multi dimensional selection register. 


Tl;Dr - It was mainly to enforce more expressive workflow and avoid confusing constructs in the specific case of selection registers. I think `BoundedQUint` is a more general abstraction that I can imagine using outside of the selection registers context (eg: to represent integers mod M) and so I think we can drop this assertion.

fdmalone:
I updated the serialization proto stuff to use the regisers dtype. Which required some intermediate boilerplate similar to what's required for IntOrSympy serialization.

qualtran/_infra/registers.py
fdmalone: I'll remove the exception.

qualtran/serialization/data_types.py
```diff
@@ -86,3 +88,83 @@ def data_type_from_proto(serialized: SerializedDataTypes) -> QDType:
             "QAny, QInt, QIntOnesComp, QUInt, BoundedQUInt, "
             "QFixedPoint"
         )
+
+
+def int_or_sympy_from_qdt_proto(val: data_types_pb2.QDataType) -> Union[int, sympy.Expr]:
+    if val.HasField('qbit'):
+        return 1
```
tanujkhattar, line 95: I think we should raise a value error if `val.HasField('qbit')` is true ? We've already checked the callsite and expect that it won't happen

qualtran/protos/data_types.proto
```diff
@@ -45,3 +47,15 @@ message QFxp {
     optional IntOrSympy num_frac = 2;
     optional bool signed = 3;
 }
+
+message QDataType {
+  oneof val {
+    QBit qbit = 1;
+    QAny qany = 2;
+    QInt qint = 3;
+    QIntOnesComp qintoc = 4;
```
tanujkhattar, line 56: maybe have `qint_ones_comp` to match the type?

qualtran/protos/data_types.proto
```diff
@@ -45,3 +47,15 @@ message QFxp {
...
+    QUInt quint = 5;
+    BoundedQUInt bquint = 6;
```
tanujkhattar, line 58: ditto
```suggestion
    BoundedQUInt bounded_quint = 6;
```

qualtran/serialization/data_types.py
```diff
@@ -86,3 +88,83 @@ def data_type_from_proto(serialized: SerializedDataTypes) -> QDType:
...
+    if val.HasField('quint'):
+        return int_or_sympy_from_proto(val.quint.bitsize)
+    if val.HasField('bquint'):
+        return int_or_sympy_from_proto(val.bquint.bitsize)
+    if val.HasField('qfxp'):
+        return int_or_sympy_from_proto(val.qfxp.bitsize)
+    raise ValueError(f"Cannot deserialize {val=}")
+
+
+def data_type_to_qdt_proto(data: QDType) -> data_types_pb2.QDataType:
```
tanujkhattar, line 111: I think we can simply update  `data_type_to_proto`  and `data_type_from_proto` methods to expect and return a `QDType` instead of `SerializedDataTypes` and not have the two new methods implemented ?

mpharrigan (commented):
lgtm

qualtran/serialization/data_types.py
fdmalone: One could call this function from outside of data_type_to_qdt_proto no?

qualtran/serialization/data_types.py
fdmalone: I guess my only question was if we ever would want the actual specific _pb2 data type (QAny, ...) vs the QDataType "wrapper".

fdmalone: Given that they're not used outside of the tests my guess is no, but maybe?

tanujkhattar: I think we'll always be dealing with `QDataType` wrapper in the proto serialization, so I vote for changing the code to deal with `QDataType` directly. 

We can consider making the existing functions internal helpers and changing your code to delegate to those helpers instead of rewriting the entire implementation. For example:

```python
def data_type_to_qdt_proto(data: QDType) -> data_types_pb2.QDataType:
    serialized_proto = data_type_to_proto(data)
    kwargs = {data_type_to_field_name[type(data)]:serialized_proto}
    return data_types_pb2.QDataType(**kwargs)

def data_type_from_qdt_proto(serialized: data_types_pb2.QDataType:) -> QDType:
    return data_type_from_proto(getattr(serialized, serialized.WhichOneof('val')))
```

qualtran/protos/data_types.proto
fdmalone: Done.

qualtran/protos/data_types.proto
fdmalone: Done.

qualtran/serialization/data_types.py
fdmalone: Done.

qualtran/serialization/data_types.py
fdmalone: I went with the delete code suggestion!

mpharrigan (approved):

qualtran/serialization/data_types.py
```diff
@@ -13,40 +13,56 @@
...
+        return int_or_sympy_from_proto(val.qint.bitsize)
+    if val.HasField('qint_ones_comp'):
+        return int_or_sympy_from_proto(val.qint_ones_comp.bitsize)
+    if val.HasField('quint'):
+        return int_or_sympy_from_proto(val.quint.bitsize)
+    if val.HasField('bounded_quint'):
+        return int_or_sympy_from_proto(val.bounded_quint.bitsize)
+    if val.HasField('qfxp'):
+        return int_or_sympy_from_proto(val.qfxp.bitsize)
+    raise ValueError(f"Cannot deserialize {val=}")
```
tanujkhattar, line 38: I think this helper is moot and should be removed, the name and interface is confusing and it's not saving us much since we are doing all the same isinstance checks in the `data_type_to_proto` function below.

fdmalone: Sorry I was a bit too eager to hit merge: https://github.com/quantumlib/Qualtran/pull/698

PR #674, opened 2024-02-16
Implements the Bloq Decomposition for modular addition in montgomery form

fpapa250:
Doesn't have classical simulation because it requires the ability to do signed addition with `Add()`. Also corrects an error in the paper where it shows a NOT controlled by y > x, but should be x > y. (I can't prove it without classical simulation, but I worked it out on paper).


qualtran/bloqs/factoring/mod_add_test.py
```diff
@@ -31,3 +34,9 @@ def test_ctrl_mod_add_k():
     counts = bloq.bloq_counts()
     ((bloq, n),) = counts.items()
     assert n == 5
+
+
+@pytest.mark.parametrize('bitsize,p', [(1, 1), (2, 3), (5, 8)])
+def test_montgomery_mod_add_decomp(bitsize, p):
+    bloq = MontgomeryModAdd(bitsize=bitsize, p=p)
+    assert_valid_bloq_decomposition(bloq)
```
tanujkhattar, line 42: Can we add a correctness test using tensor network simulation for a small case, till the time the classical simulation is not available?

fpapa250: I'm not familiar with tensor network simulation. Is there a doc or example test I can take a look at?

mpharrigan: For an effective test, one would need to implement the tensor for the new Bloq. In principle, this should be derivable from the classical action. I suggest in this instance we take out a techdebt loan and don't block this implementation on better testing infrastructure. The dividends come from the fact that it will be easier to develop and test the classical simulation and tensor stuff with these examples in the library

qualtran/bloqs/factoring/mod_add.py
```diff
@@ -138,3 +141,103 @@ def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
...
+    References:
+        [How to compute a 256-bit elliptic curve private key with only 50 million Toffoli gates](https://arxiv.org/abs/2306.08585) Fig 6a and 8
+    """
+
+    bitsize: int
+    p: int
+
+    @cached_property
+    def signature(self) -> 'Signature':
+        return Signature([Register('x', bitsize=self.bitsize), Register('y', bitsize=self.bitsize)])
```
tanujkhattar, line 171: Can we use the new dtypes here?

qualtran/bloqs/factoring/mod_add.py
```diff
@@ -138,3 +141,103 @@ def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
...
+        # constant subtraction circuit.
+        x_split = bb.split(x)
+        y_split = bb.split(y)
+        x = bb.join(np.concatenate([[junk_bit], x_split]))
+        y = bb.join(np.concatenate([[sign], y_split]))
+
+        # Perform in-place addition on quantum register y.
+        x, y = bb.add(Add(bitsize=self.bitsize + 1), a=x, b=y)
+
+        # Temporary solution to equalize the bitlength of the x and y registers for Add().
```
tanujkhattar, line 204: Can we change the signature of `Add` to support `a_bitsize` and `b_bitsize`; update tensor decomposition and open an issue that tracks updating the decomposition? So that the costs and tensor simulations / classical simulations work for aribtrary bitsizes but the decomposition works only when bitsizes are the same and that can be updated later? 

I think it's much harder to find such TODOs than fixing the decomposition later.

qualtran/bloqs/factoring/mod_add.py
tanujkhattar: Also, should we support different bitsizes for x and y? Is it nontrivial to add that support?

qualtran/bloqs/factoring/mod_add_test.py
tanujkhattar: I think it's worth overriding the `add_my_tensors` for the new Bloq at least. We can leave a TODO to fix the `on_classical_vals` and `add_my_tensors` for Bloqs that appear in the decomposition to verify that the decomposition is consistent with the Bloq action. But currently the Bloq doesn't define it's in terms of tensor network simulation and I think it's worth adding that in this PR. The code should be fairly straightforward, something like

```python
    def add_my_tensors(
        self,
        tn: 'qtn.TensorNetwork',
        tag: Any,
        *,
        incoming: Dict[str, 'SoquetT'],
        outgoing: Dict[str, 'SoquetT'],
    ):
        import quimb.tensor as qtn

        N = 2**self.bitsize
        inds = (incoming['x'], incoming['y'], outgoing['x'], outgoing['y'])
        data = np.zeros((N, N) * 2, dtype=np.complex128)
        # TODO: Add a value-to-index method on dtype to make this easier.
        for x, y in itertools.product(range(N), range(N)):
            out = self.on_classical_vals(x=x, y=y)
            data[x, y, out['x'], out['y']] = 1
        tn.add(qtn.Tensor(data=data, inds=inds, tags=[self.short_name(), tag]))
```

qualtran/bloqs/factoring/mod_add.py
fpapa250: I am going to add a new data type MontgomoryUInt which is an unsigned int but in montgomery form (https://en.wikipedia.org/wiki/Montgomery_modular_multiplication). Anything in MontgomeryUInt can be treated as UInt, but not all UInt are in Montgomery form. All mod arithmetic bloqs in the PsiQuantum paper are designed to operate on integers in montgomery form.

I didn't add different bitsizes for x and y because I'm just following the PsiQuantum paper which proposes subroutines with equal bitsizes. The added complexity of adding different bitsize capabilities would involve making sure Add() and SimpleAddConstant() support different bitsizes which really just boils down to adding it for Add().

qualtran/bloqs/factoring/mod_add.py
mpharrigan: I'm going to say this is outside of the scope of the current PR

mpharrigan (approved):

Issue #668, opened 2024-02-15
`PhaseGradientState` exponent should be float or int?


mpharrigan:
`PhaseGradientState` annotates exponent as `int` and checks that in `__pow__`, but unit tests use floating point values.

tanujkhattar:
It can be float. I'll send a fix.

However, this raises a good question - how do we handle `__pow__` for bloqs that have directional registers and also support arbitrary `exponent` ? For example, in this case should `phase_gradient_state ** -1` return an effect with LEFT register instead of a right register? What about `phase_gradient_state ** -0.5` ? 

I think in general, `bloq ** -1` and `bloq.adjoint` should be consistent. What do you think?

PR #675, opened 2024-02-16
Remove adjoint attribute from PhaseGradientState and use the default Adjoint Bloq instead

tanujkhattar:
Fixes https://github.com/quantumlib/Qualtran/issues/668

I also fixed an issue because of which Cirq simulations would fail on using `Adjoint`. I remember @mpharrigan had encountered this before but I'm unable to find the relevant issue.

### DIFF
--- a/qualtran/_infra/adjoint.py
+++ b/qualtran/_infra/adjoint.py
@@ -16,7 +16,9 @@ from functools import cached_property
 from typing import Dict, List, Set, Tuple, TYPE_CHECKING
 
 import attrs
+import cirq
 from attrs import frozen
+from numpy.typing import NDArray
 
 from .composite_bloq import _binst_to_cxns, _cxn_to_soq_dict, _map_soqs, _reg_to_soq, BloqBuilder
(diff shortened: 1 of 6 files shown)


qualtran/_infra/adjoint.py
```diff
@@ -141,6 +143,13 @@ def decompose_bloq(self) -> 'CompositeBloq':
         """The decomposition is the adjoint of `subbloq`'s decomposition."""
         return self.subbloq.decompose_bloq().adjoint()
 
+    def decompose_from_registers(
+        self, *, context: cirq.DecompositionContext, **quregs: NDArray[cirq.Qid]
+    ) -> cirq.OP_TREE:
+        if isinstance(self.subbloq, GateWithRegisters):
+            return cirq.inverse(self.subbloq.decompose_from_registers(context=context, **quregs))
+        return super().decompose_from_registers(context=context, **quregs)
```
mpharrigan, line 151: thanks

mpharrigan (approved):
sweet

PR #678, opened 2024-02-16
Unify t-complexity protocol

### LINKED ISSUE #489: Have a single `t_complexity` protocol instead of `_t_complexity_` and `t_complexity`
The two different protocols along with the special case implementation in `cirq_interop/` adds complexity and makes the fallbacks hard to maintain. The cirq-style protocol should be deprecated in favour of bloq-style `t_complexity` but we still need to import all the goodies of Cirq-style protocol (like caching for performance, fallback strategies to check if a cirq-style gate is clifford/t using  `_is_clifford_or_t` etc. 

One example of this complexity is the tests in https://github.com/quantumlib/Qualtran/pull/488. Even though we are using `ZPowGate` defined in Qualtran, the reported costs are in terms of `t` gates and `rotations` because cirq-style `t_complexity(op)`  protocol currently ignores the `gate.t_complexity` defined on Bloqs.

### LINKED ISSUE #662: t_complexity rotations
CZPowGate and ZPowGate `t_complexity` and/or `_t_complexity_()` disagree. ZPowGate is a RotationBloq and uses a formula to return t counts for the bloq, but only when using the Bloq t_complexity method (unknown what happens if you use the cirq t_complexity protocol). CZPowGate returns a rotation=1 count. 

This shows up in the test
`qualtran.bloqs.rotations.phase_gradient_test.test_phase_gradient_state`

which seemingly expects everything to be kept as a "rotation".

mpharrigan:
- switch to using cirq-style `t_complexity(x: Any)` strategies-based protocol for counting T gates
 - Remove `Bloq.t_complexity()` overridable method. For convenience `Bloq.t_complexity` just forwards to the above function
 - Add strategies for dealing with bloqs and compositebloqs
 - Move rotations to `rotations=1` and a new method on `TComplexity` if you want an estimate for everything in terms of T instead of T+rotations. 

Fixes #662 
Fixes #489

### DIFF
--- a/qualtran/bloqs/basic_gates/rotation.py
+++ b/qualtran/bloqs/basic_gates/rotation.py
@@ -12,7 +12,6 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-import abc
 from functools import cached_property
 from typing import Protocol
 
--- a/qualtran/bloqs/basic_gates/rotation.py
+++ b/qualtran/bloqs/basic_gates/rotation.py
@@ -31,13 +31,8 @@ class _HasEps(Protocol):
     eps: float
 
 
-class _RotationBloq(CirqGateAsBloqBase, metaclass=abc.ABCMeta):
-    def _t_complexity_(self: _HasEps):
-        return TComplexity(rotations=1)
-
-
 @frozen
-class ZPowGate(_RotationBloq):
+class ZPowGate(CirqGateAsBloqBase):
     r"""A gate that rotates around the Z axis of the Bloch sphere.
 
     The unitary matrix of `ZPowGate(exponent=t, global_shift=s)` is:
@@ -109,7 +104,7 @@ class CZPowGate(CirqGateAsBloqBase):
 
 
 @frozen
-class XPowGate(_RotationBloq):
+class XPowGate(CirqGateAsBloqBase):
     r"""A gate that rotates around the X axis of the Bloch sphere.
 
     The unitary matrix of `XPowGate(exponent=t, global_shift=s)` is:
@@ -158,7 +153,7 @@ class XPowGate(_RotationBloq):
 
 
 @frozen
-class YPowGate(_RotationBloq):
+class YPowGate(CirqGateAsBloqBase):
     r"""A gate that rotates around the Y axis of the Bloch sphere.
 
     The unitary matrix of `YPowGate(exponent=t)` is:
@@ -207,7 +202,7 @@ class YPowGate(_RotationBloq):
 
 
 @frozen
-class Rz(_RotationBloq):
+class Rz(CirqGateAsBloqBase):
     """Single-qubit Rz gate.
 
     Args:
@@ -233,7 +228,7 @@ class Rz(_RotationBloq):
 
 
 @frozen
-class Rx(_RotationBloq):
+class Rx(CirqGateAsBloqBase):
     angle: float
     eps: float = 1e-11
 
@@ -243,7 +238,7 @@ class Rx(_RotationBloq):
 
 
 @frozen
-class Ry(_RotationBloq):
+class Ry(CirqGateAsBloqBase):
     angle: float
     eps: float = 1e-11
 
--- a/qualtran/bloqs/basic_gates/rotation.py
+++ b/qualtran/bloqs/basic_gates/rotation.py
@@ -14,19 +14,15 @@
 
 import abc
 from functools import cached_property
-from typing import Protocol, Set, TYPE_CHECKING
+from typing import Protocol
 
 import cirq
 import numpy as np
 from attrs import frozen
 
 from qualtran import bloq_example
-from qualtran.bloqs.basic_gates.t_gate import TGate
 from qualtran.cirq_interop import CirqGateAsBloqBase
 from qualtran.cirq_interop.t_complexity_protocol import TComplexity
-
-if TYPE_CHECKING:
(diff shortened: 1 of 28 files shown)


qualtran/bloqs/rotations/phase_gradient_test.py
```diff
@@ -37,12 +37,13 @@ def test_phase_gradient_state(n: int):
...
-    assert np.allclose(
+    np.testing.assert_allclose(cirq.unitary(gate), cirq.unitary(state_prep_cirq_circuit))
+    np.testing.assert_allclose(
         cirq.unitary(gate**-1), cirq.unitary(cirq.inverse(state_prep_cirq_circuit))
     )
-    assert gate.t_complexity().rotations == n - 2
-    assert gate.t_complexity().clifford == n + 2
+    assert gate.t_complexity().t == 0
+    assert gate.t_complexity().rotations == n
+    assert gate.t_complexity().clifford == n
```
mpharrigan, line 46: @tanujkhattar is this right?

qualtran/bloqs/basic_gates/rotation.py
```diff
@@ -36,18 +32,8 @@ class _HasEps(Protocol):
...
-        # See: https://github.com/quantumlib/Qualtran/issues/219
-        # See: https://github.com/quantumlib/Qualtran/issues/217
-        num_t = int(np.ceil(1.149 * np.log2(1.0 / self.eps) + 9.2))
-        return TComplexity(t=num_t)
-
-    def build_call_graph(self: _HasEps, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
-        num_t = _RotationBloq.t_complexity(self).t
-        return {(TGate(), num_t)}
+    def _t_complexity_(self: _HasEps):
+        return TComplexity(rotations=1)
```
tanujkhattar, line 36: I think we should do `return t_complexity(self.cirq_gate)` and that should fix the issue discussed below.

qualtran/bloqs/rotations/phase_gradient_test.py
tanujkhattar: No, this is incorrect. `PhaseGradientState(n)`  decomposes into $n$ Hadamards and $n$ Z rotations of the form $Z^{2^i}$. 2 out of those $n$ Z rotations are cliffords - $Z^1$ and $Z^0.5$ and remaining are non-clifford rotations.

qualtran/cirq_interop/t_complexity_protocol.py
```diff
@@ -31,6 +33,17 @@ class TComplexity:
...
+        """Return the total number of T gates after compiling rotations"""
+
+        # TODO Determine precise clifford count and/or ignore.
+        # This is an improvement over Ref. 2 from the docstring which provides
+        # a bound of 3 log(1/eps).
+        # See: https://github.com/quantumlib/Qualtran/issues/219
+        # See: https://github.com/quantumlib/Qualtran/issues/217
+        t_per_rot = int(np.ceil(1.149 * np.log2(1.0 / eps) + 9.2))
+        return self.t + t_per_rot * self.rotations
+
```
tanujkhattar, line 46: This is fine as a first step, but we need to arrive at the right solution soon as highlighted here - https://github.com/quantumlib/Qualtran/issues/662#issuecomment-1949215534

I'm curious to see how quickly we can deprecate T-complexity in favour of Bloq counts. If it takes a while, I think it might be reasonable to update TComplexity to keep track of rotations as tuples in the meantime since it's such a fundamental part of cost analysis in the library right now.

qualtran/bloqs/rotations/phase_gradient_test.py
mpharrigan: doing the suggested change leaves t=1, rot=n-3, clifford=n+2; which works by your logic if you consider z/4 to as a T

qualtran/cirq_interop/t_complexity_protocol.py
mpharrigan: I mean you can already count rotations in that manner if you add them to the `keep` predicate

qualtran/bloqs/basic_gates/rotation_test.py
```diff
@@ -44,7 +44,7 @@ def test_rotation_gates():
     tcount = 52
     assert Rx(angle).t_complexity().t_incl_rotations() == tcount
     assert Ry(angle).t_complexity().t_incl_rotations() == tcount
-    assert Rz(angle).t_complexity().t_incl_rotations() == tcount
+    assert Rz(angle).t_complexity().t_incl_rotations() == 1
```
mpharrigan, line 47: flagging this for second review @tanujkhattar @fdmalone

tanujkhattar: LGTM. And this just makes me more suspicious that T-costs of rotations currently don't depend upon the epsilon. But that's a TODO already tracked in one of the issues so the change LGTM

mpharrigan: I think almost all of the tests use the default epsilon

tanujkhattar: Sorry, I meant rotation angle and not epsilon in my previous comment.

mpharrigan: ah, yes; that is an issue :)

PR #685, opened 2024-02-18
Improvements to `PhasingViaCostFunction` infrastructure

tanujkhattar:
Fixes https://github.com/quantumlib/Qualtran/issues/654

PR #688, opened 2024-02-18
Support symbolic analysis in rotations and phase gradient bloqs

tanujkhattar:
Currently based on top of https://github.com/quantumlib/Qualtran/pull/685

Adds a `utils.py` for to pick between numpy vs sympy functions for utilities like `log` / `ceil` etc.

### DIFF
--- a/qualtran/bloqs/utils.py
+++ b/qualtran/bloqs/utils.py
@@ -1,0 +1,39 @@
+#  Copyright 2023 Google LLC
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+import numbers
+from typing import Union
+
+import numpy as np
+import sympy
+from cirq._doc import document
+
+SymbolicFloat = Union[float, sympy.Expr]
+document(SymbolicFloat, """A floating point value or a sympy expression.""")
+
+SymbolicInt = Union[int, sympy.Expr]
+document(SymbolicFloat, """A floating point value or a sympy expression.""")
+
+
+def log2(x: SymbolicFloat) -> SymbolicFloat:
+    from sympy.codegen.cfunctions import log2
+
+    if isinstance(x, numbers.Number):
+        return np.log2(x)
+    return log2(x)
+
+
+def ceil(x: SymbolicFloat) -> SymbolicInt:
+    if isinstance(x, numbers.Number):
+        return int(np.ceil(x))
+    return sympy.ceiling(x)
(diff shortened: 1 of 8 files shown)


fdmalone:
I know from experience "utils" as a module name is going to be shot down :)

qualtran/bloqs/utils.py
```diff
@@ -0,0 +1,38 @@
...
+
+    if not isinstance(x, sympy.Basic):
+        return np.log2(x)
+    return log2(x)
+
+
+def ceil(x: SymbolicFloat) -> SymbolicInt:
+    if not isinstance(x, sympy.Basic):
+        return int(np.ceil(x))
+    return sympy.ceiling(x)
```
mpharrigan, line 38: why not always use the sympy functions? they should work on concrete values

qualtran/cirq_interop/_cirq_to_bloq.py
```diff
@@ -276,15 +276,15 @@ def _ensure_in_reg_exists(
...
     for qreg, soq in new_qreg_to_qvar.items():
         if len(in_reg_qubits) > 1 and qreg.qubits and qreg.qubits[0] in in_reg_qubits:
             assert len(qreg.qubits) == 1, "Individual qubits should have been split by now."
-            soqs_to_join.append(soq)
+            soqs_to_join[qreg.qubits[0]] = soq
         else:
             qreg_to_qvar[qreg] = soq
     if soqs_to_join:
-        qreg_to_qvar[in_reg] = bb.join(np.array(soqs_to_join))
+        qreg_to_qvar[in_reg] = bb.join(np.array([soqs_to_join[q] for q in in_reg.qubits]))
```
mpharrigan, line 287: was this a bug?

tanujkhattar: Yes, I'll write a description but this is captured in the previous PR (https://github.com/quantumlib/Qualtran/pull/685) as well. 

Essentially, the ordering of soquets to join should match the ordering of qubits in the register otherwise a 1 (`001`) can be split and joined into an 8 (`100`).

qualtran/bloqs/utils.py
tanujkhattar: It's because they return sympy types and it's hard to distinguish whether a bloq is parameterized or not (eg: using cirq.is_parameterized(val) which uses a bunch of isinstance checks)

```python
In [7]: isinstance(sympy.log(2), numbers.Number)
Out[7]: False

In [8]: type(sympy.log(2))
Out[8]: log
```

mpharrigan: Just for completeness: sympy has evaluation rules which keeps `sympy.log(2)` as symbolic because there isn't a way of representing the result with the same precision as the input. If you use floats or in the `ceiling` example: it gets evaluated.

```python
isinstance(sympy.log(2.), numbers.Number)
Out[15]: True
isinstance(sympy.ceiling(2), numbers.Number)
Out[16]: True
isinstance(sympy.ceiling(2.), numbers.Number)
Out[17]: True
```

https://docs.sympy.org/latest/guides/custom-functions.html

It might be less surprising for people if we say "for symbolic-supporting gates, the symbolic rules of sympy are used" (note: this is also how mathematica does it)

mpharrigan: I bet there will be cases where we want things to always evaluate numerically if possible but I bet there will be cases where someone will want to keep something as a sympy expression and not have it considered "parameterized". The "if it's sympy then it's a parameterized gate" logic is not quite right. 

For the first case: instead of these functions with "if" it might be better to keep things sympy until we need them to be numbers and then do the sympy evaluate-as-float button. In mathematica it's `// N` I'm sure such a thing exists in sympy too.

mpharrigan: update: it's `N` in sympy too https://docs.sympy.org/latest/modules/evalf.html

tanujkhattar:
I've pulled out the logic into separate PRs. This can now be closed.

PR #693, opened 2024-02-21
Add 15-to-1 magic state factory

NoureldinYosri:
This PR is the first of a series of PRs that add T state distillation and improve the resource estimation workflow.

In this PR I implemented the 15-to-1 as it's presented in https://arxiv.org/abs/1905.06903. A few points of note:
1. The 4th rotation in Figure 3 gets moved to after the the 9th rotation.
2. I added comments to show which rotation is which.
3. all rotation which are supposed to be unitary in Figure 3 become channels that can either do the rotation $P\frac{\pi}{8}$ where $P$ is a pauli string or overshoot to $P\frac{5\pi}{8}$, $P\frac{-\pi}{8}$, or $P\frac{3\pi}{8}$.
4. Z and X errors are created as [cirq.AsymmetricDepolarizingChannel](https://quantumai.google/reference/python/cirq/asymmetric_depolarize) channels using the `storage_error` method.
5. The probabilities in each of the channels are the same as in the [supplementary materials](https://github.com/litinski/magicstates/blob/79d99a5f21e63f992f63e644ea43a665944b4bb5/Python/onelevel15to1.py#L8) of https://arxiv.org/abs/1905.06903.
6. The `n_cycles` method in magic factories should take a physical error parameter. the CCZ model that we have https://arxiv.org/abs/1812.01238 assumes physical error rate of $1e-3$ (`Assuming a physical gate error rate of 10^−3, our CCZ factory can produce ...`)

---

This PR reintroduces MagicCount dataclass to be used exclusively with magic factories. the reason for this is that although AlgorithmSummary is a superset of MagicCount, it becomes unclear what to do when AlgorithmSummary has other counts (e.g. rotations). one way would be to error out but I think a seprate class should make things cleaner.


qualtran/surface_code/fifteen_to_one_test.py
```diff
@@ -0,0 +1,49 @@
...
+    TestCase(d_X=9, d_Z=3, d_m=3, phys_err=1e-4, p_out=9.3e-10, footprint=1150, cycles=18.1),
+    TestCase(d_X=11, d_Z=5, d_m=5, phys_err=1e-4, p_out=1.9e-11, footprint=2070, cycles=30),
+    TestCase(d_X=17, d_Z=7, d_m=7, phys_err=1e-3, p_out=4.5e-8, footprint=4620, cycles=42.6),
+]
+
+
+@pytest.mark.parametrize("test", PAPER_RESULTS)
+def test_compare_with_paper(test: TestCase):
+    factory = FifteenToOne(test.d_X, test.d_Z, test.d_m)
+    assert f'{factory.distillation_error(MagicCount(n_t=1), test.phys_err):.1e}' == str(test.p_out)
```
mpharrigan, line 47: consider using `np.assert_allclose(x, y, rtol=...)`

qualtran/surface_code/fifteen_to_one_test.py
```diff
@@ -0,0 +1,49 @@
...
+    assert round(factory.footprint(), -1) == test.footprint  # rounding to the 10s digit.
+    assert factory.n_cycles(MagicCount(n_t=1), test.phys_err) == math.ceil(test.cycles + 1e-9)
```
mpharrigan, line 49: why are you adding 1e-9? consider using `np.testing.assert_allclose`

qualtran/surface_code/fifteen_to_one_test.py
```diff
@@ -0,0 +1,49 @@
...
+    TestCase(d_X=11, d_Z=5, d_m=5, phys_err=1e-4, p_out=1.9e-11, footprint=2070, cycles=30),
+    TestCase(d_X=17, d_Z=7, d_m=7, phys_err=1e-3, p_out=4.5e-8, footprint=4620, cycles=42.6),
+]
+
+
+@pytest.mark.parametrize("test", PAPER_RESULTS)
+def test_compare_with_paper(test: TestCase):
+    factory = FifteenToOne(test.d_X, test.d_Z, test.d_m)
+    assert f'{factory.distillation_error(MagicCount(n_t=1), test.phys_err):.1e}' == str(test.p_out)
+    assert round(factory.footprint(), -1) == test.footprint  # rounding to the 10s digit.
```
mpharrigan, line 48: consider using `np.assert_allclose(x, y, atol)`

qualtran/surface_code/t_factory_utils.py
```diff
@@ -0,0 +1,91 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from typing import Sequence
+
+import cirq
+import numpy as np
+
+
+def _rotation_unitary(pauli_string: str, theta: float):
```
mpharrigan, line 21: short docstring

qualtran/surface_code/t_factory_utils.py
```diff
@@ -0,0 +1,91 @@
...
+    num_qubits = len(pauli_string)
+    u = cirq.unitary(cirq.DensePauliString(pauli_string))
+    return np.cos(theta) * np.eye(1 << num_qubits) + 1j * np.sin(theta) * u
+
+
+class PauliRotationChannel(cirq.Gate):
```
mpharrigan, line 27: ok I was confused when you were presenting today because I thought this might be something that already exists in cirq and I was confused how you were getting the error probabilities into it.

Long story short: I think this needs a different name. Like `NoisyPauliRotation` or `OverrotatingPauliString` or `PauliRotationWithError`. 

I would assume something called `PauliRotationChannel` would do a (noiseless) pauli rotation on a density matrix. In quantum information, channels are a superset of unitary operations and does not imply that the underlying operation _isn't_ unitary

qualtran/surface_code/t_factory_utils.py
```diff
@@ -0,0 +1,91 @@
...
+    r"""A channel that applies a pi/8 pauli rotation with possible overshooting to 5pi/8, -pi/8, and 3pi/8.
+
+    The channel is defined as
+
+    $$
+        \sum_{k \in \{1, 5, -1, 3\}} p_k e^{-i \frac{\pi}{8} k P} \rho e^{i \frac{\pi}{8} k P}
+    $$
+    """
+
+    def __init__(self, pauli_string, p1, p2, p3):
```
mpharrigan, line 37: type annotations and document these arguments. E.g.

```
Args:
  pauli_string: string of 'x' 'y' 'z' 'i' ...
  p1: the probability of doing a 5pi/8 rotation
```

qualtran/surface_code/t_factory_utils.py
```diff
@@ -0,0 +1,91 @@
...
+    def _circuit_diagram_info_(
+        self, args: 'cirq.CircuitDiagramInfoArgs'
+    ) -> cirq.CircuitDiagramInfo:
+        return cirq.CircuitDiagramInfo(wire_symbols=tuple(self.pauli_string))
+
+
+def storage_error(
+    kind: str, probabilities: Sequence[float], qubits: Sequence[cirq.Qid]
+) -> Sequence[cirq.Operation]:
+    r"""Creates several channels each applying the requested pauli error a single qubit.
```
mpharrigan, line 57: to a?

qualtran/surface_code/fifteen_to_one.py
```diff
@@ -0,0 +1,380 @@
...
+    QuantumErrorCorrectionSchemeSummary,
+)
+from qualtran.surface_code.t_factory_utils import PauliRotationChannel, storage_error
+
+
+@frozen
+class FifteenToOne(MagicStateFactory):
+    """15-to-1 Magic T state factory.
+
+        reference: https://arxiv.org/abs/1905.06903
```
mpharrigan, line 33: weird formatting. structure this as a references section

```
Reference:
    [Not as costly](arxiv:whatever)
```

qualtran/surface_code/fifteen_to_one.py
```diff
@@ -0,0 +1,380 @@
...
+        qec: Quantum error correction scheme being used.
+    """
+
+    d_X: int
+    d_Z: int
+    d_m: int
+    qec: QuantumErrorCorrectionSchemeSummary = FowlerSuperconductingQubits
+
+    def footprint(self) -> int:
+        return 2 * (self.d_X + 4 * self.d_Z) * 3 * self.d_X + 4 * self.d_m
```
mpharrigan, line 48: some explanation or equation number from the reference would be good here

qualtran/surface_code/fifteen_to_one.py
```diff
@@ -0,0 +1,380 @@
...
+    @lru_cache(5)
+    def p_fail(self, phys_err: float) -> float:
+        projector = np.kron(np.eye(2), np.ones((16, 16)) / 16)
+        return np.real(1 - np.trace(projector @ self._final_state(phys_err)))
+
+    @lru_cache(5)
+    def p_out(self, phys_err: float) -> float:
+        projector = np.kron(
+            np.eye(2), np.ones((16, 16)) / 16
+        )  # I \otimes ones \otimes ones \otimes ones \otimes ones / 16
```
mpharrigan, line 66: if you put this comment on its own line, the formatter will make this look nicer

qualtran/surface_code/fifteen_to_one.py
```diff
@@ -0,0 +1,380 @@
...
+        project_state = (
+            1
+            / (1 - self.p_fail(phys_err))
+            * (projector @ self._final_state(phys_err) @ projector.T.conj())
+        )
+        T_state = np.array([1, np.exp(-1j * np.pi / 4)]).reshape((1, 2)) / np.sqrt(
+            2
+        )  # |T><T| \otimes ones \otimes ones \otimes ones \otimes ones / 16
```
mpharrigan, line 74: if you put this comment on its own line, the formatter will make it look nicer

qualtran/surface_code/fifteen_to_one.py
```diff
@@ -0,0 +1,380 @@
...
+        target_density = np.kron(T_state.T.conj() @ T_state, np.ones((16, 16)) / 16)
+        return np.real(1 - np.trace(project_state @ target_density))
```
mpharrigan, line 76: does `np.real_if_close` work here? a little safer

qualtran/surface_code/fifteen_to_one.py
```diff
@@ -0,0 +1,380 @@
...
+
+    def n_cycles(self, n_magic: MagicCount, phys_err: float) -> int:
+        """The number of cycles (time) required to produce the requested number of magic states."""
+        num_t = n_magic.n_t + 4 * n_magic.n_ccz
+        return np.ceil(num_t * 6 * self.d_m / (1 - self.p_fail(phys_err)))
```
mpharrigan, line 81: a reference for the equation for the factor of `6` would be appreciated here.

Why are you dividing by the success probability? Is this supposed to take into account the number of cycles "wasted" by a bad distillation? Do other factories take this into account or do they assume everything was fine and you have to look at the failure probability to figure out this correction? The failure probability includes both heralded and unheralded errors: if you get a bad distillation and have to try again you increase the number of cycles but don't increase the error of your algorithm. If you have a data error during the distillation you don't notice, you don't increase the number of cycles, but you do increase the error rate of your actual algorithm.

qualtran/surface_code/fifteen_to_one.py
```diff
@@ -0,0 +1,380 @@
...
+    factory = cirq.Circuit.from_moments(
+        cirq.H.on_each(qs),
+        PauliRotationChannel(
+            'IZIII',
+            phys_err / 3 + 0.5 * (d_m / d_Z) * pz * d_m,
+            phys_err / 3 + 0.5 * d_Z * pm,
+            phys_err / 3,
+        )(
+            *qs
+        ),  # 1
```
mpharrigan, line 124: if you put this comment on its own line, the formatter will make it look nicer

qualtran/surface_code/fifteen_to_one.py
```diff
@@ -0,0 +1,380 @@
...
+            phys_err / 3 + 0.5 * pm * d_m,
+            phys_err / 3 + 0.5 * pm * d_m + 0.5 * (d_X + 3 * d_Z) * d_X / d_m * pm,
+            phys_err / 3,
+        )(
+            *qs
+        ),  # 7
+        *storage_error(
+            'Z', [0.5 * ((d_X + 2 * d_Z) + (d_X + 3 * d_Z)) / d_X * px * d_m, 0, 0, 0, 0], qs
+        ),
+        *storage_error(
```
mpharrigan, line 190: why do the storage errors only come after certain groups of the rotations? 

Can you factor out the storage_error operations? Looks like a lot of code duplication here. e.g.

```python

x_storage_err = get_storage_error('X')
z_storage_err = get_storage_error('Z')
cirq.Circuit([
  Pauli...
  Pauli...
  x_storage_err,
  z_storage_err,
  PauliRot
  ...
```

mpharrigan (commented):
cool stuff! It's great that we can reproduce the numbers from the paper. I left some comments and questions

qualtran/surface_code/t_factory_utils.py
NoureldinYosri: done

qualtran/surface_code/t_factory_utils.py
NoureldinYosri: done

qualtran/surface_code/fifteen_to_one_test.py
NoureldinYosri: the `n_cycles` function returns an math.ceil() of the result. the result in the paper is 30. the actualy value is slightly above 30 (and by slightly I mean ~1e-9)

qualtran/surface_code/fifteen_to_one_test.py
NoureldinYosri: I'm comparing two integers. the result in the paper was rounded to the nearest 10th digit

qualtran/surface_code/t_factory_utils.py
NoureldinYosri: used NoisyPauliRotation

qualtran/surface_code/fifteen_to_one.py
NoureldinYosri: yes, updated

qualtran/surface_code/fifteen_to_one.py
NoureldinYosri: added reference

qualtran/surface_code/fifteen_to_one.py
NoureldinYosri: added reference. yes I think this formula is the expected nubmer of attempts until accepting a state 

$$
E[\mathrm{attempts}] = \sum_{i = 1}^\infty i \times p_\mathrm{fail}^{i-1} (1 - p_\mathrm{fail}) = \frac{1}{1 - p_\mathrm{fail}}
$$

each attempt has $6 d_m$ code cycles.

---

> if you get a bad distillation and have to try again

yes this is what is caught here. the number of tries until we meeting the acceptance criteria

> If you have a data error during the distillation you don't notice, you don't increase the number of cycles, but you do increase the error rate of your actual algorithm.

That is `p_out` which I use in computing the `distillation_error`

---

from what I understand from your comment it seems $6 d_m$ would be more in line with the only other factory we have (the CCZ factory). which way do we wanna go from here?

qualtran/surface_code/fifteen_to_one.py
NoureldinYosri: they don't explain why they grouped them in this way in the paper.  the storage errors have different sets of probabilities. so they can't be extracted

qualtran/surface_code/fifteen_to_one.py
mpharrigan: I don't know what the best way forward is. Maybe just add a comment that this includes both sources of increased time and the other factory doesn't

mpharrigan (approved):

tanujkhattar:
Is this good to go? Would be great to merge this soon.

NoureldinYosri:
@tanujkhattar yes

mpharrigan:
@NoureldinYosri in the future -- create a branch on your personal fork and open a PR from there instead of pushing your branch to the main quantumlib repository; thanks!

PR #694, opened 2024-02-21
QIR Interop

skushnir123:
- Added functionality to compile a Qualtran Bloq to [QIR](https://www.qir-alliance.org/)
- Added tests including the compilation of a QFTTextbook with 10 qubits
- Added notebook showcasing how to use the Microsoft Resource Estimator with QIR generated from a QFTTextbook bloq with 20 qubits.

There is still much to do here on top of this PR. For example, right now, we don't support bloqs that allocate or deallocate new qubits. However, this PR provides working code that can convert a nontrivial bloq (QFT) and get resource estimates.


qualtran/qir_interop/bloq_to_qir.py
```diff
@@ -0,0 +1,349 @@
...
+    Returns:
+        The IR for the given index.
+    """
+    for key, val in ir_map.items():
+        if i == val:
+            return key
+    raise Exception(f"Error in find_ir_for_index: No IR for index {i}")
+
+
+def cz_power(angle: float, params: List[pyqir._native.Constant], builder, qubit_alloc):
```
skushnir123, line 129: @tanujkhattar I may have misheard you in the meeting when you said the decomposition of cz power into toffoli + power gate. Please let me know if this is incorrect.

tanujkhattar:
Samuel, do you know why the tests are failing? I think you need to recompile the dependencies after adding `azure-quantum[qiskit]` to the requirements file.

skushnir123:
> Samuel, do you know why the tests are failing? I think you need to recompile the dependencies after adding `azure-quantum[qiskit]` to the requirements file.

Yes, that is it. I just added it, so hopefully that check passes now.

skushnir123:
> > Samuel, do you know why the tests are failing? I think you need to recompile the dependencies after adding `azure-quantum[qiskit]` to the requirements file.
> 
> Yes, that is it. I just added it, so hopefully that check passes now.

But actually, even after adding azure-quantum, the notebook won't run because there is not a valid "resource_id" specified. Obviously, we don't want to put in an actual resource ID here. So I think we have two options: 1) Have the code in the notebook in comments, and tell users to uncomment to run so that the notebook checker won't fail. 2) Disable the notebook checker on this notebook.

qualtran/qir_interop/bloq_to_qir.py
```diff
@@ -0,0 +1,392 @@
...
+    Controlled-Z power gate implementation in QIR
+
+
+    Args:
+        angle: The angle of the rotation.
+        params: The parameters for the gate.
+        builder: The QIR builder to use.
+        qubit_alloc: The QIR function to allocate qubits.
+    """
+    ancilla = builder.call(qubit_alloc, [])
```
tanujkhattar, line 150: Is this the only place where we use the qubit allocator right now?

qualtran/qir_interop/bloq_to_qir.py
```diff
@@ -0,0 +1,392 @@
...
+        if bloq_instance.bloq.short_name() == 'Split':
+            reg_name = inputs['reg'].reg.name
+            for i in range(len(outputs[0])):
+                ir = soq_map[inputs['reg']][i] if inputs['reg'] in soq_map else (reg_name, i)
+                soq_map[outputs[0][i]] = [ir]
+            continue
+        if bloq_instance.bloq.short_name() == 'Join':
+            irs = [el for soq in inputs['reg'] for el in soq_map[soq]]
+            soq_map[outputs[0]] = irs
+            continue
```
tanujkhattar, line 287: We have dealt with a lot of this logic of maintaining a mapping between soquets and a flat list of qubits in the Qualtran -> Cirq conversion code. I think it may be worth seeing if we can leverage some of that here directly. 

My instinct would be to define a custom `QIRQubitManager` class that derives from `QubitManager` and keeps track of all the `qalloc` / `qfree` / `qborrow` calls and delegates to corresponding PyQIR function calls like `builder.call(qubit_alloc, [])`. 

Then, given a composite bloq `cbloq`, we can call `cbloq.to_cirq_circuit(qubit_manager=qir_qubit_manager, **input_qubits)` to generate a cirq circuit with all Bloqs. This takes care of iterative through the `cbloq`, mapping all multi-dimensional soquets etc. correctly so that we have an expanded cirq style `Operation` for each `Bloq` with a simple list of qubits that can be obtained via `op.qubits` that the bloq operators upon. Mapping this to a list of QIR function calls should then be fairly straightforward and we can hopefully get rid of `soq_map` and all the logic to maintain / update it. 

Happy to leave it as a follow-up improvement for this PR, but we should open an issue to track it.

tanujkhattar (changes requested):
This is a great start, thanks a lot! I've left a round of comments but it's looking pretty good overall. I've proposed a bunch of improvements, which should hopefully make it easy to extend the conversion logic to more complicated bloqs as well. 

Main feedback is that I think we can reuse a lot of the logic already implemented as part of Qualtran to Cirq conversion and then use the generated cirq objects to have a 1:1 mapping to QIR more easily since both QIR and Cirq expect a "flat list of qubits" style interface. 

We can make this improvement as a follow-up PR though so this comment is not blocking. We can merge once some of the other nits are addressed.

tanujkhattar:
> > > Samuel, do you know why the tests are failing? I think you need to recompile the dependencies after adding `azure-quantum[qiskit]` to the requirements file.
> > 
> > 
> > Yes, that is it. I just added it, so hopefully that check passes now.
> 
> @tanujkhattar But actually, even after adding azure-quantum, the notebook won't run because there is not a valid "resource_id" specified. Obviously, we don't want to put in an actual resource ID here. So I think we have two options: 1) Have the code in the notebook in comments, and tell users to uncomment to run so that the notebook checker won't fail. 2) Disable the notebook checker on this notebook.

We've encountered similar situations in Cirq in the past and we essentially skip testing such notebooks as part of the CI and commit them with the output stored. 

@mpharrigan Do you have any thoughts here?

skushnir123:
> This is a great start, thanks a lot! I've left a round of comments but it's looking pretty good overall. I've proposed a bunch of improvements, which should hopefully make it easy to extend the conversion logic to more complicated bloqs as well.
> 
> Main feedback is that I think we can reuse a lot of the logic already implemented as part of Qualtran to Cirq conversion and then use the generated cirq objects to have a 1:1 mapping to QIR more easily since both QIR and Cirq expect a "flat list of qubits" style interface.
> 
> We can make this improvement as a follow-up PR though so this comment is not blocking. We can merge once some of the other nits are addressed.

Sounds good, I'll make the nit changes and then can focus on the restructuring of the code to use Qualtran -> Cirq functionality.

skushnir123:
Closing this PR as it seems a Qualtran -> QIR conversion doesn't make the most sense right now. We will instead be creating a notebook showcasing how to use the open source Microsoft resource estimator with Qualtran programs.

PR #696, opened 2024-02-21
Automatically cross-check bloq counts

mpharrigan:
#### `check_equivalent_bloq_example_counts`

Check that the BloqExample has consistent bloq counts.

Bloq counts can be annotated directly via the `Bloq.build_call_graph` override.
They can be inferred from a bloq's decomposition. This function checks that both
data sources are present and that they produce the same values.

If both sources are present, and they disagree, that results in a `FAIL`. If only one source
is present, an `UNVERIFIED` result is returned. If neither are present, a `MISSING` result
is returned.

----

This adds a new check and wires it up through the `bloq_autotester` fixture. I've marked any failures as `pytest.xfail` as part of this original PR so we can fix or exempt the more complicated ones in their own PRs. You can always check all bloq examples (whether or not the developer has remembered to add a unit test using the test fixture) via the bloq-report-card notebook which now reports statistics.

There's 101 bloq examples. For the new "counts" check: 19 pass, 56 are unverified, 11 fail. The failures are generally of the flavor of: the `build_call_graph` reports callees at a low level like T counts but the decomposition uses higher-level primitives. 

I made sure the pytest error messages are helpful. We can also `xfail` specific tests by doing:

```python
def test_whatever(bloq_autotester):
  if bloq_autotester.check_name == 'counts':
    pytest.xfail('reason')
  bloq_autotester(_whatever)
```

### DIFF
--- a/qualtran/testing.py
+++ b/qualtran/testing.py
@@ -201,6 +201,23 @@ def assert_valid_bloq_decomposition(bloq: Bloq) -> CompositeBloq:
     return cbloq
 
 
+def assert_wire_symbols_match_expected(bloq: Bloq, expected_ws: List[str]):
+    """Assert a bloq's wire symbols match the expected ones.
+
+    Args:
+        bloq: the bloq whose wire symbols we want to check.
+        expected_ws: A list of the expected wire symbols.
+    """
+    ws = []
+    regs = bloq.signature
+    for i, r in enumerate(regs):
+        # note this will only work if shape = ().
+        # See: https://github.com/quantumlib/Qualtran/issues/608
+        ws.append(bloq.wire_symbol(Soquet(i, r)).text)
+
+    assert ws == expected_ws
+
+
 def execute_notebook(name: str):
     """Execute a jupyter notebook in the caller's directory.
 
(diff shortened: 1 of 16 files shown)


qualtran/testing.py
```diff
@@ -373,18 +390,91 @@ def check_bloq_example_decompose(bloq_ex: BloqExample) -> Tuple[BloqCheckResult,
...
+    has_manual_counts: bool
+    has_decomp_counts: bool
+    manual_counts = None
+    decomp_counts = None
+
+    if bloq.build_call_graph.__qualname__.startswith('Bloq.'):
+        has_manual_counts = False
+    else:
+        has_manual_counts = True
+        # TODO: don't let the author sneak in decomposition??
```
fdmalone, line 423: What does this mean?

mpharrigan: I guess I could remove the TODO, but there's an inherent limitation in checking to see if a method is overridden to see if we have an independent data source. A bloq author could override the `build_call_graph` method but just use the default behavior of counting up bloqs in the decomposition. This cross-check would report success even though the two sources of data are the same.

fdmalone (approved):

qualtran/testing.py
mpharrigan: changed the todo to a block comment explaining the consideration.

Issue #699, opened 2024-02-22
Make SelectSwapQROM more idiomatic


mpharrigan:
The figure from the paper looks like this

<img width="1547" alt="image" src="https://github.com/quantumlib/Qualtran/assets/4967059/47eca941-bb1e-4417-8d8d-3800a552cfbd">

the current decomposition looks like

<img width="873" alt="image" src="https://github.com/quantumlib/Qualtran/assets/4967059/b243923b-f991-409c-8f25-6af64260cd56">

Some more intermediate-abstraction bloqs could clean this up. For example: the list of CNOTs could be wrapped in a bloq. The decomposition currently uses the cirq interop which does a lot of splitting and joining; as well as allocating. Perhaps more thoughtfully routing the registers with e.g. `Partition` could clean things up.

PR #701, opened 2024-02-22
SelectSwapQROM maintenance

mpharrigan:
- autodoc
 - fix docstrings to be valid markdown
 - remove the use of `cirq.inverse` which produces `cirq._InverseCompositeGate` which hurts my head
 - while I was at it: I changed the control flow of the construction to be more verbose.
 - sneak in a fix to the adjoint/t_complexity shim. Note: something like an `Adjoint` bloq is incompatible with the (vanilla) t complexity protocol because it involves leaping more than one level in the call graph. 

This doesn't change the actual construction, cc #699


mpharrigan:
Supporting cirq simulation via the bloq decomposition is borked. The error message fails trying to use cirq to simulate an `Allocate` bloq; but playing around with this I also got errors about a `CNOT` having the same source and target qubit.

PR #703, opened 2024-02-23
Create a web-based UI for quantum resource estimation

NoureldinYosri:
This PR:
- Creates a web-based UI for running resource estimation that can created by running `python qualtran.surface_code.ui.py`
- A bit of refactoring was needed to unify the approach to the currently supported models (CCZ and Azure)
- I reran `dev_tools/requirements/re-pip-compile.sh` to include `dash` a `plotly` which are required for the UI.
- while models have a lot of common, some information make sense for one model but not the others. this is why the UI displays information/visualizations depending on the estimation model. for example both models benifit from a qubit utilization chart but a runtime vs qubit count is not currently supported for CCZ instead a single number"duration" gets displayed (see figures).
- The UI runs locally and doesn't need internet access.
---

A preview of the UI for quantum chemistry

GidneyFowler Model

Azure Model

---

THC FeMoco Compilation
image


tanujkhattar:
Can you please rebase this?

NoureldinYosri:
@tanujkhattar @mpharrigan this is ready for review

qualtran/surface_code/azure_cost_model.py
```diff
@@ -64,10 +64,13 @@ def minimum_time_steps(
...
-        alg.rotation_circuit_depth
-        * rotation_model.rotation_cost(eps_syn / alg.rotation_gates).t_gates
-    )
+    if alg.rotation_gates > 0:
+        rotation_cost = rotation_model.prepartion_overhead(eps_syn) + rotation_model.rotation_cost(
+            eps_syn / alg.rotation_gates
+        )
+        c_min += math.ceil(
+            alg.rotation_circuit_depth * (rotation_cost.n_t + 4 * rotation_cost.n_ccz)
+        )
```
tanujkhattar, line 73: This would have an effect of multiplying the preparation overhead of rotation with rotation circuit depth, which I don't think we want right? The preparation overhead should be an additive cost to `c_min` instead of a multiplicative cost.

qualtran/surface_code/ui.py
```diff
@@ -0,0 +1,608 @@
...
+
+
+QEC_SCHEMES = dict(get_objects([qecs], qecs.QuantumErrorCorrectionSchemeSummary))
+MAGIC_FACTORIES = dict(
+    get_objects([fifteen_to_one, ccz2t_cost_model], magic_state_factory.MagicStateFactory)
+)
+ROTATION_MODELS = dict(get_objects([rotation_cost_model], rotation_cost_model.RotationCostModel))
+SUPPORTED_ESTIMATION_MODELS = [
+    'GidneyFolwer (arxiv:1812.01238)',
+    'Beverland et al (arxiv:2211.07629)',
```
tanujkhattar, line 52: Can we make these enums / constants instead of strings? Comparing against the exact string at multiple places in the code seems very error prone.

qualtran/surface_code/ui.py
```diff
@@ -0,0 +1,608 @@
...
+    factory_cycles = factory.n_cycles(needed_magic, physical_error_rate)
+    min_num_factories = int(np.ceil(factory_cycles / c_min))
+    magic_counts = list(
+        1 + np.random.choice(min_num_factories, replace=False, size=min(min_num_factories, 5))
+    )
+    magic_counts.sort(reverse=True)
+    magic_counts = np.array(magic_counts)
+    time_steps = np.ceil(factory_cycles / magic_counts)
+    magic_counts[0] = min_num_factories
+    time_steps[0] = c_min
```
tanujkhattar, line 448: Ideally, we should define a dataclass that holds all the information required for the pie chart and runtime plots of the UI and then have methods defined in the corresponding estimation model files that return this dataclass for a given list of input parameters. We shouldn't have any logic to compute values for estimation models (eg: `total_magic` for gidney model; which simply returns `n_t / 4 + n_ccz`. This is an approximation because it's trivial to decompose 1 CCZ into 4 Ts but combining 4Ts in a logical circuit is non-trivial and cannot always be done) in the UI file. 

The UI file should strictly delegate all computation and preferably call just one function to get a populated data class which can then be plotted.

tanujkhattar (changes requested):
This is great! My main request is to further simplify the code in `ui.py` and strictly delegate to function calls, in files for appropriate estimation models, to compute quantities of interest and then plot them as part of `ui.py`.

qualtran/surface_code/ui.py
NoureldinYosri: I agree, but this is a first step. I plan to add an abstraction to resource estimation models that covers both of the models we have and make it easier to add others in the future. but this can be in followup PRs

qualtran/surface_code/ui.py
NoureldinYosri: not an enum. but constants should do the trick here.

qualtran/surface_code/ui.py
```diff
@@ -0,0 +1,608 @@
...
+from qualtran.surface_code.azure_cost_model import code_distance, logical_qubits, minimum_time_steps
+from qualtran.surface_code.ccz2t_cost_model import (
+    get_ccz2t_costs_from_grid_search,
+    iter_ccz2t_factories,
+)
+from qualtran.surface_code.magic_count import MagicCount
+from qualtran.surface_code.multi_factory import MultiFactory
+
+
+def get_objects(modules, object_type):
```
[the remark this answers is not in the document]
NoureldinYosri: done

qualtran/surface_code/azure_cost_model.py
NoureldinYosri: fixed

tanujkhattar (approved):

PR #705, opened 2024-02-23
Make `ZPow` and `CZPow` gates respect the precision `eps` in simulations

tanujkhattar:
This PR makes a few changes / improvements to the fundamental gates `ZPow` and `CZPow`

1. We add an error proportional to `eps` to the unitary of the two gates. This makes sure that simulations that rely on this precision parameter test the right thing (eg: `phase_gradient_test.py` fails if I add `eps` instead of `eps/np.pi`, the latter is the right thing to do)
2. In order to support (1), I've currently removed the `global_shift` parameter from the two gates. As a follow up, I think we should add a general gate that phases the two eigenvalues by given amounts and respects error propagation (similar to cirq.EigenGate) and then make `ZPow`, `CZPow` etc. derive from the eigen gate. This would also reduce dependence on the Cirq implementation. 
3. Update the `t_complexity` and `build_call_graph` of `CZPow` to delegate to non-controlled `ZPow` + 1 pair of AND/AND† gates.

### DIFF
--- a/qualtran/bloqs/basic_gates/rotation.py
+++ b/qualtran/bloqs/basic_gates/rotation.py
@@ -13,7 +13,7 @@
 #  limitations under the License.
 
 from functools import cached_property
-from typing import Protocol
+from typing import Set, TYPE_CHECKING
 
 import cirq
 import numpy as np
@@ -23,20 +23,16 @@ from qualtran import bloq_example
 from qualtran.cirq_interop import CirqGateAsBloqBase
 from qualtran.cirq_interop.t_complexity_protocol import TComplexity
 
-
-class _HasEps(Protocol):
-    """Protocol for typing `RotationBloq` base class mixin that has accuracy specified as eps."""
-
-    eps: float
+if TYPE_CHECKING:
+    from qualtran.resource_counting import BloqCountT, SympySymbolAllocator
 
 
 @frozen
 class ZPowGate(CirqGateAsBloqBase):
     r"""A gate that rotates around the Z axis of the Bloch sphere.
 
-    The unitary matrix of `ZPowGate(exponent=t, global_shift=s)` is:
+    The unitary matrix of `ZPowGate(exponent=t)` is:
     $$
-        e^{i \pi s t}
         \begin{bmatrix}
             1 & 0 \\
             0 & e^{i \pi t}
@@ -46,8 +42,7 @@ class ZPowGate(CirqGateAsBloqBase):
     Note in particular that this gate has a global phase factor of
     $e^{i\pi t/2}$ vs the traditionally defined rotation matrices
     about the Pauli Z axis. See `Rz` for rotations without the global
-    phase. The global phase factor can be adjusted by using the `global_shift`
-    parameter when initializing.
+    phase.
 
     Args:
         exponent: The t in gate**t. Determines how much the eigenvalues of
@@ -55,10 +50,6 @@ class ZPowGate(CirqGateAsBloqBase):
             when `gate**1` is applied will gain a relative phase of
             e^{i pi exponent} when `gate**exponent` is applied (relative to
             eigenvectors unaffected by `gate**1`).
-        global_shift: Offsets the eigenvalues of the gate at exponent=1.
-            In effect, this controls a global phase factor on the gate's
-            unitary matrix. The factor for global_shift=s is:
-                exp(i * pi * s * t)
         eps: precision for implementation of rotation.
 
     Registers:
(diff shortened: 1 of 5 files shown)


qualtran/bloqs/basic_gates/rotation.py
```diff
@@ -72,30 +63,50 @@ class ZPowGate(CirqGateAsBloqBase):
...
     def _t_complexity_(self) -> 'TComplexity':
-        return TComplexity(rotations=1)
+        if cirq.has_stabilizer_effect(self.cirq_gate):
+            return TComplexity(clifford=1)
+        return TComplexity(rotations=1, t=4, clifford=9 + 4)
+
+    def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
+        # A CZ**k can be implemented via 1 AND/AND† and 1 non-controlled Z**k
+        if cirq.has_stabilizer_effect(self.cirq_gate):
+            return super().build_call_graph(ssa)
```
mpharrigan, line 101: this is very roundabout. What are you trying to achieve here

qualtran/bloqs/basic_gates/rotation.py
```diff
@@ -72,30 +63,50 @@ class ZPowGate(CirqGateAsBloqBase):
...
+
+        from qualtran.bloqs.and_bloq import And
+
+        return {
+            (And(), 1),
+            (And().adjoint(), 1),
+            (ZPowGate(exponent=self.exponent, eps=self.eps), 1),
```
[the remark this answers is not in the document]
tanujkhattar: It raises a `DecomposeNotImplementedError ` error right now, but maybe it's worth making it explicit.

qualtran/bloqs/basic_gates/rotation.py
```diff
@@ -72,30 +63,50 @@ class ZPowGate(CirqGateAsBloqBase):
     """
 
     exponent: float = 1.0
-    global_shift: float = 0.0
     eps: float = 1e-11
 
     @cached_property
     def cirq_gate(self) -> cirq.Gate:
-        return cirq.ZPowGate(exponent=self.exponent, global_shift=self.global_shift)
+        exponent = self.exponent + (-1) ** np.random.randint(2) * self.eps / np.pi
```
mpharrigan, line 70: this can't possibly be the correct way of handling this.

What is the source of the error? If we can only do a rotation to a certain number of bits of precision we'll get a deterministic unitary effect but viewed as an error if we're trying to a rotation angle not representable exactly with that many bits of precision. There shouldn't be any `np.random` in this case and we should change `exponent` to be a fixedpoint real

If there's true "ignorance" over what angle we're actually doing: is this the correct probability distribution? coin flip between perfect and eps/pi overroration? why are there never underrotations?

If there's true ignorance and we know the correct probability distribution: this isn't a unitary gate and we'd need to switch to a density matrix simulation. 

Under no circumstance should a cached property be backed by a globally-seeded `np.random` call. This is begging for trouble.

mpharrigan (changes requested):
are you sure about this

qualtran/bloqs/basic_gates/rotation.py
mpharrigan: what do you mean "if this node is a leaf"? This is a CZPowGate

qualtran/bloqs/basic_gates/rotation.py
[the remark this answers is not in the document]
mpharrigan: that's fine

qualtran/bloqs/basic_gates/rotation.py
tanujkhattar: Do not override the `build_composite_graph` method in this case.

qualtran/bloqs/basic_gates/rotation.py
tanujkhattar: > What is the source of the error? If we can only do a rotation to a certain number of bits of precision we'll get a deterministic unitary effect but viewed as an error if we're trying to a rotation angle not representable exactly with that many bits of precision. 

In this case, I'm mainly trying to capture the error introduced due to approximate synthesis of `ZPowGate` by decomposing into a Clifford+T gateset. https://arxiv.org/pdf/1404.5320.pdf defines doing such an approximation as implementing a unitary $V^{\prime}$ instead of $V$ s.t $d(V, V^\prime) < \epsilon$

image

A similar, more generalized definition is used in https://iopscience.iop.org/article/10.1088/1367-2630/14/11/115023/pdf
image


Given an `eps`, randomly adding an error to rotation angle gives a simple way to construct a unitary $V^{\prime}$ instead of $V$ which satisfies the above mentioned distance metric. I've added a randomization to the rotation angle to make sure the callers can handle under / over rotations in both directions. 

This approximation doesn't really represent the approximation that would happen if we'd actually decompose the rotation into a Clifford + T gateset, but any caller / algorithm that can handle such a decomposition should be   able to handle this approximation as well. 

This is a proxy for a sanity check to make sure decompositions (like preparing a phase gradient state) that promise to be handle an error `eps` don't make an error when propagating that error down (eg: passing `eps/n` to ZPow instead of `eps`). 

> coin flip between perfect and eps/pi overroration? why are there never underrotations?

There are under rotations. `(-1) ** np.random.randint(2)` gives either `+1` or `-1`. So we have an under or over rotation of `eps/np.pi` always. 

> Under no circumstance should a cached property be backed by a globally-seeded np.random call. This is begging for trouble

hmm. I'm open to suggestions here. If I only update the unitary / tensor logic to include the approximate exponent then it seems problematic as well. 


I agree the correct solution would be to include some kind of a noise model that captures this ignorance OR implement the actual decomposition into Clifford + T gateset. But I thought an approximation like this may be worth having in the meantime. I can revert the change if you feel strongly about it.

tanujkhattar:
Closing this for now.

PR #706, opened 2024-02-23
Run all unit tests by default locally to match CI behavior

mpharrigan:
My understanding of events: I think https://github.com/quantumlib/Qualtran/pull/499 disabled any tests marked "slow" from the CI and one had to pass an extra flag to ever run the slow tests.

In https://github.com/quantumlib/Qualtran/pull/667 I marked a lot of things `slow` but intended them to be run as part of the CI. A check that is not run is not worth having. I didn't realize from above that the "slow" tests would get skipped.

Confusingly, I don't think they did get skipped since I passed an argument `-m 'not notebook'` to the pytest command which somehow interfered with the custom logic for skipping slow tests by default and enabled them as part of the CI.

I noticed something was up when all the slow tests were getting skipped locally when I was just trying to run all tests with no flags. 

I recommend removing the conftest.py code and using `-m 'not (slow or notebook)'` locally to do quick checks. This PR shouldn't have any effect on the CI. It will bring local executions in line with the CI. Note that the slow tests are slow #711


fdmalone:
Add the proto check too?

mpharrigan:
Trying to figure out what happened here. I think https://github.com/quantumlib/Qualtran/pull/499 disabled any tests marked "slow" from the CI and one had to pass an extra flag to ever run the slow tests.

In https://github.com/quantumlib/Qualtran/pull/667 I marked a lot of things `slow` but intended them to be run as part of the CI. A check that is not run is not worth having. I didn't realize from above that the "slow" tests would get skipped.

Confusingly, I don't think they did get skipped since I passed an argument `-m 'not notebook'` to the pytest command which somehow interfered with the custom logic for skipping slow tests by default and enabled them as part of the CI.

I noticed something was up when all the slow tests were getting skipped locally when I was just trying to run all tests with no flags. 

I recommend removing the conftest.py code and using `-m 'not (slow or notebook)'` locally to do quick checks. I don't think this PR will have any effect on the CI. It will bring local executions in line with the CI. Note that the slow tests are slow #711

mpharrigan:
> Add the proto check too?

out of scope for this PR; can be done elsewhere

fdmalone (approved):

Issue #691, opened 2024-02-20
Tensors from classical action


mpharrigan:
Provide functionality for automatically supporting the tensor simulation protocol if a bloq supports classical simulation. 

```python
tensor = np.zeros(...)
for inputs in itertools.product(reg.all_possible_inputs for reg in self.signature):
  ouputs = bloq.call_classically(inputs)
  tensor[outputs, inputs] = 1
```

We'll likely need some new methods on the quantum data types
 - getting all possible values
 - a canonical ordering of values in the domain so we can assign to zero-indexed indices in the tensors

PR #708, opened 2024-02-23
[data types] Add dtype to addition bloq

fdmalone:
We are currently using the Addition bloq for QInts and QUInts and QMontgomeryUInts. So in a step towards enforcing type checking we should add a register dtype to the bloq.

The diff is small in reality but currently is on top of #707


qualtran/bloqs/arithmetic/addition.py
```diff
@@ -96,7 +99,7 @@ def add_my_tensors(
     ):
         import quimb.tensor as qtn
 
-        N = 2**self.bitsize
+        N = 2**self.dtype.bitsize
```
mpharrigan, line 102: I think this tensor only works for unsigned ints, no?

xref #691

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -112,15 +115,15 @@ def on_classical_vals(
         self, a: 'ClassicalValT', b: 'ClassicalValT'
     ) -> Dict[str, 'ClassicalValT']:
         unsigned = True  # TODO: derive from signature
-        N = 2**self.bitsize if unsigned else 2 ** (self.bitsize - 1)
+        N = 2**self.dtype.bitsize if unsigned else 2 ** (self.dtype.bitsize - 1)
```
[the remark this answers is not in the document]
fdmalone: Done.

qualtran/bloqs/arithmetic/addition.py
fdmalone: I raise an error for now.

mpharrigan (approved):

PR #719, opened 2024-02-28
Add symbolic cost analysis support to QVR primitives and T-cost analysis to QVR notebook

tanujkhattar:
This PR supersedes https://github.com/quantumlib/Qualtran/pull/688 and adds cost analysis to `quantum_variable_rotation.ipynb` notebook. It relies on the nice assumption that running `python dev_tools/autogenerate-bloqs-notebooks-v2.py` does not overwrite any custom added cells (@mpharrigan is this by design?) so we can append tutorials style content to auto generated notebooks.

The two main plots that highlight all the nuances related to QVR using phase gradients are 

image

image


mpharrigan:
yes, it's by design

fdmalone:
I would recommend using a different linestyle for the zpow lines for contrast purposes. I'm a little confused by the values plotted, in the top plot the gold line never crosses the zpow line, but in the bottom plot it does around n ~ 60 and eps = 1e-12, which seems like it would be captured by the rightmost tick in the top plot.

tanujkhattar:
@fdmalone If I extend the top plot to eps=1e-29 then the zpow line crosses the gold line 

image


Basically, when $n=\log_2(2\pi/eps)$, cost of `zpow` QVR depends quadratically on $\log_2(1/eps)$ 

image

whereas the cost of phase gradient based QVR scales as $\log_2(1/eps) \times \text{number of set bits}(\gamma)$. The number of set bits in $\gamma$ influences the number of scaled additions we need to into the phase gradient register, which forms the right half of the expression below. When $\gamma$ is symbolic, we default back to the worse case in which we need at most $\text{gamma\\_bitsize} / 2 + 1$ additions. (Note: the plots above use actual values of gamma and plot the different scenarios for increasing number of set bits in $\gamma$).  

image

Since $\text{gamma\\_bitsize}$ also scales as $\log(1/\epsilon)$, we don't get an obvious advantage by using phase gradient QVR over ZPow QVR since both of them have a logarithmic dependence on $\log(1/\epsilon)$. For specific values of $\gamma$ however, phase gradient based QVR can be advantageous

fdmalone:
If I understand the plot correctly there is a crossover for the phase gradient scaling from some polynomial in log(1/eps) to some other polynomial in log(1/eps). Is it possible to show something like this on a plot by fitting to the data an extracting the power laws. There is a function [here](https://github.com/quantumlib/Qualtran/blob/090941f8d5697ce49cbf306056fa354690f95a03/qualtran/bloqs/chemistry/chem_tutorials.py#L75), which could be helpful.

tanujkhattar:
> If I understand the plot correctly there is a crossover for the phase gradient scaling from some polynomial in log(1/eps) to some other polynomial in log(1/eps). Is it possible to show something like this on a plot by fitting to the data an extracting the power laws. There is a function [here](https://github.com/quantumlib/Qualtran/blob/090941f8d5697ce49cbf306056fa354690f95a03/qualtran/bloqs/chemistry/chem_tutorials.py#L75), which could be helpful.

Does my previous comment help clarify the plot? I think we both posted at the same time, so let me know if I should still generate the suggested plot.

fdmalone:
resolved following some offline discussion.

fdmalone (approved):

PR #725, opened 2024-02-29
[data types] Add cast bloq

fdmalone:
It's not entirely clear to me right now to what extent we should use this bloq given #710 BUT there are cases I can see where we may want to explicitly cast a register from one type to another. One place where I updated the code was when converting to a BoundedQUInt with a non default upper limit. I implemented the decomposition via a split and a join so that classical simulation / tensor contraction works.

Depends on #724 

Screenshot 2024-02-29 at 11 48 05 AM


qualtran/bloqs/util_bloqs.py
```diff
@@ -410,3 +410,46 @@ def signature(self) -> Signature:
...
+    Args:
+        in_qdtype: Input QDType to cast from.
+        out_qdtype: Output QDType to cast to.
+
+    Registers:
+        in: input register to cast from.
+        out: input register to cast to.
+    """
+
+    in_dtype: QDType
```
tanujkhattar, line 429:
```suggestion
    inp_dtype: QDType
```

qualtran/bloqs/util_bloqs.py
```diff
@@ -410,3 +410,46 @@ def signature(self) -> Signature:
...
+            [
+                Register('x', dtype=self.in_dtype, shape=self.shape, side=Side.LEFT),
+                Register('y', dtype=self.out_dtype, shape=self.shape, side=Side.RIGHT),
+            ]
+        )
+
+    def build_composite_bloq(self, bb: BloqBuilder, x: SoquetT) -> Dict[str, 'SoquetT']:
+        split = bb.split(x)
+        y = bb.join(split, dtype=self.out_dtype)
+        return {'y': y}
```
tanujkhattar, line 452: Is the split/join even required? I think we can just override the `add_my_tensors` and not have a decomposition?  Since we don't really change anything (even the ordering of qubits) so it's really just reinterpreting the meaning of bits.

fdmalone: This was just a convenient way to avoid implementing classical sim and tensor contraction methods.

tanujkhattar: Can we add classical sim and tensor contraction, since it's one of the fundamental bloqs and it's useful to not have to decompose it for the corresponding protocol. It'd also be easier to debug what's happening (for example - it looks like right now, the "tensor decomposition" would simply be identity?)

fdmalone: Done.

tanujkhattar (approved):

PR #731, opened 2024-03-01
More basic gates

fdmalone:
Adds the S gate and the F(k,n) gate from https://arxiv.org/pdf/1902.10673.pdf (Figure 8). This is used for costing out the trotter bloqs and had a simple decomposition so thought it worthwhile adding. It also forms the basis for the fermionic fast fourier transform from that reference.

Screenshot 2024-02-29 at 10 59 48 PM


fdmalone:
Rename this gate and move it out of basic_gates?

mpharrigan:
I don't think "F gate" is known enough to be ambiguous. Googling for it doesn't return any quantum stuff. Googling T Gate does, for reference.

mpharrigan:
does this gate exist in Cirq?

mpharrigan:
`TwoBitFFFT` ?

fdmalone:
A special case is in [openfermion](https://github.com/quantumlib/OpenFermion/blob/v1.6.1/src/openfermion/circuits/primitives/ffft.py#L25) (I actually only need the special case).

mpharrigan:
I'll give @tanujkhattar the ultimate authority over organizing this where he sees fit as the de facto organization czar: but I would not object to it being in `basic_gates/` since 1) it operates on <= 2 qubits 2) (almost) exists as a gate in cirq (tehnically openfermion)

mpharrigan (approved):

tanujkhattar:
How did we arrive at the `TwoBitFFFT` name? Sorry, I haven't read the reference yet. Is it related to `QFT` by any chance?

mpharrigan:
it's a building block for the fermionic fast fourier transform hence the three Fs

mpharrigan:
image

tanujkhattar:
Where will we put the fermionic fast fourier transform ? Does it make sense to put it under `qft` @fdmalone  ?

tanujkhattar:
I think it's fine to merge under `basic_gates` for now. Philosophically, so far basic gates have been the one where we have 1:1 mapping between cirq and qualtran as well. If you think we'll create a new subpackage for fermions, then we can put it there as well

fdmalone:
Having it under QFT also makes sense I think. I am not really planning on adding a FFFT right now but this primitive gate appears elsewhere so it's helpful to have for costing / simulation.

tanujkhattar:
Let's put it under QFT and merge the PR!

PR #737, opened 2024-03-04
More improvements and bugfixes to t-complexity protocol to make things work with Adjoint

tanujkhattar:
Follow up of https://github.com/quantumlib/Qualtran/pull/736. See added tests for more bugfixes. 

One philosophical discussion - what's the difference between `DecomposeNotImplemented` and `DecomposeTypeErrror`? In Cirq, we just return `NotImplmented` if the decomposition is not implemented and I've updated the logic to map both the above errors to this behavior. 

@mpharrigan Can you check if this looks good?

### DIFF
--- a/qualtran/_infra/adjoint.py
+++ b/qualtran/_infra/adjoint.py
@@ -187,6 +187,9 @@ class Adjoint(GateWithRegisters):
         of high-level bloqs, so we need to shim in an extra `adjoint` boolean flag.
         """
         # TODO: https://github.com/quantumlib/Qualtran/issues/735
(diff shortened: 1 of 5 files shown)


mpharrigan:
philosophically: one means "it doesn't make sense to decompose this" -- it's a leaf bloq or symbolic or something. NotImplemented means "the bloq author hasn't written this yet." It shows up in the `qlt_testing.check_xxx` statuses: "missing" vs "na". these show up in the `dev_tools/bloq_report_card`

mpharrigan (approved):

PR #739, opened 2024-03-04
Fix bug in TwoBitSwap.as_cirq_op to return the out_regs dict as well

tanujkhattar:


qualtran/bloqs/basic_gates/swap.py
```diff
@@ -77,7 +77,7 @@ def as_cirq_op(
     ) -> Tuple['cirq.Operation', Dict[str, 'CirqQuregT']]:
         (x,) = x
         (y,) = y
-        return cirq.SWAP.on(x, y)
+        return cirq.SWAP.on(x, y), {'x': [x], 'y': [y]}
```
fdmalone, line 80: the type hint suggests these should be numpy arrays rather than lists right? or does it matter

tanujkhattar: It doesn't matter and we return lists at many other places. I agree it'd be nice to enforce type checking and catch the errors early on.

mpharrigan (approved):

Issue #735, opened 2024-03-04
Future and upgrades to T-Complexity protocol


tanujkhattar:
Right now, the `TComplexity` protocol is broken as described in https://github.com/quantumlib/Qualtran/pull/732

This issue is to discuss the future of TComplexity protocol since there have been informal discussions and upgrades to either deprecate it in favour of the new generalized costs framework or make improvements to it and make it better supported. 

First, I'd like to highlight that the `TComplexity` protocol; or an equivalent `LogicalCounts` protocol that captures the _most commonly used_ logical counts people care about when thinking about fault tolerant resource estimation, is extremely important and we should have an _easy_ way (read single function call) to obtain this value for any given `bloq`. The `TComplexity` / `LogicalCounts` class is also the input format to invoke the microsoft azure resource estimator. 

The current way to obtain this value without going via the t-complexity protocol is 
a) Compute `sigma` via `bloq.call_graph()`
b) Call `t_counts_from_sigma(sigma)` assuming `sigma` was computed without any fancy parameters (eg:  no truncation of the call graph using `max_depth` parameter; etc.) 

We need to have a single function call (eg: `bloq.t_complexity()` or `bloq.logical_counts()`) that gives the desired counts. Inspecting the call graph should be done only when then bloq users need to get into then weeds of how exactly the logical counts were computed. Having agreed on this premise, let's lay down some requirements (I'll refer to the protocol as TComplexity for now, but we should probably change it to `logical_counts` in the future) - 

a) The most important requirement is that the it needs to be `_fast_` for the general case. Expecting users to override `build_call_graph` should be the recommended way but if users do not provide this decomposition, we should still be able to support fast computations of logical counts. The cirq style `t_complexity` uses a global cache to store and retrieve the logical counts of bloqs. This has an advantage of reusing any structure that the bloq decomposition has and ends up being sufficient for a majority of cases. This is because a lot of the times, the very high gate counts are present because we do phase estimation on a unitary `U` which ends up repeating `U` a lot of times. 

 b) It should report more than just T-counts and should be sufficient as an "input" to our qec overhead methods defined in `bloqs/surface_code`.  Some other useful logical counts are - (a) Toffoli counts (b) qubit usage (c) Counts of (rotation, eps) pairs instead of just number of rotations (d) Rotation depth (azure model uses this) 

Design options going forward:
[...]

PR #740, opened 2024-03-04
Use `build_call_graph` instead of `decompose_bloq` to compute `TComplexity`for bloqs

tanujkhattar:
Needed in https://github.com/quantumlib/Qualtran/pull/732
Larger discussion on improving T-complexity in https://github.com/quantumlib/Qualtran/issues/735

This is strictly better because bloqs `build_call_graph` can fallback to `decompose_bloq` if user doesn't have a custom `build_call_graph` specified. The latter is often faster and more commonly specified (eg: all newly added chemistry bloqs)

### DIFF
--- a/qualtran/cirq_interop/t_complexity_protocol.py
+++ b/qualtran/cirq_interop/t_complexity_protocol.py
@@ -18,7 +18,7 @@ import attrs
 import cachetools
 import cirq
 
-from qualtran import Bloq, CompositeBloq, DecomposeNotImplementedError, DecomposeTypeError
+from qualtran import Bloq, DecomposeNotImplementedError, DecomposeTypeError
 from qualtran.cirq_interop.decompose_protocol import _decompose_once_considering_known_decomposition
 from qualtran.resource_counting.symbolic_counting_utils import ceil, log2, SymbolicFloat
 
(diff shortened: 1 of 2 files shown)


qualtran/cirq_interop/t_complexity_protocol.py
```diff
@@ -128,11 +128,15 @@ def _from_bloq_decomposition(stc: Any, fail_quietly: bool) -> Optional[TComplexi
     if not isinstance(stc, Bloq):
         return None
 
-    if isinstance(stc, CompositeBloq):
-        return _is_iterable((binst.bloq for binst in stc.bloq_instances), fail_quietly=fail_quietly)
-
     try:
-        return _from_bloq_decomposition(stc.decompose_bloq(), fail_quietly=fail_quietly)
+        _, sigma = stc.call_graph(max_depth=1)
```
mpharrigan, line 132: update the function name?

mpharrigan (approved):
consider updating the helper function name

qualtran/cirq_interop/t_complexity_protocol.py
tanujkhattar: done

PR #741, opened 2024-03-04
Add TrotterizedUnitary bloq

fdmalone:
Adds a bloq to build an arbitrary trotterized unitary from a collection of sub-bloqs which implement the individual terms in the Hamiltonian.

This will be helpful for:

1. Resource counting for higher order trotter approximations and comparing these to qubitization
2. Error propagation / resource optimization for QPE.


qualtran/bloqs/chemistry/trotter/trotterized_unitary.py
```diff
@@ -0,0 +1,121 @@
...
+        ref_sig = self.bloqs[0].signature
+        for bloq in self.bloqs[1:]:
+            assert bloq.signature == ref_sig
+
+    @cached_property
+    def signature(self) -> Signature:
+        return self.bloqs[0].signature
+
+    def build_composite_bloq(self, bb: 'BloqBuilder', **soqs: SoquetT) -> Dict[str, 'Soquet']:
+        for i, c in zip(self.indices, self.coeffs):
```
tanujkhattar, line 90: nit: Let's use `a` instead of `c` since that's what we used in the docstring
```suggestion
        for i, a in zip(self.indices, self.coeffs):
```

tanujkhattar (approved):
Left a round of suggestions, questions and comments. Looking good overall.

qualtran/bloqs/chemistry/trotter/trotterized_unitary.py
```diff
@@ -0,0 +1,121 @@
...
+
+    $$
+        H = \sum_{\gamma=1}^\Gamma H_\gamma
+    $$
+
+    then the unitary $e^{-i t H}$ can be approximately implemented via a $p$-th order product
+    formula
+
+    $$
+        S_p(t) = \prod_{v=1}^{\Upsilon}\prod_{k=1}^\Gamma e^{-it a_{v,\gamma} H_{\pi_v(\gamma)}}
```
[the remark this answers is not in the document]
mpharrigan: if it was a typo, I say we point it out somewhere in our docs

qualtran/bloqs/for_testing/ising.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+    spins = cirq.LineQubit.range(n_spins)
+    ham = get_1d_Ising_hamiltonian(spins, j_zz_strength, gamma_x_strength)
+    coeffs = np.array([term.coefficient.real for term in ham])
+    lcu_coeffs = coeffs / np.sum(abs(coeffs))
+    return lcu_coeffs
+
+
+@attrs.frozen
+class IsingXUnitary(Bloq):
```
mpharrigan, line 93: we don't want these in the `bloqs.chemistry` package?

fdmalone: The bloqs or the test functions or both?

qualtran/bloqs/chemistry/trotter/trotterized_unitary.py
```diff
@@ -0,0 +1,121 @@
...
+
+    For example, the second order Suzuki splitting would have indicies = (0, 1, 0)
+    and coeffs = (0.5, 1, 0.5), which would build
+
+    $$
+        e^{-i \frac{t}{2} H_0} e^{-i t H_1} e^{-i \frac{t}{2} H_0}
+    $$
+
+    Args:
+        bloqs: A tuple of bloqs of length $\Gamma$ which implement the unitaries for
```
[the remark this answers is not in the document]
fdmalone: Not strictly necessary no, but a little more convenient until we run into a case that needs a more convoluted set of registers. I guess we could define a high level bloq like we do for bloq encoding with an "ancilla" and "system" register and then some sugar to wrap / join the matching registers, but I don't really have a good use case right now. Maybe I should open an issue?

qualtran/bloqs/for_testing/ising.py
mpharrigan: the bloqs or both?

qualtran/bloqs/chemistry/trotter/trotterized_unitary.py
```diff
@@ -0,0 +1,121 @@
...
+    def __attrs_post_init__(self):
+        ref_sig = self.bloqs[0].signature
+        for bloq in self.bloqs[1:]:
+            assert bloq.signature == ref_sig
+
+    @cached_property
+    def signature(self) -> Signature:
+        return self.bloqs[0].signature
+
+    def build_composite_bloq(self, bb: 'BloqBuilder', **soqs: SoquetT) -> Dict[str, 'Soquet']:
```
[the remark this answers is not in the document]
fdmalone: yeah I was thinking about this. I'm not sure if this is something you'd really encounter though, there aren't really that many ways to split the Hamiltonian typically, and when you do split it it's quite bespoke (like the plaquette splitting).

qualtran/bloqs/chemistry/trotter/trotterized_unitary.py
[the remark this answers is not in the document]
fdmalone: I think the reference is a bit more general (so t can be real or imaginary) to encompass both unitary evolution and imaginary time evolution since trotterization is used for both cases. It just looked a little weird to me to omit the the $-i$ for us.

qualtran/bloqs/chemistry/trotter/trotterized_unitary.py
```diff
@@ -0,0 +1,121 @@
...
+
+    $$
+        e^{-i \frac{t}{2} H_0} e^{-i t H_1} e^{-i \frac{t}{2} H_0}
+    $$
+
+    Args:
+        bloqs: A tuple of bloqs of length $\Gamma$ which implement the unitaries for
+            each term in the Hamiltonian.
+        indices: A tuple of integers which specifies which bloq to apply when
+            forming the unitary as a product of unitaries.
```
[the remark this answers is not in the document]
fdmalone: right, I don't really like the notation as it's just more straightforward to implement what is used in practice which merges terms like you say.

qualtran/bloqs/chemistry/trotter/trotterized_unitary.py
fdmalone: Done.

qualtran/bloqs/chemistry/trotter/trotterized_unitary.py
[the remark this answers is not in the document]
fdmalone: done.

qualtran/bloqs/chemistry/trotter/trotterized_unitary.py
```diff
@@ -0,0 +1,121 @@
...
+from typing import Dict, Iterable, Union
+
+import attrs
+import sympy
+
+from qualtran import Bloq, bloq_example, BloqBuilder, BloqDocSpec, Signature, Soquet, SoquetT
+
+
+@attrs.frozen
+class TrotterizedUnitary(Bloq):
```
tanujkhattar, line 26: I think you'll have to add `TrotterizedUnitary` to the `RESOLVER_DICT`; hopefully the tests will fail since you run `bloq_autotester` and it should automatically test for serialization and raise an error.

fdmalone: Is there an automatic way to build up this dictionary?

tanujkhattar: I used the code in `dev_tools/bloq_finder.py` to `get_bloq_classes` and copy/paste to construct the dictionary; so I'm sure it can be done. The only issue is that this code is not part of the qualtran package and uses calls to `git` to get all files and recursively extract class names. 

We could add a decorator that "registers" the classes with the serialization infrastructure and therefore adds the class to the resolver dictionary. Then, we'll have to decorate all bloqs with this decorator. I'll have to think more about other potential ways to achieve this.

PR #742, opened 2024-03-04
Ignore cliffords in `t_counts_from_sigma` using `cirq.has_stabilizer_effect`

tanujkhattar:


mpharrigan (approved):
ok with this as a temporary thing. we should remove the cirq dependency for the final manifestation of the protocol

tanujkhattar:
> we should remove the cirq dependency for the final manifestation of the protocol

Sure, we can either move the implementation of the algorithm that checks whether an object is clifford or not to Qualtran (as part of one of the linalg utils) or we can go and update each bloq to return the correct counts.

PR #746, opened 2024-03-05
Update cirq to bloq interop to automatically replace basic gates with their Bloq versions

tanujkhattar:

TODO(as part of a follow-up PR): @mpharrigan Can we add a method to convert Cirq's `AbstractControlValues` to `CtrlSpec`? This would allow converting `ControlledGate` instaces to `Controlled` Bloq. There's also a nuance that both `cirq.Gate.controlled()` and `Bloq.controlled()`exist and they have different interfaces so cirq style API and bloq style API clashes for `GateWithRegisters`.

### DIFF
--- a/qualtran/cirq_interop/_cirq_to_bloq.py
+++ b/qualtran/cirq_interop/_cirq_to_bloq.py
@@ -300,18 +300,27 @@ def _gather_input_soqs(
     return qvars_in
 
 
-def _extract_bloq_from_op(op: 'cirq.Operation') -> Bloq:
-    """Get a `Bloq` out of a cirq Operation.
-
-    Unwrap BloqAsCirqGate, pass through any GateWithRegisters, and wrap
-    true cirq gates with `CirqGateAsBloq`.
-    """
+def _cirq_gate_to_bloq(gate: cirq.Gate) -> Bloq:
+    from qualtran import Adjoint
+    from qualtran.bloqs.basic_gates import (
+        CNOT,
+        CZPowGate,
+        Hadamard,
+        Rx,
+        Ry,
+        Rz,
+        TGate,
+        Toffoli,
+        TwoBitSwap,
+        XGate,
+        XPowGate,
+        YPowGate,
+        ZGate,
+        ZPowGate,
+    )
+    from qualtran.cirq_interop import CirqGateAsBloq
     from qualtran.cirq_interop._bloq_to_cirq import BloqAsCirqGate
 
-    if op.gate is None:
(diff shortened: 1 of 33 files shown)


qualtran/cirq_interop/_cirq_to_bloq.py
```diff
@@ -300,29 +300,80 @@ def _gather_input_soqs(
...
+    if gate == cirq.CNOT:
+        return CNOT()
+    if gate == cirq.TOFFOLI:
+        return Toffoli()
+    if gate == cirq.X:
+        return XGate()
+    if gate == cirq.Z:
+        return ZGate()
+    if gate == cirq.SWAP:
+        return TwoBitSwap()
```
mpharrigan, line 346: could be a dictionary lookup

qualtran/cirq_interop/_cirq_to_bloq.py
```diff
@@ -300,29 +300,80 @@ def _gather_input_soqs(
...
+    if isinstance(gate, cirq.Rx):
+        return Rx(angle=gate._rads)
+    if isinstance(gate, cirq.Ry):
+        return Ry(angle=gate._rads)
+    if isinstance(gate, cirq.XPowGate):
+        return XPowGate(exponent=gate.exponent, global_shift=gate.global_shift)
+    if isinstance(gate, cirq.YPowGate):
+        return YPowGate(exponent=gate.exponent, global_shift=gate.global_shift)
+    if isinstance(gate, cirq.ZPowGate):
+        return ZPowGate(exponent=gate.exponent, global_shift=gate.global_shift)
```
mpharrigan, line 360: these (along with CZPowGate) can probably (?) be condensed

```python
if isinstance(gate, (Rx, Ry, Rz):
  return BLOQMAP[gate.__class__](angle=gate._rads)
if isintance(gate, (XPowGate, YPowGate, ...)):
  return BLOQMAP[gate.__class__](exponent=gate.exponent, global_shirt=gate.global_shift)
```

mpharrigan (approved):
Left some ideas for performance tweaks, but I don't know if they would be faster in practice

tanujkhattar:
Instantiating a local dictionary on every function call would not be faster, but I think it's cleaner so I have done the replacement. Merging now!

mpharrigan:
can't you pull out the dictionary as a (private) global?

tanujkhattar:
I'll have to import all the basic gates, that can probably lead to circular imports, so I wanted to keep the global imports minimal. I'm sure I can find a way around if it becomes a bottleneck.

mpharrigan:
ah fair point

PR #748, opened 2024-03-06
Move `and_bloq.py` and `multi_control_multi_target.py` to new `mcmt/` folder

tanujkhattar:
Part of a larger effort to organize bloqs/


mpharrigan:
one tiny note with the "multi control multi target" thing: the current module made me think there was a bloq in it that would do multi-control multi-target cnot. But it's really "multi control" _or_ "multi target". Making it a (sub)package instead of a module might actually improve things, but I wanted to highlight this potential confusion.

mpharrigan (approved):

tanujkhattar:
> Making it a (sub)package instead of a module might actually improve things

Yeah, splitting `multi_control_multi_target.py` into smaller files is part of the plan. I didn't want this PR to blow up even more so I'll do that as a follow-up.

PR #753, opened 2024-03-07
Add resource estimation notebook for trotterized QPE

fdmalone:
This builds upon #743 (and the preceding PRs) to add a notebook which numerically and symbolically analyzes the T counts for the PLAQ hubbard bloqs (without HWP). I've kept it how I sort of wrote it chronologically but I'm open to removing some of the hand written cost code in favour of "here are the symbolic counts lets use those instead). I didn't have much luck with scipy's constrained optimization so just did a pretty basic inner/outer minimization step over two of the parameters which seems to work better than doing anything fancy.


fdmalone:
Screenshot 2024-03-07 at 2 32 07 PM
Screenshot 2024-03-07 at 2 32 15 PM
Screenshot 2024-03-07 at 2 32 24 PM
Screenshot 2024-03-07 at 2 32 34 PM
Screenshot 2024-03-07 at 2 32 44 PM

mpharrigan:
@fdmalone can you resolve the conflicts? I think most of them are vestigial from being factored out into other PRs, right?

fdmalone:
For sure

qualtran/bloqs/chemistry/trotter/hubbard/qpe_cost_optimization.ipynb
```diff
@@ -0,0 +1,574 @@
...
+    "    '\\Delta_{HT}, \\Delta_{TS}, \\Delta_{PE}, p, xi'\n",
+    ")\n",
+    "s_n_rot, s_n_t, s_n_pe = sympy.symbols('N_R, N_T, N_PE')\n",
+    "s_timestep = (s_delta_ts / s_xi) ** (1 / s_p)\n",
+    "s_eps_r = (s_delta_ht * s_timestep) / s_n_rot\n",
+    "s_n_pe = 0.76 * sympy.pi / (s_delta_pe * s_timestep)\n",
+    "s_trotter_step = build_plaq_unitary_second_order_suzuki(\n",
+    "    s_length, s_hubb_u, s_timestep, eps=s_eps_r, hubb_t=s_tau\n",
+    ")\n",
+    "# just use this cost in lieu of a QPE bloq\n",
```
tanujkhattar, line 526: In the paper they use adoptive phase estimation that uses a single control qubit but cost goes as `0.76 * pi`. We have the optimal scaling version where cost goes as `0.5 * pi` but it uses `m` control bits. We also have an open issue to implement https://arxiv.org/abs/2303.12503, which is the optimal scaling version but with only 2 control qubits. 

It'd be nice if we can either 
(a) Use a bloq (maybe just a shim for now) to get the cost for adoptive phase measurement based QPE used here so that we can add an implementation of the bloq later and keep track of the different QPE versions used and supported in cost estimates throughout the library (instead of a hardcoded formula)
(b) If (a) is too much work, then open an issue to do this at a later point and track that we are using a hardcoded formula instead of a bloq to get cost estimates for QPE here. 


I know doing (a) may require more than just a shim because of the new capabilities required in `SympySymbolAllocator` so I'm fine with (b) as well to make sure we can merge this PR quickly

qualtran/bloqs/qft/two_bit_ffft.py
```diff
@@ -73,7 +74,7 @@ class TwoBitFFFT(Bloq):
 
     k: int
     n: int
-    eps: float = 1e-10
+    eps: Union[float, sympy.Expr] = 1e-10
```
tanujkhattar, line 77: Use `SymbolicFloat`instead.

tanujkhattar (approved):
LGTM % comments

qualtran/bloqs/chemistry/trotter/hubbard/qpe_cost_optimization.ipynb
fdmalone: I think I'll add an issue for now, and add the shim as a follow up. I need to remind myself of the sympy allocation updates. It would also be nice to compare all three alternatives. I have some bandwidth for the next few weeks.

fdmalone: Done.

PR #758, opened 2024-03-08
Gate decomposition via Householder reflections

Vicara12:
This code implements the first two points in [this comment](https://github.com/quantumlib/Qualtran/issues/596#issuecomment-1922366568) from issue [#596 ](https://github.com/quantumlib/Qualtran/issues/596). It adds a bloq that decomposes an arbitrary unitary gate into householder reflections. It allows for partial specification of the gate, in which case the T-gate cost of the bloq is linearly proportional to the number of gates specified.

This code is a necessary part for the goal of implementing a MPS preparation algorithm, even though it is useful on its own for other applications.


qualtran/bloqs/chemistry/prepare_mps/decompose_gate_hr.py
```diff
@@ -0,0 +1,240 @@
...
+from qualtran.bloqs.rotations.phase_gradient import PhaseGradientState
+from qualtran.bloqs.select_and_prepare import PrepareOracle
+from qualtran.bloqs.state_preparation.state_preparation_via_rotation import (
+    StatePreparationViaRotations,
+)
+
+
+@attrs.frozen
+class DecomposeGateViaHR(Bloq):
+    r"""Given all or some columns of an unitary, it is generated via Householder reflections.
```
mpharrigan, line 34: grammar nit: what is "it"? maybe try the active voice:

> given columns of a unitary, generate the operator via Householder reflections

qualtran/bloqs/chemistry/prepare_mps/decompose_gate_hr.py
```diff
@@ -0,0 +1,240 @@
...
+    $$
+
+    even though for the purpose of making this bloq simpler to work with, the gate actually
+    implemented is
+
+    $$
+        |0\rangle\langle 1| U + |1\rangle\langle 0| U^\dagger$.
+    $$
+
+    For a detailed description of how to use this bloq refer to its tutorial.
```
mpharrigan, line 50: give the title of the notebook/page instead of "its tutorial"

qualtran/bloqs/chemistry/prepare_mps/decompose_gate_hr.py
```diff
@@ -0,0 +1,240 @@
...
+
+    Args:
+        phase_bitsize: size of the register that is used to store the rotation angles. Bigger values
+            increase the accuracy of the results.
+        gate_cols: tuple that contains, for each entry, a pair where the first element is the
+            index of the column and the second a tuple that contains the complex values of each
+            column. For example, for the gate [[a,b,],[c,d]] it would be ((0,(a,c)), (1(b,d))).
```
mpharrigan, line 57: `(1, (b, d))`

qualtran/bloqs/chemistry/prepare_mps/decompose_gate_hr.py
```diff
@@ -0,0 +1,240 @@
...
+
+    References:
+        [Synthesis of unitaries with Clifford+T circuits]
+        (https://arxiv.org/pdf/1306.3200.pdf).
+            Kliuchnikov 2013.
+    """
+    phase_bitsize: int  # number of ancilla qubits used to encode the state preparation's rotations
+    gate_cols: Tuple[
+        int, Tuple[complex, ...]
+    ]  # tuple with the columns of the gate that are specified
```
mpharrigan, line 75: so I think this is actually `Tuple[Tuple[int, Tuple[complex, ...], ...]`.

A less specific but (imo) more readable way of writing it would be `Sequence[Tuple[int, Sequence[complex]]]`

qualtran/bloqs/chemistry/prepare_mps/decompose_gate_hr.py
```diff
@@ -0,0 +1,240 @@
...
+    uncompute: bool = False
+    internal_phase_grad: bool = False
+    internal_refl_ancilla: bool = True
+
+    def __attrs_post_init__(self):
+        # at least one column has to be specified
+        assert len(self.gate_cols) > 0
+        # there can't be more columns that rows
```
mpharrigan, line 83: that -> than

qualtran/bloqs/chemistry/prepare_mps/decompose_gate_hr.py
```diff
@@ -0,0 +1,240 @@
...
+        assert len(self.gate_cols) <= len(self.gate_cols[0][1])
+        # all cols must be the same length and a power of two
```
mpharrigan, line 85: some of these comments could be better as assertion messages or `ValueError` messages

qualtran/bloqs/chemistry/prepare_mps/decompose_gate_hr.py
```diff
@@ -0,0 +1,240 @@
...
+        return (len(self.gate_cols[0][1]) - 1).bit_length()
+
+    def build_composite_bloq(
+        self, bb: BloqBuilder, *, gate_input: SoquetT, **soqs: SoquetT
+    ) -> Dict[str, SoquetT]:
+        """Extra soquets inside soqs are:
+        * phase_grad: a phase gradient state of size phase_bitsize if internal_phase_gradient
+                    is set to False
+        * refl_ancilla: a clean qubit in |0> if internal_refl_ancilla is set to False
+        """
```
mpharrigan, line 109: this can probably just be line comments `#` instead of a docstring. If you want this to be a docstring, it should follow the docstring format. 

By "extra soquets" you mean "optional soquets"?

qualtran/bloqs/chemistry/prepare_mps/decompose_gate_hr_test.py
```diff
@@ -0,0 +1,145 @@
...
+            2,
+            (
+                (0, ((-0 - 0.5j), (0.5 - 0j), (-0.5 + 0.5j), -0j)),
+                (1, ((0.5 + 0.5j), (0.5 - 0.5j), 0j, 0j)),
+                (2, (0.5j, (-0.5 + 0j), (-0.5 + 0.5j), 0j)),
+                (3, (0j, 0j, (-0 + 0j), (-1 + 0j))),
+            ),
+        ],
+    ],
+)
```
mpharrigan, line 60: super optional, but if you want the formatter to stop making this terrible, you can use

```python
# fmt: off
[ (2, (a, b)), ... ]
# fmt: on
```

and the formatter won't touch those lines

mpharrigan (commented):
Great contribution! @tanujkhattar can you take a look as well.

My major ask is: can we rename these bloqs and the module to something other than "decompose". We already use that term to mean something else. I see the reference uses the word "synthesis" whose verb would be synthesize

qualtran/bloqs/chemistry/prepare_mps/decompose_gate_hr.py
Vicara12: Not really optional, but required in some situations. For example, if an user sets the `internal_phase_grad` flag to `False` then that register has to be provided, but it is not optional in that case.

mpharrigan: optional in a cosmic sense where you can set the flag to true

qualtran/bloqs/chemistry/prepare_mps/synthesize_gate_hr.py
```diff
@@ -0,0 +1,237 @@
...
+                    bitsize=self.state_bitsize + 1, iteration_length=2 ** (self.state_bitsize)
+                ),
+            ),
+        )
+
+    @property
+    def signature(self) -> Signature:
+        return Signature.build(target_reg=self.state_bitsize + 1, phase_grad=self.phase_bitsize)
+
+    def build_composite_bloq(
```
tanujkhattar, line 186: Please implement a `build_call_graph` for symbolic analysis.

qualtran/bloqs/reflection_using_prepare.py
```diff
@@ -63,6 +65,7 @@ class ReflectionUsingPrepare(GateWithRegisters):
     """
 
     prepare_gate: PrepareOracle
+    extra_registers: Tuple[Tuple[str, int], ...] = ()
```
tanujkhattar, line 68: Maybe directly expect a `Tuple[Register, ...]` instead of ` Tuple[Tuple[str, int], ...]` ?

Also, add the `extra_registers` as the last attribute in the after `global_phase` so existing call sites that pass `*args` instead of `**kwargs` are not broken

qualtran/bloqs/reflection_using_prepare.py
```diff
@@ -74,9 +77,15 @@ def control_registers(self) -> Tuple[Register, ...]:
     def selection_registers(self) -> Tuple[Register, ...]:
         return self.prepare_gate.selection_registers
 
+    @cached_property
+    def extra_reg_objects(self) -> Tuple[Register, ...]:
+        return tuple([Register(name, QAny(bitsize)) for name, bitsize in self.extra_registers])
```
tanujkhattar, line 82: Not needed after we change the class attribute to directly store `Tuple[Register, ...]`

qualtran/bloqs/reflection_using_prepare.py
```diff
@@ -141,7 +150,9 @@ def controlled(
             and len(control_values) == 1
             and self.control_val is None
         ):
-            return ReflectionUsingPrepare(self.prepare_gate, control_val=control_values[0])
+            return ReflectionUsingPrepare(
+                self.prepare_gate, self.extra_registers, control_val=control_values[0]
+            )
```
tanujkhattar, line 155: Change this to `return attrs.evolve(self, control_val=control_values[0])` 
```suggestion
            return attrs.evolve(self, control_val=control_values[0])
```

qualtran/bloqs/reflection_using_prepare.py
```diff
@@ -51,6 +51,8 @@ class ReflectionUsingPrepare(GateWithRegisters):
 
     Args:
         prepare_gate: An implementation of `PREPARE` for state preparation.
+        extra_registers: If the prepare_gate receives extra registers that do not interfere in the
+            rotation, those should be specified there as pairs (name, bitsize)
```
tanujkhattar, line 55: Should be the last argument instead of second.

qualtran/bloqs/chemistry/prepare_mps/synthesize_gate_hr.py
```diff
@@ -0,0 +1,237 @@
...
+
+    For a detailed description of how to use this bloq refer to its tutorial:
+    synthesize_gate_hr_tutorial.ipynb.
+
+    Args:
+        phase_bitsize: size of the register that is used to store the rotation angles. Bigger values
+            increase the accuracy of the results.
+        gate_cols: tuple that contains, for each entry, a pair where the first element is the
+            index of the column and the second a tuple that contains the complex values of each
+            column. For example, for the gate [[a,b,],[c,d]] it would be ((0,(a,c)), (1,(b,d))).
```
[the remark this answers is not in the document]
tanujkhattar: If they don't have to be consecutive, let's clarify that in the docstring. Also, we can add a factory method that takes a `Dict[int, np.ndarray]` that specifies the columns OR a 2D np.ndarray that specifies the columns (if they are consecutive) and instantiates the Bloq by converting to a `Tuple[Tuple[int, Tuple[complex, ...]]]`

We can also change the bloq attribute to be `Tuple[Tuple[int, np.ndarray]]` and add a `attrs.field(eq=_convert_to_tuple)` like we do in other state preparation bloqs now

tanujkhattar (changes requested):
@Vicara12 Thanks a lot for opening this PR and apologies for the late review!

I've left a round of comments - Would you still have time to address the comments and get this PR merged? Please let us know one way or the other since this would be a very useful primitive to get in. 

Main suggestions are around making changes so we can support symbolic analysis.

qualtran/bloqs/chemistry/prepare_mps/synthesize_gate_hr.py
```diff
@@ -0,0 +1,237 @@
...
+
+    $$
+        |0\rangle\langle 1| U + |1\rangle\langle 0| U^\dagger$,
+    $$
+
+    even though for the purpose of making this bloq simpler to work with, the gate actually
+    implemented is
+
+    $$
+        |0\rangle\langle 1| U + |1\rangle\langle 0| U^\dagger$.
```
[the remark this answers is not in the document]
Vicara12: Yes, sorry. The second one should have been $|0\rangle\langle 0| U + |1\rangle\langle 1| U^\dagger$, that is, instead of flipping the ancilla it just takes it and gives it back both as |0>. This makes using the block easier because when freeing the ancilla it is projected to zero.

qualtran/bloqs/chemistry/prepare_mps/synthesize_gate_hr.py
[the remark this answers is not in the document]
Vicara12: Gate columns do not need to be consecutive, I'll specify that in the docstring!

mhucka:
@Vicara12 @tanujkhattar Just circling around to different PRs and trying to do a bit of housekeeping. For planning purposes, what is the status of this PR?

Issue #738, opened 2024-03-04
QROM controls


mpharrigan:
QROM takes a `num_controls` parameter and constructs one control register of bitsize=n. These usually function as a vector of flags, right? shouldn't it be `bitsize=1, shape=(n,)`? @tanujkhattar

Issue #677, opened 2024-02-16
Extra MultiControlPauli Cases


brempfer:
In the decompose_from_registers method for the MultiControlPauli class it would be helpful for some of MIT LL's workflow to have two more cases for len(self.cvs) = 0 or len(self.cvs) = 1, i.e. the zero case would return the target gate on the target register etc. We know this an easy fix but to avoid licensing issues we thought we would go through the proper channels.

PR #765, opened 2024-03-11
Replace `MultiControlX` bloq with `MultiControlPauli`

tanujkhattar:
`MultiControlX` was added because `MultiControlPauli` didn't support classical simulations. After the recent improvements to Cirq/Bloqs interop; `MultiControlPauli` now supports classical simulations and thus it `MultiControlX` can now be replaced with uses of `MultiControlPauli`

This PR also changes the signature of `MultiControlPauli` to accept a flat list of qubits as controls, as done in `MultiControlX` and suggested in https://github.com/quantumlib/Qualtran/issues/738


qualtran/bloqs/factoring/mod_sub.py
```diff
@@ -138,7 +138,7 @@ def build_composite_bloq(self, bb: 'BloqBuilder', x: SoquetT) -> Dict[str, 'Soqu
         for i in range(self.bitsize):
             cvs = cvs + (0,)
         x_split = bb.split(x)
-        x_split, ctrl = bb.add(MultiControlX(cvs=cvs), ctrls=x_split, x=ctrl)
+        x_split, ctrl = bb.add(MultiControlPauli(cvs=cvs), controls=x_split, target=ctrl)
```
mpharrigan, line 141: the reader used to be able to see what was being done here and now there is ambiguity. Honestly, the most readable thing would be something like `XGate().controlled(CtrlSpec(cvs))`

qualtran/bloqs/mcmt/multi_control_multi_target_pauli.py
```diff
@@ -111,43 +101,61 @@ class MultiControlPauli(GateWithRegisters):
     target_gate: cirq.Pauli = cirq.X
```
mpharrigan, line 101: does it make sense to have a default here? do you have to use cirq gates or can you use `XGate()`?

qualtran/bloqs/factoring/mod_sub.py
tanujkhattar: I agree, that would require overriding the `controlled()` method for `XGate()`, which can be done as a follow-up PR. 

The reason I am making this change is that I have plans to add more specialized and efficient decompositions for `MultiControlX`, which can also be used at the call sites instead of the more general `MultiControlPauli` that we use right now.

qualtran/bloqs/mcmt/multi_control_multi_target_pauli.py
```diff
@@ -111,43 +101,61 @@ class MultiControlPauli(GateWithRegisters):
...
 
     @cached_property
-    def signature(self) -> Signature:
-        return Signature.build(controls=len(self.cvs), target=1)
+    def signature(self) -> 'Signature':
+        ctrl = Register('controls', QBit(), shape=(len(self.cvs),))
+        target = Register('target', QBit())
+        return Signature([ctrl, target] if len(self.cvs) > 0 else [target])
```
mpharrigan, line 107: why are we supporting no controls?

qualtran/bloqs/mcmt/multi_control_multi_target_pauli.py
```diff
@@ -111,43 +101,61 @@ class MultiControlPauli(GateWithRegisters):
...
+        ctrl = f'C^{n}' if n > 2 else ['', 'C', 'CC'][n]
+        return f'{ctrl}{self.target_gate!s}'
 
     def _circuit_diagram_info_(self, _) -> cirq.CircuitDiagramInfo:
         wire_symbols = ["@" if b else "@(0)" for b in self.cvs]
         wire_symbols += [str(self.target_gate)]
         return cirq.CircuitDiagramInfo(wire_symbols=wire_symbols)
 
+    def on_classical_vals(
+        self, controls: 'ClassicalValT', target: 'ClassicalValT'
```
mpharrigan, line 140: this assumes >= 1 controls

qualtran/bloqs/mcmt/multi_control_multi_target_pauli.py
tanujkhattar: Happy to remove the default in favour of more readability. 

Right now, it needs to be a cirq gate because `Bloq.controlled` and `Gate.controlled` don't alight very well and `Gate.controlled` is nicer for `X/Y/Z` gates because it has special overrides to give back `CX/CY/CZ` and `CCX/CCY/CCZ` for `1/2` qubit case. In future, I want to accept both `XGate` and `cirq.X` here

qualtran/bloqs/mcmt/multi_control_multi_target_pauli.py
tanujkhattar: There was an explicit feature request from Lincoln lab folks. See https://github.com/quantumlib/Qualtran/issues/677

qualtran/bloqs/mcmt/multi_control_multi_target_pauli.py
tanujkhattar: @mpharrigan I've removed the default, PTAL!

qualtran/bloqs/mcmt/multi_control_multi_target_pauli.py
tanujkhattar: Added support for 0 controls and a corresponding test

tanujkhattar:
@mpharrigan This is ready for another look. I've addressed all comments. PTAL!

qualtran/bloqs/factoring/mod_sub.py
mpharrigan: isn't that a strong argument for keeping these separate

tanujkhattar: Follow up decompositions for Multi control X would be implemented as a separate new Bloq; which isn't general enough for all pauli types

This Bloq works for all pauli types so it makes sense to have it this way.

tanujkhattar:
@mpharrigan Can you take another look at this?

mpharrigan:
I don't think this is a good idea. It makes the user code more verbose; I'm wary of having a controlled bloq that supports no controls, and it sounds like we're going to have its own decomposition anyway. It's not like we pay per bloq

PR #768, opened 2024-03-11
Remove special adjoint override of `HammingWeightCompute` Bloq

tanujkhattar:
Fixes a TODO (see comment in code) and removes the special override for `adjoint` in `HammingWeightCompute`. 

The T-Complexity method is updated to accept a `adjoint` parameter and make the tests pass, but should be changed as part of https://github.com/quantumlib/Qualtran/issues/735


mpharrigan:
Do we know what the problem with cirq simulation was and how it got fixed?

Issue #396, opened 2023-10-13
Bloqs should know their namespaced name


mpharrigan:
Our standard library of bloqs is under `qualtran.bloqs`. As it grows, we'll continue to organize things into packages and modules. 

For serialization and to help some of the automatic documentation, bloqs should know their namespace.

 - is there a way to reliably do this from within python? E.g. if I have a `ModMul` object is it possible to find that it lives in `qualtran.bloqs.factoring`? I've found that `obj.__module__` depends on how you imported it; which is not great.
 - maybe we need a namespace field on bloqs? or the bloq class? how do we ensure that this namespace field matches the actual organization of the object in the class heirarchy

cc @tanujkhattar for thoughts

mpharrigan:
I don't know what issue I was running in to with just using `cls.__module__`. I can't replicate it

PR #772, opened 2024-03-12
Add a bloq classification utility

fdmalone:
In the THC notebook I added a bloq classification method. I've generalized this here to use the newly added structure for the bloqs directory to classify the T counts into buckets defined by their classification (swaps, arithmetic, ...) The user can optionally provide their own classifier which is just a dictionary mapping a classification (str) to a tuple of bloqs.

For example one can now do:

``` python
...
select = BlackBoxSelect(sel_thc)
prepare = BlackBoxPrepare(prep_thc)
block_encoding_bloq = BlackBoxBlockEncoding(select, prepare) 
from qualtran.resource_counting.classify_bloqs import classify_t_count_by_bloq_type
classify_t_count_by_bloq_type(block_encoding_bloq)
```
which will return

```
defaultdict(int,
            {'other': 0,
             'arithmetic': 2088,
             'data_loading': 45584,
             'reflection': 152,
             'rotations': 104,
             'mcmt': 16,
             'toffoli': 27556,
             'swaps': 1960})
```
Helpful for also gauging where there are some gaps in desciption (the big bunch of Toffolis is due to too little description in the call graph for Select).


qualtran/resource_counting/generalizers.py
```diff
@@ -39,6 +39,15 @@ def ignore_split_join(b: Bloq) -> Optional[Bloq]:
     return b
 
 
+def ignore_partition(b: Bloq) -> Optional[Bloq]:
+    """A generalizer that ignores Partition operations."""
+    from qualtran.bloqs.util_bloqs import Partition
+
+    if isinstance(b, Partition):
+        return None
+    return b
```
mpharrigan, line 48: I think this should just be done as part of `ignore_split_join` #777

qualtran/resource_counting/classify_bloqs.py
```diff
@@ -0,0 +1,104 @@
...
+    ignore_partition,
+    ignore_split_join,
+)
+from qualtran.resource_counting.t_counts_from_sigma import (
+    _get_all_rotation_types,
+    t_counts_from_sigma,
+)
+
+
+def _get_all_bloqs_in_module(bloq_module: str) -> Tuple[Bloq]:
```
mpharrigan, line 37: here and everywhere the type should be `Type[Bloq]`, no?

mpharrigan:
I would architect this the other way, where you can take any bloq -- look at its fully-qualified name -- and extract out its category. As written, you have to import every bloq in the library but the code will still fail on any bloq defined outside of one of the listed modules

qualtran/resource_counting/classify_bloqs.py
```diff
@@ -0,0 +1,104 @@
...
+    """Returns all classes defined in bloqs `bloq_module`."""
+    importlib.import_module(bloq_module)
+    all_members = [v for (_, v) in inspect.getmembers(sys.modules[bloq_module], inspect.isclass)]
```
mpharrigan, line 40: what if someone defines a class that isn't a subclass of `Bloq`?

mpharrigan: did you test that this isn't mis-organizing bloqs imported by other bloqs? Take a look at dev_tools/bloq_finder.py

fdmalone: When would this happen? The goal is to classify bloqs, are any bloqs not subclasses of Bloq?

mpharrigan: there might be members of `sys.modules[bloq_module]` that are classes but not bloqs that will be returned

fdmalone: On your second point, that's a good test. But it works:

``` python
hwp =  _get_all_bloqs_in_module('qualtran.bloqs.rotations.hamming_weight_phasing')
```
returns 

```
 (<class 'qualtran.bloqs.rotations.hamming_weight_phasing.HammingWeightPhasing'>, <class 'qualtran.bloqs.rotations.hamming_weight_phasing.HammingWeightPhasingViaPhaseGradient'>)
```
(i.e. no ZPowGate or any other basic gates used in that module)

The following line checks that `bloq_module` is in the list of members which prunes anything out which doesn't have 'qualtran.bloqs.hamming_weight_phasing` in it.

fdmalone: > there might be members of `sys.modules[bloq_module]` that are classes but not bloqs that will be returned

Ah, I can definitely add a check that it is a bloq.

fdmalone:
> I would architect this the other way, where you can take any bloq -- look at its fully-qualified name -- and extract out its category. As written, you have to import every bloq in the library but the code will still fail on any bloq defined outside of one of the listed modules.
> 
> #396

This seems a little more brittle but probably just a matter of checking if the module path is in the name.

qualtran/resource_counting/classify_bloqs.py
```diff
@@ -0,0 +1,88 @@
...
+        'rotations': ('qualtran.bloqs.rotations',),
+        'state_preparation': ('qualtran.bloqs.state_preparation',),
+        'data_loading': ('qualtran.bloqs.data_loading',),
+        'mcmt': ('qualtran.bloqs.mcmt',),
+        'multiplexers': ('qualtran.bloqs.multiplexers',),
+        'swaps': ('qualtran.bloqs.swap_network', 'qualtran.bloqs.basic_gates.swap'),
+        'reflection': ('qualtran.bloqs.reflection',),
+        'toffoli': ('qualtran.bloqs.basic_gates.toffoli',),
+        'tgate': ('qualtran.bloqs.basic_gates.t_gate',),
+    }
```
mpharrigan, line 42: Sorry for stringing you along, but what do you think about making the full name even more influential. Instead of a dict you can use a callable

```python
def classify(bloq_cls):
  mod_name = bloq_cls.__module__
  q, b, pkg, *_ = mod_name.split('.')
  if f'{q}.{b}' != 'qualtran.bloqs':
    return 'other'
  if bloq_cls == TGate:
    return 'tgate'
  return pkg
```

fdmalone: I just pushed an updated version. I think what I have is a bit more flexible as it won't necessarily rely on the module structure. For example, what if I want to be more specific and classify 'data_loading' into 'qrom' and 'select_swap_qrom', which would require some further logic?

fdmalone: Or for example, currently I'm combining swaps into one bucket. Granted the combination could be done using a second pass if you'd prefer?

fdmalone: @mpharrigan WDYT: should it be a two stage thing with sub-classification happening after 'pkg' level classification?

mpharrigan (approved):
sorry for holding this up. I think this is an improvement.

There will likely be merge conflict with #777 tho

PR #773, opened 2024-03-12
Implementation of an Approximate QFT algorithm

skushnir123:
In practice, running an exact QFT is often wasteful, and one can get by running an approximate QFT in which phase shifts smaller than a certain threshold are deleted. 

ApproximateQFT differentiates from the "textbook" QFT in two ways: 1) It uses the "add into phase-gradient trick" and 2) It 
truncates the phase-gradient register and therefore does smaller additions than the [PhaseGradientQFT](https://github.com/quantumlib/Qualtran/blob/main/qualtran/bloqs/qft/qft_phase_gradient.py). This has the same effect as deleting small phase-shifts in the "textbook" QFT. 

The degree to which the QFT is approximated is up to the user: ApproximateQFT is parameterized by "b" which is a function of n, the number of qubits in the QFT. The parameter b is the number of qubits in the phase gradient register. (By default b = log(n)) If b = n, this would be the standard QFT.

The asymptotic T-count of ApproximateQFT, when b = log(n), is O(8n*(log(n)^2)).


qualtran/bloqs/qft/approximate_qft.py
```diff
@@ -0,0 +1,93 @@
...
+            yield cirq.H(*quregs['q'])
+            return
+        q, phase_grad = quregs['q'], quregs['phase_grad']
+        for i in range(len(q)):
+            if i == 0:
+                yield cirq.H(q[i])
+                continue
+            addition_bitsize = min(i, len(phase_grad) - 1)
+            a, b = q[:addition_bitsize], phase_grad[: addition_bitsize + 1]
+            yield PlusEqualProduct(addition_bitsize, 1, addition_bitsize + 1).on_registers(
```
skushnir123, line 86: @tanujkhattar originally, I was using AddIntoPhaseGrad using the following line of code: `yield AddIntoPhaseGrad(addition_bitsize, addition_bitsize + 1, right_shift=1).on_registers(x=a[::-1], phase_grad=b).controlled_by(q[i])`

However, the T-count is hard to access from this construction. And after all, a controlled-addition is exactly the same thing as multiplying the summand by the control. This might be the most efficient way to implement a controlled-addition?

qualtran/bloqs/qft/approximate_qft_test.py
```diff
@@ -0,0 +1,79 @@
...
+        return {'q': q}
+
+
+@pytest.mark.parametrize('n', [2, 3, 4, 5])
+@pytest.mark.parametrize('without_reverse', [True, False])
+def test_qft_with_phase_gradient(n: int, without_reverse: bool):
+    qft_bloq = TestApproximateQFT(n, not without_reverse)
+    qft_cirq = cirq.QuantumFourierTransformGate(n, without_reverse=without_reverse)
+    np.testing.assert_allclose(cirq.unitary(qft_bloq), cirq.unitary(qft_cirq))
+    np.testing.assert_allclose(cirq.unitary(qft_bloq**-1), cirq.unitary(qft_cirq**-1))
```
tanujkhattar, line 60: Can you add tests where we pass in a smaller bitsize for the phase gradient state and compare the unitaries are "approximately equal" upto some error tolerance epsilon?

qualtran/bloqs/qft/approximate_qft_test.py
```diff
@@ -0,0 +1,79 @@
...
+
+    assert_valid_bloq_decomposition(qft_bloq)
+
+
+@pytest.mark.parametrize('n', [10, 123])
+def test_qft_text_book_t_complexity(n: int):
```
tanujkhattar, line 66: Change the name

qualtran/bloqs/qft/approximate_qft_test.py
```diff
@@ -0,0 +1,79 @@
...
+    def f(n, b):
+        t_complexity = 0
+        for i in range(1, n):
+            t_complexity += PlusEqualProduct(min(i, b - 1), 1, min(i, b - 1) + 1).t_complexity().t
+        return t_complexity
+
+    qft_t_complexity = qft_bloq.t_complexity()
+    b = math.ceil(math.log2(n))
+
+    assert qft_t_complexity.t == f(n, b) <= 8 * n * (math.log2(n) ** 2)
```
tanujkhattar, line 78: `PlusEqualProduct` has a complexity that scales as `max(a_bitsize, b_bitsize) ** 2`. This is giving you the `math.log2(n) ** 2` factor right now. 

However, a controlled addition should have a complexity that is twice that of an addition (see https://arxiv.org/pdf/1709.06648.pdf). Therefore if you use a controlled addition (which you can maybe add as a separate PR and would be a great contribution, see https://github.com/quantumlib/Qualtran/issues/778) then the cost of your `AQFT` will go down from $n\log^2(n)$ to $n\log^2(n)$

qualtran/bloqs/qft/approximate_qft.py
```diff
@@ -0,0 +1,93 @@
...
+            performing later operations on different qubits.
+
+    References:
+        [Turning Gradients into Additions into QFTs](https://algassert.com/post/1620)
+    """
+
+    bitsize: int
+    b: Callable = lambda n: math.ceil(math.log2(n))
+    with_reverse: bool = True
+
```
tanujkhattar, line 64: Can we add a class function which accepts an error tolerance `epsilon` and figures out the required phase gradient bitsize `b` using the error tolerance?

tanujkhattar (changes requested):
This is a great start. I've left a round of comments with some suggestions

qualtran/bloqs/qft/approximate_qft.py
```diff
@@ -0,0 +1,93 @@
...
+            These are technically necessary in order to perform the
+            correct effect, but can almost always be optimized away by just
+            performing later operations on different qubits.
+
+    References:
+        [Turning Gradients into Additions into QFTs](https://algassert.com/post/1620)
+    """
+
+    bitsize: int
+    b: Callable = lambda n: math.ceil(math.log2(n))
```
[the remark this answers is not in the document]
skushnir123: Done

qualtran/bloqs/qft/approximate_qft_test.py
skushnir123: Done

qualtran/bloqs/qft/approximate_qft.py
skushnir123: Added

qualtran/bloqs/qft/approximate_qft_test.py
skushnir123: Great, this is helpful. Should this be blocking for this PR? I can work on adding the controlled addition.

qualtran/bloqs/qft/approximate_qft.py
```diff
@@ -0,0 +1,121 @@
...
+import numpy as np
+from attr import field
+from numpy.typing import NDArray
+
+from qualtran import GateWithRegisters, QFxp, QUInt, Signature
+from qualtran.bloqs.arithmetic.multiplication import PlusEqualProduct
+
+
+@attrs.frozen
+class ApproximateQFT(GateWithRegisters):
```
tanujkhattar, line 28: Can you use the auto doc infrastructure to generate a corresponding tutorial notebook? You can see my recent PR on how to do this when adding a new Bloq

https://github.com/quantumlib/Qualtran/commit/6b6900f71a2f226fa5b8ee6557bcb3073897edcd

Specifically, do the following
1. Create functions in `approximate_qft.py` file that instantiate an aqft Bloq and decorate it with a `@bloq_example`
2. Create a `_CC_AQFT_DPC = BloqDocSpec` object and include it in the `autogenerate-bloqs-notebooks-v2.py` file
3. Use function from step-1 in a bloq_autotester test in `approximate_qft_test.py`
4. Run `python dev_tools/autogenerate-bloqs-notebooks-v2.py` to generate a tutorial notebook, verify the latex renders correctly and commit it.

tanujkhattar (approved):
I've pushed a few minor improvements since I know you are on vacation. Thanks for the contribution!

PR #781, opened 2024-03-12
Add bloqs to convert classical values to quantum values for some dtypes

fpapa250:
Adds functions which convert classical values to quantum registers using XGate()'s. Includes classical simulation testing for positive integers only.

### DIFF
--- a/qualtran/bloqs/classical_quantum_conversions.py
+++ b/qualtran/bloqs/classical_quantum_conversions.py
@@ -1,0 +1,180 @@
+#  Copyright 2023 Google LLC
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from functools import cached_property
+from typing import Dict
+
+from attrs import frozen
+
+from qualtran import (
+    Bloq,
+    BloqBuilder,
+    QBit,
+    QInt,
+    QIntOnesComp,
+    QUInt,
+    Register,
+    Signature,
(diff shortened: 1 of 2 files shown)


mpharrigan:
Re: naming: these operations will initialize a constant `k` if the register is in the zero state. But otherwise it does a bitwise xor. Maybe something like `XorInt` `XorUInt`

fpapa250:
> Re: naming: these operations will initialize a constant `k` if the register is in the zero state. But otherwise it does a bitwise xor. Maybe something like `XorInt` `XorUInt`

Do you want me to add that Xor functionality for the classical simulation or is the current implementation sufficient?

qualtran/bloqs/classical_quantum_conversions.py
```diff
@@ -0,0 +1,180 @@
...
+    k: int
+
+    @cached_property
+    def signature(self) -> 'Signature':
+        return Signature([Register('x', QBit())])
+
+    def on_classical_vals(self, x: 'ClassicalValT') -> Dict[str, 'ClassicalValT']:
+        assert (
+            self.k == 0 or self.k == 1
+        ), "Argument `k` should be a classical bit containing value 0 or 1"
```
mpharrigan, line 55: this sort of validation probably should be done at construction time. You can use attrs validators

```python
k: int = field(validator=lambda k: k in [0, 1])
```

or similar

mpharrigan:
Yeah, I think the classical simulation should match the actual behavior of the bloq. At the very minimum, you could check the input value and if it's anything other than zero, raise a `NotImplementedError`; but ideally you'd implement an xor.

If you want to mandate that the input is always zero, then you should avoid having an input. You can change the register to a RIGHT register. The decomposition can start with an allocation and then you can do your x gates. This would be redundant with `IntState`, though; so check that out first if you want to go that route.

fpapa250:
> Yeah, I think the classical simulation should match the actual behavior of the bloq. At the very minimum, you could check the input value and if it's anything other than zero, raise a `NotImplementedError`; but ideally you'd implement an xor.
> 
> If you want to mandate that the input is always zero, then you should avoid having an input. You can change the register to a RIGHT register. The decomposition can start with an allocation and then you can do your x gates. This would be redundant with `IntState`, though; so check that out first if you want to go that route.

Honestly I think `IntState` is probably better for this anyway. I just wasn't aware it already existed. I'll opt to use that instead.

PR #782, opened 2024-03-12
Configure `LinearDepthGreaterThan()` bloq to allow for multiple target qubits

fpapa250:
Add num_targets arg with default value 1 which allows us to have a bloq with multiple targets. Added functionality to classical simulation as well (for positive values).


qualtran/bloqs/arithmetic/comparison.py
```diff
@@ -575,26 +577,33 @@ class LinearDepthGreaterThan(Bloq):
     """
     bitsize: int
     signed: bool
+    num_targets: int = 1
```
tanujkhattar, line 580: Change to `target_bitsize` ?

tanujkhattar (changes requested):
Can you explain what's the use case for this? The interface looks very unintuitive to me.

fpapa250:
> Can you explain what's the use case for this? The interface looks very unintuitive to me.

I agree it's odd looking, but it's used in https://arxiv.org/abs/2306.08585 specifically in figure 7 for modular inverse. Open to thoughts on cleaning it up; this is what immediately made sense to me.

fdmalone:
Could you just add a CX gate controlled on the target of the original inequality target?

fpapa250:
> Could you just add a CX gate controlled on the target of the original inequality target?

Neither target will always be flipped from 0 to 1. So I don't think that would be an equivalent operation.

qualtran/bloqs/arithmetic/comparison.py
```diff
@@ -603,10 +612,11 @@ def build_composite_bloq(
             # We use a specially controlled Toffolli gate to implement GreaterThan.
             # If a is 1 and b is 0 then a > b and we can flip the target bit.
             ctrls = [a, b]
-            ctrls, target = bb.add(MultiControlX(cvs=(1, 0)), ctrls=ctrls, x=target)
+            for i in range(self.num_targets):
+                ctrls, targets[i] = bb.add(MultiControlX(cvs=(1, 0)), ctrls=ctrls, x=targets[i])
```
tanujkhattar, line 616: Doing multiple Toffoli's is more expensive than doing an `And` and use a temporary ancilla to store the output; do a multi target CX controlled the result of this ancilla; and then uncompute the ancilla using `And(uncompute=True)`

tanujkhattar:
> Neither target will always be flipped from 0 to 1. So I don't think that would be an equivalent operation.

We can always use a temporary ancilla and use that to do a multi target CX

mpharrigan:
If you look at the construction, there already is an ancilla with the correct value that needs to be cnotted out to all the outputs. It's flanked by a compute/uncompute pattern per usual. 

In an ideal world, we'd have these first-class flanking operations so you could define a partial `GreaterThan` register; maybe with a `Side.DOWN` register; and you could combine it with whatever output-like bloq you want. e.g.

```python
MultiTarget(bloq=GreaterThan(), n_targets=5)
```

but that's sadly not the world we live in yet. So you could manually wrap the existing one-target greaterthan with a multitarget version composed of

 1. do greaterthan to output one bit
 2. do multi-target cnot to copy it to the n-1 other bits

all that being said -- this `LinearDepthGreaterThan` was introduced to track the referenced paper and it already doesn't really match the signature of `GreaterThan` so it might be fine to introduce the multi-target version directly for expediency

Issue #769, opened 2024-03-11
Implement resource state preparation algorithm for phase estimation from arxiv:1805.03662


tanujkhattar:
Implement resource state preparation algorithm described in https://arxiv.org/pdf/1805.03662.pdf

<img width="1311" alt="image" src="https://github.com/quantumlib/Qualtran/assets/7863287/63ba3742-4f5f-46aa-a13e-2687004cc1df">

tanujkhattar:
Here is a quirk circuit that shows how to implement the deterministic state prep using a single round of AA 

[...]

Issue #650, opened 2024-02-08
[controlled] Use quimb tensor factorization for `Controlled` bloqs


mpharrigan:
Following #600 

Implement a more intelligent override for `add_my_tensors` that can construct a control tensor and contract it with the original bloq tensor

tanujkhattar:
Bumping this up - `Controlled` bloq is currently not usable for situations where we we'd like to do any kind of simulations. This is because we can't derive tensors for controlled atomic Bloqs. For example:

```python
In [4]: TGate().controlled().tensor_contract()
```
Raises an error

<details> 

```
---------------------------------------------------------------------------
DecomposeNotImplementedError              Traceback (most recent call last)
Cell In[4], line 1
----> 1 TGate().controlled().tensor_contract()

File ~/quantum/Qualtran/qualtran/_infra/bloq.py:247, in Bloq.tensor_contract(self)
    238 """Return a contracted, dense ndarray representing this bloq.
    239
    240 This constructs a tensor network and then contracts it according to our registers,
   (...)
    243 of (right, left) indices.
    244 """
    245 from qualtran.simulation.tensor import bloq_to_dense
--> 247 return bloq_to_dense(self)

File ~/quantum/Qualtran/qualtran/simulation/tensor/_dense.py:60, in bloq_to_dense(bloq)
     46 """Return a contracted, dense ndarray representing the composite bloq.
     47
     48 The public version of this function is available as the `Bloq.tensor_contract()`
   (...)
     57 `cbloq_to_quimb` and `TensorNetwork.to_dense` directly.
     58 """
     59 cbloq = bloq.as_composite_bloq()
---> 60 tn, _ = cbloq_to_quimb(cbloq)
     61 inds = get_right_and_left_inds(cbloq.signature)
     63 if inds:

File ~/quantum/Qualtran/qualtran/simulation/tensor/_quimb.py:99, in cbloq_to_quimb(cbloq, pos)
     86 inc_d = _cxn_to_soq_dict(
     87     bloq.signature.lefts(),
     88     incoming,
     89     get_me=lambda cxn: cxn.right,
     90     get_assign=lambda cxn: cxn.left,
     91 )
     92 out_d = _cxn_to_soq_dict(
     93     bloq.signature.rights(),
     94     outgoing,
     95     get_me=lambda cxn: cxn.left,
     96     get_assign=_assign_outgoing,
     97 )
[...]

PR #788, opened 2024-03-13
Add Bloq to prepare optimal resource state proposed by A. Luis and JPeřina (1996) for phase estimation

tanujkhattar:
Fixes https://github.com/quantumlib/Qualtran/issues/769

cc @ncrubin


qualtran/bloqs/phase_estimation/lp_resource_state.py
```diff
@@ -0,0 +1,179 @@
...
+            (ZPowGate(exponent=1 / ((2**self.bitsize + 1) * np.pi), global_shift=-0.5), 1),
+            (Hadamard(), 2 + self.bitsize),
+        }
+
+    def _t_complexity_(self) -> 'TComplexity':
+        return TComplexity(rotations=self.bitsize + 1, clifford=2 + self.bitsize)
+
+
+@attrs.frozen
+class LPResourceState(GateWithRegisters):
```
fdmalone, line 79: nit: Do we imagine having many other resource state algorithms implemented? If not Would it be better to remove the LP from the name since you've cited their contribution in the doc string?

qualtran/bloqs/phase_estimation/lp_resource_state.py
```diff
@@ -0,0 +1,179 @@
...
+        Complexity](https://arxiv.org/abs/1805.03662) Section II-B
+    """
+
+    bitsize: int
+
+    @cached_property
+    def signature(self) -> 'Signature':
+        return Signature([Register('m', QUInt(self.bitsize), side=Side.RIGHT)])
+
+    def state(self):
```
fdmalone, line 106: Should this be apart of the class? Maybe it should be called expected_state. Also docstring + return type hint.

qualtran/bloqs/phase_estimation/lp_resource_state.py
```diff
@@ -0,0 +1,179 @@
...
+    eps: float = 1e-11
+
+    @cached_property
+    def signature(self) -> 'Signature':
+        return Signature.build(m=self.bitsize, anc=1)
+
+    def short_name(self) -> str:
+        return 'LPRS'
+
+    def decompose_from_registers(
```
fdmalone, line 53: What's the general motivation for making this a GateWithRegisters rather than Bloq/build_composite_bloq? I guess below we're missing a global phase operations for bloqs right?

qualtran/bloqs/phase_estimation/lp_resource_state.py
```diff
@@ -0,0 +1,179 @@
...
+from qualtran.bloqs.mcmt import MultiControlPauli
+from qualtran.cirq_interop import CirqGateAsBloq
+from qualtran.cirq_interop.t_complexity_protocol import TComplexity
+
+if TYPE_CHECKING:
+    from qualtran.resource_counting import BloqCountT, SympySymbolAllocator
+
+
+@attrs.frozen
+class _LPRSHelper(GateWithRegisters):
```
fdmalone, line 35: Is it worthwhile making this public? It could be nice in a notebook to show this reproduces Eq. 19 for example?

qualtran/bloqs/phase_estimation/lp_resource_state.py
tanujkhattar: Yes, there are more resource state based algorithms that we'll eventually add. For example resource states based on Kaiser windows.

qualtran/bloqs/phase_estimation/lp_resource_state.py
fdmalone: It's only used for testing right?

tanujkhattar: Yes, I can remove it and move to the test file

qualtran/bloqs/phase_estimation/lp_resource_state_test.py
```diff
@@ -0,0 +1,40 @@
...
+@pytest.mark.parametrize('n', [*range(1, 14, 2)])
+def test_t_complexity(n):
+    bloq = LPResourceState(n)
+    if n == 1:
+        # n=1 fails due to https://github.com/quantumlib/Qualtran/issues/785
+        with pytest.raises(AssertionError):
+            assert_decompose_is_consistent_with_t_complexity(bloq)
+    else:
+        assert_decompose_is_consistent_with_t_complexity(bloq)
+        assert bloq.t_complexity().t + bloq.t_complexity().rotations == 7 * n + 6
```
fdmalone, line 40: Do we have an autotester for bloq_counts == decomposed_bloq_counts. If not maybe just assert the call_graph is the same as the decomposition.

fdmalone (changes requested):
Some small nits and questions. The perennial cirq / bloq decomposition question.

qualtran/bloqs/phase_estimation/lp_resource_state.py
tanujkhattar: done

qualtran/bloqs/phase_estimation/lp_resource_state.py
tanujkhattar: Done

qualtran/bloqs/phase_estimation/lp_resource_state_test.py
tanujkhattar: There is an `assert_equivalent_bloq_example_counts_for_pytest` which isn't currently enforced. I've added bloq_autotester tests so that whenever we enforce the above test, these bloqs will be included.

qualtran/bloqs/phase_estimation/lp_resource_state.py
tanujkhattar: Yes, also the controlled bloq in Qualtran isn't very functional right now for simulations; so we need to fallback to `cirq.ControlledGate`. See https://github.com/quantumlib/Qualtran/issues/650 for example

tanujkhattar:
Also, tensor contraction is significantly slower than the cirq state vector simulation, so using the Cirq API is helpful for writing meaningful tests that depend upon state vector simulation. For example:

image


My initial guess for why it happens was that we don't explicitly override `add_my_tensors` for `LPRSInterimPrep` (which is true for cirq interface as well, i.e. we don't explicitly override `_unitary_`) so the Bloq style tensor contraction tries to get a dense tensor for `LPRSInterimPrep` and use it for tensor contraction of `LPResourceState`; instead of relying on the decomposition of `LPRSInterimPrep`

image


I then tried flattening the Bloq but that just makes the simulation time even worse

image


and now you see from the snakeviz plot that the majority of the time really goes in contracting the flattened tensor; so quimb is just very slow I believe

image

fdmalone:
very concerning. LGTM.

fdmalone:
@tanujkhattar Was the paper construction correct in the end or were some updates required?

fdmalone (approved):

qualtran/bloqs/phase_estimation/lp_resource_state.ipynb
```diff
@@ -31,7 +31,217 @@
   },
   {
    "cell_type": "markdown",
-   "id": "99197479",
+   "id": "9b0e985d",
```
fdmalone: Whoops. I think you'll have to clear the cells, I didn't catch this.

PR #790, opened 2024-03-13
Improvements to `MultiControlPauli` 

tanujkhattar:
This PR makes a bunch of improvements to `MultiControlPauli`, some of which are pulled out from https://github.com/quantumlib/Qualtran/pull/765

Specifically

1. `MultiControlPauli` learns classical simulation when target is an X gate
2. A more efficient tensor contraction which doesn't default to decomposing the Bloq; which is more expensive since it uses a bunch of ancilla qubits
3. Change the signature to accept a list of qubits instead of a single n-bit register as control

### DIFF
--- a/qualtran/bloqs/mcmt/multi_control_multi_target_pauli.py
+++ b/qualtran/bloqs/mcmt/multi_control_multi_target_pauli.py
@@ -13,7 +13,7 @@
 #  limitations under the License.
 
 from functools import cached_property
-from typing import Dict, Tuple
+from typing import Any, Dict, Tuple
 
 import cirq
 import numpy as np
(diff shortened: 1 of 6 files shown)


qualtran/bloqs/mcmt/multi_control_multi_target_pauli.py
```diff
@@ -108,50 +108,84 @@ class MultiControlPauli(GateWithRegisters):
...
+        n = len(self.cvs)
+        if n <= 2:
+            pre_post_clifford = TComplexity(clifford=2 * (len(self.cvs) - sum(self.cvs)))
+            target_cost = t_complexity(self.target_gate.controlled(n))
+            return pre_post_clifford + target_cost
+        and_cost = t_complexity(MultiAnd(self.cvs))
         controlled_pauli_cost = t_complexity(self.target_gate.controlled(1))
-        and_inv_cost = t_complexity(and_gate**-1)
+        and_inv_cost = t_complexity(MultiAnd(self.cvs).adjoint())
         return and_cost + controlled_pauli_cost + and_inv_cost
```
mpharrigan, line 168: we should probably move this to `build_call_graph` and rely on the `from_call_graph` t complexity strategy since it's already structured this way.

mpharrigan (approved):
looks good, thanks

qualtran/bloqs/mcmt/multi_control_multi_target_pauli.py
tanujkhattar: done.

Issue #794, opened 2024-03-15
Update sphinx docs table-of-contents with new bloqs locations


mpharrigan:
@tanujkhattar can you update `docs/bloqs/index.rst` to reflect the new locations of all the notebooks following the bloq reorganization

mpharrigan:
I'll do this

PR #796, opened 2024-03-15
Improvements to `autogenerate-bloqs-notebooks-v2.py`

tanujkhattar:
`autogenerate-bloqs-notebooks-v2.py` got stale after the recent reorganization of Bloqs. This PR fixes bugs and adds a notebook for `Hadamard` since it was raising a warning.  

Now, running `autogenerate-bloqs-notebooks-v2.py` actually generates all the correct notebooks and does not raise an error.

### DIFF
--- a/dev_tools/qualtran_dev_tools/jupyter_autogen_v2.py
+++ b/dev_tools/qualtran_dev_tools/jupyter_autogen_v2.py
@@ -52,7 +52,7 @@ class NotebookSpecV2:
     module: ModuleType
     bloq_specs: List[BloqDocSpec]
     directory: str = field()
-    _path_stem: Optional[str] = None
+    path_stem_name: Optional[str] = None
 
     @directory.default
     def _default_directory(self) -> str:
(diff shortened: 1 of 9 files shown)


qualtran/bloqs/swap_network/swap_network.ipynb
mpharrigan, line 1: can you revert these changes? The id should generally be preserved

dev_tools/qualtran_dev_tools/jupyter_autogen_v2.py
```diff
@@ -52,17 +52,17 @@ class NotebookSpecV2:
     module: ModuleType
     bloq_specs: List[BloqDocSpec]
     directory: str = field()
-    _path_stem: Optional[str] = None
+    path_stem_name: Optional[str] = None
```
mpharrigan, line 55: why this change? attrs will strip the leading underscore to accept this as a init arg

mpharrigan:
can you also update the sphinx toc #794

tanujkhattar:
@mpharrigan I've reverted the changes `swap_network.ipynb`. Can I update the sphinx toc in a follow-up PR?

dev_tools/qualtran_dev_tools/jupyter_autogen_v2.py
tanujkhattar: That's correct, this is not needed. I've reverted the change. Thanks for flagging.

qualtran/bloqs/swap_network/swap_network.ipynb
tanujkhattar: done

mpharrigan (approved):

mpharrigan:
> Can I update the sphinx toc in a follow-up PR?
yep!

PR #798, opened 2024-03-18
Serialize all Bloqs :)

tanujkhattar:
This is a very large PR that "fixes" (almost) everything that's broken in the serialization infrastructure and adds support for serializing all Bloqs. See the newly generated Bloq report card, by clicking on the `details` tab below, that now includes serialization info.  

<details>

df_styled

</details>

8 bloq examples currently fail serialization (hardcoded in `assert_bloq_example_serialize_for_pytest` to pass tests and make sure all future bloqs are tested for serialization). The reasons are as follows - (I'll also file individual issues to track the reason for failures)
- [ ] `SelectSwapQROM` needs to be made a dataclass - we don't yet support serializing non-dataclasses but everything else works! (`thc_prep` and `prep_sparse` fail due to this)
- [ ] `ApplyGateToLthQubit` expects a callable for `nth_gate`. We should change the API so it expects something else and can be made serializable (we can't serialize arbitrary callables, so `apply_z_to_odd` fails)
- [ ] Serialization of (non-trivial) sympy expressions is broken. This leads to failure of tests for symbolic examples (`lp_resource_state_symbolic`, `modexp_symb`, `modmul_symb`). We had done custom serialization for sympy in Cirq and this would be a good follow-up project
- [ ] `modexp` and `modexp_small` raise a `ValueError: base is not invertible for the given modulus` error when tried to decompose. 

The major changes introduced by the PR are

- [ ] An expanded `RESOLVER_DICT` to map all bloq names to corresponding Bloqs. We now use their fully qualified names, along with module name, because we have repetition of Bloq names across modules; eg: `PrepareTUVSuperpositions` is in `pbc.first_quantization` and `pbc.first_quantization.projectile.`
- [ ] Updates to `bloq_report_card.py` and `qualtran/testing.py` to test serialization for all bloq_examples. 
- [ ] Make `CtrlSpec` a frozen dataclass with a custom hash function. This is mainly because we have built-in support for serializing dataclasses and it was nice to make all of this work without having to add any new protocols to the Bloq class. This change can probably be reverted if needed; but I'm in favour of making the change independently also. 
- [ ] Instead of introducing new types for serializing tuples / lists, we choose to serialize all sequences as numpy arrays. This has two implications:
  - [ ] Bloqs (mostly frozen dataclasses) that expect `Tuple` / `List` as input should now specify a `converter` so they can expect any `Sequence` as input. This is a good practice in general and helps make the deserialization work when passing in a deserialized numpy array. This was done for `PairPotential`, `QROM`, `Reflection`, `StatePreparationViaRotations`, 
  - [ ] To save / load numpy arrays we now need `allow_pickle=True`, because we wish to save/load object arrays (and not just numerical  int / float / complex arrays).  This is probably not the best solution since it hurts portability (you may not be able to load pickles created by others due to mismatch of library versions)  but maybe can be fixed as a future improvement? I'll open an issue to track this. 
- [ ] Wrapping a qualtran native Bloq in a cirq style `ControlledGate` does not play well with serialization because Cirq gates are serialized using their json serialization but Bloqs don't have a json serialization. Therefore, when we go to the cirq land (for `ControlledGate`) it's good to stay entirely in the cirq land. This was used to change decomposition of `LPRSInterimPrep` to use `cirq.Rz.controlled()` instead of `Rz.controlled()`.
- [ ] When trying to `bloq.decompose_bloq()` (eg: for `And` bloq), if the bloq is implemented using a cirq-style API of `decompose_from_registers` and uses operations which are not currently supported by the Bloq API (eg: non-gate operation like classically controlled operations), we now raise a `DecomposeNotImplementedError` instead of a `ValueError`. This helps play nicely at the call sites. 
- [ ] Updates to `bloq.proto` to include (a) ctrl_spec, (b) Register
- [ ] For many Bloqs (bloq examples), their complete decomposition was tested for the first time as part of the serialization test, and this revealed bugs. For example: 
  - [ ] `Add` bloq had a `val.num_qubits <= 1` post init check which was triggered in decomposition of one of the chemistry bloqs. I've removed this assertion.
  - [ ] `QROAMTwoRegs` black box has a signature that uses `self.data_size` but `self.data_size` is no longer a class attribute. Therefore, accessing the signature raises an error. 
  - [ ] `ApplyControlledZs` decomposes into `MultiControlPauli` but still assumes the ctrl register needs a join/split; which is no longer the case after the recent PR that changes the signature of this Bloq.


mpharrigan:
How crucial is the pickle stuff? Introducing pickle anywhere has the chance of destroying the entire point of doing our own bespoke serialization: namely interoperability (e.g. with non-python languages) and security (loading pickle files can execute arbitrary code)

tanujkhattar:
> How crucial is the pickle stuff?

It was added for convenience and will definitely be removed (we can decide whether to do it as part of this PR, or as a follow-up PR as I have mentioned in the description)

qualtran/conftest.py
```diff
@@ -75,10 +75,30 @@ def assert_equivalent_bloq_example_counts_for_pytest(bloq_ex: BloqExample):
...
+        'prep_sparse',
+        'thc_prep',
+        'modmul_symb',
+        'modexp',
+        'modexp_small',
+        'modexp_symb',
+        'apply_z_to_odd',
+        'lp_resource_state_symbolic',
+    ]:
+        return
```
mpharrigan, line 89: `pytest.xfail` might be more appropriate here

qualtran/_infra/controlled.py
```diff
@@ -29,6 +29,7 @@
     from qualtran.simulation.classical_sim import ClassicalValT
 
 
+@attrs.frozen(eq=False)
```
mpharrigan, line 32: lgtm

mpharrigan:
when you say complete decomposition -- you mean repeated decomposition as far as possible?

qualtran/serialization/bloq.py
```diff
@@ -345,8 +311,9 @@ def _bloq_to_proto(bloq: Bloq, *, bloq_to_idx: Dict[Bloq, int]) -> bloq_pb2.Bloq
     except (DecomposeTypeError, DecomposeNotImplementedError, TypeError):
         t_complexity = None
 
+    name = bloq.__module__ + "." + bloq.__class__.__qualname__
```
mpharrigan, line 314: how will we handle moving bloqs w/ backwards compatibility. You want to always serialize the latest fully-qualified name and then have multiple entries in resolverdict?

tanujkhattar:
> when you say complete decomposition -- you mean repeated decomposition as far as possible?

Yes, that's right.

mpharrigan (commented):
looks pretty good. I think I need to know what you intend for the pickle args. At one point you wanted to be strict about what types of values we allow in the args, but now you can put literally anything if you wrap it in a `[list]`.

qualtran/serialization/resolver_dict.py
```diff
@@ -0,0 +1,311 @@
...
+    "qualtran.bloqs.util_bloqs.Cast": qualtran.bloqs.util_bloqs.Cast,
+    "qualtran.bloqs.util_bloqs.Free": qualtran.bloqs.util_bloqs.Free,
+    "qualtran.bloqs.util_bloqs.Join": qualtran.bloqs.util_bloqs.Join,
+    "qualtran.bloqs.util_bloqs.Partition": qualtran.bloqs.util_bloqs.Partition,
+    "qualtran.bloqs.util_bloqs.Split": qualtran.bloqs.util_bloqs.Split,
+}
+
+
+def add_to_resolver_dict(*bloqs: Type[Bloq]):
+    RESOLVER_DICT.update({bloq.__module__ + '.' + bloq.__qualname__: bloq for bloq in bloqs})
```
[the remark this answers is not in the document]
tanujkhattar: Add a fully qualified name of the bloq to the resolver dict, instead of asking users to do this. The goal is to let users update the resolver_dict with custom types they may have defined outside of the library (eg: in a jupyter notebook, in their own project etc.) This is currently invoked from tests as well.

qualtran/protos/bloq.proto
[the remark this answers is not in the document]
tanujkhattar: Not renumbering anymore.

qualtran/testing.py
```diff
@@ -488,3 +488,50 @@ def check_equivalent_bloq_example_counts(bloq_ex: BloqExample) -> Tuple[BloqChec
...
+    try:
+        bloq_roundtrip = bloqs_from_proto(bloq_lib)[0]
+    except Exception as e:
+        raise BloqCheckException.fail('DeSerialization Failed:\n' + str(e)) from e
+
+    try:
+        assert bloq == bloq_roundtrip
+    except AssertionError as e:
+        raise BloqCheckException.fail(
+            f'Roundtrip equality failed.\n{bloq=}\n{bloq_roundtrip=}\n{bloq == bloq_roundtrip}'
```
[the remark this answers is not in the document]
tanujkhattar: Rasing from exception now and removed last line since it's always False.

qualtran/conftest.py
tanujkhattar: done

tanujkhattar:
@mpharrigan @fdmalone I've updated the code so now we save numpy arrays only for numerical types without picking (`allow_pickle=False`). The only place I was using this earlier was to save tuple of registers; I now use `registers` object for that. 

I think this PR is good to go now and we can make further improvements in follow up PRs. PTAL.

qualtran/serialization/bloq.py
tanujkhattar: That's one way. Or we can introduce a `_qualtran_namespace_` protocol, similar to Cirq's `_json_namespace_`, and use that instead of explicit fully qualified names. This line would then change to `bloq._namespace_() + "." + bloq.__class__.__qualname__`. 

Adding a new protocol to the Bloq class can be done in a follow-up PR if needed.

mpharrigan (approved):
a definite improvement!

PR #802, opened 2024-03-18
move `SU2RotationGate` to `basic_gates/`

anurudhp:


anurudhp:
@tanujkhattar one question: `cirq.PhasedXZGate` and `SU2RotationGate` differ by a global phase, i.e. $SU_2(\theta, \phi, \lambda) = e^{i(\phi+\lambda-\theta)} \text{PhasedXZ}(x=\frac{2\theta}{\pi}, z=1-\frac{\lambda+\phi}\pi, a=\frac{\lambda}\pi - 0.5)$. (this only matters when using controlled GSQP inside a larger algorithm).

Should we still map them ignoring global phase? Another idea could be to just used `PhasedXZ` in GQSP rotations, and apply a single global phase correction at the end.

qualtran/bloqs/basic_gates/su2_rotation.py
```diff
@@ -0,0 +1,107 @@
...
+    $$
+
+    References:
+        [Generalized Quantum Signal Processing](https://arxiv.org/abs/2308.01501)
+            Motlagh and Wiebe. (2023). Equation 7.
+    """
+
+    theta: float
+    phi: float
+    lambd: float
```
mpharrigan, line 53: lambda

mpharrigan:
What was the conclusion from the meeting? Were we going to remove the decomposition and leave this as an "atomic" bloq? And add its tensor. Also: re: global phase: we said that it needs to match, so if `cirq.PhasedXZ` isn't up to the task we can leave that out.

anurudhp:
@mpharrigan I dropped the PhasedXZ conversion.

About the decomposition, I am fine with either case. Perhaps its still useful to have a decomposition in terms of Pauli rotations in case someone wants to generate circuits?

qualtran/bloqs/basic_gates/su2_rotation.py
anurudhp: unfortunately as `lambda` is a python keyword, we can't use it here. I could maybe use `lambda_` instead?

mpharrigan: oh interesting. maybe make a note (comment) about this reason

anurudhp:
@tanujkhattar @mpharrigan could one of you review this PR? thanks!

I also have two questions:
- Should I make the bloq atomic?
- Is there anything specific to be done to support symbolic params?

qualtran/bloqs/basic_gates/su2_rotation_test.py
```diff
@@ -0,0 +1,69 @@
...
+    random_state = np.random.default_rng(42)
+
+    for _ in range(20):
+        theta, phi, lambd, global_shift = random_state.random(size=4) * 2 * np.pi
+        gate = SU2RotationGate(theta, phi, lambd, global_shift)
+
+        np.testing.assert_allclose(gate.tensor_contract(), gate.rotation_matrix)
+
+
+def test_su2_rotation_gates(bloq_autotester):
```
tanujkhattar, line 43: Let's also add a `test_assert_decompose_is_consistent_with_t_complexity`  and call `assert_decompose_is_consistent_with_t_complexity`

qualtran/bloqs/basic_gates/su2_rotation.py
```diff
@@ -0,0 +1,158 @@
...
+    def pretty_name(self) -> str:
+        return 'SU_2'
+
+    def wire_symbol(self, soq: 'Soquet') -> 'WireSymbol':
+        return TextBox(
+            f"{self.pretty_name()}({self.theta}, {self.phi}, {self.lambd}, {self.global_shift})"
+        )
+
+    def _t_complexity_(self) -> TComplexity:
+        return TComplexity(rotations=1)
```
tanujkhattar, line 133: Let's change the T-complexity to 3 rotations instead.

qualtran/bloqs/basic_gates/su2_rotation.py
```diff
@@ -0,0 +1,158 @@
...
+            )
+        )
+
+    def build_composite_bloq(self, bb: 'BloqBuilder', q: 'SoquetT') -> Dict[str, 'SoquetT']:
+        q = bb.add(ZPowGate(exponent=2, global_shift=0.5 + self.global_shift / (2 * np.pi)), q=q)
+        q = bb.add(ZPowGate(exponent=1 - self.lambd / np.pi, global_shift=-1), q=q)
+        q = bb.add(Ry(angle=2 * self.theta), q=q)
+        q = bb.add(ZPowGate(exponent=-self.phi / np.pi, global_shift=-1), q=q)
+        return {'q': q}
+
```
tanujkhattar, line 123: Can we change the decomposition to use 3 rotations instead also return `TComplexity(rotations=3)` ? 


We currently use a cost of `ceil(1.149 * log2(1.0 / eps) + 9.2)` for synthesizing 1 rotation (used in `TComplexity`). This formula is derived from https://arxiv.org/pdf/1404.5320.pdf 

In the paper above, they show that any 2qubit rotation can written using 3 Z rotations 

image


and then they give an algorithm to synthesize Z rotations using clifford + T gates 

image


with the average cost scaling as `ceil(1.149 * log2(1.0 / eps) + 9.2)`

image

qualtran/bloqs/basic_gates/su2_rotation.py
```diff
@@ -0,0 +1,158 @@
...
+                tags=[self.short_name(), tag],
+            )
+        )
+
+    def build_composite_bloq(self, bb: 'BloqBuilder', q: 'SoquetT') -> Dict[str, 'SoquetT']:
+        q = bb.add(ZPowGate(exponent=2, global_shift=0.5 + self.global_shift / (2 * np.pi)), q=q)
+        q = bb.add(ZPowGate(exponent=1 - self.lambd / np.pi, global_shift=-1), q=q)
+        q = bb.add(Ry(angle=2 * self.theta), q=q)
+        q = bb.add(ZPowGate(exponent=-self.phi / np.pi, global_shift=-1), q=q)
+        return {'q': q}
```
tanujkhattar, line 122: Since the signature of this gate does not depend upon the symbolic parameters (`theta`, `phi`, `lambd`, `global_shift`); the bloq should work for symbolics automatically. 

Often, it's helpful to also override the `build_call_graph` method so you can take care of propagating symbolics correctly there and not worry about symbolic parameters in the `build_composite_bloq` because symbolic analysis would usually be carried out via the call graph; especially when signature depends on symbolic parameters. 

One slight improvement we can make here is that when the bloq is parameterized (i.e. any of `theta`, `phi`, `lambd`, `global_shift` is symbolic); we can use `sympy.pi` instead of `np.pi` in the decomposition. 

You can do a `pi = sympy.pi if self._is_parameterized_() else np.pi` and then use `pi` in the decomposition formulas.

tanujkhattar (changes requested):
Answered all your questions and requested a few changes. Looking great overall. 


> Should I make the bloq atomic?

No

> Is there anything specific to be done to support symbolic params?

Suggested minor improvements

qualtran/bloqs/basic_gates/su2_rotation.py
```diff
@@ -0,0 +1,158 @@
...
+
+    References:
+        [Generalized Quantum Signal Processing](https://arxiv.org/abs/2308.01501)
+        Motlagh and Wiebe. (2023). Equation 7.
+    """
+
+    theta: float
+    phi: float
+    lambd: float  # cannot use `lambda` as it is a python keyword
+    global_shift: float = 0
```
tanujkhattar, line 63: Let's accept a parameter `eps` for precision and pass `eps / 3` to the 3 rotation subbloqs in the decomposition.

qualtran/bloqs/basic_gates/su2_rotation.py
```diff
@@ -0,0 +1,158 @@
+#  Copyright 2023 Google LLC
```
tanujkhattar, line 1: Also it's 2024 now, let's update the copyright notice everywhere :)

qualtran/bloqs/basic_gates/su2_rotation_test.py
anurudhp: added

qualtran/bloqs/basic_gates/su2_rotation.py
anurudhp: I added `sympy.pi` support as you suggested. Now we don't need to override `build_call_graph` anymore right?

qualtran/bloqs/basic_gates/su2_rotation.py
tanujkhattar: You can return a `CirqGateAsBloq(cirq.GlobalPhaseGate())` instead to make it explicit. We should eventually add a native global phase Bloq to Qualtran but this should work in the meantime

anurudhp: Added in https://github.com/quantumlib/Qualtran/pull/828

qualtran/bloqs/basic_gates/su2_rotation.py
tanujkhattar: Yes, that's correct.

qualtran/bloqs/basic_gates/su2_rotation.py
anurudhp: The `GlobalPhase` op would also need an `eps` right? Although this matters only when controlled. I currently used `eps/4` for the four ops.

qualtran/cirq_interop/t_complexity_protocol.py
```diff
@@ -94,6 +94,11 @@ def _is_clifford_or_t(stc: Any, fail_quietly: bool) -> Optional[TComplexity]:
     if not isinstance(stc, (cirq.Gate, cirq.Operation)):
         return None
 
+    if isinstance(stc, cirq.GlobalPhaseGate) or (
+        isinstance(stc, cirq.Operation) and isinstance(stc.gate, cirq.GlobalPhaseGate)
+    ):
+        return TComplexity()
```
tanujkhattar, line 100: I think ideally this shouldn't be required but the T-complexity protocol is being updated in PR https://github.com/quantumlib/Qualtran/pull/825 so this hack can be removed there. 

We should update the `CIRQ_TYPE_TO_BLOQ_MAP` in `_cirq_gate_to_bloq` to return `GlobalPhase` bloq when it sees a `cirq.GlobalPhaseGate`. That would be in-line with other substitutions that we do for basic gates.

tanujkhattar (approved):
LGTM % one final suggestion to update the `_cirq_gate_to_bloq` to return `GlobalPhase` bloq when it sees a `cirq.GlobalPhaseGate`. 

Feel free to merge after that.

qualtran/cirq_interop/t_complexity_protocol.py
anurudhp: `_cirq_gate_to_bloq` already converts `GlobalPhase` (merged with https://github.com/quantumlib/Qualtran/pull/828)
https://github.com/quantumlib/Qualtran/blob/4dffc97de898b92509934c593e6ffb9b77b9337d/qualtran/cirq_interop/_cirq_to_bloq.py#L347-L348

The only neccessary hack here is `isinstance(stc, cirq.Operation) and isinstance(stc.gate, cirq.GlobalPhaseGate)` because cirq decompose seems to convert it to a `global_phase_operation` instead of a gate. I updated the code now

anurudhp:
`_cirq_gate_to_bloq` already converts to `GlobalPhase` (merged with https://github.com/quantumlib/Qualtran/pull/828)
https://github.com/quantumlib/Qualtran/blob/4dffc97de898b92509934c593e6ffb9b77b9337d/qualtran/cirq_interop/_cirq_to_bloq.py#L347-L348

PR #806, opened 2024-03-20
Remove cirq_ft_jupyter_autogen_fatories.py and update stale notebooks.

fdmalone:


dev_tools/qualtran_dev_tools/cirq_ft_jupyter_autogen_factories.py
```diff
@@ -1,83 +0,0 @@
...
-    from qualtran.bloqs.multiplexers.select_pauli_lcu import SelectPauliLCU
-
-    target_bitsize = 4
-    us = ['XIXI', 'YIYI', 'ZZZZ', 'ZXYZ']
-    us = [cirq.DensePauliString(u) for u in us]
-    selection_bitsize = int(np.ceil(np.log2(len(us))))
-    return SelectPauliLCU(selection_bitsize, target_bitsize, select_unitaries=us)
-
-
-def _make_StatePreparationAliasSampling():
```
tanujkhattar, line 68: Can we add bloq examples for all these classes as well? The construction code can simply be copied over from here.

fdmalone: Done. I've updated the title to reflect that adding these and updating the notebook spec revealed a bunch of stale notebooks which I've added. The diff is pretty straightforward though..

fdmalone:
@tanujkhattar PTAL

mpharrigan:
wow!

fdmalone:
> wow!

Surprised it still exists?

mpharrigan:
end of an era :)

mpharrigan (approved):

Issue #722, opened 2024-02-29
[data types] bb./splits/joins/... should preserve dtypes. 


fdmalone:
Data types for splits and joins should be added to decompositions.

Following #707 Bloqs using bb.split/join/allocate should be updated to use appropriate types.

``` python
from qualtran.bloqs.swap_network import SwapWithZero
from qualtran.drawing import show_bloq
bloq = SwapWithZero(3, 3, 4)
show_bloq(bloq.decompose_bloq(), type='dtype')
```
 
![Screenshot 2024-02-28 at 4 52 15 PM](https://github.com/quantumlib/Qualtran/assets/12097876/4a1bcede-4539-48c1-ab4c-ae87aa1fcad7)

PR #808, opened 2024-03-21
QDTypes learn `to_bits` and `from_bits`

tanujkhattar:
This is useful to support classical and tensor simulations and support conversion of control values to Cirq. I'll demonstrate usage in follow-up PRs. 

cc @fdmalone

### DIFF
--- a/qualtran/_infra/data_types.py
+++ b/qualtran/_infra/data_types.py
@@ -54,6 +54,7 @@ from typing import Any, Iterable, Union
 import attrs
 import numpy as np
 import sympy
+from fxpmath import Fxp
 from numpy.typing import NDArray
 
 
@@ -69,6 +70,14 @@ class QDType(metaclass=abc.ABCMeta):
     def get_classical_domain(self) -> Iterable[Any]:
         """Yields all possible classical (computational basis state) values representable
         by this type."""
+
+    @abc.abstractmethod
+    def to_bits(self, x) -> Iterable[int]:
+        """Yields individual bits corresponding to binary representation of x"""
+
+    @abc.abstractmethod
+    def from_bits(self, *bits: int):
+        """Combine individual bits to form x"""
 
     @abc.abstractmethod
     def assert_valid_classical_val(self, val: Any, debug_str: str = 'val'):
@@ -110,6 +119,16 @@ class QBit(QDType):
         if not (val == 0 or val == 1):
             raise ValueError(f"Bad {self} value {val} in {debug_str}")
 
+    def to_bits(self, x) -> Iterable[int]:
+        """Yields individual bits corresponding to binary representation of x"""
+        self.assert_valid_classical_val(x)
+        return (int(x),)
+
+    def from_bits(self, *bits: int) -> int:
+        """Combine individual bits to form x"""
+        assert len(bits) == 1
+        return bits[0]
+
     def assert_valid_classical_val_array(self, val_array: NDArray[int], debug_str: str = 'val'):
         if not np.all((val_array == 0) | (val_array == 1)):
             raise ValueError(f"Bad {self} value array in {debug_str}")
@@ -127,6 +146,14 @@ class QAny(QDType):
 
     def get_classical_domain(self) -> Iterable[Any]:
         raise TypeError(f"Ambiguous domain for {self}. Please use a more specific type.")
+
+    def to_bits(self, x) -> Iterable[Any]:
+        # TODO: Raise an error once usage of `QAny` is minimized across the library
+        return QUInt(self.bitsize).to_bits(x)
+
+    def from_bits(self, *bits: int) -> int:
+        # TODO: Raise an error once usage of `QAny` is minimized across the library
+        return QUInt(self.bitsize).from_bits(*bits)
 
     def assert_valid_classical_val(self, val, debug_str: str = 'val'):
         pass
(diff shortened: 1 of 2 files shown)


qualtran/_infra/data_types.py
```diff
@@ -190,6 +232,24 @@ def __attrs_post_init__(self):
     def num_qubits(self):
         return self.bitsize
 
+    def to_bits(self, x: int) -> Iterable[int]:
+        """Yields individual bits corresponding to binary representation of x"""
+        from qualtran.cirq_interop.bit_tools import iter_bits
```
fdmalone, line 237: We should probably move this to somewhere outside of cirq_interop. There should probably be some consolidation with https://github.com/quantumlib/Qualtran/blob/main/qualtran/simulation/classical_sim.py#L40

qualtran/_infra/data_types.py
```diff
@@ -354,6 +439,22 @@ def num_int(self) -> Union[int, sympy.Expr]:
     def fxp_dtype_str(self) -> str:
         return f'fxp-{"us"[self.signed]}{self.bitsize}/{self.num_frac}'
 
+    @property
+    def _fxp_dtype(self) -> Fxp:
+        return Fxp(None, dtype=self.fxp_dtype_str)
+
+    def to_bits(self, x: Union[float, Fxp]) -> Iterable[int]:
+        """Yields individual bits corresponding to binary representation of x"""
+        self.assert_valid_classical_val(x)
```
fdmalone, line 448: This needs to be implemented still. Is it worth doing now?

qualtran/_infra/data_types_test.py
```diff
@@ -207,3 +207,62 @@ def test_single_qubit_consistency():
...
+    with pytest.raises(ValueError):
+        QInt(4).to_bits(10)
+
+    # QUInt
+    quint4 = QUInt(4)
+    assert list(quint4.to_bits(10)) == [1, 0, 1, 0]
+    assert quint4.from_bits(*quint4.to_bits(10)) == 10
+    for x in range(16):
+        assert quint4.from_bits(*quint4.to_bits(x)) == x
+    with pytest.raises(ValueError):
```
fdmalone, line 230: Test raises for negative value.

qualtran/_infra/data_types.py
fdmalone: I think it would potentially change the logic in the next line. For example, if isinstance(x, Fxp) should we expect x to have the same _fxp_dtype (so no truncating occurs). For the float case I'm not really sure how one should handle it, perhaps we should only accept an Fxp float as input and not attempt to approximate it?

qualtran/_infra/data_types.py
```diff
@@ -70,6 +71,14 @@ def get_classical_domain(self) -> Iterable[Any]:
         """Yields all possible classical (computational basis state) values representable
         by this type."""
 
+    @abc.abstractmethod
+    def to_bits(self, x) -> Iterable[int]:
+        """Yields individual bits corresponding to binary representation of x"""
+
+    @abc.abstractmethod
+    def from_bits(self, *bits: int):
```
fdmalone, line 79: It seems a little more symmetric to me to have `bits: Iterable[int]` everywhere, to match the return type of to_bits.

qualtran/_infra/data_types.py
fdmalone: Although, doing the "casting" from float to Fxp is very convenient...

fdmalone (changes requested):
Very nice. Some questions / comments. I think we need to consolidate bit_iter tools (although this could be a follow up). The qualtran tools do restrict integers to be less than 64 bits, I'm not sure how fundamental that is.

qualtran/_infra/data_types.py
tanujkhattar: Having `bits: Iterable[int]` is annoying because then I won't be able to subscript into `bits`. I can change it to `bits: Sequence[int]` instead and return a `List[int]` instead of `Iterable[int]` from `to_bits` method so the return value is compatible with `from_bits`.

qualtran/_infra/data_types.py
tanujkhattar: I've removed dependence on `cirq_interop` for now so that all bit manipulation code is consolidated in this file. Let's do the consolidation in follow-up PRs?

qualtran/_infra/data_types.py
tanujkhattar: done

qualtran/_infra/data_types.py
tanujkhattar: I've implemented the validation logic for `QIntOnesComp` and `QFxp`. However, implementing `assert_valid_classical_val` for `QFxp` opens a new can of worms because `Fxp` is currently not propagated correctly across the Bloqs; and is basically treated as a `QUint` in the classical simulation infrastructure; because of which the assertions fail in multiple bloq tests.

I don't want to open this can of worms as part of this PR. I have local updates to bloqs like `Join`, `Split` etc; all of which currently completely ignore `Fxp` and treat all classical data as unsigned integers; but we can do those updates as a follow-up PR

tanujkhattar:
@fdmalone I've addressed all your comments, PTAL!

qualtran/_infra/data_types.py
fdmalone: I opened #811 to track

qualtran/_infra/data_types.py
fdmalone: That makes sense. Is this because of #722? I can prioritize if it's the case.

fdmalone:
I like this.

fdmalone (approved):

PR #817, opened 2024-03-22
Propagate split dtype to join

### LINKED ISSUE #722: [data types] bb./splits/joins/... should preserve dtypes.

fdmalone:
Fixes #722:

Adds back types to joins to ensure no loss of information when a split is followed by a join. I added a severity level to how strictly we may want to enforce type checking:
1. Strict -> dtype_a == dtype_b
2. Any -> QAny(bitsize) into other types is ok.
3. Loose (current): 1 and 2 + some [unpleasant integer / fxp casting](https://github.com/quantumlib/Qualtran/blob/main/qualtran/_infra/data_types.py#L584) necessary for phasing via cost function (there's a uint x register and a QFxp register wiring into each other.)

I added an autotester to colorize the check to show how strictly enforced our type checking currently is, and updated the cirq interop to handle the splits and joins there

- [x] Fix obvious examples.
- [x] Add autotester to track any finalization type mismatches.
- [x] cirq to bloq spin / joins


fdmalone:
Pass == dtype_ is preserved the whole way
Unverified ~ Wiring QAny into A typed register OR casting (Should I add some more colors / categorization @mpharrigan ). It would be nice to have QAny into something colored nicer than Casting
Fail == type checking failed which shouldn't occur since we do this at runtime during the decomposition for egregious type checking failures. 

Screenshot 2024-03-22 at 3 43 11 PM

fdmalone:
image

fdmalone:
^^ I'm not actually sure the cirq-interop splits and joins are correct looking and the picture.

tanujkhattar:
> I'm not actually sure the cirq-interop splits and joins are correct looking and the picture.


@fdmalone  What are your concerns? One concern I see is that the `QBit()` wire going into the `x` register of `pg += x >> 3` should be a `QFxp(1, 1)` which has a different interpretation than `QBit()`. 

I think we default to using `QBit()` for single bit values but I'm not sure if we should do that always. Definitely not do that here.

fdmalone:
Maybe it's unrelated, I was concerned by the integer-like QFxp(4, 0) splitting into a QFxp(2,2), but I think this fine.

fdmalone:
@tanujkhattar PTAL
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="4885pt" height="542pt" viewBox="0.00 0.00 4885.00 542.00">

fdmalone:
Safer casting of single qubits into QFxp(1,1) and back to QBit:

Screenshot 2024-04-05 at 5 11 04 PM

qualtran/cirq_interop/_cirq_to_bloq.py
```diff
@@ -230,7 +248,10 @@ def _ensure_in_reg_exists(
     all_mapped_qubits = {q for qreg in qreg_to_qvar for q in qreg.qubits}
     qubits_to_allocate: List[cirq.Qid] = [q for q in in_reg.qubits if q not in all_mapped_qubits]
     if qubits_to_allocate:
-        qreg_to_qvar[_QReg(qubits_to_allocate)] = bb.allocate(len(qubits_to_allocate))
+        n_alloc = len(qubits_to_allocate)
+        qreg_to_qvar[
+            _QReg(qubits_to_allocate, dtype=QBit() if n_alloc == 1 else QAny(n_alloc))
+        ] = bb.allocate(len(qubits_to_allocate))
```
tanujkhattar, line 254:
```suggestion
        ] = bb.allocate(n_alloc)
```

qualtran/cirq_interop/_cirq_to_bloq.py
```diff
@@ -252,11 +275,27 @@ def _ensure_in_reg_exists(
     for qreg, soq in new_qreg_to_qvar.items():
         if len(in_reg_qubits) > 1 and qreg.qubits and qreg.qubits[0] in in_reg_qubits:
             assert len(qreg.qubits) == 1, "Individual qubits should have been split by now."
-            soqs_to_join[qreg.qubits[0]] = soq
+            # Need to split single-qubit Fxp registers which are ambiguous,
+            # these will be appropriately joined later.
+            if isinstance(qreg.dtype, QFxp):
+                soqs_to_join[qreg.qubits[0]] = bb.split(soq=soq)[0]
```
tanujkhattar, line 281: Should we do a cast here instead? Why do we need to do this only for `QFxp` ? The input register of `Join` expects an input register of type `QBit()` so I imagine `soqs_to_join[qreg.qubits[0]] = soq` should correspond to a soquet of type `QBit` for every soquet that needs to be joined but we need to insert a special case only for `QFxp` ?

qualtran/testing.py
```diff
@@ -535,3 +549,51 @@ def check_bloq_example_serialize(bloq_ex: BloqExample) -> Tuple[BloqCheckResult,
         return BloqCheckResult.ERROR, f'{bloq_ex.name}: {e}'
 
     return BloqCheckResult.PASS, ''
+
+
+def assert_bloq_example_preserves_types(bloq_ex: BloqExample) -> Tuple[BloqCheckResult, str]:
+    """ """
```
tanujkhattar, line 555: Please write the docstring

tanujkhattar (changes requested):
Left a round of comments. Looking good overall. Main question is regarding the special casing for `QFxp` in `_ensure_in_reg_exists`

qualtran/bloqs/swap_network/swap_with_zero_test.py
```diff
@@ -32,12 +32,6 @@
 random.seed(12345)
 
 
-def _make_SwapWithZero():
-    from qualtran.bloqs.swap_network import SwapWithZero
-
-    return SwapWithZero(selection_bitsize=3, target_bitsize=64, n_target_registers=5)
```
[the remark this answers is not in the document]
fdmalone: right.

qualtran/cirq_interop/_cirq_to_bloq.py
fdmalone: Yeah you're right. Cast is much better.

qualtran/cirq_interop/_cirq_to_bloq.py
```diff
@@ -275,17 +275,15 @@ def _ensure_in_reg_exists(
...
             else:
                 soqs_to_join[qreg.qubits[0]] = soq
         elif len(in_reg_qubits) == 1 and qreg.qubits and qreg.qubits[0] in in_reg_qubits:
-            # Need to split single-qubit Fxp registers which are ambiguous,
-            # these will be appropriately joined later.
-            if isinstance(qreg.dtype, QFxp):
-                soqs_to_join[qreg.qubits[0]] = bb.split(soq=soq)[0]
+            # Cast single QBit registers to the appropriate single-bit register dtype.
+            if not isinstance(in_reg.dtype, QBit):
+                qreg_to_qvar[in_reg] = bb.add(Cast(QBit(), in_reg.dtype), reg=soq)
```
tanujkhattar, line 286: Do we need an assertion here? Is guaranteed that `soq` will always be of type `QBit` ?

tanujkhattar (approved):
LGTM % one final question

qualtran/cirq_interop/_cirq_to_bloq.py
fdmalone: I added an assertion.

qualtran/cirq_interop/_cirq_to_bloq.py
```diff
@@ -282,10 +282,12 @@ def _ensure_in_reg_exists(
...
             # Cast single QBit registers to the appropriate single-bit register dtype.
-            assert isinstance(
-                soq.reg.dtype, QBit
-            ), f"Found non-QBit type register which shouldn't happen: {soq.reg.name} {soq.reg.dtype}"
             if not isinstance(in_reg.dtype, QBit):
+                err_msg = (
+                    "Found non-QBit type register which shouldn't happen: "
+                    f"{soq.reg.name} {soq.reg.dtype}"
+                )
+                assert isinstance(soq.reg.dtype, QBit), err_msg
```
tanujkhattar: The assertion probably does belon above since in both the if / else cases we assume that `soq.reg.dtype` is of type `QBit` ?

fdmalone: Reverted

tanujkhattar (approved):

PR #820, opened 2024-03-22
Replace string references with a dataclass

NoureldinYosri:
We had a discussion before about the `reference` attribute should be like, at the time we decided to keep as just a string. I think we should move it being a dataclass.

I also updated the `surface_code.__init__` to expose the primatives defined in the module and the README.md file to explain how to invoke the resource estimation UI.


README.md
```diff
@@ -32,3 +32,9 @@ You can also install the latest tagged release using `pip`:
 You can also install the latest state of the main branch:
 
     pip install git+https://github.com/quantumlib/Qualtran
+
+## Resource Estimation UI
+Qualtran provides a UI for estimating the resources (qubits, magic states, runtime, ..etc) required to run a quantum algorithm. The UI can be run locally by running:
```
mpharrigan, line 37: GUI? 

resources -> physical resources

..etc -> etc.

qualtran/surface_code/fifteen_to_one.py
```diff
@@ -378,6 +380,10 @@ def _build_factory(
     return factory
 
 
-FifteenToOne733 = FifteenToOne(7, 3, 3, reference='https://arxiv.org/abs/1905.06903')
+FifteenToOne733 = FifteenToOne(
+    7, 3, 3, reference=Reference(source='https://arxiv.org/abs/1905.06903')
```
mpharrigan, line 384: extreme nit: you can probably get rid of the `reference=` kwarg now since it's obvious from the (new) class name

qualtran/surface_code/reference.py
```diff
@@ -0,0 +1,45 @@
...
+
+from attrs import frozen
+
+
+@frozen
+class Reference:
+    """A reference to a source material.
+
+    Attributes:
+        source: A link to a paper/article (e.g. https://link-to-paper).
```
mpharrigan, line 25: should we be more specific with this attribute? if it's supposed to be a link then `url` is likely more appropriate

mpharrigan (commented):
some nits

README.md
NoureldinYosri: done

qualtran/surface_code/fifteen_to_one.py
NoureldinYosri: not really, there is another parameter before reference

qualtran/surface_code/reference.py
NoureldinYosri: done

NoureldinYosri:
@mpharrigan ptal

mpharrigan (approved):

PR #824, opened 2024-03-25
using controlled phase grad

skushnir123:
Small PR to now use a controlled version of AddIntoPhaseGrad


qualtran/bloqs/rotations/phase_gradient.py
```diff
@@ -183,21 +197,28 @@ def scaled_val(self, x: int) -> int:
         x_fxp = _fxp(x / 2**x_width, x_width).like(_fxp(0, self.phase_bitsize)).astype(float)
         return int(x_fxp.astype(float) * 2**self.phase_bitsize)
 
-    def apply(self, x: int, phase_grad: int) -> Union[int, Iterable[int]]:
-        out = self.on_classical_vals(x=x, phase_grad=phase_grad)
-        return out['x'], out['phase_grad']
+    def apply(self, ctrl: int, x: int, phase_grad: int) -> Union[int, Iterable[int]]:
```
tanujkhattar, line 200: You'll have to change this to `*args` because `apply` will be called with only 2 parameters with `self.control` is False.

qualtran/bloqs/rotations/phase_gradient.py
```diff
@@ -150,27 +152,39 @@ class AddIntoPhaseGrad(GateWithRegisters, cirq.ArithmeticGate):
         [Compilation of Fault-Tolerant Quantum Heuristics for Combinatorial Optimization](https://arxiv.org/abs/2007.07391)
         Appendix A: Addition for controlled rotations
     """
+
     x_bitsize: int
     phase_bitsize: int
     right_shift: int = 0
     sign: int = +1
+    controlled: bool = False
```
tanujkhattar, line 160: We've used `controlled: Optional[bool] = None` in the past to denote 3 cases
a) None -  no controls
b) 1 - Controlled with value 1
c) 0 - Controlled with value 0

qualtran/bloqs/rotations/phase_gradient.py
```diff
@@ -183,21 +197,28 @@ def scaled_val(self, x: int) -> int:
...
-        out = self.on_classical_vals(x=x, phase_grad=phase_grad)
-        return out['x'], out['phase_grad']
+    def apply(self, ctrl: int, x: int, phase_grad: int) -> Union[int, Iterable[int]]:
+        out = self.on_classical_vals(ctrl=ctrl, x=x, phase_grad=phase_grad)
+        return out['ctrl'], out['x'], out['phase_grad']
 
-    def on_classical_vals(self, x: int, phase_grad: int) -> Dict[str, 'ClassicalValT']:
-        phase_grad_out = (phase_grad + self.sign * self.scaled_val(x)) % (2**self.phase_bitsize)
-        return {'x': x, 'phase_grad': phase_grad_out}
+    def on_classical_vals(self, ctrl: int, x: int, phase_grad: int) -> Dict[str, 'ClassicalValT']:
```
tanujkhattar, line 204: ditto

qualtran/bloqs/rotations/phase_gradient.py
```diff
@@ -183,21 +197,28 @@ def scaled_val(self, x: int) -> int:
...
+        )
+        return {'ctrl': ctrl, 'x': x, 'phase_grad': phase_grad_out}
 
     def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
-        num_toffoli = self.phase_bitsize - 2
-        return {(Toffoli(), num_toffoli)}
+        if self.controlled == False:
+            num_toffoli = self.phase_bitsize - 2
+            return {(Toffoli(), num_toffoli)}
+        return {(TGate(), 8 * self.phase_bitsize)}
```
tanujkhattar, line 214: If `num_toffoli = self.phase_bitsize-2`, the call graph should decompose into `num_toffoli` $\text{And}$ and $\text{And}^\dagger$'s for the non-controlled case and `2 * num_toffoli` of each for the controlled case. 


The fact that we need twice the number of $\text{And}$ / $\text{And}^\dagger$ comes from https://arxiv.org/pdf/1709.06648.pdf (see primitives for non-controlled additions and controlled additions below)


image


image


And the fact that we need `self.phase_bitsize - 2` instead of `self.phase_bitsize - 1` comes from the optimization described in Fig 13 of https://arxiv.org/pdf/2007.07391.pdf where they get rid of the last carry bit utilizing the fact that target register is a phase gradient register

qualtran/bloqs/rotations/phase_gradient.py
```diff
@@ -50,6 +51,7 @@ class PhaseGradientUnitary(GateWithRegisters):
 
     The implementation simply decomposes into $n$ (controlled-) rotations, one on each qubit.
     """
+
```
tanujkhattar, line 54: Revert unrelated change

qualtran/bloqs/rotations/phase_gradient.py
```diff
@@ -150,27 +152,39 @@ class AddIntoPhaseGrad(GateWithRegisters, cirq.ArithmeticGate):
         [Compilation of Fault-Tolerant Quantum Heuristics for Combinatorial Optimization](https://arxiv.org/abs/2007.07391)
         Appendix A: Addition for controlled rotations
     """
+
     x_bitsize: int
     phase_bitsize: int
     right_shift: int = 0
     sign: int = +1
+    controlled: Optional[bool] = None
```
tanujkhattar, line 160: Please update the docstring to include the `controlled` arg in the `Args` and `Registers` section. Also change to `controlled: Optional[int] = None` s.t. `controlled=0` means control should be active when ctrl register stores value `0`. I think right now, your implementation below works only for the case when control register stores value `1` (i.e. controlled=True)

qualtran/bloqs/rotations/phase_gradient.py
```diff
@@ -183,16 +197,33 @@ def scaled_val(self, x: int) -> int:
...
         return int(x_fxp.astype(float) * 2**self.phase_bitsize)
 
-    def apply(self, x: int, phase_grad: int) -> Union[int, Iterable[int]]:
+    def apply(self, *args) -> Union[int, Iterable[int]]:
+        if self.controlled:
+            ctrl, x, phase_grad = args
+            out = self.on_classical_vals(ctrl=ctrl, x=x, phase_grad=phase_grad)
+            return out['ctrl'], out['x'], out['phase_grad']
+
+        x, phase_grad = args
```
tanujkhattar, line 206: For all the changes made to `AddIntoPhaseGrad`, you should add tests to `phase_gradient_test.py` to verify the correctness of these changes.

qualtran/bloqs/rotations/phase_gradient.py
```diff
@@ -183,16 +197,33 @@ def scaled_val(self, x: int) -> int:
...
+            )
+            return {'ctrl': ctrl, 'x': x, 'phase_grad': phase_grad_out}
+
         phase_grad_out = (phase_grad + self.sign * self.scaled_val(x)) % (2**self.phase_bitsize)
         return {'x': x, 'phase_grad': phase_grad_out}
 
     def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
         num_toffoli = self.phase_bitsize - 2
+        if self.controlled:
+            return {(TGate(), 2 * num_toffoli)}
```
tanujkhattar, line 225:
```suggestion
            return {(Toffoli(), 2 * num_toffoli)}
```

qualtran/bloqs/qft/approximate_qft_test.py
```diff
@@ -70,7 +69,6 @@ def test_approximate_qft_exact(n: int, without_reverse: bool):
     qft_bloq = TestApproximateQFT(n, n, not without_reverse)
     qft_cirq = cirq.QuantumFourierTransformGate(n, without_reverse=without_reverse)
     np.testing.assert_allclose(cirq.unitary(qft_bloq), cirq.unitary(qft_cirq))
-    np.testing.assert_allclose(cirq.unitary(qft_bloq**-1), cirq.unitary(qft_cirq**-1))
```
tanujkhattar, line 73: Why was this removed? Does this assertion not pass anymore?

tanujkhattar (changes requested):
Please add more tests

qualtran/bloqs/qft/approximate_qft_test.py
skushnir123: added back after adding inverse implementation for AddIntoPhaseGrad

qualtran/bloqs/rotations/phase_gradient.py
skushnir123: I think this was the automatic formatter

skushnir123: Should I still revert?

tanujkhattar: nope, that's fine.

PR #825, opened 2024-03-26
Maintenance of "strategies" in the T complexity protocol

mpharrigan:
This PR cleans up some of the T complexity protocol code and associated tests. The ultimate goal is to reduce dependency on explicit overrides of the `_t_complexity_()` method and have most of the logic in the protocol strategies. 

The changes in this PR:

 - remove unused/unrespected `fail_quietly` argument from `t_complexity`. In general, this wasn't used. We'll stick with the normal Python way of using exceptions. You can tell why exceptions were invented because this flag needed to be plumbed through many functions.
 - The Cirq t-complexity equivalence checker compares an explicit annotation with t complexity derived from a decomposition. I changed this to explicitly start from the `_is_iterable`. Given the order of strategies, it could very easily be decomposing the decomposition a second time. 
 - The Cirq t-complexity equivalence checker compares an explicit annotation with t complexity derived from a decomposition. `Bloq.t_complexity()` is not an independent source of t complexity annotation, as it is no longer overridden anywhere and just calls `t_complexity()` which has many strategies available to it. I changed this function to only look for `_t_complexity_` to avoid a false sense of security in our tests. 
 - This revealed some not-really-testing-anything tests. I switched to `qlt_testing.assert_equivalent_bloq_counts`, which is a small shim over the one for testing `BloqExample`s. 
 - The strategies are executed in order and short-circuit return. There were bugs where some of the later strategies would cause an exception but this never showed up due to the short-circuit return. I fixed these. 
 - I re-named all the strategy private functions to follow the `_from_xxx`. This makes more sense. 

In a small number of cases (addition, modular addition) -- I removed the `_t_complexity_` method from the bloq where it was redundant with the `build_call_graph` annotation. If the code wasn't completely trivial, I moved it to the test file and added a check to the unit tests. 

In future work, I'd like to continue to inspect the `_t_complexity_()` methods and see which ones are needed and which can be handled by the two main strategies: `_from_build_call_graph` or `_from_directly_countable`.


tanujkhattar:
One high level comment - One of the main reasons to have the logical count annotations (i.e. `_t_complexity_` or equivalent) even when decomposition / call graph is present was to keep things performant. We used to write tests to verify that decomposition is consistent with `_t_complexity_` annotations and then use the annotated formulas when costing out large circuits. While replacing it with `build_call_graph`; we should keep an eye on performance for large circuits and make sure it doesn't introduce any regression.

mpharrigan:
Agreed on performance. I have a script that runs t_complexity on all the bloq examples and times them (_and_ I clear the cache between runs). I saw no regression mucking with the strategies to do all three for each bloq example. For this PR, I'm not going to change the order of strategies. I'll update the description with what it actually does.

mpharrigan:
I went through and computed the t complexity of all bloq examples in the library on `main` and again on this branch. I asserted that all the values are equivalent. These are the bloq examples where the duration increased by more than 1e-3 seconds and the amount of extra duration

```
add_k_symb 	 0.00269042s
black_box_block_bloq 	 0.00229562s
chebyshev_poly 	 0.00195442s
prep_power_two_proj 	 0.0662807s
pair_potential 	 0.00469533s
potential_energy 	 0.00294783s
state_prep_via_rotation 	 0.001054s
```

mpharrigan:
@tanujkhattar can you take a look?

qualtran/cirq_interop/t_complexity_protocol.py
```diff
@@ -170,65 +178,55 @@ def _get_hash(val: Any, fail_quietly: bool = False):
...
         if ret is not None:
             break
     return ret
 
 
 @cachetools.cached(cachetools.LRUCache(128), key=_get_hash, info=True)
 def _t_complexity_for_gate_or_op(
-    gate_or_op: Union[cirq.Gate, cirq.Operation, Bloq], fail_quietly: bool
+    gate_or_op: Union[cirq.Gate, cirq.Operation, Bloq]
 ) -> Optional[TComplexity]:
```
tanujkhattar, line 194: This currently fails when `gate_or_op` is a `cirq.GateOperation` which wraps a `GateWithRegisters` that doesn't have an explicit annotation. This is because we currently fetch the underlying `gate` from an `op` via an `isinstance(stc, cirq.Operation)` only in `_from_explicit_annotation` 

I suggest we should add a top level check in `_t_complexity_for_gate_or_op` and run all the strategies on the underlying gate when `gate_or_op` is an `op`. 

```
    if isinstance(gate_or_op, cirq.Operation) and gate_or_op.gate is not None:
        return _t_complexity_from_gate(gate_or_op.gate)
```

qualtran/cirq_interop/t_complexity_protocol.py
```diff
@@ -133,22 +141,22 @@ def _from_bloq_build_call_graph(stc: Any, fail_quietly: bool) -> Optional[TCompl
         return None
     ret = TComplexity()
     for bloq, n in sigma.items():
-        r = t_complexity(bloq, fail_quietly=fail_quietly)
+        r = t_complexity(bloq)
```
tanujkhattar, line 144: We need to change the `_, sigma = stc.call_graph(max_depth=1)` line above to the following; so that the `t_complexity` protocol doesn't encounter cirq gates for which an equivalent Bloq exists. 

```
from qualtran.resource_counting.generalizers import cirq_to_bloqs

_, sigma = stc.call_graph(max_depth=1, generalizer=cirq_to_bloqs)
```

tanujkhattar:
The current bloq examples are mostly pretty small. In https://github.com/quantumlib/Qualtran/pull/838 I've added a bloq_example for phase estimation of qubitized walk operator of a 20x20 hubbard model hamiltonian. In the phase estimation notebook, there are two cells now (after my linked PR is merged)

image


The output on main for the two cells are 
"""
CPU times: user 8.44 s, sys: 2.06 s, total: 10.5 s
Wall time: 8.08 s
T-count:   1.85053e+10
Rotations: 3.77494e+07
Cliffords: 8.31144e+10


CPU times: user 8 s, sys: 57.6 ms, total: 8.06 s
Wall time: 8.01 s
T-count:   1.85053e+10
Rotations: 3.77497e+07
Cliffords: 8.31144e+10
"""

The output on your PR for the two cells
"""
CPU times: user 5.12 s, sys: 1.74 s, total: 6.87 s
Wall time: 4.4 s
T-count:   1.85053e+10
Rotations: 3.77494e+07
Cliffords: 8.31144e+10


CPU times: user 7.94 s, sys: 69.4 ms, total: 8.01 s
Wall time: 7.96 s
T-count:   1.85053e+10
Rotations: 3.77497e+07
Cliffords: 8.31144e+10
"""


The good thing is that the time taken by `bloq.t_complexity` didn't regress with your PR. The bad thing is that the time taken by `bloq.t_complexity` is twice the time taken by the cirq API. And after your PR, the time taken by cirq API is the same as that of the Bloq API. 

It would be nice if 
(a) We can figure out and document what's causing the 2x regression in the Bloq API
(b) We can do (a) later but then not regress the Cirq API by 2x as part of this PR


I'll be happy to merge the PR if you can do some investigation for (a) and document it. A `2x` regression is fine if we know it's source. I just don't want to be in a situation where we loose the ability of doing this investigation because the Cirq API timings would be the same as the bloq timings after this PR

mpharrigan:
I'll investigate

mpharrigan:
it's falling back to the decomposition for `ApplyGateToLthQubit` and `SelectedMajoranaFermion` because `nth_operation_callback` is not implemented. Implementing those two methods takes the runtime down to ~200 ms.

I didn't investigate the decomposition, but 40-50% of the time (per snakeviz) is `_cirq_gate_to_bloq`. which isn't too surprising since we need to fall through a bunch of if statements and we create a bunch of gates each call because of circular import considerations.

qualtran/cirq_interop/t_complexity_protocol.py
mpharrigan: thanks, done

mpharrigan:
@tanujkhattar ptal

the slowdown is from the cirq decomposition -> compositebloq in large decompositions, which would be a non issue with #844

qualtran/cirq_interop/jupyter_tools.py
```diff
@@ -62,7 +62,7 @@ def circuit_with_costs(circuit: 'cirq.AbstractCircuit') -> 'cirq.AbstractCircuit
     """Annotates each operation in the circuit with its T-complexity cost."""
 
     def _map_func(op: cirq.Operation, _):
-        t_cost = t_complexity_protocol.t_complexity(op)
+        t_cost = t_complexity_protocol.t_complexity(op.gate)
```
tanujkhattar, line 65: This shouldn't be required now since getting the t-complexity out from a `GateOperation` should be supported after the recent bugfix as per my comment below. Can you verify and revert since we still have some logic to get T-complexity of operations that don't have a gate (like classically controlled ops)?

tanujkhattar (approved):

PR #831, opened 2024-03-27
Add more data blocks

NoureldinYosri:
This PR adds the datablocks from [A Game of Surface Codes: Large-Scale Quantum Computing with Lattice Surgery](https://arxiv.org/abs/1808.02892) and at the same time is a step towards unifying the abstraction for physical resource estimation.


qualtran/surface_code/data_block.py
```diff
@@ -41,6 +43,10 @@ def footprint(self, n_algo_qubits: int) -> int:
     def data_error(self, n_algo_qubits: int, n_cycles: int, phys_err: float) -> float:
         """The error associated with storing data on `n_algo_qubits` for `n_cycles`."""
 
+    @abc.abstractmethod
+    def n_timesteps_to_consume_a_magic_state(self) -> float:
+        """The worst case number of timesteps needed to consume a magic state."""
```
mpharrigan, line 48: what is a timestep?

qualtran/surface_code/ccz2t_cost_model.py
```diff
@@ -189,7 +189,9 @@ def get_ccz2t_costs(
     )
     failure_prob = distillation_error + data_error
     footprint = factory.footprint() + data_block.footprint(n_algo_qubits=n_algo_qubits)
-    duration_hr = (cycle_time_us * n_cycles) / (1_000_000 * 60 * 60)
+    duration_hr = (cycle_time_us * n_cycles * data_block.n_timesteps_to_consume_a_magic_state()) / (
```
mpharrigan, line 192: is this correct? if it takes e.g. 7 steps (7*d cycles) to make one magic state and 7 steps to consume a magic state the total time is still just n_cycles * cycle_time

qualtran/surface_code/data_block.py
NoureldinYosri: Screenshot from 2024-03-29 17-33-42

qualtran/surface_code/ccz2t_cost_model.py
NoureldinYosri: hah, we define `n_cycle` to be the total number of surface code cycles so it's `time_steps * $d$ * n_timesteps_to_consume_magic_state`. I will update the PR accordingly.

NoureldinYosri: actually we need to take the `max`. I renamed `n_cycle` to `n_generation_cycles` and redefined `n_cycle` as `max(n_generation_cycles, n_consumption_cycles)`

qualtran/surface_code/data_block.py
NoureldinYosri: I changed timesteps to cycles and did the needed conversions for consistancy.

qualtran/surface_code/data_block.py
```diff
@@ -76,3 +83,93 @@ def data_error(self, n_algo_qubits: int, n_cycles: int, phys_err: float) -> floa
...
+    the data batches are lined in the first and last row with the middle row being
+    an ancilla region. This lowers the memory footprint of the block at the cost of an
+    increased number of cycles to consume a magic state.
+
+    References:
+        [A Game of Surface Codes](https://arxiv.org/abs/1808.02892)
+        page 7, figure 9
+    """
+
+    routing_overhead: float = 0.5
```
mpharrigan, line 105: Document all the args in each of the concrete subclasses to aid the user; and you can describe why routing overhead is 0.5

When constructing a `CompactDataBlock`, you can specify a non-default `routing_overhead`. is this desired?

mpharrigan (approved):

qualtran/surface_code/data_block.py
NoureldinYosri: the routing overhead for the compact class is different from the others. the others have routing overhead that is just a number * n_qubits, but for the compac class it's `n + sqrt(8n)`

NoureldinYosri: I will add a validator for that

mpharrigan: seems weird tho -- it should be a property that doesn't show up at all in the constructor. I know it's because of the interaction between attrs and inheritance but there's probably a more elegant way

NoureldinYosri: I modified my comment, I will use `field(init=False, default=42)` instead

NoureldinYosri: this way it won't show up in init

Issue #793, opened 2024-03-14
Quimb tensor contraction is significantly slower than cirqs state vector simulation


tanujkhattar:
See https://github.com/quantumlib/Qualtran/pull/788#issuecomment-1996292451 for more details

I'm not sure what's the fix right now, but this is definitely a blocker to transition away from Cirq and use Qualtran native interfaces when writing Bloqs.

mpharrigan:
I'll take a look

mpharrigan:
```
lp = LPResourceState(10)

%%time
lp.tensor_contract()
CPU times: user 24.6 s, sys: 3.82 s, total: 28.5 s
Wall time: 4.18 s

from qualtran.cirq_interop.testing import GateHelper
circuit = GateHelper(lp).circuit

%%time
circuit.final_state_vector()
CPU times: user 59.3 ms, sys: 9.04 ms, total: 68.3 ms
Wall time: 70.1 ms

%%time
circuit.unitary()
CPU times: user 4.17 s, sys: 1.86 s, total: 6.03 s
Wall time: 3.24 s
```

so right off the top, a large proportion of the wall time difference is due to comparing apples and oranges. Asking cirq for the unitary is the same order of magnitude; although the cpu time is worse for `tensor_contract()` and the wall time is still slower for quimb

mpharrigan:
It's worth pointing out that using the cirq simulation for `LPResourceState` still makes calls to `Hadamard().tensor_contract()` and `XGate().tensor_contract()` which may be slowing down the cirq simulation vs a pure cirq implementation. Is this because of the recent change to proactively change cirq gates into bloqs if the bloqs exist in qualtran? Something worth investigating later is if this slows down the cirq simulation

tanujkhattar:
>  large proportion of the wall time difference is due to comparing apples and oranges

From a user perspective, I don't think we are comparing apples and oranges. `LPResourceState` is a Bloq with only a RIGHT register. So `bloq.tensor_contract()` is expected to return a state vector and not a unitary

The same query to cirq is `circuit.final_state_vector()`, so the comparison makes sense from an API perspective I think. 


> Is this because of the recent change to proactively change cirq gates into bloqs if the bloqs exist in qualtran? 

[...]

PR #832, opened 2024-03-28
`GateWithRegisters.controlled()` now returns `BloqAsCirqGate(Controlled())` instead of `cirq.Controlled` gate.

tanujkhattar:
This is a step towards using the new `Controlled` bloq instead of `cirq.Controlled` gate throghout the library. Eventually, we'd to update all the call sites so that `GateWithRegisters.controlled` follows the same api as `Bloq.controlled` but as a step towards this direction; this PR ensures that we move away from using `cirq.Controlled` gate and always stay in Qualtran land. 

Note: This further regressed the CI time from ~17m to ~19m due to https://github.com/quantumlib/Qualtran/issues/793 and the fact that `Controlled` bloq is not a `GateWithRegisters`, it cannot use the underlying unitary even when it wraps a `GateWithRegisters` bloq, when doing cirq style simulations. Using the `Controlled` bloq in cirq-style simulations forces us to go via tensor contraction. Making it a `GateWithRegisters` would circumvent this but this can be addressed in follow-up PRs if needed. cc @mpharrigan

### DIFF
--- a/qualtran/_infra/gate_with_registers.py
+++ b/qualtran/_infra/gate_with_registers.py
@@ -318,7 +318,6 @@ class GateWithRegisters(Bloq, cirq.Gate, metaclass=abc.ABCMeta):
         control_values=None,
         control_qid_shape: Optional[Tuple[int, ...]] = None,
     ) -> 'cirq.Gate':
-        # Multiple inheritance: use the `cirq.Gate` method.
         from qualtran.cirq_interop import BloqAsCirqGate
 
         controlled_gate = cirq.ControlledGate(
--- a/qualtran/_infra/gate_with_registers.py
+++ b/qualtran/_infra/gate_with_registers.py
@@ -21,6 +21,7 @@ from numpy.typing import NDArray
 
 from qualtran._infra.bloq import Bloq, DecomposeNotImplementedError, DecomposeTypeError
 from qualtran._infra.composite_bloq import CompositeBloq
+from qualtran._infra.controlled import Controlled, CtrlSpec
 from qualtran._infra.quantum_graph import Soquet
 from qualtran._infra.registers import Register, Side
 
@@ -318,12 +319,42 @@ class GateWithRegisters(Bloq, cirq.Gate, metaclass=abc.ABCMeta):
         control_qid_shape: Optional[Tuple[int, ...]] = None,
     ) -> 'cirq.Gate':
         # Multiple inheritance: use the `cirq.Gate` method.
-        return cirq.Gate.controlled(
+        from qualtran.cirq_interop import BloqAsCirqGate
+
+        controlled_gate = cirq.ControlledGate(
             self,
             num_controls=num_controls,
             control_values=control_values,
             control_qid_shape=control_qid_shape,
         )
+        ctrl_spec = CtrlSpec.from_cirq_cv(controlled_gate.control_values)
+        return BloqAsCirqGate(Controlled(self, ctrl_spec))
+
+    def _unitary_(self):
+        return NotImplemented
+
+    def add_my_tensors(
+        self,
+        tn: 'qtn.TensorNetwork',
+        tag: 'Any',
+        *,
+        incoming: Dict[str, 'SoquetT'],
+        outgoing: Dict[str, 'SoquetT'],
+    ):
+        if not self._unitary_.__qualname__.startswith('GateWithRegisters.'):
+            from qualtran.cirq_interop._cirq_to_bloq import _add_my_tensors_from_gate
+
+            _add_my_tensors_from_gate(
+                self,
+                self.signature,
+                self.short_name(),
+                tn,
+                tag,
+                incoming=incoming,
+                outgoing=outgoing,
+            )
+        else:
+            return super().add_my_tensors(tn, tag, incoming=incoming, outgoing=outgoing)
 
     def _circuit_diagram_info_(self, args: cirq.CircuitDiagramInfoArgs) -> cirq.CircuitDiagramInfo:
         """Default diagram info that uses register names to name the boxes in multi-qubit gates.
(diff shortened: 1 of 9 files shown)


qualtran/_infra/gate_with_registers.py
```diff
@@ -318,12 +319,42 @@ def controlled(
         control_qid_shape: Optional[Tuple[int, ...]] = None,
     ) -> 'cirq.Gate':
         # Multiple inheritance: use the `cirq.Gate` method.
-        return cirq.Gate.controlled(
+        from qualtran.cirq_interop import BloqAsCirqGate
```
mpharrigan, line 322: update the comment above

qualtran/cirq_interop/_bloq_to_cirq_test.py
```diff
@@ -160,75 +160,22 @@ def test_bloq_as_cirq_gate_left_register():
...
-                              │
-targets[3][0]: ───────────────×(y)────────────────────
-                              │
-targets[3][1]: ───────────────×(y)────────────────────
-                              │
-targets[3][2]: ───────────────×(y)────────────────────
-''',
-    )
+@attrs.frozen
+class TestSwapWithZero(Bloq):
```
tanujkhattar, line 164: This is moot and I'll delete it.

qualtran/_infra/gate_with_registers.py
```diff
@@ -318,12 +319,42 @@ def controlled(
...
+
+        controlled_gate = cirq.ControlledGate(
             self,
             num_controls=num_controls,
             control_values=control_values,
             control_qid_shape=control_qid_shape,
         )
+        ctrl_spec = CtrlSpec.from_cirq_cv(controlled_gate.control_values)
```
mpharrigan, line 330: So we make the gate so we can get the corresponding `CtrlSpec` (and throw out the gate). Is this a kludge or is this fine? If kludge: open an issue to track simplifying

qualtran/cirq_interop/_bloq_to_cirq.py
```diff
@@ -145,13 +146,15 @@ def _decompose_(self, qubits: Sequence[cirq.Qid]) -> cirq.OP_TREE:
         return self._decompose_with_context_(qubits)
 
     def _unitary_(self):
-        if self._num_qubits_() > 3:
-            # Prefer decomposition for large bloqs
-            return NotImplemented
```
tanujkhattar, line 150: This had to be removed because it leads us to a situation where if we have a leaf bloq without a decomposition which acts on > 3 qubits but does override the `add_my_tensors` method; the cirq wrapper does not derive a unitary from it and raises an error which is not ideal

qualtran/_infra/gate_with_registers.py
```diff
@@ -318,12 +319,42 @@ def controlled(
...
+        return BloqAsCirqGate(Controlled(self, ctrl_spec))
+
+    def _unitary_(self):
+        return NotImplemented
```
mpharrigan, line 334: why this change? was this not the default?

qualtran/cirq_interop/cirq_interop.ipynb
```diff
@@ -455,55 +455,6 @@
     "out_quregs"
    ]
   },
-  {
-   "cell_type": "markdown",
-   "id": "896e0392-3f1f-4965-bb13-0a541581ec25",
-   "metadata": {},
-   "source": [
-    "## Test `Cirq -> Bloqs -> Cirq` roundtrip using `QROM`\n",
```
mpharrigan, line 463: is this removal on purpose?

qualtran/_infra/gate_with_registers.py
tanujkhattar: Eventually, I'd like to change the `def controlled` method to use the bloq style `def controlled(ctrl_spec)` instead of cirq style `def controlled(self, num_controls, control_values, ...)` so this simplification wouldn't be required.

qualtran/bloqs/util_bloqs.py
```diff
@@ -471,3 +471,41 @@ def on_classical_vals(self, reg: int) -> Dict[str, 'ClassicalValT']:
 
     def _t_complexity_(self) -> 'TComplexity':
         return TComplexity()
+
+
+@frozen
+class Power(Bloq):
```
mpharrigan, line 477: is this addition on purpose? factor into its own PR? is it used elsewhere in the control stuff?

mpharrigan (approved):

qualtran/bloqs/util_bloqs.py
tanujkhattar: Yes, this addition is on purpose and was required to pass tests (eg: for phase estimation, where we use powers of controlled gate). I can factor into it's own PR but this PR will be blocked on the other PR being merged. I also need to add tests for this bloq

qualtran/cirq_interop/cirq_interop.ipynb
tanujkhattar: Yeah, we've come a long way since this was written. The cirq interop probably needs to be updated now but maybe as a follow up. 


The idea is that now `BloqAsCirqGate` should be used to wrap only bloqs which are not already cirq gates (I've added an assertion to the constructor). `QROM` is a `GateWithRegisters` so should never be wrapped in a `BloqAsCirqGate`. This section was written when `QROM` was implemented in Cirq-FT and was not already a Bloq. 

Other examples of wrapping cirq-only gates in `CirqGateAsBloq` and bloq-only bloqs in `BloqAsCirqGate` are still part of the notebook.

qualtran/bloqs/util_bloqs.py
mpharrigan: no need to factor

qualtran/cirq_interop/_bloq_to_cirq.py
mpharrigan: sounds good

Issue #829, opened 2024-03-27
support serializing complex values


anurudhp:
Ran into this error when serializing `GlobalPhase` bloq (https://github.com/quantumlib/Qualtran/pull/828/)
```
E           qualtran.testing.BloqCheckException: Serialization Failed:
E           Cannot serialize 1j of unknown type <class 'complex'>
```

PR #833, opened 2024-03-28
serialize `complex` bloq arguments

anurudhp:
fixes https://github.com/quantumlib/Qualtran/issues/829


tanujkhattar (approved):
This is great. Can we remove the `GlobalPhase`bloq from not serializable list as part of this PR as well, to show that the bloq can now be serialized given the new capabilities?

cc @Epsilon1024  FYI

Issue #785, opened 2024-03-13
`t_complexity(cirq.CCX)` and `t_complexity(Toffoli())` don't match


tanujkhattar:
```python
In [7]: t_complexity(cirq.CCX)
Out[7]: TComplexity(t=7, clifford=10, rotations=0)

In [9]: t_complexity(Toffoli())
Out[9]: TComplexity(t=4, clifford=0, rotations=0)
```

brempfer:
It would be nice to know exactly which Toffoli is being used in decompositions (without looking at source or decomposing fully) - for example when decomposing a MCX gate one can use Toffoli() which has t=4 because each Toffoli that is computed is subsequently uncomputed. However, when decomposing a controlled SWAP() one must use cirq.CCX which has t=7 because there is only one CCX gate in the decomposition to a Fredkin gate.

mpharrigan:
Yeah, all the qualtran bloqs have a specific decomposition and the t count must match that. We don't have (direct) control over the decompositions in cirq.

brempfer:
As far as I can tell Toffoli() is supposed to be implementing the full CCX gate. This requires T-count: 7 and Clifford: 8 by standard construction (no ancilla). However, according to the reference in Toffoli() source code you can utilize two ancilla and measurement control to yield T-count: 4, Clifford: 14. This is what is implemented in mcmt.MultiControlPauli((1,1), target_gate=cirq.X). 

In most decompositions that need them CCX gates appear in pairs which means they are actually compute and uncompute AND gates implemented in mcmt.and_bloq.MultiAnd(). This is why mcmt.MultiControlPauli((1,1), target_gate=cirq.X) yields the correct count for a full CCX gate. In my work I am always worried about decomposition to the Toffoli level and therefore, in my decompositions, fully specify with MultiAnd gates or cirq.CCX. I would prefer to have Toffoli() return T-count: 7, Clifford: 8 with no ancilla so I can use it instead of cirq.CCX as the bloqs will play better with decomposing into base gatesets. Alternatively, if one wants a lower T-count they can call mcmt.MultiControlPauli((1,1), target_gate=cirq.X) instead of Toffoli() and then there is no need to use cirq.CCX. 

TLDR: I want to differentiate compute/uncompute AND with Toffoli(). My suggestion would be to have Toffoli() use no ancilla and return TComplexity(t=7, clifford=8, rotations=0).

brempfer:
[...]

PR #834, opened 2024-03-28
Update decomposition of MultiControlPauli to use AND/AND† for num_controls=2

tanujkhattar:
Part of fixing https://github.com/quantumlib/Qualtran/issues/785

The `MultiControlPauli` should stay true to it's description of using `n - 1` clean ancillas for `n` controls and using $\text{And} / \text{And}^{\dagger}$  ladders with a singly controlled pauli at the end.


mpharrigan:
What was happening before? It was using `MultiAnd` with two control values? Should that be an error or somehow force a replacement with `[TwoBit]And`?

tanujkhattar:
It was using cirq.CCX, cirq.CCZ etc for 2 control case

qualtran/bloqs/mcmt/multi_control_multi_target_pauli.py
```diff
@@ -123,20 +123,25 @@ def decompose_from_registers(
         self, *, context: cirq.DecompositionContext, **quregs: NDArray['cirq.Qid']
     ) -> cirq.OP_TREE:
         controls, target = quregs.get('controls', np.array([])), quregs['target']
-        if len(self.cvs) <= 2:
+        if len(self.cvs) < 2:
```
mpharrigan, line 126: do you still need `controls.flatten()`? Does this work for len(self.cvs) == 0?

mpharrigan (commented):
Can you take the opportunity to flesh out the docstring for this class a bit. In particular: which part of the reference does this correspond to?

mpharrigan:
Does this actually use the fact that the gate is a Pauli? Or does it work for any single qubit gate?

qualtran/bloqs/mcmt/multi_control_multi_target_pauli.py
tanujkhattar: Yes and Yes. We have tests already for `len(self.cvs) == 0` which are passing after this PR as well.

tanujkhattar:
> Can you take the opportunity to flesh out the docstring for this class a bit. In particular: which part of the reference does this correspond to?

@mpharrigan I've added more details to the docstring. 

> 
Does this actually use the fact that the gate is a Pauli? Or does it work for any single qubit gate?

It works for any single qubit gate. In a follow-up PR we can factor out the logic into a generic `MultiControlBloq` that delegates to a singly controlled Bloq between an $\text{AND} / \text{AND}^\dagger$ ladder

mpharrigan (approved):

PR #836, opened 2024-03-29
One true jupyter autogen module

mpharrigan:
- Remove the last remaining bloqs documented under the "old" v1 system
 - Remove the v1 system
 - Put the "v2" code back into the `qualtran_dev_tools.jupyter_autogen` one true module

I kept the commits meaningful to aid in review.


fdmalone:
End of an era :(

dev_tools/autogenerate-bloqs-notebooks-v2.py
```diff
@@ -50,7 +50,7 @@
 
 from qualtran_dev_tools.bloq_finder import get_bloqdocspecs
 from qualtran_dev_tools.git_tools import get_git_root
-from qualtran_dev_tools.jupyter_autogen_v2 import NotebookSpecV2, render_notebook
+from qualtran_dev_tools.jupyter_autogen import NotebookSpecV2, render_notebook
```
fdmalone, line 53: Should we also do NotebookSpecV2 -> NotebookSpec change, or do you want to keep it this way?

fdmalone (approved):

dev_tools/autogenerate-bloqs-notebooks-v2.py
mpharrigan: i'll do that in a follow-up so I don't create any merge conflicts with in-flight things

PR #838, opened 2024-03-31
Bloq example for phase estimation of Qubitized Quantum Walk of 20x20 Hubbard Model

tanujkhattar:
Currently on top of https://github.com/quantumlib/Qualtran/pull/837

Adds a `bloq_example` for phase estimation of a Qubitized quantum walk operator for 20x20 hubbard model. This should be used as a benchmark for testing potential performance regressions as part of updating the T-complexity infrastructure as done in https://github.com/quantumlib/Qualtran/pull/825

Also fixes a few bugs as part of the update.


mpharrigan (approved):
you fell for my (well intentioned) trap :) yes, I think we should have the things we actually care about as `bloq_examples` so it can be included in folks' "check everything at once" scripts

PR #844, opened 2024-04-01
Unary iteration: nth_operation_callgraph for more subclasses

mpharrigan:
This speeds up gate counting from a couple seconds to a couple hundred milliseconds on the hubbard model example.

### DIFF
--- a/qualtran/bloqs/multiplexers/apply_gate_to_lth_target.py
+++ b/qualtran/bloqs/multiplexers/apply_gate_to_lth_target.py
@@ -14,7 +14,7 @@
 
 import itertools
 from functools import cached_property
-from typing import Callable, Sequence, Tuple
+from typing import Callable, Sequence, Set, Tuple
 
 import attrs
 import cirq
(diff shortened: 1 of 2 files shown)


qualtran/bloqs/multiplexers/apply_gate_to_lth_target.py
```diff
@@ -104,6 +106,10 @@ def nth_operation(  # type: ignore[override]
         target_idx = int(np.ravel_multi_index(selection_idx, selection_shape))
         return self.nth_gate(*selection_idx).on(target[target_idx]).controlled_by(control)
 
+    def nth_operation_callgraph(self, **selection_regs_name_to_val) -> Set['BloqCountT']:
+        selection_idx = tuple(selection_regs_name_to_val[reg.name] for reg in self.selection_regs)
+        return {(_cirq_gate_to_bloq(self.nth_gate(*selection_idx)), 1)}
```
tanujkhattar, line 111: The `self.target_gate` and `self.nth_gate` should be controlled gates here and below.

tanujkhattar (changes requested):

qualtran/bloqs/multiplexers/apply_gate_to_lth_target.py
tanujkhattar: I think this would be a good opportunity to revisit these gates, use the new controlled infrastructure, create a bloq example and fix whatever is failing.

mpharrigan:
This is a proof of principle for investigating #825. I'll make as draft and "contributions welcome" if someone wants to take this on fully

mpharrigan:
Has this been done?

Issue #840, opened 2024-04-01
No explicit tests for `PRGAViaPhaseGradient`


ncrubin:
The documentation and testing is very light for this bloq.  Can this be expanded? it would seem it only has system testing via state-prep via rotations.

Reference:

https://github.com/quantumlib/Qualtran/blob/main/qualtran/bloqs/state_preparation/state_preparation_via_rotation_test.py

contains no tests for the mentioned bloq.

ncrubin:
@Vicara12

Vicara12:
@ncrubin I added a test on PR #846, which contains both an example of usage and also the formula for converting the desired angle into the value to be loaded in the QROM (which is what the bloq receives). Let me know if you have further feedback!

PR #846, opened 2024-04-02
Tests for PRGAViaPhaseGradient

Vicara12:
This PR adds explicit testing for the PRGAViaPhaseGradient bloq, which can also be useful as a reference on its usage. It solves issue #840.


ncrubin (approved):
Thanks for the updated test.  

For the future it would be good to have a specific rotation-by-phasegrad since this is bundling with a QROM build. the rotate by controlled addition (which is in here) is necessary for lots and lots of things.

tanujkhattar:
> For the future it would be good to have a specific rotation-by-phasegrad

This is encapsulated in `QvrPhaseGradient` ?

ncrubin:
I don't think so.  It is very close to what one would need for rotation by phase gradient given quantum data. The documentation for `QvrPhaseGradient` indicates just phasing. Control is never listed which you would need. Also an additional multiplication step by `gamma` seems specific to the reference in the documentation.  `QvrPhaseGradient` looks like a specific example of something that I would imagine a rotation by phasegradient.

ncrubin (changes requested):
There seems to be a problem with the costs for `AddIntoPhaseGradient`

```
    add_into_phasegrad_gate = AddIntoPhaseGrad(
        x_bitsize=4,
        phase_bitsize=4,
        right_shift=0,
        sign=1,
        controlled=0
    )
    print(add_into_phasegrad_gate._t_complexity_())
    add_into_phasegrad_gate = AddIntoPhaseGrad(
        x_bitsize=4,
        phase_bitsize=4,
        right_shift=0,
        sign=1,
        controlled=1
    )
    print(add_into_phasegrad_gate._t_complexity_())
    
    ```
both return the same cost despite one being controlled and the other not

ncrubin:
I am mistaken.  In https://arxiv.org/pdf/2007.07391.pdf Appendix A there is an improvement over the naive controlled addition which would be double the complexity.  The improvement enables control with only clifford gates and thus b_grad - 2 Toffoli.

ncrubin (approved):

ncrubin:
@Vicara12 can you merge this?

PR #847, opened 2024-04-02
Add bloq can take in different size registers for a and b

skushnir123:


tanujkhattar:
@NoureldinYosri Can you please take a look?

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -74,18 +75,26 @@ class Add(Bloq):
...
+    a_dtype: Union[QInt, QUInt, QMontgomeryUInt] = field()
+    b_dtype: Union[QInt, QUInt, QMontgomeryUInt] = field()
 
-    @dtype.validator
-    def _dtype_validate(self, field, val):
+    @a_dtype.validator
+    def _a_dtype_validate(self, field, val):
+        if not isinstance(val, (QInt, QUInt, QMontgomeryUInt)):
+            raise ValueError("Only QInt, QUInt and QMontgomerUInt types are supported.")
+        if isinstance(val.num_qubits, sympy.Expr):
```
NoureldinYosri, line 85: >         if isinstance(val.num_qubits, sympy.Expr):
>              return
this check does nothing

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -74,18 +75,26 @@ class Add(Bloq):
...
+            return
+
+    @b_dtype.validator
+    def _b_dtype_validate(self, field, val):
         if not isinstance(val, (QInt, QUInt, QMontgomeryUInt)):
             raise ValueError("Only QInt, QUInt and QMontgomerUInt types are supported.")
         if isinstance(val.num_qubits, sympy.Expr):
```
NoureldinYosri, line 92: same as above

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -114,16 +124,17 @@ def decompose_bloq(self) -> 'CompositeBloq':
...
+        N = 2**b_bitsize if unsigned else 2 ** (b_bitsize - 1)
         return {'a': a, 'b': int(math.fmod(a + b, N))}
 
     def short_name(self) -> str:
         return "a+b"
 
     def _circuit_diagram_info_(self, _) -> cirq.CircuitDiagramInfo:
-        wire_symbols = ["In(x)"] * self.dtype.bitsize
-        wire_symbols += ["In(y)/Out(x+y)"] * self.dtype.bitsize
+        wire_symbols = ["In(x)"] * self.b_dtype.bitsize
```
NoureldinYosri, line 136: shouldn't this be a_dtype?

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -137,10 +148,18 @@ def wire_symbol(self, soq: 'Soquet') -> 'WireSymbol':
...
-            yield CNOT().on(anc[depth - 1], inp[depth])
+            if inp[depth] is not None:
+                yield CNOT().on(anc[depth - 1], inp[depth])
+            else:
+                # If inp[depth] = None, we treat it as a |0>,
+                # and therefore applying CNOT().on(anc[depth - 1], inp[depth])
+                # essentially "copies" anc[depth - 1] into inp[depth]
+                # in the classical basis. So therefore, on future operations,
+                # we can use anc[depth - 1] in its place.
+                inp[depth] = anc[depth - 1]
```
NoureldinYosri, line 162: this statement is potentially dangerous if the caller assumes that `_left_building_block` doesn't modify its input which is reasonable since its current implementation doesn't.

consider using a a variable instead

```py3

if ...:
   ...
   control = inp[depth]
else:
   control = anc[depth - 1]
```

This will make the logic clearer and remove this condition from below
```py3
            if anc[depth - 1] != inp[depth]:
```

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -152,63 +171,81 @@ def _right_building_block(self, inp, out, anc, depth):
...
+                yield CNOT().on(inp[depth], out[depth])
             yield from self._right_building_block(inp, out, anc, depth - 1)
 
     def decompose_from_registers(
         self, *, context: cirq.DecompositionContext, **quregs: NDArray[cirq.Qid]
     ) -> cirq.OP_TREE:
         # reverse the order of qubits for big endian-ness.
         input_bits = quregs['a'][::-1]
+        # Make len(input_bits) = len(output_bits) by appending Nones
+        input_bits = np.append(input_bits, [None] * (self.b_dtype.bitsize - self.a_dtype.bitsize))
```
NoureldinYosri, line 191: since the assumption is that `self.b_dtype.bitsize >= self.a_dtype.bitsize`, please add a validator for that

NoureldinYosri (changes requested):

qualtran/bloqs/arithmetic/addition.py
skushnir123: Ya this was there before so I just didn't remove it because I wasn't sure but I'll remove it now.

qualtran/bloqs/arithmetic/addition.py
skushnir123: Ah yes.

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -74,18 +75,26 @@ class Add(Bloq):
         [Halving the cost of quantum addition](https://arxiv.org/abs/1709.06648)
     """
 
-    dtype: Union[QInt, QUInt, QMontgomeryUInt] = field()
+    a_dtype: Union[QInt, QUInt, QMontgomeryUInt] = field()
```
NoureldinYosri, line 78: I don't think you need the ` = field()` part

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -151,9 +171,12 @@ def _right_building_block(self, inp, out, anc, depth):
             return
         else:
             yield CNOT().on(anc[depth - 1], anc[depth])
-            yield And().adjoint().on(inp[depth], out[depth], anc[depth])
-            yield CNOT().on(anc[depth - 1], inp[depth])
-            yield CNOT().on(inp[depth], out[depth])
+            if depth < len(inp):
+                yield And().adjoint().on(inp[depth], out[depth], anc[depth])
+                yield CNOT().on(anc[depth - 1], inp[depth])
```
NoureldinYosri, line 176: what happens if `depth = 0`?

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -62,9 +62,10 @@ class Add(Bloq):
...
     Args:
-        bitsize: Number of bits used to represent each integer. Must be large
+        dtype: Quantum datatype used to represent each integer. Must be large
             enough to hold the result in the output register of a + b, or else it simply
-            drops the most significant bits.
+            drops the most significant bits. One can either provide a single Quantum datatype
+            to be used for both a and b or a 2-element tuple of Quantum datatypes for a and b.
 
     Registers:
         a: A bitsize-sized input register (register a above).
```
NoureldinYosri, line 71: for `a` replace bitsize-sized with a_dtype.bitsize-sized and same for `b`

NoureldinYosri (commented):
almost there, a few more changes

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -74,18 +75,26 @@ class Add(Bloq):
         [Halving the cost of quantum addition](https://arxiv.org/abs/1709.06648)
     """
 
-    dtype: Union[QInt, QUInt, QMontgomeryUInt] = field()
+    a_dtype: Union[QInt, QUInt, QMontgomeryUInt] = field()
+    b_dtype: Union[QInt, QUInt, QMontgomeryUInt] = field()
```
NoureldinYosri, line 79: actually keep the ` = field()` and make this  parameter assumes the value of the `a_dtype` when not provided
```py3
    b_dtype: Union[QInt, QUInt, QMontgomeryUInt] = field()
   
    @b_dtype.default
     def _b_dtype_default(self):
           return self.a_dtype
```

This will  make your change backward compatible and greatly reduce the size of your PR (you will probably be able to undo all of the changes to the old _test.py and elsewhere)

qualtran/bloqs/arithmetic/addition.py
skushnir123: depth starts off = 1

qualtran/simulation/xcheck_classical_quimb_test.py
```diff
@@ -42,7 +42,7 @@ def test_toffoli():
 
 
 def test_add():
-    add = Add(QUInt(bitsize=5))
+    add = Add(QUInt(bitsize=5), QUInt(bitsize=5))
```
NoureldinYosri, line 45: can you undo your change to this file and other test files to make sure the change is backward compatible?

qualtran/bloqs/arithmetic/addition.py
skushnir123: Done

qualtran/simulation/xcheck_classical_quimb_test.py
skushnir123: Yup I just finished going through and changing everything back.

qualtran/bloqs/arithmetic/addition.py
NoureldinYosri: please add a gaurd on the `depth - 1` statements to make the code safer, e.g.

```py3
if depth:
   yield CNOT().on(anc[depth - 1], inp[depth])
```

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -62,30 +62,43 @@ class Add(Bloq):
     Implements $U|a\rangle|b\rangle \rightarrow |a\rangle|a+b\rangle$ using $4n - 4 T$ gates.
 
     Args:
-        bitsize: Number of bits used to represent each integer. Must be large
+        dtype: Quantum datatype used to represent each integer. Must be large
```
NoureldinYosri, line 65: this docstring needs modification. the class now has two Args `a_dtype` and `b_dtype` not a single `dtype`

NoureldinYosri (commented):
almost there, two more comments

qualtran/bloqs/arithmetic/addition.py
skushnir123: What about adding an assert statement at the beginning of the _left_building_block? Because adding an if statement every time I have anc[depth - 1] will make the code kinda messy in my opinion.

NoureldinYosri: okay but make it more informative

```py3
if depth < 1:
   raise ValueError(f"{depth=} is not a positive integer")
```

skushnir123: Ok I added a more informative message along with the assert

NoureldinYosri (approved):

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -162,29 +191,30 @@ def decompose_from_registers(
...
     def _t_complexity_(self):
-        n = self.dtype.bitsize
+        n = self.b_dtype.bitsize
         num_clifford = (n - 2) * 19 + 16
         num_toffoli = n - 1
         return TComplexity(t=4 * num_toffoli, clifford=num_clifford)
 
     def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
-        n = self.dtype.bitsize
+        n = self.b_dtype.bitsize
```
mpharrigan, line 217: why are we using the second registers' bitsize?

skushnir123: This the register that we are adding into and is always >= first register size.

PR #848, opened 2024-04-02
[GQSP] Reorganize code into module `qsp` 

anurudhp:


docs/bloqs/index.rst
```diff
@@ -29,6 +29,7 @@ Bloqs Library
     basic_gates/toffoli.ipynb
     basic_gates/hadamard.ipynb
     basic_gates/cnot.ipynb
+    basic_gates/su2_rotation.ipynb
```
tanujkhattar, line 32: @mpharrigan  was adding `su2_rotation` to the `Rotations:` section in https://github.com/quantumlib/Qualtran/pull/843

qualtran/bloqs/for_testing/atom.py
```diff
@@ -111,3 +111,63 @@ def add_my_tensors(
 
     def short_name(self) -> str:
         return 'op'
+
+
+@frozen(repr=False)
+class TestGWRAtom(GateWithRegisters):
+    """An atomic gate with registers useful for generic testing and demonstration.
```
tanujkhattar, line 118:
```suggestion
    """An atomic gate that derives from `GateWithRegisters` which useful for testing.
    
    The single qubit gate has a unitary effect corresponding to a 2x2 identity matrix. 
```

qualtran/bloqs/qsp/generalized_qsp.py
```diff
@@ -274,16 +340,51 @@ def decompose_from_registers(
...
         if self.negative_power > 0:
-            counts.add((self.U.adjoint().controlled(), min(degree, self.negative_power)))
+            counts.add((Controlled(self.U.adjoint(), CtrlSpec()), min(degree, self.negative_power)))
 
         return counts
+
+
+@bloq_example
+def _gqsp() -> GeneralizedQSP:
+    from qualtran.bloqs.for_testing.atom import TestGWRAtom
```
tanujkhattar, line 360: I think we should use the gates defined in `for_testing/` only in tests. For examples, it's better to use commonly used basic gates; maybe add a `Identity` bloq if you need it for the example or use a `ZPow` or `XPow` gate ?

qualtran/bloqs/qsp/generalized_qsp.py
```diff
@@ -274,16 +340,51 @@ def decompose_from_registers(
...
+    )
+    return gqsp_with_negative_power
+
+
+@bloq_example
+def _gqsp_with_large_negative_power() -> GeneralizedQSP:
+    from qualtran.bloqs.for_testing.atom import TestGWRAtom
+
+    gqsp_with_large_negative_power = GeneralizedQSP.from_qsp_polynomial(
+        TestGWRAtom(), (0.5, 0, 0.5), negative_power=5
```
tanujkhattar, line 381: Is `5` large enough? Does it support higher negative powers as well?

tanujkhattar (changes requested):

qualtran/bloqs/qsp/generalized_qsp.py
anurudhp: It works for any negative power. When `negative_power` is more than the degree, the circuit has U^\dagger terms, which I wanted to showcase with the example.

qualtran/bloqs/qsp/generalized_qsp.py
anurudhp: Ah yes that makes sense, I'll just use `XGate` here.

docs/bloqs/index.rst
anurudhp: ah yes, will rebase on top of that

qualtran/bloqs/qsp/generalized_qsp.py
```diff
@@ -274,16 +345,51 @@ def decompose_from_registers(
...
     def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
-        degree = len(self.P)
+        degree = len(self.P) - 1
 
-        counts = {(rotation, 1) for rotation in self.signal_rotations}
+        counts = set(Counter(self.signal_rotations).items())
 
         if degree > self.negative_power:
-            counts.add((self.U.controlled(control_values=[0]), degree - self.negative_power))
+            counts.add((Controlled(self.U, CtrlSpec(cvs=0)), degree - self.negative_power))
```
tanujkhattar, line 353: This should eventually be `self.U.controlled()` but the controlled infrastructure right now is in a messy state so it's fine to do this. Please open an issue to track this however.

tanujkhattar (approved):

PR #849, opened 2024-04-02
Remove qpe_hubbard tests

Epsilon1024:
These tests are very large and will cause problems when running. These
tests are now disabled.


dev_tools/qualtran_dev_tools/bloq_finder.py
```diff
@@ -121,10 +121,16 @@ def get_bloq_examples() -> List[BloqExample]:
     bloqs_root = reporoot / 'qualtran/bloqs'
     paths = get_bloq_module_paths(bloqs_root)
 
+    # These examples are too large to run.
+    REJECTED_BEXAMPLES = set(
+        ["qubitization_qpe_hubbard_model_large", "qubitization_qpe_hubbard_model_small"]
+    )
+
```
tanujkhattar, line 128: Can we just remove them from the serialization test instead of removing completely?

mpharrigan: +1; `get_bloq_examples` should get all the bloq examples. If there are specific places where they shouldn't be used, you can intercept there.

Epsilon1024: Fixed

mpharrigan:
```python
def test_whatever(bloq_autotester):
  if bloq_autotester.check_name == 'serialization':
    pytest.skip('too big')

  bloq_autotester(_whatever_example)
```

Epsilon1024:
> ```python
> def test_whatever(bloq_autotester):
>   if bloq_autotester.check_name == 'serialization':
>     pytest.skip('too big')
> 
>   bloq_autotester(_whatever_example)
> ```

I gave this a try using [test_qubitization_qpe_bloq_autotester](https://github.com/quantumlib/Qualtran/blob/37250fbcaaf9223d8d969e4e584045616a3a0011/qualtran/bloqs/phase_estimation/qubitization_qpe_test.py#L30) as the "test_whatever" method.  I noticed two issues with this:
1) test_qubitization_qpe_bloq_autotester is never called from [get_bloq_report_card](https://github.com/quantumlib/Qualtran/blob/37250fbcaaf9223d8d969e4e584045616a3a0011/dev_tools/qualtran_dev_tools/bloq_report_card.py#L100).  The stack goes from get_bloq_report_card -> [record_for_bloq_example](https://github.com/quantumlib/Qualtran/blob/37250fbcaaf9223d8d969e4e584045616a3a0011/dev_tools/qualtran_dev_tools/bloq_report_card.py#L84) ->[check_bloq_example_serialize](https://github.com/quantumlib/Qualtran/blob/37250fbcaaf9223d8d969e4e584045616a3a0011/qualtran/testing.py#L515)->[assert_bloq_example_serialize](https://github.com/quantumlib/Qualtran/blob/37250fbcaaf9223d8d969e4e584045616a3a0011/qualtran/testing.py#L493)

It looks like this won't prevent this test from running.

2) The other thing I noticed is that [_qubitization_qpe_hubbard_model_large](https://github.com/quantumlib/Qualtran/blob/37250fbcaaf9223d8d969e4e584045616a3a0011/qualtran/bloqs/phase_estimation/qubitization_qpe.py#L183) doesn't seem to be called by an autotester.  However, [_qubitization_qpe_hubbard_model_small](https://github.com/quantumlib/Qualtran/blob/37250fbcaaf9223d8d969e4e584045616a3a0011/qualtran/bloqs/phase_estimation/qubitization_qpe.py#L164) is called by an [autotester]((https://github.com/quantumlib/Qualtran/blob/37250fbcaaf9223d8d969e4e584045616a3a0011/qualtran/bloqs/phase_estimation/qubitization_qpe_test.py#L31))

mpharrigan:
`get_bloq_report_card` is called ad-hoc by the developer outside of the CI flow. Re: (2) you have to "opt in" to including the auto-tests as part of the `pytest` unit tests, which _is_ run as part of the CI flow

Epsilon1024:
> `get_bloq_report_card` is called ad-hoc by the developer outside of the CI flow. Re: (2) you have to "opt in" to including the auto-tests as part of the `pytest` unit tests, which _is_ run as part of the CI flow

Gotcha.  Should we address how to remove it in `get_bloq_report_card` as well?

Epsilon1024:
I don't think these changes are needed any longer.  We can close this PR.

Issue #854, opened 2024-04-03
Show size of registers in `musical_score` diagrams of Bloqs


tanujkhattar:
The musical score diagrams currently have wires for each register but the wires don't have any data type / size information. I'd like to see the size of the wires added on to the musical score diagrams. 

For example:
I want to see something like 
<img width="817" alt="image" src="https://github.com/quantumlib/Qualtran/assets/7863287/a8e6b713-86e3-4db1-9424-a3a33c562f86">

instead of 

<img width="638" alt="image" src="https://github.com/quantumlib/Qualtran/assets/7863287/cb034de8-1ed4-4627-b012-b6e98b4f8789">

Notice `x[7]` instead of `x`  in the first figure. This becomes more useful for complicated diagrams like 

<img width="1086" alt="image" src="https://github.com/quantumlib/Qualtran/assets/7863287/32d7241e-2695-4554-887f-3f2f416caf5b">

Here it's very important to know that phase gradient register size is 7, and the information is not visible if we don't make the suggested change 

To generate these diagrams, I changed `_soq_to_symb` to return `return Text(soq.pretty() + f'[{soq.reg.bitsize}]')` instead of `return Text(soq.pretty())` when `soq.binst` is a dangling soquet. It would be nicer if we can also show the information about quantum data types somehow on the wire (eg: for fixed point registers, both `bitsize` and `nfrac` become important)

PR #857, opened 2024-04-03
Show qdtype info in musical score diagrams

tanujkhattar:
Fixes https://github.com/quantumlib/Qualtran/issues/854

Currently based on top of https://github.com/quantumlib/Qualtran/pull/856 since I imagine this PR can involve more discussion than https://github.com/quantumlib/Qualtran/pull/856

Can also default to show only the size instead of `str(dtype)` if people think this is cluttering the output too much.

### DIFF
--- a/qualtran/drawing/musical_score.py
+++ b/qualtran/drawing/musical_score.py
@@ -382,6 +382,7 @@ class TextBox(WireSymbol):
 @frozen
 class Text(WireSymbol):
     text: str
+    fontsize: int = 10
 
     def draw(self, ax, x, y):
         ax.text(
@@ -389,7 +390,7 @@ class Text(WireSymbol):
             -y,
             self.text,
             transform=ax.transData,
-            fontsize=10,
+            fontsize=self.fontsize,
             ha='center',
             va='center',
             bbox={'lw': 0, 'fc': 'white'},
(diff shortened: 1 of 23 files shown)


mpharrigan (approved):
Easy way to get at least the dtypes of the dangling wires. 

Nit: I think of bracketed numbers as indexing in to something rather than the size of something. My preference would be to use `/` 

`x / QInt(5)`. 

or maybe parentheses

qualtran/drawing/musical_score.py
```diff
@@ -476,7 +477,7 @@ def _soq_to_symb(soq: Soquet) -> WireSymbol:
 
     # Use text (with no box) for dangling register identifiers.
     if isinstance(soq.binst, DanglingT):
-        return Text(soq.pretty())
+        return Text(soq.pretty() + f'[{soq.reg.dtype!s}]', fontsize=8)
```
mpharrigan, line 480: is !s necessary? isn't that the default?

tanujkhattar: you are right, it's not needed. Removed.

tanujkhattar:
Change to `/` instead of `[]` as suggested, so it'll now view as `x / QInt(5)`

PR #859, opened 2024-04-04
 Add Fxp classical simulation 

fdmalone:
Adds:
* classical simulation for the Fxp multiplication bloqs + refactors the error analysis notebook to use these
* Updates phase gradient classical simulation to use Fxp types
* Uses correct QDType bit converters.

Problems:
* Phase gradient tests are way too slow with Fxp classical simulation.

Built upon #817


fdmalone:
cc @tanujkhattar, I need to profile why classical simulation is so slow + an alternative to caching BUT the big problem right now is that some of the tests which use the phasegradient  have impossibly slow tests. Might be good to get the ball rolling though, was a painful PR. I can also split it up potentially since there are a few things going on.

fdmalone:
My sneaking suspicion is doing the bitwise algorithm for the paper is slow, @tanujkhattar did you check if Fxp multiplication yielded the same thing as what's in the reference.

PR #864, opened 2024-04-08
Controlled-Addition implementation

skushnir123:


mpharrigan:
This should be a separate bloq since there's more conditional `if`-ed code than shared code

skushnir123:
> This should be a separate bloq since there's more conditional `if`-ed code than shared code

Ok I changed this to be another separate bloq

qualtran/bloqs/arithmetic/controlled_addition.py
```diff
@@ -0,0 +1,240 @@
...
+            return {'ctrl': ctrl, 'a': a, 'b': b}
+        else:
+            return {'ctrl': ctrl, 'a': a, 'b': int(math.fmod(a + b, N))}
+
+    def short_name(self) -> str:
+        return "a+b"
+
+    def _circuit_diagram_info_(self, _) -> cirq.CircuitDiagramInfo:
+        if self.controlled is not None:
+            wire_symbols = ["In(ctrl)"]
```
fdmalone, line 144: Double check this is the correct wire symbol for ctrl. Typically I read In as input, which I guess the ctrl value is but it's sort of a separate concept.

qualtran/bloqs/arithmetic/controlled_addition.py
```diff
@@ -0,0 +1,240 @@
...
+        yield CNOT().on(ancillas[-1], output_bits[-1])
+        if len(input_bits) == len(output_bits):
+            yield MultiControlPauli((1, 1), cirq.X).on(control, input_bits[-1], output_bits[-1])
+            yield CNOT().on(ancillas[-1], output_bits[-1])
+        # right part of Fig.4
+        yield from self._right_building_block(
+            input_bits, output_bits, ancillas, control, self.b_dtype.bitsize - 2
+        )
+        yield And().adjoint().on(input_bits[0], output_bits[0], ancillas[0])
+        yield MultiControlPauli((1, 1), cirq.X).on(control, input_bits[0], output_bits[0])
```
fdmalone, line 221: Just a Toffoli right? Could this be another And gate?

qualtran/bloqs/arithmetic/controlled_addition_test.py
```diff
@@ -0,0 +1,71 @@
...
+    )
+    assert cirq.circuit_diagram_info(gate).wire_symbols == expected_wire_symbols
+
+
+@pytest.mark.parametrize("n", [*range(3, 10)])
+def test_addition_gate_counts_controlled(n: int):
+    add = ControlledAdd(QUInt(n), controlled=1)
+    qlt_testing.assert_valid_bloq_decomposition(add)
+    assert add.t_complexity() == add.decompose_bloq().t_complexity()
+    assert add.bloq_counts() == add.decompose_bloq().bloq_counts(generalizer=ignore_split_join)
```
fdmalone, line 71: Another thing to check is the formula for the expected value (addition is n - 1 toffolis, controlled addition is n toffolis).

fdmalone (changes requested):
Thanks for adding this, it should be very helpful. Some some small suggestions / comments.

qualtran/bloqs/arithmetic/controlled_addition.py
skushnir123: I am pretty sure that there is a wire_symbol test:
    expected_wire_symbols = (
        ("In(ctrl)",) + ("In(x)",) * num_bits_a + ("In(y)/Out(x+y)",) * num_bits_b
    )
    assert cirq.circuit_diagram_info(gate).wire_symbols == expected_wire_symbols

qualtran/bloqs/arithmetic/controlled_addition.py
skushnir123: I used MultiControlPauli because this dynamically uses the available free ancillas. Or was that not what you were asking?

qualtran/bloqs/arithmetic/controlled_addition_test.py
skushnir123: I added a test for the Toffoli count, although its not exactly n toffolis.

mpharrigan: where's the test?

qualtran/bloqs/arithmetic/controlled_addition.py
```diff
@@ -0,0 +1,237 @@
...
+            return {'ctrl': ctrl, 'a': a, 'b': int(math.fmod(a + b, N))}
+
+    def short_name(self) -> str:
+        return "a+b"
+
+    def _circuit_diagram_info_(self, _) -> cirq.CircuitDiagramInfo:
+        wire_symbols = ["In(ctrl)"]
+        wire_symbols += ["In(x)"] * self.a_dtype.bitsize
+        wire_symbols += ["In(y)/Out(x+y)"] * self.b_dtype.bitsize
+        return cirq.CircuitDiagramInfo(wire_symbols=wire_symbols)
```
mpharrigan, line 146: lets remove this and just use the `wire_symbol` method, which will need to be updated following #919

qualtran/bloqs/arithmetic/controlled_addition_test.py
skushnir123: assert add.t_complexity().t == t_count line 74

qualtran/bloqs/arithmetic/controlled_addition.py
skushnir123: done

qualtran/bloqs/arithmetic/controlled_addition.py
skushnir123: this method was removed per Matthew's comment

qualtran/bloqs/arithmetic/controlled_addition_test.py
mpharrigan: I thought controlled addition was supposed to be `n` toffolis (per above). The formula you're comparing against is 8*(n-1) + 4. What are these coefficients? If we assume 4 T gates per toffoli, this is a factor of two different

tanujkhattar: Controlled addition takes twice as many T / Toffoli's as a non-controlled addition. See https://arxiv.org/pdf/1709.06648 for reference. 

The confusion is being caused because of the way the code is structured (decomposition and `build_composite_graph` is fine, `_t_complexity_` should be refactored to have `4 * n_and` instead of `8 * n_toffoli`) - 1 $\text{And} / \text{And}^{\dagger}$ pair is hiding inside the `MultiControlPauli` bloq which corresponds to the rightmost $\text{And}  - CNOT - \text{And}^{\dagger}$ in Fig (a)

image

fdmalone: ^^ my bad, it's ~ + 1 more toffolis per bit, I misremembered.

skushnir123: Ok, I fixed it to be based on the number of and pairs. I think what I have now should be correct.

NoureldinYosri (approved):
I'm going to merge this now and in a followup PR I'll add the notebook, documentation, classical action, and more tests

Issue #801, opened 2024-03-18
Symbolic analysis for bloqs that expect a sequence of objects as inputs but only care about the length of the sequence for symbolic costs


tanujkhattar:
I couldn't find a way to define a sequence (list / tuple) using sympy where the length of the sequence is a symbolic parameter. This is often useful to specify costs of Bloqs which expect sequences as inputs but the call graph really only depends upon the length of this sequence. This is useful for a variety of Bloqs, like 

1. Bloqs that expect control values (`cv: Tuple[int, ...]`) or `CtrlSpec`. 
2. `GQSP` - https://github.com/quantumlib/Qualtran/pull/669#discussion_r1529250665
3. `QROM` - The T-cost depends mainly on the iteration length of the data while the clifford cost depends upon the actual values of the data
(I'm sure there are more Bloqs that we should append to this list)

Let's figure out a way to support this. One way would be to have a custom wrapper class for sequences that can be instantiated via either just `shape` (which can be symbolic) or the actual data (in which case `shape` can be derived). Open to other suggestions. 

cc @mpharrigan

mpharrigan:
Yeah, we could have a class that just has a shape and either raises an error if you try to index into it or yields sympy symbols like `Shaped(name='x', shape=(3,))[2] = Symbol('x_2')` 

Do we want to wrap our data or just accept `Union[NDArray, Shaped]`. This is consistent with how we generally do `n: Union[int, sympy.Expr]`

mpharrigan:
implemented in #867

PR #867, opened 2024-04-11
Symbolic `Shaped`

anurudhp:
Discussion: https://github.com/quantumlib/Qualtran/issues/801

### DIFF
--- a/qualtran/bloqs/hamiltonian_simulation/hamiltonian_simulation_by_gqsp.ipynb
+++ b/qualtran/bloqs/hamiltonian_simulation/hamiltonian_simulation_by_gqsp.ipynb
@@ -80,7 +80,7 @@
     "This therefore block-encodes $e^{-iHt}$ in the block where the signal qubit and selection registers are all $|0\\rangle$.\n",
     "\n",
     "#### References\n",
-    " - [Generalized Quantum Signal Processing](https://arxiv.org/abs/2308.01501).     Motlagh and Wiebe. (2023). Theorem 7, Corollary 8. \n",
+    " - [Generalized Quantum Signal Processing](https://arxiv.org/abs/2308.01501). Motlagh and Wiebe. (2023). Theorem 7, Corollary 8. \n",
     "\n",
     "#### Parameters\n",
     " - `walk_operator`: qubitization walk operator of $H$ constructed from SELECT and PREPARE oracles.\n",
@@ -129,6 +129,26 @@
    ]
   },
   {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "11baa846",
+   "metadata": {
+    "cq.autogen": "HamiltonianSimulationByGQSP.symbolic_hamsim_by_gqsp"
+   },
+   "outputs": [],
+   "source": [
+    "import sympy\n",
+    "\n",
+    "from qualtran.bloqs.hubbard_model import get_walk_operator_for_hubbard_model\n",
+    "\n",
+    "walk_op = get_walk_operator_for_hubbard_model(2, 2, 1, 1)\n",
+    "t, alpha, inv_eps = sympy.symbols(\"t alpha N\")\n",
+    "symbolic_hamsim_by_gqsp = HamiltonianSimulationByGQSP(\n",
+    "    walk_op, t=t, alpha=alpha, precision=1 / inv_eps\n",
+    ")"
+   ]
+  },
+  {
    "cell_type": "markdown",
    "id": "a6e29295",
    "metadata": {
@@ -148,8 +168,8 @@
    "outputs": [],
    "source": [
     "from qualtran.drawing import show_bloqs\n",
-    "show_bloqs([hubbard_time_evolution_by_gqsp],\n",
-    "           ['`hubbard_time_evolution_by_gqsp`'])"
+    "show_bloqs([hubbard_time_evolution_by_gqsp, symbolic_hamsim_by_gqsp],\n",
+    "           ['`hubbard_time_evolution_by_gqsp`', '`symbolic_hamsim_by_gqsp`'])"
    ]
   },
   {
@@ -171,21 +191,42 @@
    },
    "outputs": [],
    "source": [
-    "from qualtran.resource_counting.generalizers import ignore_split_join\n",
-    "hubbard_time_evolution_by_gqsp_g, hubbard_time_evolution_by_gqsp_sigma = hubbard_time_evolution_by_gqsp.call_graph(max_depth=1, generalizer=ignore_split_join)\n",
+    "from qualtran.resource_counting.generalizers import generalize_rotation_angle, ignore_split_join, ignore_alloc_free\n",
+    "hubbard_time_evolution_by_gqsp_g, hubbard_time_evolution_by_gqsp_sigma = hubbard_time_evolution_by_gqsp.call_graph(generalizer=[ignore_split_join, ignore_alloc_free, generalize_rotation_angle])\n",
     "show_call_graph(hubbard_time_evolution_by_gqsp_g)\n",
     "show_counts_sigma(hubbard_time_evolution_by_gqsp_sigma)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "7045d204b9fcd6f1",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from qualtran.resource_counting.generalizers import generalize_rotation_angle, ignore_split_join, ignore_alloc_free\n",
+    "_, symbolic_hamsim_by_gqsp_sigma = symbolic_hamsim_by_gqsp.call_graph(max_depth=2, generalizer=[ignore_split_join, ignore_alloc_free, generalize_rotation_angle])\n",
+    "show_counts_sigma(symbolic_hamsim_by_gqsp_sigma)"
    ]
   }
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "Python 3 (ipykernel)",
    "language": "python",
    "name": "python3"
   },
   "language_info": {
-   "name": "python"
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.11.8"
   }
  },
  "nbformat": 4,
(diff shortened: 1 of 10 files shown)


anurudhp:
@mpharrigan @tanujkhattar could you take a look at this?

qualtran/bloqs/qubitization_walk_operator.py
```diff
@@ -103,8 +105,12 @@ def decompose_from_registers(
...
         reflect_op = self.reflect.on_registers(**reflect_reg)
         for _ in range(self.power):
-            yield select_op
-            yield reflect_op
+            if self.is_adjoint:
+                yield cirq.inverse(reflect_op)
+                yield cirq.inverse(select_op)
+            else:
+                yield select_op
+                yield reflect_op
```
anurudhp, line 113: @tanujkhattar there was no bloq adjoint for `QubitizationWalkOperator`, so I fixed it manually. Ideally this should be fixed properly when refactoring `QubitizationWalkOperator`.

qualtran/bloqs/qsp/generalized_qsp.py
```diff
@@ -338,12 +342,17 @@ def signature(self) -> Signature:
...
-        P: Sequence[complex],
+        P: Union[Sequence[complex], Shaped],
         *,
         negative_power: int = 0,
         verify: bool = False,
         verify_precision=1e-7,
     ) -> 'GeneralizedQSP':
+        import cirq
+
+        if cirq.is_parameterized(P):
```
mpharrigan, line 353: do we need cirq for this?

qualtran/bloqs/qubitization_walk_operator.py
mpharrigan: why does the default fallback implementation of `Bloq.adjoint()` not work here?

qualtran/resource_counting/symbolic_counting_utils.py
```diff
@@ -24,6 +25,14 @@
 document(SymbolicFloat, """A floating point value or a sympy expression.""")
 
 
+@frozen
+class Shaped:
+    shape: tuple[SymbolicInt, ...] = field(validator=validators.instance_of(tuple))
```
mpharrigan, line 30: Could you write a motivating docstring? This is a qualtran-specific concept that needs to be thoroughly documented: what it is, why we need it, how you might use it

qualtran/resource_counting/symbolic_counting_utils.py
```diff
@@ -68,7 +77,19 @@ def smax(*args):
...
+    return min(*args)
+
+
 def acos(x: SymbolicFloat) -> SymbolicFloat:
     if not isinstance(x, sympy.Basic):
         return np.arccos(x)
     return sympy.acos(x)
+
+
+def length(x: Union[Sized, Shaped]) -> SymbolicInt:
```
mpharrigan, line 92: I think we need a better naming convention. To someone reading qualtran code, seeing a `length` function could be very confusing: what is this function? why not `len`? 

With the min function, you named it `smin` which I think is a good idea to disambiguate it from `min`. maybe `slen`. Another alternative would be to switch to using a namespaced name at the call site:

```python
import qualtran.resource_counting.symbolic_counting_utils as qlt_symb

n = qlt_symb.length(xxx)
```

but of course we'd still have to use names that don't clash with builtin `min`, `len`; so maybe the former convention is better.

qualtran/resource_counting/symbolic_counting_utils_test.py
```diff
@@ -42,3 +51,10 @@ def test_bit_length():
         assert x.bit_length() == bit_length(x)
         assert x.bit_length() == bit_length(x + 0.5)
         assert x.bit_length() == bit_length(x + 0.0001)
+
+
+@pytest.mark.parametrize("shape", [(4,), (1, 2), (1, 2, 3)])
+def test_shaped(shape: tuple[int, ...]):
+    shaped = Shaped(shape=shape)
```
mpharrigan, line 58: how about a test where the shape elements are symbols?

mpharrigan (commented):
very cool. Some polish things in the comments

qualtran/bloqs/hamiltonian_simulation/hamiltonian_simulation_by_gqsp.ipynb
```diff
@@ -80,7 +80,13 @@
     "This therefore block-encodes $e^{-iHt}$ in the block where the signal qubit and selection registers are all $|0\\rangle$.\n",
     "\n",
     "#### References\n",
-    " - [Generalized Quantum Signal Processing](https://arxiv.org/abs/2308.01501).     Motlagh and Wiebe. (2023). Theorem 7, Corollary 8.\n"
+    " - [Generalized Quantum Signal Processing](https://arxiv.org/abs/2308.01501).     Motlagh and Wiebe. (2023). Theorem 7, Corollary 8. \n",
```
mpharrigan, line 83: In the docstring, we're now (lightly) enforcing a particular format for our references. Can you remove the additional indentation from the second line of the reference please? (if there is more than one reference: separate them by a blank line)

qualtran/bloqs/qubitization_walk_operator.py
anurudhp:
```py
from qualtran.bloqs.qsp.generalized_qsp import GeneralizedQSP
from qualtran.bloqs.qubitization_walk_operator_test import get_walk_operator_for_1d_ising_model

W = get_walk_operator_for_1d_ising_model(2, 1e-4)
G = GeneralizedQSP.from_qsp_polynomial(W, (0.5, 0, 0.5), negative_power=1)
G.decompose_bloq()
```
produces
```
ValueError: Cannot control channel with non-unitary operators: Adjoint(subbloq=QubitizationWalkOperator(select=SelectPauliLCU(selection_bitsize=2, target_bitsize=2, select_unitaries=(cirq.DensePauliString('ZZ', coefficient=(2+0j)), cirq.DensePauliString('XI', coefficient=(-1+0j)), cirq.DensePauliString('IX', coefficient=(-1+0j))), control_val=None), prepare=StatePreparationAliasSampling(selection_registers=(Register(name='selection', dtype=BoundedQUInt(bitsize=2, iteration_length=3), shape=(), side=<Side.THRU: 3>),), alt=array([0, 0, 0]), keep=array([   0, 3072, 3072]), mu=12), control_val=None, power=1))
```

Because QubitizationWalkOperator still uses cirq style API to provide the decompositions.

qualtran/bloqs/qsp/generalized_qsp.py
anurudhp: actually no, I updated the `is_symbolic` function to mirror the implementation of `cirq.is_parameterized`.

qualtran/resource_counting/symbolic_counting_utils.py
anurudhp: used `slen` to match the existing symbolic functions.

qualtran/bloqs/hamiltonian_simulation/hamiltonian_simulation_by_gqsp.ipynb
anurudhp: oops forgot to fix in the last PR, fixed now.

qualtran/bloqs/qubitization_walk_operator.py
anurudhp: this has been reverted now, the default implementation works after overriding `Adjoint.controlled`

qualtran/bloqs/hamiltonian_simulation/hamiltonian_simulation_by_gqsp.py
```diff
@@ -180,18 +172,6 @@ def build_composite_bloq(self, bb: 'BloqBuilder', **soqs: 'SoquetT') -> Dict[str
 
         return soqs
 
-    def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
```
mpharrigan, line 183: what happened to this?

qualtran/bloqs/qsp/generalized_qsp.py
mpharrigan: sweeeeeeeeet

qualtran/resource_counting/symbolic_counting_utils.py
```diff
@@ -24,8 +25,45 @@
...
+
+    Returns:
+        True if any argument is either a sympy object,
+        or implements the `is_symbolic` method which returns True.
+    """
+
+    if len(args) != 1:
+        return any(is_symbolic(arg) for arg in args)
+
+    (arg,) = args
```
mpharrigan, line 58: handle `len(args)==0`?

qualtran/resource_counting/symbolic_counting_utils.py
```diff
@@ -24,8 +25,45 @@
...
+    if isinstance(arg, sympy.Basic):
```
mpharrigan, line 59: I think we should be checking `isinstance(arg, sympy.Expr)`, right?

qualtran/bloqs/hamiltonian_simulation/hamiltonian_simulation_by_gqsp.py
anurudhp: We don't need a specialization anymore, it works automatically through `GeneralizedQSP`'s call graph.

Before this, there was no way to build GQSP without the exact polynomial, so I had to override it here.

qualtran/resource_counting/symbolic_counting_utils.py
anurudhp: to my surprise, it seems `any(...)` for an empty generator returns `False`, so line 56 automatically handles this case

qualtran/resource_counting/symbolic_counting_utils.py
anurudhp: I'm not sure actually, I just copied it from the existing code. 

It seems boolean expressions are instances of `sympy.Basic` but not `sympy.Expr` (https://docs.sympy.org/latest/explanation/glossary.html#term-Expr), which we may want to use for symbolic controls?

qualtran/resource_counting/symbolic_counting_utils.py
anurudhp: I attempted to explain it a bit, could you check how it looks?

qualtran/resource_counting/symbolic_counting_utils.py
mpharrigan: Thanks. Interesting reading. Happy keeping this as Basic for now

qualtran/resource_counting/symbolic_counting_utils.py
mpharrigan: ah got it, thanks

mpharrigan (approved):
Very useful contribution for more symbolic costings!

Issue #791, opened 2024-03-13
How bloqs choose how to display themselves


mpharrigan:
You may want to print out a string representation of a bloq or control how it is displayed in a diagram. This issue lays out the current state of how a bloq can inform the library how it wishes to be displayed and future improvements.

## `__repr__`

The `__repr__` method on a Python class gives a string representation of the object. We strongly advocate defining bloqs as `attrs` frozen classes, which will give you a good repr automatically. In fact: you have to jump through a minor hoop to override the attrs repr. Custom classes should define a repr that matches the style of the attrs repr. The repr should be exceedingly close to valid Python code you can paste into a terminal and get a convincing facsimile of the object in question. 

## `__str__`

The `__str__` method on a Python class also gives a string representation of an object; but it can be more tailored to human readers (whereas the repr should be pretty close to executable Python code). Historically, we've relied on the Python/`attrs` default where we use the repr for `__str__`. 

I think `__str__` and `pretty_name` have significant overlap in how/where they're used, and we should probably rely on the `__str__` method for "a reasonably concise,  reasonably informative description of the bloq", see below.

### graphviz

The graphviz drawer for composite bloqs uses `Bloq.short_name()` for the bloq title and `Soquet.pretty()` for each soquet. This uses `Register.name`. Following #728, this should probably hook in to the wire_symbols

### musical score

The musical score drawer uses `Bloq.short_name()` to float a title of the bloq over its wire symbols and `Bloq.wire_symbols` to control the text or symbols used for each soquet.

### `Bloq.short_name()`

As you can see: `Bloq.short_name` is primarily used as the title of bloqs in diagrams where space is indeed at a premium. 
By default it is the (potentially truncated) name of the bloq class and doesn't contain any information about bloq attribute values (by default).

 - We should introduce the assumption that this name will always be adjacent to soquet labels/wire symbols. This means it can omit information that is obvious from the soquet labels. We can make this optional if a title for the bloq is completely redundant #728 
 - Note: `short_name` is also used to tag the quimb tensors. We can probably just remove this
 - I would like to rename this method to make it more obvious _how it is used_ rather than the length of the string it produces. Maybe `diagram_title` or something. Its name should be harmonious with `wire_symbols` or whatever we potentially rename that to.

## Call Graph diagram info

The call graph code currently uses `Bloq.pretty_name()` as a title and then some "bespoke" logic for adding a list of attribute name,value pairs. 
[...]

PR #869, opened 2024-04-11
Call graph drawer

mpharrigan:
- factor core graphviz interop into a private abstract base class. User-developers can subclass this for complete control over the drawing of call graphs.
 - Introduce a new, opt-in version that uses `Bloq.__str__()` per https://github.com/quantumlib/Qualtran/issues/791 and can include additional per-bloq information in a tabular format (e.g. costs)
 - Keep the existing `GraphvizCounts` unmodified so we can migrate bloqs to using `__str__` without blowing up the docs / existing diagrams. This will be removed/migrated in follow-up work


mpharrigan:
@fdmalone @tanujkhattar can one of you take a look

tanujkhattar (approved):

mpharrigan:
image

work in progress, but that's what the nodes-are-tables looks like

PR #871, opened 2024-04-12
bug fix for format_duration function

anneriet:
The `format_duration` function in `ui.py` converts the input incorrectly, I have rewritten the function so that it (should) convert the input correctly.

As an example: in the original `format_duration`, an input of 25 ms gives the output as 1.04166666667 **days**, which is obviously incorrect.

### DIFF
--- a/qualtran/surface_code/ui.py
+++ b/qualtran/surface_code/ui.py
@@ -403,21 +403,21 @@ def format_duration(duration: Sequence[float]) -> Tuple[str, Sequence[float]]:
     Finds the best unit to report `duration` and assumes that `duration` is initially in us.
     """
     unit = 'us'
-    if duration[0] > 1000:
+    if duration[0] > 86400_000_000:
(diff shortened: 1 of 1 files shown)


anurudhp:
suggestion: one could also use https://humanfriendly.readthedocs.io/en/latest/api.html#humanfriendly.format_timespan
```py
humanfriendly.format_timespan(t, max_units=1)
```

anneriet:
Totally fine by me, there are many different ways to (correctly) convert the timespan into something more human-readable, I just wanted to report this bug in a constructive way!

mpharrigan:
@NoureldinYosri 

can you or @NoureldinYosri add a simple unit test to sanity check a couple of values?

mpharrigan:
I'll merge this for now since it's a big bug. @NoureldinYosri to follow up with a test

NoureldinYosri (approved):

PR #875, opened 2024-04-15
More improvements to QVR Bloqs and their analysis

tanujkhattar:
This PR makes a few more improvements to QVR bloqs and their analysis notebook and documentation. Specifically, we now take into consideration the asymptotic improvement in `QvrZPow` due to ignoring small angle rotations in order to synthesize rotations upto accuracy $\epsilon$. 

This leads to this updated plot where you can now see that for a given accuracy $eps = 1e-12$, `QvrZPow` asymptotes at a specific cost instead of continuing to scale with $n$. This is because once $n$ becomes large enough, we can truncate small angle rotations

image


qualtran/bloqs/rotations/quantum_variable_rotation.ipynb
```diff
@@ -142,7 +154,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "32069e11",
+   "id": "c141bb25",
```
mpharrigan, line 157: do you know how all these ids got changed? would it be possible to revert these?

mpharrigan (approved):

qualtran/bloqs/rotations/quantum_variable_rotation.ipynb
tanujkhattar: I regenerated the notebook since the docstrings are updated, maybe that triggered the id change? I'm not sure. Is it worth manually reverting it?

PR #876, opened 2024-04-16
Mark tests slow

tanujkhattar:
Mark tests as slow for state prep alias sampling and add a quick version. For reflection using prepare, reduce the range of tests.


mpharrigan (approved):
Thanks!

Issue #865, opened 2024-04-09
`GateWithRegisters.controlled` does not work with call graph


anurudhp:
https://github.com/quantumlib/Qualtran/pull/848#discussion_r1558277485

Replacing `Controlled(self.U, CtrlSpec(cvs=0))` with `self.U.controlled(control_values=[0])` in GQSP results in:

```
AttributeError: 'BloqAsCirqGate' object has no attribute 'pretty_name'
```

Minimal code to reproduce error:
```py
from qualtran.bloqs.basic_gates import XPowGate
from qualtran.bloqs.qsp.generalized_qsp import GeneralizedQSP
from qualtran.drawing import show_call_graph

gqsp = GeneralizedQSP.from_qsp_polynomial(XPowGate(), (0.5, 0.5))
g, _ = gqsp.call_graph(max_depth=1)
show_call_graph(g)
```

mpharrigan:
I can't reproduce this on `main`. There was a recent PR merged that moved `controlled` to the bloqs one; maybe @tanujkhattar knows more

anurudhp:
It still crashes for me on main, on replacing
https://github.com/quantumlib/Qualtran/blob/43e3032f0f35b427b3bde984d39a2a7cb749974b/qualtran/bloqs/qsp/generalized_qsp.py#L404
with
```py
counts.add((self.U.controlled(control_values=[0]), degree - self.negative_power))
```
(in `build_call_graph`)

Issue #686, opened 2024-02-18
Divergent behaviour of controlled: `GateWithRegisters.controlled` vs. `Bloq.controlled`


anurudhp:
Ran into this issue while testing GQSP's call graph. Minimal code to reproduce:
```py
import numpy as np
from qualtran.bloqs.generalized_qsp_test import RandomGate

U = RandomGate.create(bitsize=1, random_state=np.random.RandomState(42))
print(U.call_graph()) # works
print(U.controlled().call_graph()) # fails
```
output:
```
(<networkx.classes.digraph.DiGraph object at 0x7fc8491550d0>, {RandomGate(bitsize=1, matrix=array([[ 0.27702459-0.13059069j, -0.76350488+0.56856287j],
       [ 0.36122517+0.88074958j,  0.1700831 +0.25469254j]])): 1})

Traceback (most recent call last):
  File "test.py", line 6, in <module>
    print(U.controlled().call_graph())
          ^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'ControlledGate' object has no attribute 'call_graph'
```

mpharrigan:
I'll look at this. Since `cirq.Gate` and `Bloq`both have a method named `controlled` we have to pick the most useful default. 

A workaround if you have access to the source code for the problematic gate is to call the correct one

```python
class MyGWR(GateWithRegisters):
  ...

  def controlled(self, *args, **kwargs):
    return Bloq.controlled(*args, **kwargs)
```

anurudhp:
Thanks!

So I was using this for Hamiltonian simulation by GQSP, which uses QubitizationWalkOperator. @tanujkhattar, I believe it would be a major refactor to make it and all its dependencies Bloqs?

fpapa250:
Running into this error in https://github.com/quantumlib/Qualtran/pull/780. Specifically with `MontgomeryModInv()` bloq whose call graph involves `Add().controlled()` and `LinearDepthGreaterThan().controlled()`.

Code to reproduce:
`from qualtran.drawing import show_call_graph
from qualtran.bloqs.factoring.mod_mul import MontgomeryModInv

bloq = MontgomeryModInv(bitsize=5, p=6)

graph, sigma = bloq.call_graph()
show_call_graph(graph)`

mpharrigan:
reflecting on this, we should probably change the default `GateWithRegisters.controlled` to be the bloqs one and then use the workaround for cases when we need the cirq behavior

[...]

PR #879, opened 2024-04-16
More improvements to the Controlled infrastructure

tanujkhattar:
Fixes https://github.com/quantumlib/Qualtran/issues/865
Fixes https://github.com/quantumlib/Qualtran/issues/686

The improvements are -

- `GateWithRegisters.controlled()` now returns `Controlled` bloq
- `cirq_gate_to_bloq` now maps a `cirq.Controlled` gates to `Controlled` bloq.
- `Controlled` bloq is now a `GateWithRegisters`. This is required because we want to support cirq style simulations for circuits containing `Controlled(Rx())` bloqs and `BloqAsCirqGate(Controlled())` is not sufficient because we need special logic to derive the unitary matrix for controlled bloqs (eg: when subbloq is a GateWithRegisters but doesn't have only THRU registers). 

After this PR, we should never need to invoke the `cirq.Controlled` gates in either cirq-style or bloq-style API in Qualtran and we should always use `Controlled` bloq in all cases. All existing uses of `cirq.Controlled` gates would be updated to `Controlled` bloq when invoking the bloq-style API using the `cirq_gate_to_bloq` mapping. 

This is the final PR in my series of PRs to improve the controlled infrastructure. cc @mpharrigan

### DIFF
--- a/qualtran/_infra/adjoint.py
+++ b/qualtran/_infra/adjoint.py
@@ -150,6 +150,13 @@ class Adjoint(GateWithRegisters):
             return cirq.inverse(self.subbloq.decompose_from_registers(context=context, **quregs))
         return super().decompose_from_registers(context=context, **quregs)
 
+    def _circuit_diagram_info_(self, args: 'cirq.CircuitDiagramInfoArgs'):
+        sub_info = cirq.circuit_diagram_info(self.subbloq, args, default=NotImplemented)
+        if sub_info is NotImplemented:
+            return NotImplemented
+        sub_info.exponent *= -1
+        return sub_info
+
     def supports_decompose_bloq(self) -> bool:
         """Delegate to `subbloq.supports_decompose_bloq()`"""
         return self.subbloq.supports_decompose_bloq()
--- a/qualtran/_infra/adjoint.py
+++ b/qualtran/_infra/adjoint.py
@@ -150,7 +150,9 @@ class Adjoint(GateWithRegisters):
             return cirq.inverse(self.subbloq.decompose_from_registers(context=context, **quregs))
         return super().decompose_from_registers(context=context, **quregs)
 
-    def _circuit_diagram_info_(self, args: 'cirq.CircuitDiagramInfoArgs'):
+    def _circuit_diagram_info_(
+        self, args: 'cirq.CircuitDiagramInfoArgs'
+    ) -> cirq.CircuitDiagramInfo:
         sub_info = cirq.circuit_diagram_info(self.subbloq, args, default=NotImplemented)
         if sub_info is NotImplemented:
             return NotImplemented
--- a/qualtran/_infra/controlled.py
+++ b/qualtran/_infra/controlled.py
@@ -34,7 +34,7 @@ from numpy.typing import NDArray
 
 from .bloq import Bloq
 from .data_types import QBit, QDType
-from .gate_with_registers import GateWithRegisters, merge_qubits
+from .gate_with_registers import GateWithRegisters
 from .registers import Register, Side, Signature
 
 if TYPE_CHECKING:
--- a/qualtran/_infra/controlled.py
+++ b/qualtran/_infra/controlled.py
@@ -418,7 +418,8 @@ class Controlled(GateWithRegisters):
                 cirq.ControlledGate(self.subbloq, control_values=self.ctrl_spec.to_cirq_cv())
             )
         if all(reg.side == Side.THRU for reg in self.subbloq.signature):
-            # subbloq has only THRU registers, so the unitary is well defined.
+            # subbloq has only THRU registers, so the tensor contraction corresponds
+            # to a unitary matrix.
             return self.tensor_contract()
         # Unable to determine the unitary effect.
         return NotImplemented
--- a/qualtran/_infra/controlled.py
+++ b/qualtran/_infra/controlled.py
@@ -34,6 +34,7 @@ from numpy.typing import NDArray
 
 from .bloq import Bloq
 from .data_types import QBit, QDType
+from .gate_with_registers import GateWithRegisters, merge_qubits
 from .registers import Register, Side, Signature
 
 if TYPE_CHECKING:
@@ -278,7 +279,7 @@ def _get_nice_ctrl_reg_names(reg_names: List[str], n: int) -> Tuple[str, ...]:
 
 
 @attrs.frozen
-class Controlled(Bloq):
+class Controlled(GateWithRegisters):
     """A controlled version of `subbloq`.
 
     This meta-bloq is part of the 'controlled' protocol. As a default fallback,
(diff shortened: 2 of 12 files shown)


qualtran/_infra/gate_with_registers.py
```diff
@@ -311,23 +310,40 @@ def on_registers(
...
-
-        controlled_gate = cirq.ControlledGate(
-            self,
-            num_controls=num_controls,
-            control_values=control_values,
-            control_qid_shape=control_qid_shape,
-        )
-        ctrl_spec = CtrlSpec.from_cirq_cv(controlled_gate.control_values)
-        return BloqAsCirqGate(Controlled(self, ctrl_spec))
+    ) -> 'GateWithRegisters':
```
anurudhp, line 326: Would it be useful to provide overloaded type hints?
```py
    # pylint: disable=arguments-renamed
    @overload
    def controlled(
        self,
        num_controls: Optional[int] = None,
        control_values=None,
        control_qid_shape: Optional[Tuple[int, ...]] = None,
    ) -> 'GateWithRegisters':
        ...

    # pylint: disable=signature-differs
    @overload
    def controlled(self, ctrl_spec: 'CtrlSpec') -> 'GateWithRegisters':
        ...

    # pylint: disable=arguments-renamed
    def controlled(self, num_controls=None, control_values=None, control_qid_shape=None):
```

Right now it's difficult to understand the usage without looking at the function body.

qualtran/_infra/adjoint.py
```diff
@@ -150,6 +150,13 @@ def decompose_from_registers(
             return cirq.inverse(self.subbloq.decompose_from_registers(context=context, **quregs))
         return super().decompose_from_registers(context=context, **quregs)
 
+    def _circuit_diagram_info_(self, args: 'cirq.CircuitDiagramInfoArgs'):
```
mpharrigan, line 153: return type annotation

qualtran/_infra/controlled.py
```diff
@@ -410,6 +411,18 @@ def add_my_tensors(
...
         tn.add(qtn.Tensor(data=data, inds=out_ind + in_ind, tags=[self.short_name(), tag]))
 
+    def _unitary_(self):
+        if isinstance(self.subbloq, GateWithRegisters):
+            # subbloq is a cirq gate, use the cirq-style API to derive a unitary.
+            return cirq.unitary(
+                cirq.ControlledGate(self.subbloq, control_values=self.ctrl_spec.to_cirq_cv())
+            )
+        if all(reg.side == Side.THRU for reg in self.subbloq.signature):
+            # subbloq has only THRU registers, so the unitary is well defined.
```
mpharrigan, line 421: pycharm tells me this is supposed to be "well-defined"

qualtran/_infra/gate_with_registers.py
mpharrigan: yeah, I think this needs a big docstring that describes the two modes of operation

qualtran/bloqs/phase_estimation/text_book_qpe.py
```diff
@@ -220,13 +222,14 @@ def decompose_from_registers(
...
+        target_qubits = merge_qubits(self.target_registers, **target_quregs)
 
-        phase_qubits = quregs['qpe_reg']
+        phase_qubits: Sequence[cirq.Qid] = quregs['qpe_reg'].flatten()
 
         yield self.ctrl_state_prep.on(*phase_qubits)
         for i, qbit in enumerate(phase_qubits[::-1]):
-            yield cirq.pow(unitary_op.controlled_by(qbit), 2**i)
+            yield Power(self.unitary.controlled(), 2**i).on(qbit, *target_qubits)
+            # yield cirq.pow(unitary_op.controlled_by(qbit), 2**i)
```
mpharrigan, line 232: remove old code

mpharrigan (approved):
looks good. approval but please fix the issues raised

qualtran/bloqs/qsp/generalized_qsp.py
```diff
@@ -382,14 +374,15 @@ def decompose_from_registers(
...
         for signal_rotation in self.signal_rotations[1:]:
             if num_inverse_applications > 0:
                 # apply C-U^\dagger
-                yield self.U.adjoint().on_registers(**quregs).controlled_by(signal_qubit)
+                yield self.U.adjoint().controlled().on(signal_qubit, *flat_qubits_for_u)
                 num_inverse_applications -= 1
             else:
                 # apply C[0]-U
-                yield self.U.on_registers(**quregs).controlled_by(signal_qubit, control_values=[0])
+                yield self.U.controlled(control_values=[0]).on(signal_qubit, *flat_qubits_for_u)
```
anurudhp, line 385: Should we default to using qualtran style controls `CtrlSpec(cvs=0)` instead of cirq style `control_values=[0]`?

qualtran/_infra/gate_with_registers.py
tanujkhattar: Added a detailed docstring and the overloads. Also added tests to verify the different cases raise errors.

qualtran/_infra/adjoint.py
tanujkhattar: done

qualtran/_infra/controlled.py
tanujkhattar: reworded

qualtran/bloqs/phase_estimation/text_book_qpe.py
tanujkhattar: Reverted the file so now there are no changes and the original implementation uses bloqs instead of cirq gates.

qualtran/bloqs/qsp/generalized_qsp.py
tanujkhattar: Reverted this change because `self.U.on_registers(**quregs).controlled_by(signal_qubit)` now uses the `Controlled` bloq instead of cirq gate. The `op.controlled_by` still expects the cirq-style API since we don't have a custom `GWROperation` where we can override the `controlled_by` to expect the bloq-style API as well. But I think it's fine right now to do things this way

tanujkhattar:
@mpharrigan I'm ready to merge this but feel free to take another look since I've made some minor upgrades. Using `gwr.controlled().on_registers()` and `gwr.on_registers().controlled_by` are now equivalent and both result `GateOperation(Controlled(gwr))`  where the wrapped gate is a the `Controlled` bloq. 

Tl;Dr - the discrepancy in usage from a users perspective that I described in Qualtran sync today is now gone and users don't need to worry about the intricacies when using the Cirq or Bloq style APIs.

qualtran/_infra/gate_with_registers.py
```diff
@@ -311,39 +323,142 @@ def on_registers(
...
+        to generate a CtrlSpec from the cirq-style API and thus easily support both Cirq & Bloq
+        APIs. For example
+
+        >>> class CustomGWR(GateWithRegisters):
+        >>>     def controlled(self, *args, **kwargs) -> 'Bloq':
+        >>>         ctrl_spec = self._get_ctrl_spec(*args, **kwargs)
+        >>>         # Use ctrl_spec to construct a controlled version of `self`.
+
+        Args:
+            num_controls:
```
mpharrigan, line 359: ...

mpharrigan:
where is the logic for `controlled_by`?

tanujkhattar:
> where is the logic for controlled_by?

No override is needed because of the way Cirq and Qualtran works. `GateOperation.controlled_by` delegates to `Gate.controlled()` by default, which now uses `GateWithRegisters.controlled()` which returns a `Controlled(GateWithRegisters)` which is a cirq gate and gets wrapped in the `GateOperation(controlled_bloq)`. 

This was true today morning as well, I just realized it when making these changes and thus added tests to `gate_with_registers_test.py` to make this concrete.

qualtran/_infra/gate_with_registers.py
tanujkhattar: Fixed

tanujkhattar:
Merging this now

PR #889, opened 2024-04-23
First pass of type checking in Qualtran

dstrain115:
- Fixes many mypy issues in Qualtran.

Some highlights:
- Disable override type checking since overrides or variable kwargs are not appreciated by mypy
- Remove return type for BloqBuilder.add since many places assume it returns a tuple.
- Make Soquet a np.generic so NDArray[Soquet] works
- Ignore NDArray[cirq.Qid] type issues

### DIFF
--- a/dev_tools/qualtran_dev_tools/reference_docs.py
+++ b/dev_tools/qualtran_dev_tools/reference_docs.py
@@ -94,11 +94,11 @@ def mixin_custom_template(template_name: str) -> Type:
     return _CustomTemplateMixin
 
 
-class MyModulePageBuilder(mixin_custom_template('module'), ModulePageBuilder):
+class MyModulePageBuilder(mixin_custom_template('module'), ModulePageBuilder):  # type: ignore[misc]
     """Use a custom template for module pages."""
 
 
-class MyClassPageBuilder(mixin_custom_template('class'), ClassPageBuilder):
+class MyClassPageBuilder(mixin_custom_template('class'), ClassPageBuilder):  # type: ignore[misc]
     """Use a custom template for class pages.
 
     Additionally, this will re-sort the class members (i.e. methods) to match
@@ -120,11 +120,11 @@ class MyClassPageBuilder(mixin_custom_template('class'), ClassPageBuilder):
         )
 
 
-class MyFunctionPageBuilder(mixin_custom_template('function'), FunctionPageBuilder):
+class MyFunctionPageBuilder(mixin_custom_template('function'), FunctionPageBuilder):  # type: ignore[misc]
     """Use a custom template for function pages."""
 
 
-class MyTypeAliasPageBuilder(mixin_custom_template('type_alias'), TypeAliasPageBuilder):
+class MyTypeAliasPageBuilder(mixin_custom_template('type_alias'), TypeAliasPageBuilder):  # type: ignore[misc]
     """Use a custom template for type alias pages."""
 
 
--- a/dev_tools/requirements/deps/mypy.txt
+++ b/dev_tools/requirements/deps/mypy.txt
@@ -1,3 +1,3 @@
 # the mypy dependency file
-mypy~=0.991.0
+mypy
 mypy-protobuf
--- a/qualtran/_infra/controlled_test.py
+++ b/qualtran/_infra/controlled_test.py
@@ -11,7 +11,7 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-from typing import Dict, List, Tuple
+from typing import Dict, List, Tuple, TYPE_CHECKING
 
 import attrs
 import cirq
@@ -48,6 +48,9 @@ from qualtran.bloqs.mcmt import And
 from qualtran.cirq_interop.testing import GateHelper
 from qualtran.drawing import get_musical_score_data
 from qualtran.drawing.musical_score import Circle, SoqData, TextBox
+
+if TYPE_CHECKING:
+    from qualtran import SoquetT
 
 
 def test_ctrl_spec():
@@ -374,7 +377,7 @@ class TestCtrlStatePrepAnd(Bloq):
         and_ctrl = [bb.add(one_or_zero[cv]) for cv in self.and_ctrl]
 
         ctrl_soqs = bb.add_d(cbloq, **ctrl_soqs, ctrl=and_ctrl)
-        out_soqs = [*ctrl_soqs.pop('ctrl'), ctrl_soqs.pop('target')]
+        out_soqs = np.asarray([*ctrl_soqs.pop('ctrl'), ctrl_soqs.pop('target')])
 
         for reg, cvs in zip(cbloq.ctrl_regs, self.ctrl_spec.cvs):
             for idx in reg.all_idxs():
--- a/qualtran/_infra/quantum_graph.py
+++ b/qualtran/_infra/quantum_graph.py
@@ -16,6 +16,7 @@
 
 from typing import Tuple, TYPE_CHECKING, Union
 
+import numpy as np
 from attrs import field, frozen
 
 if TYPE_CHECKING:
@@ -79,7 +80,7 @@ def _to_tuple(x: Union[int, Tuple[int, ...]]) -> Tuple[int, ...]:
 
 
 @frozen
-class Soquet:
+class Soquet(np.generic):
     """One half of a connection.
 
     Users should not construct these directly. They should be marshalled
--- a/qualtran/bloqs/basic_gates/rotation.py
+++ b/qualtran/bloqs/basic_gates/rotation.py
@@ -13,10 +13,11 @@
 #  limitations under the License.
 
 from functools import cached_property
-from typing import Protocol
+from typing import Protocol, Union
 
 import cirq
 import numpy as np
+import sympy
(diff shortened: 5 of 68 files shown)


qualtran/_infra/controlled_test.py
```diff
@@ -374,7 +377,7 @@ def build_composite_bloq(self, bb: 'BloqBuilder') -> Dict[str, 'SoquetT']:
         and_ctrl = [bb.add(one_or_zero[cv]) for cv in self.and_ctrl]
 
         ctrl_soqs = bb.add_d(cbloq, **ctrl_soqs, ctrl=and_ctrl)
-        out_soqs = [*ctrl_soqs.pop('ctrl'), ctrl_soqs.pop('target')]
+        out_soqs = np.asarray([*ctrl_soqs.pop('ctrl'), ctrl_soqs.pop('target')])
```
mpharrigan, line 380: here's another philosophical question. Really, the return value is supposed to be a numpy array. But we've coded the library to still work if it's a list (by sprinkling defensive `np.asarray` calls throughout). Is it worth keeping some of these badly-typed examples in the tests to keep those code paths exercised?

dev_tools/qualtran_dev_tools/reference_docs.py
```diff
@@ -120,11 +120,11 @@ def __init__(self, page_info):
         )
 
 
-class MyFunctionPageBuilder(mixin_custom_template('function'), FunctionPageBuilder):
+class MyFunctionPageBuilder(mixin_custom_template('function'), FunctionPageBuilder):  # type: ignore[misc]
     """Use a custom template for function pages."""
 
 
-class MyTypeAliasPageBuilder(mixin_custom_template('type_alias'), TypeAliasPageBuilder):
+class MyTypeAliasPageBuilder(mixin_custom_template('type_alias'), TypeAliasPageBuilder):  # type: ignore[misc]
```
mpharrigan, line 127: for my edification: does `[misc]` refer to a specific mypy error? class of error?

qualtran/_infra/controlled_test.py
dstrain115: I think that if you want to officially support Lists, then the function declaration should state it.

dev_tools/qualtran_dev_tools/reference_docs.py
dstrain115: misc is for a bunch of random stuff that are relatively rare errors.  In this case, it is because mypy doesn't allow dynamically created mix-ins as base classes.  My understanding is that this is fine but is a limitation of mypy, but this is starting to get into territory I am not super familiar with.

qualtran/bloqs/basic_gates/rotation.py
```diff
@@ -263,7 +264,7 @@ class Rz(CirqGateAsBloqBase):
         of z-rotations](https://arxiv.org/pdf/1403.2975.pdf).
     """
 
-    angle: float
+    angle: Union[sympy.expr, float]
```
mpharrigan, line 267: is this supposed to be `Expr`?

qualtran/bloqs/mcmt/and_bloq_test.py
```diff
@@ -186,6 +186,8 @@ def signature(self) -> 'Signature':
     def build_composite_bloq(
         self, bb: 'BloqBuilder', q0: 'SoquetT', q1: 'SoquetT'
     ) -> Dict[str, 'SoquetT']:
+        assert isinstance(q0, Soquet)
+        assert isinstance(q1, Soquet)
```
mpharrigan, line 190: is this sort of thing going to be generally necessary? presumably one could also change the method signature to assign these the types `Soquet` instead of `SoquetT`, right?

qualtran/_infra/quantum_graph.py
```diff
@@ -79,7 +80,7 @@ def _to_tuple(x: Union[int, Tuple[int, ...]]) -> Tuple[int, ...]:
 
 
 @frozen
-class Soquet:
+class Soquet(np.generic):
```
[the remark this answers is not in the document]
dstrain115: I am not super clear on this.  I think so?  The documentation is also unclear, but my reading was just that you had to inherit from it:

`Class from which most (all?) numpy scalar types are derived. For consistency, exposes the same API as [ndarray](https://numpy.org/doc/1.15/reference/generated/numpy.ndarray.html#numpy.ndarray), despite many consequent attributes being either “get-only,” or completely irrelevant. This is the class from which it is strongly suggested users should derive custom scalar types.`

qualtran/bloqs/mcmt/and_bloq_test.py
dstrain115: Yes, a lot of these are since SoquetT can be a numpy array of Soquet.  Changing the type of the return signature would also fix this.  

I'm not sure I really have the bandwidth to figure whether this is possible on a case-by-case basis with 1400+ mypy errors to sort through and will generally opt for the simple solution unless reviewers point out cases where we want to change it.

qualtran/bloqs/state_preparation/state_preparation_via_rotation_test.py
```diff
@@ -64,8 +64,9 @@ def test_state_prep_via_rotation(bloq_autotester):
     ],
 )
 def test_exact_state_prep_via_rotation_(phase_bitsize: int, state_coefs: Tuple[complex, ...]):
+    # https://github.com/python/mypy/issues/5313
```
mpharrigan, line 67: we make heavy use of attrs converters, so this is one to look out for. A lot of places, our classes need a sequence of something and need to be hashable, so we convert to `Tuple[T, ...]` and I've noticed its very hard to get the Pycharm type checker and pyright (which are the two I've played around with before) to accept this.

qualtran/bloqs/util_bloqs.py
```diff
@@ -166,9 +165,7 @@ def add_my_tensors(
     def on_classical_vals(self, reg: 'NDArray[np.uint8]') -> Dict[str, int]:
         return {'reg': bits_to_ints(reg)[0]}
 
-    def get_ctrl_system(
-        self, ctrl_spec: Optional['AbstractCtrlSpec'] = None
-    ) -> Tuple['Bloq', 'AddControlledT']:
+    def get_ctrl_system(self, ctrl_spec=None) -> Tuple['Bloq', 'AddControlledT']:
```
mpharrigan, line 168: what was wrong with the annotation that used to be here?

dev_tools/qualtran_dev_tools/reference_docs.py
mpharrigan: Makes sense. This is in the dev_tools, so less essential to keep everything perfect and static

qualtran/_infra/quantum_graph.py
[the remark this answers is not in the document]
mpharrigan: I read that too :) I guess many of the attributes will be completely irrelevant in our case

qualtran/bloqs/mcmt/and_bloq_test.py
mpharrigan: What do you mean by changing the return type signature? It's currently `Dict[str, 'SoquetT']` which means these can be single soquets or arrays thereof, no?

dstrain115: I mean your original suggestion "presumably one could also change the method signature to assign these the types Soquet instead of SoquetT"

qualtran/bloqs/util_bloqs.py
dstrain115: AbstractCtrlSpec doesn't exist?  Was this refactored?  I couldn't find it anywhere.  (Unless I missed it)

qualtran/bloqs/state_preparation/state_preparation_via_rotation_test.py
dstrain115: Yeah, this is annoying that mypy craps out with this.  It looks like it doesn't understand the auto-converter.  It looked fine to me, so I have just been type: ignoring these.

qualtran/bloqs/util_bloqs.py
mpharrigan: OH -- yeah that would be a problem. I squashed `CtrlSpec` and `AbstractCtrlSpec` into `CtrlSpec` during that PR but must have missed this

dev_tools/requirements/deps/mypy.txt
```diff
@@ -1,3 +1,3 @@
 # the mypy dependency file
-mypy~=0.991.0
-mypy-protobuf
\ No newline at end of file
+mypy
```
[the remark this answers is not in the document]
dstrain115: Done.

qualtran/_infra/quantum_graph.py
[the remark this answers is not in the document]
dstrain115: Actually, I am going to remove this for now.  It seems to give an error "TypeError: cannot create 'Soquet' instances".  Will investigate further in a different PR.

qualtran/bloqs/basic_gates/rotation.py
dstrain115: Done.

qualtran/bloqs/util_bloqs.py
dstrain115: Changed to Optional['CtrlSpec']

PR #890, opened 2024-04-24
`QubitizationWalkOperator` add attribute: sum of LCU coefficients

### LINKED ISSUE #868: QubitizationWalkOperator: add normalization factor to the API
The complexity of algorithms using a qubitization walk operator usually depends on the sum of coefficients (called `lambda`)
https://github.com/quantumlib/Qualtran/blob/a23da98bfe5a8edd36c4ca47f878b4ca25a35f5b/qualtran/bloqs/qubitization_walk_operator.py#L52-L53

For example, in hamiltonian simulation by GQSP, the complexity is $O(\lambda t + ...)$ ($\lambda$ is `alpha` in the code below)
https://github.com/quantumlib/Qualtran/blob/a23da98bfe5a8edd36c4ca47f878b4ca25a35f5b/qualtran/bloqs/hamiltonian_simulation/hamiltonian_simulation_by_gqsp.py#L96-L99

cc @tanujkhattar

anurudhp:
fixes #868 

The sum of coefficients determines the complexity of hamiltonian simulation. This will also enable us to propagate symbolic values to the prepare gate.

@tanujkhattar could you review this? Also it would be great if someone with qchem expertise checks whether the sum of coeffs for the Hubbard gate is correct.


qualtran/bloqs/qubitization_walk_operator.py
```diff
@@ -64,6 +67,7 @@ class QubitizationWalkOperator(GateWithRegisters):
 
     select: SelectOracle
     prepare: PrepareOracle
+    sum_of_ham_coeffs: Optional[float]
```
fdmalone, line 70: The default value here should be 1 or something else?

fdmalone:
It's typically called lambda in the literature which is problematic in python as a variable name I know. Maybe sum_of_lcu_coeffs?

qualtran/bloqs/qubitization_walk_operator.py
anurudhp: I didn't set a default to avoid accidentally forgetting it while building walk ops

But as the `sum_of_ham_coeffs` attribute is not needed in the decomposition/call-graph, it won't break anything. It's only needed for Hamiltonian simulation, in which case it needs to be correctly calculated.

fdmalone: right, I think Optional[float] = None in that case.

anurudhp: fixed

anurudhp: would it be better to not have a default, and let the user explicitly specify `None` when they don't know the value? (just to make sure it's not accidentally omitted)

tanujkhattar:
Is `QubitizationWalkOperator` the right place to add this attribute? Should we instead add it to the `PrepareOracle` and just export it as a property in the `QubitizationWalkOperator` ?

anurudhp:
Yes I think that makes sense. That way we can also implicitly compute this value because we usually pass the exact coeffs to Prepare (which are normalized there).

For ex. it would simplify this (we can drop L93) https://github.com/quantumlib/Qualtran/blob/91b80b1ed561829eb14c66a1af1c385074929fe7/qualtran/bloqs/chemistry/ising/hamiltonian.py#L92-L94

anurudhp:
@tanujkhattar PTAL

qualtran/bloqs/qubitization_walk_operator.py
```diff
@@ -92,6 +94,11 @@ def signature(self) -> Signature:
     def reflect(self) -> ReflectionUsingPrepare:
         return ReflectionUsingPrepare(self.prepare, control_val=self.control_val, global_phase=-1)
 
+    @cached_property
+    def sum_of_lcu_coefficients(self) -> Optional[float]:
+        r"""value of $\lambda$, i.e. sum of the coefficients $w_l$."""
```
tanujkhattar, line 99: Let's clarify by writing "sum of absolute values of coefficients. " instead. In this case, the docstring of the class does clarify that each $w_l$ is a positive integer so it's just a sum of coefficients.

qualtran/bloqs/select_and_prepare.py
```diff
@@ -63,14 +66,15 @@ class PrepareOracle(GateWithRegisters):
     r"""Abstract base class that defines the API for a PREPARE Oracle.
 
     Given a set of coefficients $\{c_0, c_1, ..., c_{N - 1}\}, the PREPARE oracle is used to encode
-    the coefficients as amplitudes of a state $|\Psi\rangle = \sum_{i=0}^{N-1}c_{i}|i\rangle$ using
-    a selection register $|i\rangle$. In order to prepare such a state, the PREPARE circuit is also
-    allowed to use a junk register that is entangled with selection register.
+    the coefficients as amplitudes of a state $|\Psi\rangle = \sum_{i=0}^{N-1} \sqrt{\frac{c_{i}}{\lambda}} |i\rangle$
+    where $\lambda = \sum_i c_i$, using a selection register $|i\rangle$. In order to prepare such
```
tanujkhattar, line 70:
```suggestion
    where $\lambda = \sum_i | c_i |$, using a selection register $|i\rangle$. In order to prepare such
```

qualtran/bloqs/select_and_prepare.py
```diff
@@ -86,3 +90,11 @@ def junk_registers(self) -> Tuple[Register, ...]:
     @cached_property
     def signature(self) -> Signature:
         return Signature([*self.selection_registers, *self.junk_registers])
+
+    @property
+    def l1_norm_of_coeffs(self) -> Optional['SymbolicFloat']:
+        r"""Sum of the coefficients $c_i$.
```
tanujkhattar, line 96:
```suggestion
        r"""Sum of the absolute values of coefficients $c_i$.
```

tanujkhattar (approved):
LGTM % nits

qualtran/bloqs/qubitization_walk_operator.py
anurudhp: fixed

qualtran/bloqs/select_and_prepare.py
anurudhp: fixed

qualtran/bloqs/select_and_prepare.py
anurudhp: fixed

Issue #652, opened 2024-02-09
Throw an error if bloq has adjoint as a variable?


fdmalone:
We still have a lot of bloqs accepting adjoints as arguments (many of these are my fault). We should  clean these up and maybe throw an error through the bloq autotesting?

mpharrigan:
How many bloqs still have `adjoint` as a class attribute?

PR #893, opened 2024-04-24
Remove adjoint attribute from chemistry bloqs

fdmalone:
Removes a lot of adjoint variables from chemistry bloqs in favour if is_adjoint when necessary or defers to adjoint protocol elsewhere.


fdmalone:
Fixes #652

mpharrigan (approved):
Thanks Fionn!

PR #896, opened 2024-04-25
override FastDataBlock.n_logical_qubits

NoureldinYosri:


mpharrigan (approved):
consider adding an abstract base method

PR #897, opened 2024-04-25
Rename `RandomGate` -> `MatrixGate`

anurudhp:
The name `RandomGate` doesn't make sense because we can instantiate with a matrix of choice as well.
This also renames `RandomGate.create(...)` -> `MatrixGate.random(...)` which makes more sense.

@mpharrigan @tanujkhattar PTAL


mpharrigan (approved):
Makes sense to me

mpharrigan:
can you resolve merge conflicts, then I'll merge this in

anurudhp:
fixed

PR #899, opened 2024-04-25
Costs

mpharrigan:
Tracking PR

 - [x] preliminaries #900 
 - [x] costing framework https://github.com/quantumlib/Qualtran/pull/913
 - [x] Bloq/gate counts
 - [x] Qubit counts
 - [x] Success probability
 - [ ] Docs and niceties

PR #900, opened 2024-04-25
[Costs] get_bloq_callee_counts and call graph organization

mpharrigan:
- factor out `get_bloq_callee_counts`. Getting direct callees is a useful thing to be able to do. 
 - Organize modules. Existing module is renamed to "_call_graph" and made private. 

Preliminaries for #899


mpharrigan:
@tanujkhattar ptal

tanujkhattar (approved):

PR #891, opened 2024-04-24
Mypy type issues part 2

dstrain115:
- Continue fixing mypy type issues.

Some highlights include:
- Fix bloq_example decorator to pass through types
- Add custom converters for attrs tuple conversion to aid typing of attrs tuples with automatic converters.

PR #901, opened 2024-04-25
Fix coefficient type for `GlobalPhase`, and convert from cirq correctly

anurudhp:
The gate now supports symbolic phase coefficients.

The `_cirq_gate_to_bloq` was changed in #891 to only consider `float|complex`. This PR fixes the type in `GlobalPhase`, so that every `cirq.GlobalPhaseGate` can be converted to it.

### DIFF
--- a/qualtran/bloqs/basic_gates/global_phase.py
+++ b/qualtran/bloqs/basic_gates/global_phase.py
@@ -12,10 +12,11 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Union
 
 import attrs
 import cirq
+import sympy
 from attrs import frozen
 
 from qualtran import bloq_example, BloqDocSpec, DecomposeTypeError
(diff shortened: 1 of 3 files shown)


qualtran/bloqs/basic_gates/global_phase.py
```diff
@@ -46,7 +47,14 @@ def decompose_bloq(self) -> 'CompositeBloq':
...
 
     def adjoint(self) -> 'GlobalPhase':
-        return attrs.evolve(self, coefficient=complex(self.coefficient).conjugate())
+        from qualtran.resource_counting.symbolic_counting_utils import is_symbolic
+
+        coefficient = (
+            sympy.conjugate(self.coefficient)
+            if is_symbolic(self.coefficient)
+            else complex(self.coefficient).conjugate()
+        )
```
mpharrigan, line 56: worth adding as a helper function to `symbolic_counting_utils`?

mpharrigan (approved):

qualtran/bloqs/basic_gates/global_phase.py
anurudhp: moved it there, and added a new type alias `SymbolicComplex`

PR #902, opened 2024-04-26
Clean up and maintenance of checks

mpharrigan:
- consistent names and docstrings
 - skip slow bloq_examples from bloq report card


dev_tools/qualtran_dev_tools/bloq_report_card.py
```diff
@@ -110,6 +110,9 @@ def get_bloq_report_card(
         bclasses = get_bloq_classes()
     if bexamples is None:
         bexamples = get_bloq_examples()
+        # Default exclusions: pass explicit bexamples to override.
+        skips = ['qubitization_qpe_hubbard_model_small', 'qubitization_qpe_hubbard_model_large']
```
fdmalone, line 114: are these slow or why are they skipped?

fdmalone (approved):

dev_tools/qualtran_dev_tools/bloq_report_card.py
fdmalone: nvm, it's in the PR description 🤦

mpharrigan: they're slow

PR #903, opened 2024-04-28
Remove attribute `power` and default to `Power`/`GateWithRegisters.__pow__`

anurudhp:
- Add `_circuit_diagram_info_` for `Controlled` and `Power` to generate the same custom diagrams as before


anurudhp:
@tanujkhattar PTAL

qualtran/bloqs/mean_estimation/mean_estimation_operator_test.py
```diff
@@ -282,8 +283,9 @@ def test_mean_estimation_operator_consistent_protocols():
         _ = mean_gate.controlled(num_controls=2)
 
     # Test with_power
-    assert mean_gate.with_power(5) ** 2 == MeanEstimationOperator(
-        code, arctan_bitsize=arctan_bitsize, power=10
+    assert (
+        Power(mean_gate, 5) ** 2
```
tanujkhattar, line 287: Does this work?

```suggestion
        (mean_gate ** 5) ** 2
```

qualtran/_infra/controlled.py
```diff
@@ -462,3 +462,20 @@ def as_cirq_op(
...
+
+        if info is None:
+            return super()._circuit_diagram_info_(args)
+
+        wire_symbols = []
+        for cv in self.ctrl_spec.cvs:
+            for bit in cv.ravel():
+                wire_symbols.append('@' if bit else '@(0)')
+        wire_symbols += info.wire_symbols
+        return cirq.CircuitDiagramInfo(wire_symbols=wire_symbols)
```
tanujkhattar, line 481: Add tests for this?

tanujkhattar (approved):
LGTM % comments

qualtran/bloqs/mean_estimation/mean_estimation_operator_test.py
anurudhp: Yep this also works. I actually added these tests for `Power` in `util_bloqs_test.py`, so maybe we can remove this one here?

qualtran/bloqs/util_bloqs.py
```diff
@@ -514,3 +515,22 @@ def build_composite_bloq(self, bb: 'BloqBuilder', **soqs: 'SoquetT') -> Dict[str
 
     def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
         return {(self.bloq, self.power)}
+
+    def __pow__(self, power) -> 'Power':
+        bloq = self.bloq.adjoint() if power < 0 else self.bloq
+        return Power(bloq, self.power * abs(power))
```
mpharrigan, line 521: should we check for integer?

anurudhp: added a check in `__attrs_post_init__`

qualtran/_infra/controlled.py
anurudhp: turns out this is circular, reverted the change in GWR and instead only calling this in Controlled

mpharrigan (approved):

qualtran/_infra/controlled.py
anurudhp: added tests

qualtran/_infra/controlled.py
```diff
@@ -462,3 +462,18 @@ def as_cirq_op(
...
+            sub_info = cirq.circuit_diagram_info(self.subbloq, args, None)
+            if sub_info is not None:
+                cv_info = cirq.circuit_diagram_info(self.ctrl_spec.to_cirq_cv())
+
+                return cirq.CircuitDiagramInfo(
+                    wire_symbols=(*cv_info.wire_symbols, *sub_info.wire_symbols),
+                    exponent=sub_info.exponent,
+                )
+
+        return _wire_symbol_to_cirq_diagram_info(self, args)
```
anurudhp, line 479: One change after the approval: I had to add a special case here because `_wire_symbol_to_cirq_diagram_info` doesn't recurse into the subgate. eg. `XPowGate(0.25).controlled()` was rendering `X` instead of `X^0.25`. 

I added tests now to check that this works correctly

anurudhp:
@tanujkhattar @mpharrigan PTAL

PR #904, opened 2024-04-28
Fix custom controlled versions of `SelectOracle`s and `QubitizationWalkOperator`s

### LINKED ISSUE #886: Should not override `.controlled` for custom controlled implementations
Bloq authors should always override `get_ctrl_system` and not `controlled`. I think only the old cirq-ft bloqs ( `QubitizationWalkOperator` and related) override `controlled` directly, which must be upgraded.

```py
from qualtran import Bloq
from qualtran.bloqs.for_testing.random_select_and_prepare import RandomPrepareOracle
from qualtran.bloqs.reflection_using_prepare import ReflectionUsingPrepare

prepare = RandomPrepareOracle.create(3, random_state=42)
R = ReflectionUsingPrepare(prepare)

class Wrapped(Bloq):
    @property
        return R.signature

    def build_composite_bloq(self, bb: 'BloqBuilder', **soqs: 'SoquetT') -> dict[str, 'SoquetT']:
        return bb.add_d(R, **soqs)

cb_direct = R.controlled().decompose_bloq()
cb_wrapped = Wrapped().controlled().decompose_bloq().flatten_once(lambda _: True)
```

```py
show_bloq(cb_direct)
```
direct

```py
show_bloq(cb_wrapped)
```
wrapped

More detailed notebook with above code: https://github.com/anurudhp/Qualtran/blob/2024-04-issue-controlled/controlled_issue.ipynb

anurudhp:
On top of #903 

fixes #886

### DIFF
--- a/qualtran/_infra/gate_with_registers.py
+++ b/qualtran/_infra/gate_with_registers.py
@@ -27,6 +27,7 @@ from typing import (
     Union,
 )
 
+import attrs
 import cirq
 import numpy as np
 from numpy.typing import NDArray
@@ -477,10 +478,6 @@ class GateWithRegisters(Bloq, cirq.Gate, metaclass=abc.ABCMeta):
         bloq. Bloqs authors can declare their own, custom controlled versions by overriding
         `Bloq.get_ctrl_system` in the bloq.
 
-        If overriding the `GWR.controlled()` method directly, Bloq authors can use the
-        `self._get_ctrl_spec` helper to construct a `CtrlSpec` object from the input parameters of
-        `GWR.controlled()` and use it to return a custom controlled version of this Bloq.
-
 
         Args:
             num_controls: Cirq style API to specify control specification -
@@ -548,3 +545,51 @@ class GateWithRegisters(Bloq, cirq.Gate, metaclass=abc.ABCMeta):
 
         wire_symbols[0] = self.__class__.__name__
         return cirq.CircuitDiagramInfo(wire_symbols=wire_symbols)
+
+
+def get_ctrl_system_for_single_qubit_controlled(
+    bloq: GateWithRegisters, ctrl_spec: Optional['CtrlSpec'] = None
+) -> Tuple['Bloq', 'AddControlledT']:
+    """Wrapper to correctly wire a custom implementation of a single-qubit controlled version of a gate.
+
+    For any gate that supports a single-qubit controlled version
+    with an API that has `control_val` and `control_registers`.
+    Add the `get_ctrl_system` implementation as shown below:
+
+    .. code:: python
+
+        class MyGate(GateWithRegisters):
+            control_val: Optional[int]
+
+            @property
+            def control_registers() -> Tuple[Register, ...]:
+                ...
+
+            def get_ctrl_system(
+                self, ctrl_spec: Optional['CtrlSpec'] = None
+            ) -> Tuple['Bloq', 'AddControlledT']:
+                from qualtran._infra.gate_with_registers import get_ctrl_system_for_single_qubit_controlled
+
+                return get_ctrl_system_for_single_qubit_controlled(self, ctrl_spec)
+    """
+    if ctrl_spec is None:
+        ctrl_spec = CtrlSpec()
+
+    assert hasattr(bloq, 'control_val')
+    assert hasattr(bloq, 'control_registers')
+
+    if bloq.control_val is None and ctrl_spec.shapes in [((),), ((1,),)]:
+        cbloq = attrs.evolve(bloq, control_val=int(ctrl_spec.cvs[0].item()))
+        (ctrl_reg,) = cbloq.control_registers
+
+        def adder(
+            bb: 'BloqBuilder', ctrl_soqs: Sequence['SoquetT'], in_soqs: dict[str, 'SoquetT']
+        ) -> tuple[Iterable['SoquetT'], Iterable['SoquetT']]:
+            soqs = {ctrl_reg.name: ctrl_soqs[0]} | in_soqs
+            soqs = bb.add_d(cbloq, **soqs)
+            ctrl_soqs = [soqs.pop(ctrl_reg.name)]
+            return ctrl_soqs, soqs.values()
+
+        return cbloq, adder
+
+    raise NotImplementedError(f'Cannot create a controlled version of {bloq} with {ctrl_spec=}.')
--- a/qualtran/_infra/gate_with_registers.py
+++ b/qualtran/_infra/gate_with_registers.py
@@ -568,7 +568,7 @@ class SpecializedSingleQubitControlledGate(GateWithRegisters):
         self, control_val: int
     ) -> 'SpecializedSingleQubitControlledGate':
         """Override this to provide a custom controlled bloq"""
-        return attrs.evolve(self, control_val=control_val)
+        return attrs.evolve(self, control_val=control_val)  # type: ignore[misc]
 
     def get_ctrl_system(
         self, ctrl_spec: Optional['CtrlSpec'] = None
--- a/qualtran/_infra/gate_with_registers.py
+++ b/qualtran/_infra/gate_with_registers.py
@@ -557,7 +557,7 @@ class SpecializedSingleQubitControlledGate(GateWithRegisters):
     Example usage:
 
         @attrs.frozen
-        class MyGate(SpecializedSingleQubitControlledMixin, GateWithRegisters):
+        class MyGate(SpecializedSingleQubitControlledGate, GateWithRegisters):
             control_val: Optional[int] = None
 
             @property
--- a/qualtran/_infra/gate_with_registers.py
+++ b/qualtran/_infra/gate_with_registers.py
@@ -549,7 +549,7 @@ class SpecializedSingleQubitControlledGate(GateWithRegisters):
     Example usage:
 
         @attrs.frozen
-        class MyGate(SpecializedSingleQubitControlledGate, GateWithRegisters):
+        class MyGate(SpecializedSingleQubitControlledGate):
             control_val: Optional[int] = None
 
             @property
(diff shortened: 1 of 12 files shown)


qualtran/bloqs/hubbard_model.py
```diff
@@ -219,28 +219,12 @@ def decompose_from_registers(
...
-            and self.control_val is None
-        ):
-            return SelectHubbard(self.x_dim, self.y_dim, control_val=control_values[0])
-        raise NotImplementedError(
-            f'Cannot create a controlled version of {self} with control_values={control_values}.'
-        )
+    def get_ctrl_system(
+        self, ctrl_spec: Optional['CtrlSpec'] = None
+    ) -> Tuple['Bloq', 'AddControlledT']:
+        from qualtran._infra.gate_with_registers import get_ctrl_system_for_single_qubit_controlled
```
fdmalone, line 225: Why is the import happening inside?

anurudhp: ah, this should be at top level.

I haven't yet decided if this is the proper way to implement this. Somehow the `control_val` interface is a bit ad-hoc and can't be type-checked properly. I'm open to suggestions for this interface.

qualtran/_infra/gate_with_registers.py
```diff
@@ -549,3 +546,51 @@ def _circuit_diagram_info_(self, args: cirq.CircuitDiagramInfoArgs) -> cirq.Circ
...
+def get_ctrl_system_for_single_qubit_controlled(
+    bloq: GateWithRegisters, ctrl_spec: Optional['CtrlSpec'] = None
+) -> Tuple['Bloq', 'AddControlledT']:
+    """Wrapper to correctly wire a custom implementation of a single-qubit controlled version of a gate.
+
+    For any gate that supports a single-qubit controlled version
+    with an API that has `control_val` and `control_registers`.
+    Add the `get_ctrl_system` implementation as shown below:
+
+    .. code:: python
```
mpharrigan, line 560: we actually render all this as markdown, so this directive won't work properly. I think the indented section should be rendered as code; you might need backticks if you want to say "python"

qualtran/_infra/gate_with_registers.py
```diff
@@ -549,3 +546,51 @@ def _circuit_diagram_info_(self, args: cirq.CircuitDiagramInfoArgs) -> cirq.Circ
...
         return cirq.CircuitDiagramInfo(wire_symbols=wire_symbols)
+
+
+def get_ctrl_system_for_single_qubit_controlled(
+    bloq: GateWithRegisters, ctrl_spec: Optional['CtrlSpec'] = None
+) -> Tuple['Bloq', 'AddControlledT']:
+    """Wrapper to correctly wire a custom implementation of a single-qubit controlled version of a gate.
+
+    For any gate that supports a single-qubit controlled version
+    with an API that has `control_val` and `control_registers`.
```
mpharrigan, line 557: can you be more specific than "with an api that has..."

I think you require that the bloq has attributes named `control_val` and `control_registers`, right? Does this need to be a `GateWithRegisters` like the annotation?

qualtran/_infra/gate_with_registers.py
```diff
@@ -549,3 +546,51 @@ def _circuit_diagram_info_(self, args: cirq.CircuitDiagramInfoArgs) -> cirq.Circ
...
+            bb: 'BloqBuilder', ctrl_soqs: Sequence['SoquetT'], in_soqs: dict[str, 'SoquetT']
+        ) -> tuple[Iterable['SoquetT'], Iterable['SoquetT']]:
+            soqs = {ctrl_reg.name: ctrl_soqs[0]} | in_soqs
+            soqs = bb.add_d(cbloq, **soqs)
+            ctrl_soqs = [soqs.pop(ctrl_reg.name)]
+            return ctrl_soqs, soqs.values()
+
+        return cbloq, adder
+
+    raise NotImplementedError(f'Cannot create a controlled version of {bloq} with {ctrl_spec=}.')
```
mpharrigan, line 596: for this to properly implement the `get_ctrl_system` method, it should delegate to the base class method if no specific construction is available. (This base method will wrap the bloq in `Controlled` metabloq).

qualtran/_infra/gate_with_registers.py
```diff
@@ -549,3 +546,51 @@ def _circuit_diagram_info_(self, args: cirq.CircuitDiagramInfoArgs) -> cirq.Circ
...
+    """
+    if ctrl_spec is None:
+        ctrl_spec = CtrlSpec()
+
+    assert hasattr(bloq, 'control_val')
+    assert hasattr(bloq, 'control_registers')
+
+    if bloq.control_val is None and ctrl_spec.shapes in [((),), ((1,),)]:
+        cbloq = attrs.evolve(bloq, control_val=int(ctrl_spec.cvs[0].item()))
+        (ctrl_reg,) = cbloq.control_registers
```
mpharrigan, line 584: so this assumes that `control_registers` is actually a property that depends on the `control_val` attribute, right? please document and/or code this more defensively

qualtran/bloqs/mean_estimation/mean_estimation_operator.py
```diff
@@ -126,36 +125,36 @@ def decompose_from_registers(
...
+
+            def adder(
+                bb: 'BloqBuilder', ctrl_soqs: Sequence['SoquetT'], in_soqs: dict[str, 'SoquetT']
+            ) -> tuple[Iterable['SoquetT'], Iterable['SoquetT']]:
+                soqs = {ctrl_reg.name: ctrl_soqs[0]} | in_soqs
+                soqs = bb.add_d(cbloq, **soqs)
+                ctrl_soqs = [soqs.pop(ctrl_reg.name)]
+                return ctrl_soqs, soqs.values()
+
+            return cbloq, adder
```
mpharrigan, line 149: you should probably make it so it's possible to factor this out as well, right? we should probably pick what the `control_val`/`cv` argument is called

mpharrigan (commented):
This is a good idea. The current PR needs a little polish

qualtran/_infra/gate_with_registers.py
```diff
@@ -549,3 +546,51 @@ def _circuit_diagram_info_(self, args: cirq.CircuitDiagramInfoArgs) -> cirq.Circ
...
+            ) -> Tuple['Bloq', 'AddControlledT']:
+                from qualtran._infra.gate_with_registers import get_ctrl_system_for_single_qubit_controlled
+
+                return get_ctrl_system_for_single_qubit_controlled(self, ctrl_spec)
+    """
+    if ctrl_spec is None:
+        ctrl_spec = CtrlSpec()
+
+    assert hasattr(bloq, 'control_val')
+    assert hasattr(bloq, 'control_registers')
```
mpharrigan, line 580: I don't know if this is a better or worse idea, but: did you consider making this a mixin?

anurudhp: this does make the code cleaner, and enables type-checking. I made a class `SpecializedSingleQubitControlledGate` (similar to the design of `UnaryIterationGate`), please have a look

qualtran/bloqs/mean_estimation/mean_estimation_operator.py
anurudhp: fixed this now, uses the same code as the others

qualtran/_infra/gate_with_registers.py
anurudhp: fixed

qualtran/_infra/gate_with_registers.py
anurudhp: updated the design to use a mixin class

qualtran/_infra/gate_with_registers.py
anurudhp: oops, fixed now

qualtran/_infra/gate_with_registers.py
anurudhp: added a check that raises TypeError

mpharrigan (approved):
cool

mpharrigan:
@dstrain115 can you advise on how to fix the type issues

dstrain115:
I think you can add "#type: ignore" to all of them.  The first is because it doesn't know whether it is an attrs class, so it can't attrs.evolve it.

The others are because qualtran and cirq disagree on the definition of "controlled" and the classes inherit from both.   The current order should prefer Qualtran's definition to cirq's which seems fine to me.

anurudhp:
> The others are because qualtran and cirq disagree on the definition of "controlled" and the classes inherit from both. The current order should prefer Qualtran's definition to cirq's which seems fine to me.

I added `type: ignore` to the class definition line. It's strange that the issue does not show up in `GateWithRegisters`, but only here. For example:
```py
class SpecializedSingleQubitControlledGate(GateWithRegisters):
  ...

class QubitizationWalkOperator(SpecializedSingleQubitControlledGate, GateWithRegisters): # <- type check fails here
  ...
```

dstrain115:
I'm not totally sure.  It's something with the multiple inheritance and the function definitions not matching.

qualtran/_infra/gate_with_registers.py
```diff
@@ -548,3 +545,62 @@ def _circuit_diagram_info_(self, args: cirq.CircuitDiagramInfoArgs) -> cirq.Circ
...
+    """Add a specialized single-qubit controlled version of a Bloq.
+
+    `control_val` is an optional single-bit control. When `control_val` is provided,
+     the `control_registers` property should return a single named qubit register,
+     and otherwise return an empty tuple.
+
+    Example usage:
+
+        @attrs.frozen
+        class MyGate(SpecializedSingleQubitControlledGate, GateWithRegisters):
```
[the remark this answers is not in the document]
anurudhp: fixed - used the second suggestion

PR #906, opened 2024-04-29
Mypy issues part 4

dstrain115:
This PR features many fixes to get sympy.Expr to type correctly in a bunch of places.

Another change was adding iteration_length_or_zero() so we can safely get iteration_length from dtypes even if they don't have one defined.


qualtran/_infra/composite_bloq.py
```diff
@@ -831,7 +830,9 @@ def from_signature(
         initial_soqs: Dict[str, SoquetT] = {}
         for reg in signature:
             if reg.side & Side.LEFT:
-                initial_soqs[reg.name] = bb.add_register_from_dtype(reg)
+                register = bb.add_register_from_dtype(reg)
+                if register is not None:
+                    initial_soqs[reg.name] = register
```
mpharrigan, line 835: I think we should probably assert register is not None (?). Since the `if` says that it's a LEFT (or THRU) register -- it should definitely be present in initial_soqs

qualtran/bloqs/arithmetic/comparison.py
```diff
@@ -253,10 +266,10 @@ def _decomposition():
         else:
             yield from _decomposition()
 
-    def adjoint(self) -> 'Bloq':
+    def adjoint(self) -> Self:
```
mpharrigan, line 269: what version of python does this require?

qualtran/bloqs/arithmetic/comparison.py
```diff
@@ -233,7 +246,7 @@ def _cswap(control: cirq.Qid, a: cirq.Qid, b: cirq.Qid, aux: cirq.Qid) -> cirq.O
             yield cirq.CNOT(aux, a)
             yield cirq.CNOT(a, b)
 
-        def _decomposition():
+        def _decomposition() -> Generator[cirq.Operation, None, None]:
```
mpharrigan, line 249: Iterator[cirq.Operation]?

dstrain115: Done.

qualtran/bloqs/arithmetic/comparison.py
mpharrigan: one day :)

qualtran/testing.py
```diff
@@ -567,6 +567,7 @@ def assert_bloq_example_serializes(bloq_ex: BloqExample) -> None:
         raise BloqCheckException.fail(
             f'Roundtrip equality failed.\n{bloq=}\n{bloq_roundtrip=}\n'
         ) from e
+    return BloqCheckResult.PASS, ''
```
mpharrigan, line 570: this isn't right, right? the annotation and docs say this should return None (like an assertion)

mpharrigan (approved):
Thanks! approval, but please fix the thing I noted !

qualtran/testing.py
dstrain115: Done.

PR #908, opened 2024-04-30
Draw latex diagrams for bloqs using QPIC

tanujkhattar:
Adds support to draw latex diagrams for Bloqs using `qpic`. 

Here are some pictures generated by 
image

image

image

image

Many other improvements can be made. For example we use a square box instead of left / right triangles (by setting `shape=>` or `shape=<` in the `G` gates) because elongated triangles look bad. The musical score diagrams added by @mpharrigan use an arrow shape of the form `=>` where the width of the horizontal part `=` scales with length of the text but the head of the arrow `>` stays constant. qpic doesn't have an out of the box option to do this but I think we can achieve something similar using `op="valid_tickz_code_to_draw_an_arrow"` where the `valid_tickz_code_to_draw_an_arrow` needs to be figured out. I spent sometime playing around but gave up. We can do this as a follow up

Another potential improvement is the logic in line manager to allocate lines to given Soquets. In many diagrams shown above, the joined lines are placed below the original split lines even when they are equivalent. 

cc @mpharrigan @ncrubin @fdmalone


qualtran/drawing/_show_funcs.py
```diff
@@ -84,3 +86,12 @@ def show_flame_graph(*bloqs: 'Bloq', **kwargs):
     """Display hiearchical decomposition and T-complexity costs as a Flame Graph."""
     svg_data = get_flame_graph_svg_data(*bloqs, **kwargs)
     IPython.display.display(IPython.display.SVG(svg_data))
+
+
+def show_bloq_via_qpic(bloq: 'Bloq', width: int = 1000, height: int = 400):
```
mpharrigan, line 91: this should likely just be a different argument for `type` in `show_bloq`

qualtran/bloqs/hubbard_model.ipynb
mpharrigan, line 1: revert since this notebook isn't actually changed

qualtran/bloqs/util_bloqs.py
```diff
@@ -113,7 +113,7 @@ def wire_symbol(self, soq: 'Soquet') -> 'WireSymbol':
         if soq.reg.shape:
             text = f'[{", ".join(str(i) for i in soq.idx)}]'
             return directional_text_box(text, side=soq.reg.side)
-        return directional_text_box(' ', side=soq.reg.side)
+        return directional_text_box('S', side=soq.reg.side)
```
mpharrigan, line 116: motivate this change? presumably the splitting action should be implied by the circuit topology

qualtran/drawing/qpic_diagram.py
```diff
@@ -0,0 +1,332 @@
...
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+"""Classes for drawing latex diagrams for bloqs with QPIC - https://github.com/qpic/qpic.
+
+QPIC is not a dependency of Qualtran and must be manually installed by users via
+`pip install qpic`.
```
mpharrigan, line 18: why?

qualtran/drawing/qpic_diagram.py
```diff
@@ -0,0 +1,332 @@
...
+
+from qualtran import DanglingT, LeftDangle, QBit, Side, Soquet
+from qualtran.drawing.musical_score import _soq_to_symb, Circle, ModPlus, WireSymbol
+
+if TYPE_CHECKING:
+    from qualtran import Bloq, Connection, Signature
+
+
+def _wire_name_prefix_for_soq(soq: Soquet) -> str:
+    return soq.pretty()
```
mpharrigan, line 37: docstring? do you anticipate this being something other than `soq.pretty()`?

qualtran/bloqs/mcmt/and_bloq.py
```diff
@@ -160,7 +160,7 @@ def add_my_tensors(
 
     def wire_symbol(self, soq: 'Soquet') -> 'WireSymbol':
         if soq.reg.name == 'target':
-            return directional_text_box('∧', side=soq.reg.side)
+            return directional_text_box('&', side=soq.reg.side)
```
mpharrigan, line 163: motivate this change? can qpic not do unicode? document this restriction somewhere in the wire symbol world. (The previous symbol was the unicode character for "LOGICAL AND")

qualtran/drawing/qpic_diagram.py
```diff
@@ -0,0 +1,332 @@
...
+
+
+def _format_label_text(label: str, scale: Optional[float] = None) -> str:
+    replacements = {'&': r'\&', '_': r'\_', '<': r'$<$', '>': r'$>$', '⨁': r'$\oplus$'}
+    for key, val in replacements.items():
+        label = label.replace(key, val)
+    scale_prefix = r'\scalebox{' + str(scale) + r'}{' if scale else ''
```
mpharrigan, line 44: `if scale is not None` or do you want scale of 0 to mean default?

mpharrigan: unnecessary r'' strings here

qualtran/drawing/qpic_diagram.py
```diff
@@ -0,0 +1,332 @@
...
+    scale_suffix = r'}' if scale else ''
+    return r'\textrm{' + scale_prefix + str(label) + scale_suffix + r'}'
```
mpharrigan, line 46: you call `str` on label but are already calling `.replace` on it, so this duck typing won't add anything

qualtran/drawing/qpic_diagram.py
```diff
@@ -0,0 +1,332 @@
...
+
+if TYPE_CHECKING:
+    from qualtran import Bloq, Connection, Signature
+
+
+def _wire_name_prefix_for_soq(soq: Soquet) -> str:
+    return soq.pretty()
+
+
+def _format_label_text(label: str, scale: Optional[float] = None) -> str:
```
mpharrigan, line 40: I'd personally find this more readable if there were two seperate code paths for scale vs no scale

```python
if scale is not None:
  return rf'{\scalebox{scale}{label}}
else:
  return rf'{label}
```

qualtran/drawing/qpic_diagram.py
```diff
@@ -0,0 +1,332 @@
...
+    replacements = {'&': r'\&', '_': r'\_', '<': r'$<$', '>': r'$>$', '⨁': r'$\oplus$'}
+    for key, val in replacements.items():
+        label = label.replace(key, val)
+    scale_prefix = r'\scalebox{' + str(scale) + r'}{' if scale else ''
+    scale_suffix = r'}' if scale else ''
+    return r'\textrm{' + scale_prefix + str(label) + scale_suffix + r'}'
+
+
+def _soq_symbol_to_text(symbol: WireSymbol) -> str:
+    assert hasattr(symbol, 'text')
```
mpharrigan, line 50: update type annotation to be a union of the texty symbols?

qualtran/drawing/qpic_diagram.py
```diff
@@ -0,0 +1,332 @@
...
+    symbol_text = symbol.text.strip()
+    if not symbol_text:
+        symbol_text = 'SJ'
```
mpharrigan, line 53: ??

qualtran/drawing/qpic_diagram.py
```diff
@@ -0,0 +1,332 @@
...
+    for key, val in replacements.items():
+        label = label.replace(key, val)
+    scale_prefix = r'\scalebox{' + str(scale) + r'}{' if scale else ''
+    scale_suffix = r'}' if scale else ''
+    return r'\textrm{' + scale_prefix + str(label) + scale_suffix + r'}'
+
+
+def _soq_symbol_to_text(symbol: WireSymbol) -> str:
+    assert hasattr(symbol, 'text')
+    symbol_text = symbol.text.strip()
```
mpharrigan, line 51: why do we need to strip?

qualtran/drawing/qpic_diagram.py
```diff
@@ -0,0 +1,332 @@
...
+        symbol_text = 'SJ'
+    return _format_label_text(symbol_text)
+
+
+def _soq_symbol_width(symbol: WireSymbol) -> int:
+    assert hasattr(symbol, 'text')
+    symbol_text = symbol.text.strip()
+    if not symbol_text:
+        symbol_text = 'SJ'
+    return 5 + 5 * len(symbol_text)
```
mpharrigan, line 62: explain these values? Does this actually work? Does qpic use monospaced fonts?

qualtran/drawing/qpic_diagram.py
```diff
@@ -0,0 +1,332 @@
...
+    allocates a new wire with prefix `_wire_name_prefix_for_soq(soq)` and an integer suffix that
+    corresponds to the smallest integer which does not correspond to an allocated wire.
+    """
+
+    def __init__(self):
+        self._soq_to_wire_name_tuple: Dict[Soquet, Tuple[str, int]] = {}
+        self._alloc_wires_with_prefix: Dict[str, Set[int]] = defaultdict(set)
+
+    def _wire_name_tuple_for_soq(self, soq: Soquet) -> Tuple[str, int]:
+        prefix = _wire_name_prefix_for_soq(soq)
```
mpharrigan, line 78: make this a (static) method instead of a free function?

qualtran/drawing/qpic_diagram.py
```diff
@@ -0,0 +1,332 @@
...
+            # A RIGHT wire on an intermediate Bloq should start as off and be turned on
+            # by the Bloq.
+            self.wires += [f'{wire_name} W off']
+
+    def add_connection(self, cxn: 'Connection'):
+        if cxn.left not in self.soq_map:
+            self._alloc_wire_for_soq(cxn.left)
+        self.soq_map[cxn.right] = self.soq_map[cxn.left]
+
+    def add_bloq(self, bloq: 'Bloq', pred: List['Connection'], succ: List['Connection']) -> None:
```
mpharrigan, line 171: bloq isn't used

qualtran/drawing/qpic_diagram.py
```diff
@@ -0,0 +1,332 @@
...
+    def soq_to_wirename(self, soq) -> str:
+        assert soq in self._soq_to_wire_name_tuple
+        return self._wire_name_tuple_to_str(self._soq_to_wire_name_tuple[soq])
+
+    def soq_to_wirelabel(self, soq) -> str:
+        return _format_label_text(self.soq_to_wirename(soq))
+
+
+class QpicCircuit:
+    """Builds data corresponding to the input specification of a QPIC diagram"""
```
mpharrigan, line 108: can you document this class and its public methods or make this class or its methods private? document public attributes

qualtran/drawing/qpic_diagram.py
```diff
@@ -0,0 +1,332 @@
...
+
+    def __init__(self, signature: 'Signature'):
+        self.soq_map = {}
```
mpharrigan, line 111: what is soq_map? Why isn't this handled by the wire manager?

qualtran/drawing/qpic_diagram.py
```diff
@@ -0,0 +1,332 @@
...
+                add_soq(cxn.left)
+        self.gates += [' '.join(targets + controls)]
+
+
+def get_qpic_data(bloq: 'Bloq', file_path: Union[None, pathlib.Path, str] = None) -> List[str]:
+    """Get the input data that can be used to draw a latex diagram for `bloq` using `qpic`.
+
+    Args:
+        bloq: Bloqs to draw the latex diagram for
+        file_path: If specified, the output is stored at the file.
```
mpharrigan, line 197: "otherwise the data is returned"

qualtran/drawing/qpic_diagram.py
```diff
@@ -0,0 +1,332 @@
...
+
+    Returns:
+        A list of strings representing the input to `qpic` utility to draw a latex diagram
+        for bloq. If specified, the output is written to `file_path`.
+    """
+    cbloq = bloq.as_composite_bloq()
+    qpic_circuit = QpicCircuit(cbloq.signature)
```
mpharrigan, line 204: What do you think about removing all the business logic from the QpicCircuit constructor? i.e.: in `__init__` just initialize the member variables; then in this function you can call

```python
qpic_circuit.set_up_signature(cbloq.signature)
for binst, pred, succ in cbloq.iter_bloqnections():
  qpic_circuit.add_bloq(pred, succ)
```

qualtran/drawing/qpic_diagram.py
```diff
@@ -0,0 +1,332 @@
...
+    for binst, pred, succ in cbloq.iter_bloqnections():
+        qpic_circuit.add_bloq(binst.bloq, pred, succ)
+    data = qpic_circuit.wires + qpic_circuit.gates
```
mpharrigan, line 207: make a convenience property on QpicCircuit `.data` ?

qualtran/drawing/qpic_diagram.py
```diff
@@ -0,0 +1,332 @@
...
+
+def qpic_input_to_diagram(
+    qpic_file: Union[pathlib.Path, str],
+    output_file: Union[None, pathlib.Path, str] = None,
+    output_type: str = 'pdf',
+) -> str:
+    r"""Invoke `qpic` script to generate output diagram of type qpic/tex/pdf/png.
+
+    Outputs one of the following files, based on `output_type`:
+     - base_file_path + '.qpic' - Copies the input qpic_file to output_file.
```
mpharrigan, line 231: base_file_path isn't an argument

qualtran/drawing/qpic_diagram.py
```diff
@@ -0,0 +1,332 @@
...
+        https://tug.org/applications/pdftex/ for more details on how to install `pdflatex`.
+     - base_file_path + '.png' - Uses `convert` tool to convert pdf to png. See
+        https://imagemagick.org/ for more details on how to install `convert`.
+
+    Args:
+        qpic_file: Path to file containing input that should be passed to the `qpic` script.
+        output_file: Optional path to the output where generated diagram should be saved. Defaults to
+            qpic_file.with_suffix(f".{output_type}")
+        output_type: Format of the diagram generated using qpic. Supported output types are one of
+            ['qpic', 'tex', 'pdf', 'png']
```
mpharrigan, line 243: what happens if output_file extension and output_type conflict?

mpharrigan:
The font size looks large compared to e.g. the size of control-circles. Is there a way to control the font size?

qualtran/bloqs/util_bloqs.py
tanujkhattar: Using `symbol.text` leads to an empty symbol which `qpic` doesn't like for labels of the gate. I can probably hack around if needed

qualtran/drawing/qpic_diagram.py
tanujkhattar: License issues. QPIC has a [GPL-3.0 license](https://github.com/qpic/qpic#GPL-3.0-1-ov-file) license which is restricted but I confirmed with the open source team that having a dependency like this (where we don't explicit depend on the tool but can produce output the tool can consume) is fine.

qualtran/drawing/qpic_diagram.py
tanujkhattar: Maybe, I just wanted to keep it abstracted away so if we find a diagram where `soq.pretty` is not suitable and we want to use something else, we know where to make the change

qualtran/bloqs/hubbard_model.ipynb
tanujkhattar: done

qualtran/bloqs/mcmt/and_bloq.py
tanujkhattar: reverted

qualtran/drawing/qpic_diagram.py
tanujkhattar: removed

qualtran/drawing/qpic_diagram.py
tanujkhattar: `_soq_to_symb` returns a `WireSymbol` instead of an overloaded type based on soquet. Refactored and added an assertion to make mypy happy

qualtran/drawing/qpic_diagram.py
tanujkhattar: removed

qualtran/drawing/qpic_diagram.py
tanujkhattar: one or more blank spaces are considered equivalent in the qpic code. Therefore, I want to remove any blankspaces before / after symbol_text (eg: `' '` for split / join) before processing.

qualtran/drawing/qpic_diagram.py
tanujkhattar: Yes, it works. The values are empirical and the goal is to rescale the width of rectangle based on the length of the label

qualtran/drawing/qpic_diagram.py
tanujkhattar: wire manager is responsible to allocate / deallocate new wires. `soq_map` is used to track the lifetime of a wire. In Cirq, wires have names but we don't have an equivalent notion of "wires" here. 

soq_map essentially maps every soquet to the leftmost soquet which has a wire originating and it has not been split / joined / deallocated yet.

qualtran/drawing/qpic_diagram.py
tanujkhattar: done

qualtran/drawing/qpic_diagram.py
tanujkhattar: done

qualtran/drawing/qpic_diagram.py
tanujkhattar: done

qualtran/drawing/qpic_diagram.py
tanujkhattar: updated

tanujkhattar:
> The font size looks large compared to e.g. the size of control-circles. Is there a way to control the font size?

Changed the `scale` argument to `0.8` by default so all fontsizes are now smaller by default. Also made a bunch of other improvements. I've updated the pictures in my original PR comment. PTAL

tanujkhattar:
@mpharrigan This is ready for another look.

qualtran/drawing/qpic_diagram.py
tanujkhattar: I use temporary files with the right extension as intermediate files to store results and move them to the output file when the right file is computed, so it doesn't matter what the extension of the output file is

qualtran/drawing/qpic_diagram.py
mpharrigan: do you want to make a note that you can install it with pip but this module will invoke the qpic executable via subprocessing?

qualtran/drawing/_show_funcs.py
mpharrigan: @tanujkhattar ^

tanujkhattar: I consider this to still be pretty experimental so wanted to keep it separate, but I can add it to `show_bloq` as well if you think it's fine.

mpharrigan: well let's not make it the default yet, but the "spirit" of the show functions is to provide the simplest possible entry point to getting something to show up in a notebook. I think the `type` parameter is perfect for this case

mpharrigan:
The diagrams look really nice. I think the slightly smaller font  classes them up

mpharrigan (approved):
LGTM modulo merging the show function

PR #915, opened 2024-05-01
Mypy issues part 5

dstrain115:
- Solves all of the remaining mypy issues
- Mostly confusion about symbolic expressions and whether things should be numpy arrays or not


qualtran/bloqs/basic_gates/su2_rotation.py
```diff
@@ -70,6 +70,12 @@ def signature(self) -> Signature:
 
     @cached_property
     def rotation_matrix(self) -> NDArray[np.complex_]:
+        if isinstance(self.lambd, sympy.Expr):
+            raise ValueError(f'Symbolic lambda not allowed: {self.lambd}')
+        if isinstance(self.phi, sympy.Expr):
+            raise ValueError(f'Symbolic phi not allowed: {self.phi}')
+        if isinstance(self.theta, sympy.Expr):
+            raise ValueError(f'Symbolic theta not allowed: {self.theta}')
```
anurudhp, line 78:
```suggestion
        if self.is_symbolic():
            raise ValueError(f'cannot compute rotation matrix for symbolic gate {self}')
```

dstrain115: So, unfortunately, this change, while more concise, will not satisfy mypy.  The automatic type checker can't figure out that, after self.is_symbolic() the parameters are no longer expressions.  I was playing around with trying to make is_symbolic a typing.TypeGuard, but I couldn't make it work correctly given the current code structure.

One alternative, if you feel strongly, is to change to is_symbolic, then add  `# type:ignore` later, with a comment that this is guarded by is_symbolic.

anurudhp: Ah that's interesting! I think then the current code you wrote is fine. 

Btw shouldn't you also check `self.global_shift`?

qualtran/_infra/composite_bloq.py
```diff
@@ -1009,8 +1013,10 @@ def add_from(self, bloq: Bloq, **in_soqs: SoquetInT) -> Tuple[SoquetT, ...]:
                 in_soqs[k] = np.asarray(v)
 
         # Initial mapping of LeftDangle according to user-provided in_soqs.
+        # TODO: add checking if this can be a sequence
```
mpharrigan, line 1016: what does this mean? add checking?

qualtran/bloqs/basic_gates/su2_rotation.py
mpharrigan: can you open an issue describing your future plans? unless you're going to do that imminently

qualtran/bloqs/chemistry/trotter/trotterized_unitary.py
```diff
@@ -102,9 +102,11 @@ def __attrs_post_init__(self):
     def signature(self) -> Signature:
         return self.bloqs[0].signature
 
-    def build_composite_bloq(self, bb: 'BloqBuilder', **soqs: SoquetT) -> Dict[str, 'Soquet']:
+    def build_composite_bloq(self, bb: 'BloqBuilder', **soqs: SoquetT) -> Dict[str, 'SoquetT']:
         for i, a in zip(self.indices, self.coeffs):
-            soqs |= bb.add_d(attrs.evolve(self.bloqs[i], angle=2 * a * self.timestep), **soqs)
+            # Bloqs passed in are supposed to be attrs dataclasses per docs
+            # It would be nice to somehow specify that self.bloqs are both bloqs and AttrsInstance
```
mpharrigan, line 108: is it worth opening an issue for this?

mpharrigan (approved):
wooo

qualtran/bloqs/basic_gates/su2_rotation.py
dstrain115: Added global_shift.
I have no further plans.  I have no idea how to typeguard this correctly. I'm not sure it's possible.

qualtran/bloqs/chemistry/trotter/trotterized_unitary.py
dstrain115: No, I think it is blocked on features needed for mypy which look like it's not happening.
I will leave this comment in there in case anyone who is a type-guru wants to fix it, but otherwise not worth it, in my opinion.

qualtran/_infra/composite_bloq.py
dstrain115: I'm not sure.  I'll just remove it.

PR #916, opened 2024-05-01
Changed controlled attribute to is_controlled

dstrain115:
- Change controlled attribute to is_controlled or controlled_by.
- This attribute shadows the controlled() function inherited by GateWithRegisters which is confusing and type-unsafe.


mpharrigan (approved):
we usually use the term "control_value" or "cvs" but lgtm

PR #918, opened 2024-05-02
one hot encoding

skushnir123:
Simple bloq to implement a one-hot encoding. This will be used in a subroutine of QROM Inverse (PR forthcoming)


qualtran/bloqs/data_loading/one_hot_encoding.py
```diff
@@ -0,0 +1,54 @@
...
+from attr import field
+from numpy._typing import NDArray
+
+from qualtran import GateWithRegisters, QAny, QUInt, Signature
+from qualtran.bloqs.basic_gates import TwoBitCSwap
+
+
+@attrs.frozen
+class OneHotEncoding(GateWithRegisters):
+    """
```
fdmalone, line 25: One line description here, directly after the """ (no new line)

qualtran/bloqs/data_loading/one_hot_encoding.py
```diff
@@ -0,0 +1,54 @@
...
+    One-hot encode a binary integer into a target register.
+
+    Registers:
+        a: an unsigned integer
+        b: the target to one-hot encode.
+
+    References:
+        [Windowed quantum arithmetic](https://arxiv.org/pdf/1905.07682.pdf)
```
fdmalone, line 33: This reference looks malformed.

qualtran/bloqs/data_loading/one_hot_encoding.py
```diff
@@ -0,0 +1,54 @@
...
+        Figure 4]
+    """
+
+    binary_reg_size: int = field()
```
fdmalone, line 37: binary_bitsize? We use bitsize a lot throughout the rest of the library.

fdmalone: Is the field necessary here?

qualtran/bloqs/data_loading/one_hot_encoding_test.py
```diff
@@ -0,0 +1,57 @@
...
+        a = quregs['a']
+        b = quregs['b']
+        binary_repr = list(iter_bits(self.integer, self.size))[::-1]
+        for i in range(self.size):
+            if binary_repr[i] == 1:
+                yield cirq.X(a[i])
+        yield OneHotEncoding(binary_reg_size=self.size).on_registers(a=a, b=b)
+
+
+@pytest.mark.parametrize('integer', list(range(8)))
```
fdmalone, line 47: Maybe add another loop over bitsize here

qualtran/bloqs/data_loading/one_hot_encoding_test.py
```diff
@@ -0,0 +1,57 @@
...
+    size: int = field()
+
+    @property
+    def signature(self) -> 'Signature':
+        return Signature.build_from_dtypes(a=QUInt(self.size), b=QUInt(2**self.size))
+
+    def decompose_from_registers(
+        self, *, context: cirq.DecompositionContext, **quregs: NDArray[cirq.Qid]
+    ) -> cirq.OP_TREE:
+        a = quregs['a']
```
fdmalone, line 38: I would be careful here with the endianness. I think you should be reversing the qubit order here and not in your test (iter_bits). Qualtran assumes a big endian convention.

qualtran/bloqs/data_loading/one_hot_encoding_test.py
```diff
@@ -0,0 +1,57 @@
...
+
+@attrs.frozen
+class OneHotEncodingTest(GateWithRegisters):
+    integer: int = field()
+    size: int = field()
+
+    @property
+    def signature(self) -> 'Signature':
+        return Signature.build_from_dtypes(a=QUInt(self.size), b=QUInt(2**self.size))
+
```
fdmalone, line 34: I think it would be worthwhile to add on_classical_vals and add_my_tensors methods

fdmalone (changes requested):
Some small nits and suggestions. Looking good!

qualtran/bloqs/data_loading/one_hot_encoding.py
skushnir123: Are you saying a new line in between """ and the description? I don't currently see one.

qualtran/bloqs/data_loading/one_hot_encoding.py
skushnir123: updated name and removed field()

qualtran/bloqs/data_loading/one_hot_encoding_test.py
skushnir123: It gets extremely slow for anything more than 3, so that's why I kept it at 3.

qualtran/bloqs/data_loading/one_hot_encoding_test.py
skushnir123: Done

qualtran/bloqs/data_loading/one_hot_encoding_test.py
skushnir123: The action of one-hot encoding when the second register isn't in the zero state is not well-defined so the on_classical_vals and  add_my_tensors will be kind of unnatural. Is there a way to define the action only a subset of the basis states?

qualtran/bloqs/data_loading/one_hot_encoding.py
mpharrigan: the text and the triple-quotes should be on the same line

qualtran/bloqs/data_loading/one_hot_encoding.py
```diff
@@ -0,0 +1,53 @@
...
+    Registers:
+        a: an unsigned integer
+        b: the target to one-hot encode.
+
+    References:
+        [Windowed quantum arithmetic](https://arxiv.org/pdf/1905.07682.pdf)
+        Figure 4
+    """
+
+    binary_bitsize: int
```
mpharrigan, line 36: Can we just call this `bitsize`? that's what we do elsewhere. the "bi" in "bitsize" means "bi"nary

qualtran/bloqs/data_loading/one_hot_encoding.py
mpharrigan: can you please link to the arxiv abstract rather than the pdf, thanks!

qualtran/bloqs/data_loading/one_hot_encoding.py
mpharrigan: I guess this is to contrast with the unary bitsize. Maybe it's worth disambiguating. I defer to your judgement. But please document this attribute in an `Args:` section in the class docstring

qualtran/bloqs/data_loading/one_hot_encoding_test.py
mpharrigan: - the classical simulation will always operate on computational basis states, so that should be appropriate here
 - I guess if the "target" isn't in the zero state, you'll get random things swapped around
     - one option is to change the output to a RIGHT register and allocate it inside the bloq. 
     - the other would be to raise an error in the classical simulation if the second register isn't in the zero state; but really -- this sounds like your abstraction is wrong and it should be a RIGHT register

qualtran/bloqs/data_loading/one_hot_encoding_test.py
```diff
@@ -0,0 +1,58 @@
...
+        yield OneHotEncoding(binary_bitsize=self.size).on_registers(a=a, b=b)
+
+
+@pytest.mark.parametrize('integer', list(range(8)))
+def test_one_hot_encoding(integer):
+    bitsize = 3
+    gate = OneHotEncodingTest(integer, bitsize)
+    qubits = cirq.LineQubit.range(bitsize + 2**bitsize)
+    op = gate.on_registers(a=qubits[:bitsize], b=qubits[bitsize:])
+    circuit0 = cirq.Circuit(op)
```
mpharrigan, line 53: can you document either here or in the decomposition what you're testing; namely: you are testing that the pattern of X's inverts the action of the `OneHotEncoding` bloq

qualtran/bloqs/data_loading/one_hot_encoding_test.py
skushnir123: Ok this is what I was thinking. Is add_my_tensors still a thing when the bloq isn't unitary? (as it wouldn't be if I made a right register.)

qualtran/bloqs/data_loading/one_hot_encoding.py
skushnir123: done

qualtran/bloqs/data_loading/one_hot_encoding_test.py
skushnir123: Just added on_classical_vals, should we also add add_my_tensors (also don't know if it applies given the non-unitary nature of this bloq now) @fdmalone

skushnir123: nevermind added

qualtran/bloqs/data_loading/one_hot_encoding_test.py
skushnir123: Done

fdmalone (approved):

qualtran/bloqs/data_loading/one_hot_encoding.py
```diff
@@ -0,0 +1,82 @@
...
+    def add_my_tensors(
+        self,
+        tn: 'qtn.TensorNetwork',
+        tag: Any,
+        *,
+        incoming: Dict[str, 'SoquetT'],
+        outgoing: Dict[str, 'SoquetT'],
+    ):
+        _add_my_tensors_from_gate(
+            self, self.signature, self.short_name(), tn, tag, incoming=incoming, outgoing=outgoing
```
mpharrigan, line 63: `self.short_name()` needs to be updated to e.g. `self.pretty_name()`

qualtran/bloqs/data_loading/one_hot_encoding.py
```diff
@@ -0,0 +1,82 @@
...
+        self, *, context: cirq.DecompositionContext, **quregs: NDArray[cirq.Qid]  # type: ignore[type-var]
+    ) -> cirq.OP_TREE:
+        a = quregs['a'][::-1]
+        b = quregs['b']
+
+        op_tree: List[cirq.Operation] = []
+        op_tree.append(cirq.X(b[0]))
+        for i in range(len(a)):
+            for j in range(2**i):
+                op_tree.append(TwoBitCSwap().on_registers(ctrl=a[i], x=b[j], y=b[2**i + j]))
```
tanujkhattar, line 81: You can use `cirq.CSWAP` here and the it'll automatically ge tconverted to `TwoBitCSwap` whenever the bloq-api is used due to the `_cirq_gate_to_bloq` logic.

qualtran/bloqs/data_loading/one_hot_encoding.py
```diff
@@ -0,0 +1,82 @@
...
+from numpy._typing import NDArray
+
+from qualtran import GateWithRegisters, QAny, QUInt, Register, Side, Signature, SoquetT
+from qualtran.bloqs.basic_gates import TwoBitCSwap
+from qualtran.cirq_interop._cirq_to_bloq import _add_my_tensors_from_gate
+from qualtran.simulation.classical_sim import ClassicalValT
+
+
+@attrs.frozen
+class OneHotEncoding(GateWithRegisters):
```
tanujkhattar, line 28: Isn't this same as the `SwapWithZero` gate we already have?

tanujkhattar: You can essentially write a one hot encoding bloq by doing an X(q[0]) and then call `SwapWithZero` with `a` as control and `b` as target. 
The part after the first `X` gate is `SwapWithZero` bloq. 

image


Also, `SwapWithZero` now supports N-dimensional registers so you can also do a one-hot for a multi dimensional input / output register. For example, if you have a 3d data where the input is 3 selection registers (`a_x`, `a_y`, `a_z`) and target is a 3D encoding where you would set `target[x][y][z] = 1` when `a_x = x`, `a_y = y` and `a_z = z`; you can do this again by setting `target[0][0][0] = 1` and calling a 3D swap-with-zero.

qualtran/bloqs/data_loading/one_hot_encoding.py
tanujkhattar: In fact, let's change the decomposition and `build_call_graph` to delegate to `SwapWithZero` bloq

tanujkhattar (changes requested):
I'll mark this as request changes following my comments above.

qualtran/bloqs/data_loading/one_hot_encoding.py
fdmalone: There's no issue with swap with zero destroying the rest of the "b" register here? I've always been confused by when SwapWithZero is ok to use.

tanujkhattar: The |B> register is assumed to be |0> in this case. The circuit after the first X gate is the same circuit as used for SwapWithZero

fdmalone: cool

mhucka:
@skushnir123 @tanujkhattar Just circling around to different PRs and trying to do a bit of housekeeping. For planning purposes, what is the status of this PR?

Issue #728, opened 2024-02-29
Better musical score and graph drawing control for registers via `wire_symbol`


mpharrigan:
- change the `wire_symbol` method to explicitly take a `Register` and `idx: Tuple[int, ...]` rather than one `Soquet` since the symbol shouldn't depend on the `binst`.
 - Eliminate or rename or merge `short_name` to explicitly compliment the wire symbols. This string should serve (solely) as a "title" to be written above the wire symbols and should be included if and only if the wire symbols don't already imply the gate. E.g. it's not necessary to write "CNOT" over something that's clearly a CNOT. Practically: we can rename the existing method and let it return `None` _or_ we can use a sentinel value to the `wire_symbol` method (e.g. `reg=None`) to imply that we're asking for the (optional) title.
 - Single source of truth for drawing: modify the `GraphvizDrawer` for drawing composite bloq graphs to use the `wire_symbol` method and values to label the soquet rows in the table for each bloq. Use the title method for the title, and again omit if not required. Practically: we'll need a new method on `WireSymbol` to return a textual equivalent of the symbol

mpharrigan:
For the third point, since we don't actually draw the symbols in the graphviz drawer: I propose instead to follow #791 where we should have meaningful, concise `__str__` methods on all the bloqs and use that

mpharrigan:
Ok, I just actually read the issue that I linked (and wrote) and it seems in all these places I seemed pretty convinced that the diagrams should use the wire_symbols for composite bloq diagrams (graphviz and musical_score). Let's try that. We can move the graphviz drawer to use `str()` later if it's too ambiguous otherwise

dstrain115:
This has been completed.  See above PRs for details.

PR #919, opened 2024-05-02
Change wire_symbol to use Register and idx

dstrain115:
- Change wire_symbol to use Register and idx since the wire_symbol should never need to use the binst of a Soquet.

Part of #728

### DIFF
--- a/qualtran/testing.py
+++ b/qualtran/testing.py
@@ -31,7 +31,6 @@ from qualtran import (
     LeftDangle,
     RightDangle,
     Side,
-    Soquet,
 )
 from qualtran._infra.composite_bloq import _get_flat_dangling_soqs
 from qualtran._infra.data_types import check_dtypes_consistent, QDTypeCheckingSeverity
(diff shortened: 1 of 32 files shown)


qualtran/testing.py
```diff
@@ -236,9 +235,7 @@ def assert_wire_symbols_match_expected(bloq: Bloq, expected_ws: List[str]):
     ws = []
     regs = bloq.signature
     for i, r in enumerate(regs):
-        # note this will only work if shape = ().
-        # See: https://github.com/quantumlib/Qualtran/issues/608
-        ws.append(bloq.wire_symbol(Soquet(i, r)).text)
+        ws.append(bloq.wire_symbol(r, (i,)).text)
```
mpharrigan, line 238: I don't think this fixes the linked issue. Can you restore the comment

mpharrigan (approved):
Thanks for this! I suspect this may have been a little tedious

qualtran/testing.py
dstrain115: Done.  If you have a good example, I can fix the comment too.

mpharrigan: how about `Toffoli()`

PR #907, opened 2024-04-29
[mod_arithmetic] Organize modular addition

mpharrigan:
- new top-level module `mod_arithmetic` that will contain modular arithmetic.
 - Move things from factoring/ and addition/ to this module
 - Simplify naming: `[C][Mod]Op[K]` 

More to come after this

PR #923, opened 2024-05-06
Elliptic curve cryptography algorithm (top-down)

mpharrigan:
This PR outlines how phase estimation of the elliptic curve addition operation can be used to find an elliptic curve private key . This is shor's for the discrete log problem.

This PR has a top-down implementation. The compilation boils down to modular arithmetic primitives, which I'm organizing in a "bottom up" fashion in prs like https://github.com/quantumlib/Qualtran/pull/907


dev_tools/autogenerate-bloqs-notebooks-v2.py
```diff
@@ -339,6 +340,21 @@
...
         directory=f'{SOURCE_DIR}/bloqs/factoring',
     ),
+    NotebookSpecV2(
+        title='Elliptic Curve Cryptography',
+        module=qualtran.bloqs.factoring.ecc,
+        bloq_specs=[
+            qualtran.bloqs.factoring.ecc._find_ecc_private_key._ECC_BLOQ_DOC,
+            qualtran.bloqs.factoring.ecc._ec_phase_estimate_r._EC_PE_BLOQ_DOC,
+            qualtran.bloqs.factoring.ecc._ec_add_r._ECC_ADD_R_BLOQ_DOC,
+            qualtran.bloqs.factoring.ecc._ec_add_r.EC_WINDOW_ADD_BLOQ_DOC,
```
fdmalone, line 350: are you missing the leading underscore here?

qualtran/bloqs/arithmetic/__init__.py
```diff
@@ -39,3 +39,5 @@
     SumOfSquares,
 )
 from qualtran.bloqs.arithmetic.sorting import BitonicSort, Comparator
+
+from ._shims import *
```
fdmalone, line 43: import * is sort of discouraged right?

qualtran/bloqs/factoring/ecc/_ec_add.py
```diff
@@ -0,0 +1,99 @@
+#  Copyright 2023 Google LLC
```
fdmalone, line 1: What's going on with the module naming here? Why are they all private modules?

qualtran/bloqs/factoring/ecc/_ec_add_r.py
```diff
@@ -0,0 +1,166 @@
...
+        [Quantum resource estimates for computing elliptic curve discrete logarithms](https://arxiv.org/abs/1706.06752).
+        Roetteler et. al. 2017. Algorithm 1 and Figure 10.
+
+        [https://github.com/microsoft/QuantumEllipticCurves/blob/dbf4836afaf7a9fab813cbc0970e65af85a6f93a/MicrosoftQuantumCrypto/EllipticCurves.qs#L456](QuantumQuantumCrypto).
+        `DistinctEllipticCurveClassicalPointAddition`.
+
+    """
+
+    n: int
+    R: ECPoint
```
fdmalone, line 62: Are we allowing a departure from pep8 naming convention?

qualtran/bloqs/factoring/ecc/_ec_point_test.py
```diff
@@ -0,0 +1,75 @@
...
+        (None, None),
+        (15, 13),
+        (2, 10),
+        (8, 3),
+        (12, 1),
+        (6, 6),
+        (5, 8),
+    ]
+
+    print()
```
fdmalone, line 69: Remove print?

qualtran/bloqs/factoring/ecc/_find_ecc_private_key.py
```diff
@@ -0,0 +1,129 @@
...
+    def curve_a(self) -> int:
+        if self.base_point.curve_a != self.public_key.curve_a:
+            raise ValueError("Inconsistent curve parameters in the two points.")
+        return self.base_point.curve_a
+
+    def build_composite_bloq(self, bb: 'BloqBuilder') -> Dict[str, 'SoquetT']:
+        x = bb.add(IntState(bitsize=self.n, val=self.base_point.x))
+        y = bb.add(IntState(bitsize=self.n, val=self.base_point.y))
+
+        x, y = bb.add(ECPhaseEstimateR(n=self.n, point=self.base_point), x=x, y=y)
```
fdmalone, line 98: Seems very easy to break ECC.

qualtran/bloqs/mod_arithmetic/_shims.py
```diff
@@ -0,0 +1,250 @@
...
+            (Sub(self.n), 1),
+            (Add(QUInt(self.n)), 1),
+            (CNOT(), 1),
+            (ModDbl(self.n, self.mod), 1),
+            (CHalf(self.n), 1),
+            (CSwap(self.n), 2),
+            (CNOT(), 1),
+        ]
+        summer = defaultdict(lambda: 0)
+        for bloq, n in listing:
```
fdmalone, line 114: This is because you can accidentally have duplicate bloqs in the listing which won't get counted if you just convert directly into a set right? This caught be out before.

qualtran/bloqs/mod_arithmetic/_shims.py
```diff
@@ -0,0 +1,250 @@
+#  Copyright 2023 Google LLC
+#
```
fdmalone, line 2: Can you add a module level docstring explaining the need for _shims?

fdmalone (changes requested):
_LGTM % k

dev_tools/autogenerate-bloqs-notebooks-v2.py
mpharrigan: ah good catch

qualtran/bloqs/arithmetic/__init__.py
mpharrigan: made explicit. these shims will be removed in follow-on PRs (replaced with real implementations)

qualtran/bloqs/factoring/ecc/_ec_point_test.py
mpharrigan: thanks

qualtran/bloqs/factoring/ecc/_find_ecc_private_key.py
mpharrigan: It's two phase estimations, michael. What could it cost? 10$?

qualtran/bloqs/mod_arithmetic/_shims.py
mpharrigan: added a note, but the listing follows the paper figure which indeed could have duplicate "keys"

qualtran/bloqs/factoring/ecc/_ec_add_r.py
mpharrigan: hmm I think in this case it probably helps more than it hurts. Something like `ec_point` is unnecessarily long and `r` doesn't match the convention in the paper(s) this module follows. Happy to revisit / change later when more of the bloqs have landed

qualtran/bloqs/mod_arithmetic/_shims.py
mpharrigan: added

qualtran/bloqs/factoring/ecc/_ec_add.py
mpharrigan: I figured this would be an extra nudge to import from e.g. `qualtran.bloqs.factoring.ecc` rather than the specific module (and I'm keeping to a relatively small number of bloqs per module). I guess we don't do this anywhere else, though

mpharrigan: changed

fdmalone (approved):

PR #925, opened 2024-05-07
QROAM stub

fdmalone:
In the interim before we have a bone fide qroam bloq with decomposition (xref #368), it's important to capture the costs in a better way that what is in https://github.com/quantumlib/Qualtran/blob/main/qualtran/bloqs/chemistry/black_boxes.py#L73, i.e. the bloq has a proper signature and can be used with existing chemistry bloqs in their decompositions. This would "fix" our current costs which rely on SelectSwapQROM and results a ~ 2x discrepancy from the literature values.

TODO:

- [ ] delete / replace the QROAM bloq in black_boxes. The downside to this is I would need to fake some input date for some bloqs which don't accept hamiltonian coefficients yet.

### DIFF
--- a/qualtran/bloqs/data_loading/qroam.py
+++ b/qualtran/bloqs/data_loading/qroam.py
@@ -63,7 +63,7 @@ def find_optimal_log_block_size(
         k = 0.5 * np.log2(iteration_length)
 
         def value(kk: List[int]):
-            return iteration_length / np.power(2, kk) + target_bitsize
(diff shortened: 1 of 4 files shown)


fdmalone:
I'll close this in favor of @tanujkhattar suggestion which is:

1. Add clean ancilla version of SelectSwapQROM which will fix the factor of 2 cost difference.
2. Add a QROM adjoint bloq which uses table lookups for QROM and SelectSwapQROM^ (xref: #950).

SelectSwapQROM.adjoint() would then return QROM^

PR #926, opened 2024-05-07
Enable mypy CI check

dstrain115:
- This enables the mypy CI check
- It also fixes the mypy issues introduced in the past few days.
- Also fixes items that occured from mypy 1.9.0 to 1.9.1
-   (mypy 1.9.1 is apparently more strict about the return type of generators)

### DIFF
--- a/qualtran/_infra/bloq.py
+++ b/qualtran/_infra/bloq.py
@@ -31,6 +31,7 @@ if TYPE_CHECKING:
         BloqBuilder,
         CompositeBloq,
         CtrlSpec,
+        GateWithRegister,
         Register,
         Signature,
         Soquet,
@@ -167,7 +168,7 @@ class Bloq(metaclass=abc.ABCMeta):
         bb, initial_soqs = BloqBuilder.from_signature(self.signature, add_registers_allowed=False)
         return bb.finalize(**bb.add_d(self, **initial_soqs))
 
-    def adjoint(self) -> 'Adjoint':
+    def adjoint(self) -> 'GateWithRegister':
         """The adjoint of this bloq.
 
         Bloq authors can override this method in certain circumstances. Otherwise, the default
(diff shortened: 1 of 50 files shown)


qualtran/_infra/bloq.py
```diff
@@ -166,7 +167,7 @@ def as_composite_bloq(self) -> 'CompositeBloq':
         bb, initial_soqs = BloqBuilder.from_signature(self.signature, add_registers_allowed=False)
         return bb.finalize(**bb.add_d(self, **initial_soqs))
 
-    def adjoint(self) -> 'Bloq':
+    def adjoint(self) -> 'Adjoint':
```
anurudhp, line 170: Is this correct? `bloq.adjoint()` in general need not return an `Adjoint` right?

dstrain115: How about GateWithRegisters?  There are quite a bit of places where there is `bloq.adjoint() ** 2` which requires a GateWithRegisters.  (Bloq has no pow)

qualtran/_infra/bloq.py
```diff
@@ -166,7 +168,7 @@ def as_composite_bloq(self) -> 'CompositeBloq':
         bb, initial_soqs = BloqBuilder.from_signature(self.signature, add_registers_allowed=False)
         return bb.finalize(**bb.add_d(self, **initial_soqs))
 
-    def adjoint(self) -> 'Bloq':
+    def adjoint(self) -> 'GateWithRegisters':
```
mpharrigan, line 171: ??

dstrain115: This is because Bloq has no `__pow__ ` but we often do bloq.adjoint() ** 2 or similar.
Alternatives welcome.  I can also cast the bloq.adjoint() to GateWithRegisters?

dstrain115: Actually, I fixed it in GateWithRegisters instead.  Reverted this.

qualtran/_infra/bloq.py
dstrain115: Update: reverted.  Found a different way to fix it.

mpharrigan (approved):

PR #927, opened 2024-05-07
Simplify `cbloq.to_cirq_circuit()` with sensible defaults and add `cbloq.to_cirq_circuit_and_quregs()`

tanujkhattar:
More often than not, we construct a composite bloq corresponding to an algorithm / primitive in Qualtran and then wish to convert it to a cirq circuit. The `CompositeBloq.to_cirq_circuit` api is unnecessarily complex for this use case. 

This PR refactors the api s.t. the previous API for `CompositeBloq.to_cirq_circuit()` is now renamed to  `CompositeBloq.to_cirq_circuit_and_quregs()` and the new `CompositeBloq.to_cirq_circuit()` simply returns a `cirq.FrozenCircuit`. It also doesn't need you to provide the input `quregs` and can use sensible defaults using `get_named_qubits(self.signature.lefts())`. 

Therefore, to get a cirq circuit from a composite bloq, users can now simply do `circuit = cbloq.to_cirq_circuit()` for a reasonable default and can provide the input quregs explicitly or call `cbloq.to_cirq_circuit_and_quregs` for more advanced use cases.


qualtran/_infra/controlled_test.py
```diff
@@ -124,7 +126,7 @@ def _test_cirq_equivalence(bloq: Bloq, gate: cirq.Gate):
     bloq = Controlled(Swap(5), CtrlSpec(qdtypes=QUInt(4), cvs=0b0101))
     quregs = get_named_qubits(bloq.signature)
     ctrl, x, y = quregs['ctrl'], quregs['x'], quregs['y']
-    circuit1, _ = bloq.decompose_bloq().to_cirq_circuit(None, **quregs)
+    circuit1, _ = bloq.decompose_bloq().to_cirq_circuit_and_quregs(None, **quregs)
```
mpharrigan, line 129: do you wanna go through these (where the second return value is ignored) and use the new method?

tanujkhattar: done

mpharrigan (approved):

Issue #608, opened 2024-01-27
Better wire symbol testing


fdmalone:
Added a function in #602 which has the limitation that it only works if the register's shape=(). wire_symbol method probably shouldn't use the binst field.

mpharrigan:
In #791, there's an open plan to remove the `binst` field.

PR #928, opened 2024-05-07
Improve wire symbol testing

dstrain115:
- Allow wire symbol testing to accept non-trivial shapes (multi-dimensional registers)
- Allow wire symbol testing to accept WireSymbols without text
- Add wire symbol testing to Toffoli()

Fixes: #608


mpharrigan (approved):
cool, thanks!

PR #930, opened 2024-05-08
Calculate complementary polynomial for QSP using flip convolution.

Epsilon1024:
Calculate complementary polynomial using Eq. 60 of the [paper](https://arxiv.org/abs/2308.01501)


Epsilon1024:
I'm currently having an issue with the precision when checking the polynomial pairs on the unit circle.  The regular QSP test uses a precision of 1e-7.  This method runs a little bit less precise (I last measured around 1.2e-7 before the test failed.)

The issue seems to be with the minimizer.  Currently, I'm not passing an argument to the "jac" parameter so it is estimating the gradient.  The next thing I can think of to try would be to find a function to more accurately estimate the gradient (unless I can solve it directly) and pass that to the minimizer.  

What are your thoughts on this, @anurudhp?

qualtran/bloqs/qsp/fast_qsp_test.py
```diff
@@ -0,0 +1,33 @@
...
+
+# @pytest.mark.parametrize("degree", [4, 5])
+@pytest.mark.parametrize("degree", [5])
+def test_complementary_polynomial_quick(degree: int):
+    random_state = np.random.RandomState(42)
+
+    for _ in range(2):
+        P = random_qsp_polynomial(degree, random_state=random_state)
+        Q = fast_complementary_polynomial(P, verify=True)
+        check_polynomial_pair_on_random_points_on_unit_circle(P, Q, random_state=random_state)
```
anurudhp, line 33:
```suggestion
        check_polynomial_pair_on_random_points_on_unit_circle(P, Q, random_state=random_state, rtol=precision)
```
(see comment above on `precision`)

qualtran/bloqs/qsp/fast_qsp_test.py
```diff
@@ -0,0 +1,33 @@
...
+
+from .fast_qsp import fast_complementary_polynomial
+from .generalized_qsp_test import (
+    check_polynomial_pair_on_random_points_on_unit_circle,
+    random_qsp_polynomial,
+)
+
+
+# @pytest.mark.parametrize("degree", [4, 5])
+@pytest.mark.parametrize("degree", [5])
```
anurudhp, line 26: does it work for larger degrees? how well does the precision scale with the degree?

qualtran/bloqs/qsp/fast_qsp_test.py
anurudhp, line 1: could you also add a test that verifies the GQSP unitaries? i.e. similar to
https://github.com/quantumlib/Qualtran/blob/6cd49f11b256daaba759cb2acebeff2e8899b1d6/qualtran/bloqs/qsp/generalized_qsp_test.py#L179-L191
(but compute `Q` explicitly using your new method and pass it to `verify_generalized_qsp`)

qualtran/bloqs/qsp/generalized_qsp_test.py
```diff
@@ -68,6 +68,10 @@ def check_polynomial_pair_on_random_points_on_unit_circle(
 
     for _ in range(n_points):
         z = np.exp(random_state.random() * np.pi * 2j)
+        # DO NOT SUBMIT: Logs
+        a = 1 - (np.abs(P(z)) ** 2 + np.abs(Q(z)) ** 2)
+        print(a)
+
```
anurudhp, line 74: remove this?

qualtran/bloqs/qsp/fast_qsp.py
```diff
@@ -0,0 +1,84 @@
...
+        self.conv_p_negative = self.complex_conv_by_flip_conj(poly.real, poly.imag) * -1
+        self.conv_p_negative[poly.shape[0] - 1] = 1 - np.linalg.norm(poly) ** 2
+
+    def normalize(self, input_poly, granularity=8):
+        P = np.pad(input_poly, (0, 2**granularity - input_poly.shape[0]))
+        ft = np.fft.fft(P)
+
+        # Normalize P
+        P_norms = np.abs(ft)
+        return input_poly / np.max(P_norms)
```
anurudhp, line 30: why are we dividing by the maximum here? (in case it's explained in the paper, add a comment linking to the para/equation etc. in the paper)

qualtran/bloqs/qsp/fast_qsp.py
anurudhp, line 1: please add type hints and docstrings to all functions

qualtran/bloqs/qsp/fast_qsp.py
```diff
@@ -0,0 +1,84 @@
...
+import numpy as np
+from scipy.optimize import minimize
+
+
+class FastQSP:
+    def __init__(self, poly):
+        self.conv_p_negative = self.complex_conv_by_flip_conj(poly.real, poly.imag) * -1
+        self.conv_p_negative[poly.shape[0] - 1] = 1 - np.linalg.norm(poly) ** 2
+
+    def normalize(self, input_poly, granularity=8):
```
anurudhp, line 24: please add types and docstring. could you explain what `granularity` is?

qualtran/bloqs/qsp/fast_qsp.py
```diff
@@ -0,0 +1,84 @@
...
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import numpy as np
+from scipy.optimize import minimize
+
+
+class FastQSP:
```
anurudhp, line 19: add class docstring

qualtran/bloqs/qsp/fast_qsp.py
```diff
@@ -0,0 +1,84 @@
...
+
+    qsp = FastQSP(poly)
+
+    minimizer = minimize(
+        qsp.loss_function,
+        q_initial_normalized,
+        method="L-BFGS-B",
+        tol=TOLERANCE,
+        bounds=[(-1, 1)],
+        options={"maxcor": 50, "eps": 1e-15, "maxls": 20},
```
anurudhp, line 81: what is `eps` here? is it for the gradient?

qualtran/bloqs/qsp/fast_qsp.py
```diff
@@ -0,0 +1,84 @@
...
+        real_conv = conv_real_part - conv_imag_part
+        imag_conv = conv_real_imag + conv_imag_real
+
+        # Combine to form the complex result
+        return real_conv + 1j * imag_conv
+
+
+def fast_complementary_polynomial(poly, verify=True, granularity=8):
+    DTYPE = np.complex128
+    TOLERANCE = 1e-12
```
anurudhp, line 67: what is this tolerance for? is it the gap b/w the output and the actual value?

qualtran/bloqs/qsp/fast_qsp.py
```diff
@@ -0,0 +1,84 @@
...
+
+        # Compute real and imaginary part of the convolution
+        real_conv = conv_real_part - conv_imag_part
+        imag_conv = conv_real_imag + conv_imag_real
+
+        # Combine to form the complex result
+        return real_conv + 1j * imag_conv
+
+
+def fast_complementary_polynomial(poly, verify=True, granularity=8):
```
anurudhp, line 65: it seems `verify` is not being used?

qualtran/bloqs/qsp/fast_qsp.py
```diff
@@ -0,0 +1,84 @@
...
+    DTYPE = np.complex128
+    TOLERANCE = 1e-12
+    poly = np.array(poly, dtype=DTYPE)
+    np.random.seed(42)
```
anurudhp, line 69: if you need randomness, take the seed as an input parameter. You can use something like https://github.com/quantumlib/Cirq/blob/main/cirq-core/cirq/value/random_state.py

anurudhp (changes requested):
very cool! I left some comments (the docstring ones are useful but not urgent).

Could you add some more tests - larger degrees and verifying the actual GQSP unitaries? I think tweaking the precision is okay for now, say if it works with `1e-5` or something. We can use a more powerful optimizer later on to get more precise results, as long as the optimization (loss function) is correct

Epsilon1024:
> very cool! I left some comments (the docstring ones are useful but not urgent).
> 
> Could you add some more tests - larger degrees and verifying the actual GQSP unitaries? I think tweaking the precision is okay for now, say if it works with `1e-5` or something. We can use a more powerful optimizer later on to get more precise results, as long as the optimization (loss function) is correct

I'll add in the rest of the tests (I initially forgot to make this PR as a draft - there are a few things I'm planning to add/clean up).  

I just noticed that while the 4th order polynomials fit with 1e-5 tolerance, the larger order polynomials have a much looser tolerance which is not great.  (A 5th order gave me an error in the 1e-3 range).  

We might need to address this sooner rather than later.

Epsilon1024:
Update:  I was able to get the error below 1e-7 by tweaking some parameters.  (I used a different miniization method than was used by the authors of the paper, and I set the gradient to be calculated by a 3 point estimation when the default was 2).  

The issue that I'm facing now is that the tests that check for all real coefficents are giving a higher error on the unit circle check.

 The calculation seems to be a bit different depending on if we want only real coefficients.  According to the [authors' github] (https://github.com/Danimhn/GQSP-Code/blob/main/FindingQ.ipynb), the flip convolution is performed differently for real and complex polynomials.  It appears that the minimizer they used (and I assume also the minimizer I am using) won't work for complex numbers.  As a result, they deconstruct the real and imaginary components (which is why two different methods are being used).  

My understanding is that there can be multiple Q polynomials for any given P (which would mean that for an all-real P, we can have both a complex Q and an all-real Q).  Please correct me if this is wrong.

Tomorrow, I plan to look into this further and see if I can bring the error down for the real coefficients.  If so, then after some minor cleanup, I should be ready to send this out.

Epsilon1024:
I also found that the test_generalized_real_qsp_with_symbolic_signal_matrix fails with the new qsp method.  The test will run 10 random polynomials of each degree.  Oddly enough, most of them pass the test (within a tolerance of 1e-5).  However, 1 of these polynomials will produce an error in the 1e-1 range.  This seems to only happen with 10th degree polynomials.  This one does not seem to be a precision issue like the others, but seems to be a bug.  However, I can't find anything obvious in the symbolic_qsp_matrix method that could cause the error.  The P and Q polynomial pairs pass the unit circle check just before the matrix is made.

anurudhp:
> I also found that the test_generalized_real_qsp_with_symbolic_signal_matrix fails with the new qsp method. The test will run 10 random polynomials of each degree. Oddly enough, most of them pass the test (within a tolerance of 1e-5). However, 1 of these polynomials will produce an error in the 1e-1 range. This seems to only happen with 10th degree polynomials. This one does not seem to be a precision issue like the others, but seems to be a bug. However, I can't find anything obvious in the symbolic_qsp_matrix method that could cause the error. The P and Q polynomial pairs pass the unit circle check just before the matrix is made.

Could you add an issue? You can extract the failing `P, Q` from your run and report the bug. Also, does this failing polynomial pair pass `test_generalized_qsp_with_complex_poly_on_random_unitaries`? 

The "symbolic gqsp" checker only computes an upper-bound on the error, which _may_ be larger than the actual error (not sure).
https://github.com/quantumlib/Qualtran/blob/abfbfab95b35a7594b7eb8d109aa295ea787da19/qualtran/bloqs/qsp/generalized_qsp_test.py#L296-L297

Epsilon1024:
There are two small followups I will be adding after this PR is submitted:
1) This version does not normalize P before calculating Q.  (From my understanding the first version of the default QSP did not normalize P, but this function was added later).  I will be adding the normalization in a follow-up PR (that is nearly finished).

2) The main challenge that I faced in this PR was setting a high tolerance for the case with all real polynomials.  I found the  method in this PR performs comparatively to that in the code provided by the authors of the paper.  I will write out my findings in a notebook to go along with this.

qualtran/bloqs/qsp/fast_qsp.py
```diff
@@ -0,0 +1,141 @@
...
+
+        conv_real_imag = np.convolve(real_part, imag_flip, mode="full")
+        conv_imag_real = np.convolve(imag_part, real_flip, mode="full")
+
+        # Compute real and imaginary part of the convolution
+        real_conv = conv_real_part - conv_imag_part
+        imag_conv = conv_real_imag + conv_imag_real
+
+        # Combine to form the complex result
+        return real_conv + 1j * imag_conv
```
anurudhp, line 83: can be static?

qualtran/bloqs/qsp/fast_qsp.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def fast_complementary_polynomial(
+    P: Union[NDArray[np.number], Sequence[complex]],
+    only_reals: bool = False,
+    tolerance: float = 1e-10,
+):
+    """
+    Computes the Q polynomial given P
```
anurudhp, line 92: docstring nit, see comment on FastQSP docstring above

qualtran/bloqs/qsp/fast_qsp.py
```diff
@@ -0,0 +1,141 @@
...
+
+    Computes polynomial $Q$ of degree at-most that of $P$, satisfying
+
+        $$ \abs{P(e^{i\theta})}^2 + \abs{Q(e^{i\theta})}^2 = 1 $$
+
+    using the flip convolution method described in Eq(60).  This
+    a replacement for the complementary_polynomial in the
```
anurudhp, line 99: I'd say it's an alternative rather than a replacement

qualtran/bloqs/qsp/fast_qsp.py
anurudhp: bump :D I think the current best practice is to use `np.random.Generator` (refactor is also underway in cirq: https://github.com/quantumlib/Cirq/pull/6566). 

Add an argument `rng: Optional[np.random.Generator] = None`, and then you can initialize one if it isn't provided: 
```py
if rng is None:
    rng = np.random.default_rng()
```

And use `rng` instead of `np.random` from this point on.

qualtran/bloqs/qsp/fast_qsp.py
```diff
@@ -0,0 +1,141 @@
...
+            the desired tolerance used by the code that calls this method.
+
+    References:
+    [Generalized Quantum Signal Processing](https://arxiv.org/abs/2308.01501)
+        Motlagh and Wiebe. (2023). Equation 60.
+    """
+    np.random.seed(42)
+    if only_reals:
+        poly = np.array(P, dtype=np.float64)
+        q_initial = np.random.randn(poly.shape[0])
```
anurudhp, line 129: see `rng` comment above

qualtran/bloqs/qsp/fast_qsp.py
```diff
@@ -0,0 +1,141 @@
...
+    else:
+        poly = np.array(P, dtype=np.complex128)
+        q_initial = np.random.randn(poly.shape[0] * 2)
```
anurudhp, line 132: see `rng` comment above

qualtran/bloqs/qsp/fast_qsp.py
```diff
@@ -0,0 +1,141 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+from typing import Sequence, Union
+
+import numpy as np
+from numpy._typing import NDArray
+from scipy.optimize import minimize
+
+
+class FastQSP:
```
anurudhp, line 21: Rename class to something like `FastComplementaryQSPHelper`?

qualtran/bloqs/qsp/fast_qsp_test.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+@pytest.mark.slow
+@pytest.mark.parametrize(
+    "degree, num_tests, precision", [(5, 20, 2e-5), (10, 20, 2e-5), (20, 5, 2e-5), (30, 3, 2e-5)]
+)
+def test_complementary_polynomial(degree: int, num_tests: int, precision: float):
+    random_state = np.random.RandomState(42)
+
+    results = []
```
anurudhp, line 60: remove this?

qualtran/bloqs/qsp/fast_qsp_test.py
```diff
@@ -0,0 +1,127 @@
...
+    for _ in range(num_tests):
+        P = random_qsp_polynomial(degree, random_state=random_state)
+        Q = fast_complementary_polynomial(P)
+        results.append(np.array(Q))
```
anurudhp, line 64: remove this?

qualtran/bloqs/qsp/fast_qsp_test.py
```diff
@@ -0,0 +1,127 @@
...
+        check_polynomial_pair_on_random_points_on_unit_circle(
+            P, Q, random_state=random_state, rtol=precision
+        )
+
+
+@pytest.mark.slow
+@pytest.mark.parametrize("degree, num_tests", [(2, 10), (3, 10), (4, 10), (5, 10), (10, 10)])
+def test_fast_real_qsp_with_symbolic_signal_matrix(degree: int, num_tests: int):
```
anurudhp, line 72:
```suggestion
def test_fast_qsp_on_random_unitaries(degree: int, num_tests: int):
```

qualtran/bloqs/qsp/fast_qsp_test.py
```diff
@@ -0,0 +1,127 @@
...
+        (2, 2, 0, 1e-5),
+        (2, 2, 1, 1e-5),
+        (2, 2, 2, 1e-5),
+        (1, 5, 2, 1e-4),
+        (1, 5, 0, 1e-4),
+        (3, 5, 0, 1e-4),
+        (3, 5, 2, 1e-4),
+        (2, 20, 1, 1e-1),
+    ],
+)
```
anurudhp, line 117: I think you can parametrize `bitsize` separately because the complexity of computing Q and verifying P,Q only add up (with bitsize affecting only the latter). You can just use `@pytest.mark.parametrize("bitsize", [1, 2, 3])`

anurudhp (approved):
Looks good! Mostly docstring+type-hint nits. Also please fix the `rng` comment, best practice is to use `np.random.Generator`.

Epsilon1024:
As a note: I ran some checks to see if normalizing P before running the tests on fast_complementary_polynomial would give us more accurate results.  From my checks, it performs slightly worse than not being normalized.  This tells me that the precision problem is not related to normalization.

tanujkhattar (commented):
@anurudhp @Epsilon1024 Can we merge this and continue improvements in follow-up PRs? Maybe open an issue to track the potential improvements?

qualtran/bloqs/qsp/fast_qsp.py
```diff
@@ -0,0 +1,141 @@
...
+            complex, an error will be thrown.
+        tolerance: The allowable tolerance for finding the minimum of the
+            qsp loss function. In general, this number should be at least 1/10 of
+            the desired tolerance used by the code that calls this method.
+
+    References:
+    [Generalized Quantum Signal Processing](https://arxiv.org/abs/2308.01501)
+        Motlagh and Wiebe. (2023). Equation 60.
+    """
+    np.random.seed(42)
```
anurudhp, line 126: fix hardcoder random seed

anurudhp:
@tanujkhattar good to go % hardcoded random seed

qualtran/bloqs/qsp/fast_qsp.py
Epsilon1024: Yep.  I'm running the last test right now and am about to update this PR.  It should be ready to submit in the next 15 mins.

qualtran/bloqs/qsp/fast_qsp.py
```diff
@@ -0,0 +1,156 @@
...
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+from typing import Sequence, Union
+
+import numpy as np
+from numpy._typing import NDArray
```
anurudhp, line 17:
```suggestion
from numpy.typing import NDArray
```

anurudhp (approved):
LGTM!

tanujkhattar (approved):

PR #934, opened 2024-05-08
Remove short_name method

dstrain115:
- Replace with wire_symbol(reg=None)

For item 2 in #728


qualtran/_infra/controlled.py
```diff
@@ -433,11 +433,13 @@ def _unitary_(self):
...
-    def wire_symbol(self, reg: Register, idx: Tuple[int, ...] = tuple()) -> 'WireSymbol':
+    def wire_symbol(self, reg: Optional[Register], idx: Tuple[int, ...] = tuple()) -> 'WireSymbol':
+        from qualtran.drawing import Text
+
+        if reg is None:
+            return Text(f'C[{self.subbloq.wire_symbol(reg=None)}]')
         if reg.name not in self.ctrl_reg_names:
             # Delegate to subbloq
-            print(self.subbloq)
-            print(type(self.subbloq))
```
mpharrigan, line 440: whoops

mpharrigan:
One of the motivations was that the author of the `wire_symbol` method could declare that "the symbols are sufficient to tell what bloq this is, I don't need an additional label in the diagrams". For example: writing "CNOT" over a circle and modplus in the musical score diagram just adds clutter. Follow on work would aim to try to make some of the symbols more expressive. For example: the Add bloq could have its symbols be `x` and `x+y` and then get rid of the label.

Could you add support for this use case? both in the types and docs and in the musical score drawer?

dstrain115:
> One of the motivations was that the author of the `wire_symbol` method could declare that "the symbols are sufficient to tell what bloq this is, I don't need an additional label in the diagrams". For example: writing "CNOT" over a circle and modplus in the musical score diagram just adds clutter. Follow on work would aim to try to make some of the symbols more expressive. For example: the Add bloq could have its symbols be `x` and `x+y` and then get rid of the label.
> 
> Could you add support for this use case? both in the types and docs and in the musical score drawer?

Would returning `Text('')` be sufficient here?  (wire_symbol() should return a WireSymbol)

I will probably leave modifying the symbols of specific gates to a different PR (also, will only do obvious ones, like arithmetic operators, since I am not super familiar with all of these bloqs)

mpharrigan:
I sortof defer to you on how to signal "no label", but whatever it is: it should be documented and the `musical_score` drawer should avoid drawing the label

dstrain115:
@mpharrigan I think this is ready for review.  Please let me know if I missed anything.

mpharrigan (approved):

PR #882, opened 2024-04-17
Add walk operator example for sparse chem ham.

fdmalone:
Working on #841, adds a walk operator example for the sparse hamiltonian which fits the usual select and prepare construction. Just adds an example and some autodoc. ~~I have an organizational / doc question: I added a BloqExample for building the walk operator BUT this reproduces the docstring for QubitizationWalkOperator in sparse.ipynb which isn't ideal. Would it be better to add this as an example in the notebook for QubitizationWalkOperator, I feel this might make the notebook a bit crowded if we keep adding examples there?~~

Currently there is a test failing for serialization which I don't really understand yet because Prepare and Select pass the serialization test? Any idea @tanujkhattar?

I also fixed the latex for QubitizationWalkOperator.

PR #938, opened 2024-05-10
Fix chemistry mypy error

fdmalone:
#882 prematurely merged itself before the mypy check passed. Fix it here.


mpharrigan (approved):

fdmalone:
apparently not

fdmalone:
very odd, runnning mypy locally doesn't raise an error.

mpharrigan:
getting the exact version of mypy is annoyingly important

fdmalone:
I have the version in dev env 🤷 (1.10.0)

Issue #935, opened 2024-05-09
`t_complexity` doesn't work for StatePreparationAliasSampling


mstechly:
When I run this:

```
bloq = StatePreparationAliasSampling.from_lcu_probs([0.25, 0.5, 0.25], probability_epsilon=0.05)
bloq.t_complexity()
```

I get the following error:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/mstechly/.pyenv/versions/2024-05-07-qualtran-conversion/lib/python3.9/site-packages/qualtran/_infra/bloq.py", line 420, in t_complexity
    return t_complexity(self)
  File "/Users/mstechly/.pyenv/versions/2024-05-07-qualtran-conversion/lib/python3.9/site-packages/qualtran/cirq_interop/t_complexity_protocol.py", line 240, in t_complexity
    ret = _t_complexity_for_gate_or_op(stc)
  File "/Users/mstechly/.pyenv/versions/2024-05-07-qualtran-conversion/lib/python3.9/site-packages/cachetools/__init__.py", line 675, in wrapper
    v = func(*args, **kwargs)
  File "/Users/mstechly/.pyenv/versions/2024-05-07-qualtran-conversion/lib/python3.9/site-packages/qualtran/cirq_interop/t_complexity_protocol.py", line 224, in _t_complexity_for_gate_or_op
    return _t_complexity_from_strategies(gate_or_op, strategies)
  File "/Users/mstechly/.pyenv/versions/2024-05-07-qualtran-conversion/lib/python3.9/site-packages/qualtran/cirq_interop/t_complexity_protocol.py", line 204, in _t_complexity_from_strategies
    ret = strategy(stc)
  File "/Users/mstechly/.pyenv/versions/2024-05-07-qualtran-conversion/lib/python3.9/site-packages/qualtran/cirq_interop/t_complexity_protocol.py", line 163, in _from_bloq_build_call_graph
    r = t_complexity(bloq)
  File "/Users/mstechly/.pyenv/versions/2024-05-07-qualtran-conversion/lib/python3.9/site-packages/qualtran/cirq_interop/t_complexity_protocol.py", line 240, in t_complexity
    ret = _t_complexity_for_gate_or_op(stc)
  File "/Users/mstechly/.pyenv/versions/2024-05-07-qualtran-conversion/lib/python3.9/site-packages/cachetools/__init__.py", line 675, in wrapper
    v = func(*args, **kwargs)
[...]

Issue #911, opened 2024-04-30
Notebook tests are flaky?


fdmalone:
The notebook tests in the CI seem to randomly fail every so often. Rerunning them seems to work. Any idea what's going on. Heres's a recent traceback:

```
[...]

PR #944, opened 2024-05-10
Fix `t_complexity` for symbolic ham. sim. by gqsp example

anurudhp:
- fixes the first bug reported in https://github.com/quantumlib/Qualtran/issues/935#issuecomment-2102719963 - adds a simplified override for the call graph of a symbolic HamSimbyGQSP bloq.
- introduces `_BookkeepingBloq` for utility bloqs that are used for book-keeping (i.e. cast, split/join, alloc/free)

### DIFF
--- a/qualtran/bloqs/hamiltonian_simulation/hamiltonian_simulation_by_gqsp.py
+++ b/qualtran/bloqs/hamiltonian_simulation/hamiltonian_simulation_by_gqsp.py
@@ -12,7 +12,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 from functools import cached_property
-from typing import cast, Dict, Tuple, TYPE_CHECKING, Union
+from typing import cast, Dict, Set, Tuple, TYPE_CHECKING, Union
 
 import numpy as np
 from attrs import field, frozen
@@ -29,6 +29,7 @@ from qualtran.symbolics import is_symbolic, Shaped, SymbolicFloat, SymbolicInt
 
 if TYPE_CHECKING:
     from qualtran import BloqBuilder, SoquetT
+    from qualtran.resource_counting import BloqCountT, SympySymbolAllocator
 
 
 @frozen
@@ -176,6 +177,21 @@ class HamiltonianSimulationByGQSP(GateWithRegisters):
 
         return soqs
 
+    def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
+        if self.is_symbolic():
+            from qualtran.bloqs.basic_gates.su2_rotation import SU2RotationGate
+
+            d = self.degree
+            return {
+                (self.walk_operator.prepare, 1),
+                (self.walk_operator.prepare.adjoint(), 1),
+                (self.walk_operator.controlled(control_values=[0]), d),
+                (self.walk_operator.adjoint().controlled(), d),
+                (SU2RotationGate.arbitrary(ssa), 2 * d + 1),
+            }
+
+        return super().build_call_graph(ssa)
+
 
 @bloq_example
 def _hubbard_time_evolution_by_gqsp() -> HamiltonianSimulationByGQSP:
@@ -192,9 +208,8 @@ def _symbolic_hamsim_by_gqsp() -> HamiltonianSimulationByGQSP:
 
     from qualtran.bloqs.hubbard_model import get_walk_operator_for_hubbard_model
 
-    walk_op = get_walk_operator_for_hubbard_model(2, 2, 1, 1)
-
-    t, inv_eps = sympy.symbols("t N")
+    tau, t, inv_eps = sympy.symbols(r"\tau t \epsilon^{-1}", positive=True)
+    walk_op = get_walk_operator_for_hubbard_model(2, 2, tau, 4 * tau)
     symbolic_hamsim_by_gqsp = HamiltonianSimulationByGQSP(walk_op, t=t, precision=1 / inv_eps)
     return symbolic_hamsim_by_gqsp
 
--- a/qualtran/bloqs/hubbard_model.py
+++ b/qualtran/bloqs/hubbard_model.py
@@ -66,6 +66,7 @@ from qualtran.bloqs.select_and_prepare import PrepareOracle, SelectOracle
 from qualtran.bloqs.state_preparation.prepare_uniform_superposition import (
     PrepareUniformSuperposition,
 )
+from qualtran.symbolics.math_funcs import acos, ssqrt
 
 if TYPE_CHECKING:
     from qualtran.symbolics import SymbolicFloat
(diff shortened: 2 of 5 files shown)


qualtran/bloqs/hamiltonian_simulation/hamiltonian_simulation_by_gqsp.py
```diff
@@ -197,9 +213,8 @@ def _symbolic_hamsim_by_gqsp() -> HamiltonianSimulationByGQSP:
 
     from qualtran.bloqs.hubbard_model import get_walk_operator_for_hubbard_model
 
-    walk_op = get_walk_operator_for_hubbard_model(2, 2, 1, 1)
-
-    t, inv_eps = sympy.symbols("t N")
+    tau, t, inv_eps = sympy.symbols(r"\tau t \epsilon^{-1}", positive=True)
```
fdmalone, line 216: would epsilon be a better variable here since you're just passing in 1 / inv_eps anyway?

qualtran/bloqs/hubbard_model.py
```diff
@@ -303,12 +303,14 @@ def signature(self) -> Signature:
     def decompose_from_registers(
         self, *, context: cirq.DecompositionContext, **quregs: NDArray[cirq.Qid]
     ) -> Iterator[cirq.OP_TREE]:
+        from qualtran.resource_counting.symbolic_counting_utils import acos, ssqrt
```
fdmalone, line 306: Will have to update your imports if #941  goes in first.

qualtran/bloqs/util_bloqs.py
```diff
@@ -50,8 +50,31 @@
     from qualtran.simulation.classical_sim import ClassicalValT
 
 
+class UtilBloq(Bloq, metaclass=abc.ABCMeta):
```
fdmalone, line 53: is ABCMeta doing anything here? You're not specifying any abstractmethods.

qualtran/bloqs/util_bloqs.py
```diff
@@ -50,8 +50,31 @@
...
+    """Base class for util bloqs.
+
+    1. Trivial controlled - pass through the control register.
+    2. Do not affect T complexity.
+    """
+
+    def get_ctrl_system(
```
fdmalone, line 60: Would this me more appropriate as a mixin similar to SpecializedSingleQubitControlledGate?

qualtran/bloqs/util_bloqs.py
```diff
@@ -420,13 +406,14 @@ def _t_complexity_(self) -> 'TComplexity':
...
+class Cast(UtilBloq):
     """Cast a register from one n-bit QDType to another QDType.
 
 
     Args:
-        in_qdtype: Input QDType to cast from.
-        out_qdtype: Output QDType to cast to.
+        inp_dtype: Input QDType to cast from.
+        out_dtype: Output QDType to cast to.
+        shape: shape of the register to cast.
```
fdmalone, line 416: good catch.

fdmalone (changes requested):
LGTM % I'm slightly uneasy about the util bloq inheritance.

qualtran/bloqs/hamiltonian_simulation/hamiltonian_simulation_by_gqsp.py
anurudhp: Ah this is my failed attempt to fix sympy messing up `log` terms inside `Order`.

```py
from sympy import Symbol, log
from qualtran.resource_counting import big_O

x = Symbol('x', positive=True)
print(big_O(log(x)))
print(big_O(log(1 / x)))
print(big_O(log(log(x))))
```
outputs
```
O(log(1/x), (x, oo))
O(log(1/x), (x, oo))
O(log(-log(1/x)), (x, oo))
```

I'll keep track of this in an issue for now, and simplify the code to use `epsilon`.

qualtran/bloqs/hubbard_model.py
anurudhp: fixed

qualtran/bloqs/util_bloqs.py
anurudhp: Right now they are equivalent. The only difference is that `SpecializedSingleQubitControlledGate` has a method marked as `abc.abstractmethod` (i.e. `control_registers`), so it seems we don't need to explicitly specify `metaclass=abc.ABCMeta`. But here we have to.

qualtran/bloqs/hamiltonian_simulation/hamiltonian_simulation_by_gqsp.py
anurudhp: Ah no, I remember now, `sympy.Order` only supports multivariate limits to the same point, so I picked the symbol `1/eps -> inf` (instead of `eps -> 0`)

anurudhp:
@fdmalone @tanujkhattar ptal (notebook test failure is unrelated (#911))

qualtran/bloqs/util_bloqs.py
mpharrigan, line 53: I think this is reasonable, but please

 - Rename to `BookkeepingBloq`, which is more specific than "util" (a name I despise -- we can change the module later :). This also makes the properties make sense: a bookkeeping bloq is trivially controlled and does notn affect the t complexity
 - make private, i.e. `_BookkeepingBloq`

qualtran/bloqs/util_bloqs.py
```diff
@@ -50,8 +50,31 @@
     from qualtran.simulation.classical_sim import ClassicalValT
 
 
+class UtilBloq(Bloq, metaclass=abc.ABCMeta):
+    """Base class for util bloqs.
+
+    1. Trivial controlled - pass through the control register.
+    2. Do not affect T complexity.
+    """
```
mpharrigan, line 58: grammar & style in this docstring.

qualtran/bloqs/util_bloqs.py
mpharrigan: to be clear: they're not equivalent. This implementation _ignores_ ctrl soquets whereas the other will plug in a single-bit control to an expected location

mpharrigan (commented):
style things

mpharrigan:
Can you update the PR description to summarize the changes present in this PR

qualtran/bloqs/util_bloqs.py
anurudhp: Ah I just meant they are equivalent in design, i.e. not a mixin but a subclass of Bloq

qualtran/bloqs/util_bloqs.py
anurudhp: renamed to `_BookkeepingBloq`

qualtran/bloqs/util_bloqs.py
anurudhp: polished the docstring, how does it look now?

PR #945, opened 2024-05-10
Support symbolic parameters in QROM bloq

tanujkhattar:
Part of a larger effort to refresh all unary iteration based bloqs and add symbolic support. Will also form the base of refactoring SelectSwapQROM and fix cost discrepancies with QROAM. 

As a follow up PR, i'll update the `unary_iteration` bloq to support symbolics in general and also its other derived bloqs.

Based on top off https://github.com/quantumlib/Qualtran/pull/942


qualtran/bloqs/data_loading/qrom.py
```diff
@@ -70,41 +74,81 @@ class QROM(UnaryIterationGate):
...
+        assert isinstance(self.selection_bitsizes, tuple)
+        assert isinstance(self.target_bitsizes, tuple)
 
     @classmethod
-    def build(cls, *data: ArrayLike, num_controls: int = 0) -> 'QROM':
-        _data = [np.array(d, dtype=int) for d in data]
-        selection_bitsizes = tuple((s - 1).bit_length() for s in _data[0].shape)
+    def build_from_data(cls, *data: ArrayLike, num_controls: SymbolicInt = 0) -> 'QROM':
+        _data = np.array([np.array(d, dtype=int) for d in data])
+        selection_bitsizes = tuple((s - 1).bit_length() for s in _data.shape[1:])
```
fdmalone, line 119: should this raise an error if data is symbolic?

qualtran/bloqs/data_loading/qrom.py
```diff
@@ -70,41 +74,81 @@ class QROM(UnaryIterationGate):
...
+        num_controls: SymbolicInt = 0,
+    ) -> 'QROM':
+        data_shape = (
+            (data_len_or_shape,) if isinstance(data_len_or_shape, int) else data_len_or_shape
         )
-        assert all(
-            t >= int(np.max(d)).bit_length() for t, d in zip(self.target_bitsizes, self.data)
+        if not isinstance(target_bitsizes, tuple):
+            target_bitsizes = (target_bitsizes,)
+        _data = Shaped((len(target_bitsizes),) + data_shape)
```
fdmalone, line 142: why is len(target_bitsizes,) prepended to the shape? I was a bit confused by the notebook when it said the bloq was:
QROM(data_or_shape=Shaped(shape=(2, N, M)), selection_bitsizes=(ceiling(log2(N - 1)), ceiling(log2(M - 1))), target_bitsizes=(b1, b2), num_controls=c). Why is the shape (2, N, M)?

fdmalone (approved):
Very nice. I had a question about the shape in symbolic qrom.

fdmalone:
The bitsizes are a bit unfortunate here:
Screenshot 2024-05-11 at 5 22 50 AM

Do we have an issue to convert symbolics to math? (Is it possible for show_bloqs)

fdmalone:
Maybe my jupyter is whacky but the MSD also looks off:
Screenshot 2024-05-11 at 5 39 43 AM

fdmalone:
nvm just need to make it big:


Screenshot 2024-05-11 at 5 42 23 AM

qualtran/bloqs/data_loading/qrom.py
tanujkhattar: We have the type annotation of `ArrayLike` (so mypy will complain if you pass in an object of `Shaped`) and we pass it to `np.array(d, dtype=int)` which will raise an error if `d` cannot be converted to an array of integers. So I think this is fine as is.

qualtran/bloqs/data_loading/qrom.py
tanujkhattar: I've improved the docstrings to clarify what's going on. Tl;Dr is that `data_or_shape` is now an ndarray (or shaped object) of shape `(L, S_1, S_2, ..., S_K)` where `L = number of datasets to load` and `(S_1, S_2, ..., S_K)` is the shape of each dataset to be loaded.

tanujkhattar:
> Do we have an issue to convert symbolics to math? (Is it possible for show_bloqs)

I'm not sure, but +1 that it'll be a useful feature  - Maybe @mpharrigan  knows more.

mpharrigan:
what do you mean convert it to math?

tanujkhattar:
Render a latex expression $\lceil x \rceil$ instead of 'ceiling(x)'

mpharrigan:
The graphviz dot/svg exporter can't render latex math. There exists https://ctan.org/pkg/dot2tex?lang=en which would -- in theory -- render latex math from a graphviz format

The matplotlib circuit drawer can render latex math, because matplotlib can render latex math. You have to enclose it in `$`. 

To get latex math from a sympy expression, you use the `latex()` printing function https://docs.sympy.org/latest/tutorials/intro-tutorial/printing.html

Issue #947, opened 2024-05-13
Importing qualtran is very slow (~3.2s)


anurudhp:
```py
import timeit

t = timeit.timeit('import qualtran', number=1)
print(t)
```
averages around `3.2s`. Perhaps some dependency is causing this.

anurudhp:
I ran `python -X importtime -c "import qualtran"` (after #949), it seems `import cirq` takes up most of the time:

```
import time: self [us] | cumulative | imported package
...
import time:      1377 |    1072314 |     cirq
import time:      4683 |    1076997 |   qualtran._infra.gate_with_registers
...
import time:       349 |    1564730 | qualtran
```

mpharrigan:
It would be nice if we could defer importing cirq "as needed" as well. Ideally, this would only happen if you were using a bloq that wraps a cirq gate or uses cirq functionality somehow (or if using any of the cirq interop). It's blocked by the fact that `GateWithRegisters` is in the top-level namespace and descends from `cirq.Gate`.

PR #949, opened 2024-05-13
fix slow import: only import `quimb.tensor` when needed

anurudhp:
partially addresses #947: reduces `import qualtran` time from ~3.2s to ~1.4s


anurudhp:
@mpharrigan ptal

mpharrigan:
is it accurate to say that the change is only strictly required in `controlled.py`, since all the other locations aren't actually pulled in when you `import qualtran`.?

anurudhp:
> is it accurate to say that the change is only strictly required in `controlled.py`, since all the other locations aren't actually pulled in when you `import qualtran`.?

Yep. The remaining changes speed up `import qualtran.bloqs.basic_gates` by ~2s.

mpharrigan (approved):

Issue #818, opened 2024-03-22
Greedy vs Simple Qubit Ancilla Manager decompose bug with prepareuniformsuperposition


akurlej:
Greedy qubit manager fails with a duplicate QID on a CNOT in the ```LessThanConstant``` , SimpleQubitManager works without issue. 

Issue occurs with current release of qualtran (0.2.0)

Code to reproduce:
```
import cirq
from qualtran.bloqs.prepare_uniform_superposition import PrepareUniformSuperposition

n=3
target = cirq.NamedQubit.range((n-1).bit_length(),prefix='target')
control = [cirq.NamedQubit('control')]
circuit_error = cirq.inverse(cirq.Circuit(PrepareUniformSuperposition(n,cvs=(0,)).on_registers(ctrl=control,target=target)))

#Greedy fails, simple works
gqm = cirq.GreedyQubitManager(prefix="_ancilla", maximize_reuse=True)
#gqm = cirq.SimpleQubitManager(prefix="_ancilla")
cirq.decompose(circuit_error,context = cirq.DecompositionContext(gqm))
```

rroodll:
@mpharrigan is there any update on this one? Thanks
btw - sent an email to your outlook addr when you get a chance

tanujkhattar:
@akurlej @rroodll This is now fixed. Please let us know if you still run into any issues. Thanks for reporting and apologies it took so long to fix!

rroodll:
Thanks @tanujkhattar! This is great. One of us will pull the latest and give it a test.

PR #963, opened 2024-05-17
Bugfixes in cirq interop

tanujkhattar:
A stab at fixing https://github.com/quantumlib/Qualtran/issues/818

This is still a draft. More investigations are needed to figure out exactly what's causing the bug. Based on my investigations so far, during the interop, there can be multiple issues:

- In `_cirq_style_decompose_from_decompose_bloq`, we call `cbloq.to_cirq_circuit_and_quregs` and then explicitly call `qm.qfree` on the output qubits. Most likely, this is the place which is causing the bug. It's because for fancier qubit managers like `GreedyQubitManager`, us explicitly calling the `qm.qfree` messes up the state of the qubit manager and doesn't play well with the cirq <-> bloq <-> cirq <-> bloq etc. interop. For example:
image

Other potential issues (may or may not be the case):
- `decompose_from_cirq_style_method` does not accept a `context` and always uses a `SimpleQubitManager` irrespective of what the user has supplied. This may or may not be fine?
- `Alloc` and `Free` bloqs should define a `as_cirq_op` instead of getting wrapped in a `BloqAsCirqGate`.

### DIFF
--- a/qualtran/_infra/adjoint.py
+++ b/qualtran/_infra/adjoint.py
@@ -142,12 +142,15 @@ class Adjoint(GateWithRegisters):
         """The decomposition is the adjoint of `subbloq`'s decomposition."""
         return self.subbloq.decompose_bloq().adjoint()
 
-    def decompose_from_registers(
-        self, *, context: cirq.DecompositionContext, **quregs: NDArray[cirq.Qid]  # type: ignore[type-var]
-    ) -> cirq.OP_TREE:
-        if isinstance(self.subbloq, GateWithRegisters):
-            return cirq.inverse(self.subbloq.decompose_from_registers(context=context, **quregs))
-        return super().decompose_from_registers(context=context, **quregs)
+    # def decompose_from_registers(
+    #     self, *, context: cirq.DecompositionContext, **quregs: NDArray[cirq.Qid]  # type: ignore[type-var]
+    # ) -> cirq.OP_TREE:
+    #     if isinstance(self.subbloq, GateWithRegisters) or hasattr(
+    #         self.subbloq, 'decompose_from_registers'
+    #     ):
+    #         yield cirq.inverse(self.subbloq.decompose_from_registers(context=context, **quregs))
+    #     else:
+    #         yield super().decompose_from_registers(context=context, **quregs)
 
     def _circuit_diagram_info_(
         self, args: 'cirq.CircuitDiagramInfoArgs'
--- a/qualtran/bloqs/mcmt/and_bloq.py
+++ b/qualtran/bloqs/mcmt/and_bloq.py
@@ -320,6 +320,7 @@ class MultiAnd(Bloq):
     def decompose_from_registers(
         self, *, context: cirq.DecompositionContext, **quregs: NDArray[cirq.Qid]
     ) -> Iterator[cirq.OP_TREE]:
+        assert 'junk' in quregs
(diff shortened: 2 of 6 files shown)


mpharrigan:
Did you see the test failure in #961 ? is there a chance this could be related? Changing SwapWithZero to go through `decompose_from_cirq_style_method` actually causes incorrect behavior of `SelectSwapQROM` under simulation. it looks qubit-allocate-y

mpharrigan:
what's the status of this?

tanujkhattar:
I didn't have cycles to work on this

anurudhp:
On deeper inspection, it looks like `GreedyQubitManager` is indeed propagated correctly, but `_cbloq_to_cirq_circuit` seems to destroy ordering:
https://github.com/quantumlib/Qualtran/blob/cc4e3bcf1729cb31dcaa22591610ebb552e82d2b/qualtran/cirq_interop/_bloq_to_cirq.py#L270
In the example with two alloc-cnot-free, the topological ordering first processes the two allocs, then cnot, free, cnot, free. So the greedy qubit manager is unable to reuse the qubits. As `bloq.decompose_bloq` does not preserve/have any moment structure, I doubt this is avoidable. 

One way around it would be to use intermediate bloqs to decompose to, for example:

```py
import cirq
from attrs import frozen
from qualtran import Bloq, BloqBuilder, GateWithRegisters, Signature, SoquetT
from qualtran.bloqs.basic_gates import CNOT
from qualtran.cirq_interop.testing import GateHelper


@frozen
class MultiAlloc(Bloq):
    rounds: int = 2

    @property
        return Signature.build(q=1)

    def build_composite_bloq(
        self, bb: "BloqBuilder", q: "SoquetT"
    ) -> dict[str, "SoquetT"]:
        for _ in range(self.rounds):
            a = bb.allocate(1)
            a, q = bb.add(CNOT(), ctrl=a, target=q)
            bb.free(a)

        return {"q": q}

@frozen
class MultiAllocLayered(Bloq):
    rounds: int = 2

    @property
        return Signature.build(q=1)

    def build_composite_bloq(
            self, bb: "BloqBuilder", q: "SoquetT"
    ) -> dict[str, "SoquetT"]:
        for _ in range(self.rounds):
            q = bb.add(MultiAlloc(1), q=q)

        return {"q": q}


def visualize(bloq: Bloq | GateWithRegisters):
    print(bloq)
    gate = GateHelper(
        gate=bloq,
        context=cirq.DecompositionContext(
            cirq.GreedyQubitManager(prefix="_greedy", maximize_reuse=True)
        ),
    )
    print(gate.circuit)
    print()
    print(gate.decomposed_circuit)
    print()


if __name__ == "__main__":
    visualize(MultiAlloc())
    visualize(MultiAllocLayered())
```

```
MultiAlloc
q: ───q───

_greedy_0: ───alloc────@───free─────────────────────────
                       │
_greedy_1: ───alloc────┼────────────────────@───free────
                       │                    │
q: ───────────Y^-0.5───@───Y^0.5───Y^-0.5───@───Y^0.5───

MultiAllocLayered
q: ───q───

_greedy_0: ───alloc────@───free────alloc────@───free────
                       │                    │
q: ───────────Y^-0.5───@───Y^0.5───Y^-0.5───@───Y^0.5───
```

tanujkhattar:
This is now merged in a separate PR. Closing the draft now.

PR #967, opened 2024-05-20
Refactor block encoding bloqs to use abstract base class.

fdmalone:
- [x] Refactor block encoding bloqs to use an abstract base class to allow for using different constructions of block encodings other than LCUs. ~I also simplified the construction of the higher level block encoding to use attrs to automatically convert a SelectOracle and PrepareOracle into their "black boxes" to enable higher level abstractions.~
- [ ] ~Make SF and DF inherit from BlockEncoding and define the appropriate signal state reflections.~
- [ ] ~Update QubitizationWalkOperator to expect a block encoding rather than Sel and Prep.~

cc @tanujkhattar: does this sound sensible? If so I'll continue down this path. The goal is to enable phase estimation of all chemistry bloqs, not just those which follow the more standard LCU form. It is possible that I could further refactor SF and DF to build the sub block encodings but I'll leave that as a follow up.


qualtran/bloqs/block_encoding.py
```diff
@@ -182,13 +186,46 @@ def pretty_name(self) -> str:
         return 'Prep'
 
 
+class BlockEncoding(Bloq):
+    r"""Abstract interface for an arbitrary block encoding.
+
+    Users must provide methods for defining the ancilla, junk, and system registers,
+    and a method for defining the signal state $|G\rangle$. See module level
+    docstrings for more discussion.
```
tanujkhattar, line 194: Can we add more docstrings that explain the terms like "a", "signa" etc.? It woul probably mean copying some of the definitions from the linked reference but I think its worth doing

qualtran/bloqs/block_encoding.py
```diff
@@ -182,13 +186,46 @@ def pretty_name(self) -> str:
...
+        """The junk bitsize (additional ancilla registers not reflected on)."""
+
+    @property
+    @abc.abstractmethod
+    def system_bitsize(self) -> int:
+        """The system bitsize for applying the controlled (signal) unitary to."""
+
+    @property
+    @abc.abstractmethod
+    def reflection_using_signal_state(self):
```
tanujkhattar, line 218: I think this should just return a Bloq that prepares the initial state instead of providing a reflection around it? The "reflection around signal state"  is a feature of qubitization that doesn't necessarily need to be part of the block encoding base class.

fdmalone: It's called signal state in the qubitization reference, but I'm not wedded to the name. Do you know if there's something better? Yeah, I originally just had signal_state, but it seemed more convnenient for qubitization but you're right, it's something we can push to the walk operator to just get $|G\rangle$ from B to prepare the reflection.

qualtran/bloqs/block_encoding.py
tanujkhattar: I think this is a better reference for us to base the block encoding base class upon - https://arxiv.org/pdf/1804.01973

The block encoding base class should expose - (a) the state (or the registers and a bloq to prepare the state) `|G>` (b) the normalization constant `alpha` (in LCU, this will be the sum of coefficients) and (c) An error parameter `epsilon` -- these are parameters (usually denoted as `(a, alpha, epsilon)` are often used to define a block encoding). 

The other interesting case of assymetric block encodings where we have a different sized input (a) / output (b) registers can be described by our base class by just returning (one or more) registers which are not all necessarily THRU registers (but it'd be nice to verify that this is the case once we get to it). 

I don't think we need "junk" registers etc. ?  The registers to reflect upon would just correspond to the registers corresponding to `|G>` state and for block encodings which are not directly an LCU (eg SF / DF) the |G> state would be |0> and the registers would correspond to the "reflection registers" I believe. 

Does this make sense?

fdmalone: epsilon is the precision for state preparation in this case?

fdmalone: The block encoding still needs a signature though, should it just be ancilla and system?

fdmalone: or selection and system.

fdmalone: @tanujkhattar I took another stab at it

qualtran/bloqs/block_encoding.py
```diff
@@ -270,11 +362,11 @@ class ChebyshevPolynomial(Bloq):
...
 
     References:
-        [Quantum computing enhanced computational catalysis](
-            https://arxiv.org/abs/2007.14460). Page 45; Theorem 1.
+        [Quantum computing enhanced computational catalysis](https://arxiv.org/abs/2007.14460).
+            Page 45; Theorem 1.
     """
 
-    block_encoding: BlackBoxBlockEncoding
+    block_encoding: BlockEncoding
```
tanujkhattar, line 369: This currently assumes that `G` in `BlockEncoding` is identity but there's no check anywhere. It's worth highlighting in the docstring at least that it's the case. 

Also, I think the standard way to construct chebyshev polynomial is to take the nth power of a (qubitization) walk operator.   

For example - 

1. Lemma 16 from https://arxiv.org/pdf/1511.02306
image

2. Also from https://quantum-journal.org/papers/q-2019-07-12-163/pdf/

image

 
 I belive Guang Hao uses this construction of alternating between B and B^{\daggar} to save one reflection instead of directly using W^{n} and optimize constant factors? 

Either way, it'll be nice to track this in an issue so we can extend the bloq to support arbitrary block encodings in a follow-up PR.

tanujkhattar (commented):
This is looking good! Do you want to merge this and continue the refactoring or continue the refactoring in this PR?  Either way sounds good to me. I'll approve if you want to merge this once you mark it ready.

qualtran/bloqs/block_encoding.py
```diff
@@ -199,18 +264,41 @@ class BlackBoxBlockEncoding(Bloq):
...
+    selection registers (i.e.  the $l$ registers above) and junk registers which
+    are extra qubits needed by state preparation but not controlled upon during
+    SELECT.
+
+    Here $|G\rangle = \mathrm{PREPARE}|0\rangle$.
+
     Args:
         select: The bloq implementing the `SelectOracle` interface.
-        prepare: The bloq implementing the `SelectOracle` interface.
+        prepare: The bloq implementing the `PrepareOracle` interface.
```
tanujkhattar, line 278: nit: Update args

qualtran/bloqs/block_encoding.py
```diff
@@ -199,18 +264,41 @@ class BlackBoxBlockEncoding(Bloq):
...
+            block encoding is called a standard form encoding there.
+
+        [The power of block-encoded matrix powers: improved regression techniques via faster Hamiltonian simulation](https://arxiv.org/abs/1804.01973)
+            Definition 3 page 8.
     """
 
-    select: BlackBoxSelect
-    prepare: BlackBoxPrepare
+    alpha: SymbolicFloat
+    epsilon: SymbolicFloat
```
tanujkhattar, line 295: The `epsilon` in the LCU Block Encoding can typically be derived from the `epsilon` parameter used for state preparation in the `prepare` oracle. But since `prepare` currently doesn't expose an `epsilon` parameter; we can create an issue to track this and handle it later.

qualtran/bloqs/block_encoding.py
```diff
@@ -235,7 +323,11 @@ def signature(self) -> Signature:
         )
 
     def pretty_name(self) -> str:
-        return 'B[V]'
+        return 'U[H]'
```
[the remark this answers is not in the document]
fdmalone: I did too, but then was sort of mixing notation, I'll revert back

qualtran/bloqs/block_encoding.py
```diff
@@ -182,13 +194,66 @@ def pretty_name(self) -> str:
...
+
+    A $(\alpha, a, \epsilon) block encoding of an s-qubit operator $H$ if it obeys:
+
+    $$
+        \equiv \lVert H - \alpha (\langle G|_a\otimes I_s U |G\rangle_a \otimes I_s) \rVert
+        \le \epsilon,
+    $$
+
+    where $a$ is an ancilla register and $s$ is the system register, $U$ is a unitary sometimes
+    called a signal oracle and encodes $H$ in its top right corner, while $|G\rangle_a$ is
```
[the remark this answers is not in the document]
fdmalone: Done.

qualtran/bloqs/block_encoding.py
```diff
@@ -182,13 +194,66 @@ def pretty_name(self) -> str:
...
+
+    @property
+    @abc.abstractmethod
+    def junk_bitsize(self) -> int:
+        """An additional junk register typically used for Prepare which is not reflected on."""
+
+    @property
+    @abc.abstractmethod
+    def system_bitsize(self) -> int:
+        """The system bitsize for applying the controlled (signal) unitary to."""
```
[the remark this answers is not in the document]
fdmalone: Fixed. I was being a bit loose here with assuming SELECT would be applied to the system register (controlled off of the ancilla).

qualtran/bloqs/block_encoding.py
```diff
@@ -182,13 +194,66 @@ def pretty_name(self) -> str:
...
+
+    @property
+    @abc.abstractmethod
+    def selection_bitsize(self) -> int:
+        """The bitsize for the register `a` registers above"""
+
+    @property
+    @abc.abstractmethod
+    def junk_bitsize(self) -> int:
+        """An additional junk register typically used for Prepare which is not reflected on."""
```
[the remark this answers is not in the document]
fdmalone: Yeah it helps you combine many different registers (the selection registers for Prepare might have 10 different wires) into a single register of the same total bitsize. It's helpful for writing higher level algorithms which don't care about the details of these 10 registers. Like the chebyshev polynomial or the block encoding which would just be defined on ancilla (+junk) and system registers, decomposing further (zooming in) would reveal the details of the big registers.

I can do that sure.

qualtran/bloqs/block_encoding.py
fdmalone: Right, I would need to add a Identity PrepareOracle which I think I'll do in a follow up. I'll add a note / issue.

qualtran/bloqs/block_encoding.py
fdmalone: It's potentially a combination of epsilons right, if there are multi-stage state preparations. Also I imagine rotation synthesis is also packed into this epsilon?

qualtran/bloqs/block_encoding.py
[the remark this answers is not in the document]
fdmalone: I think we want to define the block encoding to be high level so that the signature should be 3 registers for system, junk and selection rather than a list of registers and use the partition block as is.

qualtran/bloqs/block_encoding.py
fdmalone: Added #984

qualtran/bloqs/block_encoding.py
fdmalone: Added #985

qualtran/bloqs/block_encoding.py
fdmalone: Wait, it doesn't assume anything about $G$ in the block encoding, the reference provides an implementation which just has B (2|0><0|-1) B^, i.e. always reflection about zero no?

tanujkhattar: it assumes that H is block encoded in the top-left corner; i.e `G` is identity ?

image

fdmalone: You're right here's the earlier definition:

fdmalone:
Ok, I took way too long going back and forth but the block should work in the verbose and "black box" way (i.e hiding complexity with a partition). I think this will enable us to write diagrams similar to those in say the linear T paper:

Screenshot 2024-05-22 at 3 06 39 PM

fdmalone:
I will follow up with the updates to QubitizationWalkOperator / chemistry block encodings.

fdmalone:
@tanujkhattar PTAL, this is good to go I think

qualtran/bloqs/block_encoding.py
```diff
@@ -182,69 +191,247 @@ def pretty_name(self) -> str:
...
+    Users must specify:
+    1. the normalization constant $\alpha \ge \lVert A \rVert$, where
+        $\lVert \cdot \rVert$ denotes the spectral norm.
+    2. the precision to which the block encoding is to be prepared ($\epsilon$).
+
+    Developers must provide a method to return a bloq to prepare $|G\rangle$.
+
+    References:
+        [Hamiltonian Simulation by Qubitization](https://quantum-journal.org/papers/q-2019-07-12-163/)
+            Sec 2 and 3 for introduction and definition of terms.
```
mpharrigan, line 232: can you add first author et. al. and the year to the references?

qualtran/bloqs/block_encoding.py
```diff
@@ -182,69 +191,247 @@ def pretty_name(self) -> str:
...
+    In general, given an $s$-qubit operator $H$ then the $(s+a)$-qubit unitary $U$ is
+    a $(\alpha, a, \epsilon)$-block encoding of $H$ if it satisfies:
+
+    $$
+        \lVert H - \alpha (\langle G|_a\otimes I_s U |G\rangle_a \otimes I_s) \rVert
+        \le \epsilon,
+    $$
+
+    where $a$ is an ancilla register and $s$ is the system register, $U$ is a unitary sometimes
+    called a signal oracle, $\alpha \ge \lVert H\rVert$ (where $\lVert \cdot \rVert$
```
mpharrigan, line 206: nit: say what alpha is rather than just its magnitude. e.g. below you call it "normalization constant alpha"

qualtran/bloqs/block_encoding.py
```diff
@@ -182,69 +191,247 @@ def pretty_name(self) -> str:
...
 @attrs.frozen
-class BlackBoxBlockEncoding(Bloq):
-    r"""Black box implementation of a block encoding using SELECT and PREPARE.
+class LCUBlockEncoding(BlockEncoding):
+    r"""LCU based block encoding using SELECT and PREPARE oracles.
 
     Builds the block encoding via
     $$
-        \mathcal{B}[V] = \mathrm{PREPARE}^\dagger \cdot \mathrm{SELECT} \cdot \mathrm{PREPARE},
+        B[H] = \mathrm{SELECT}
```
mpharrigan, line 269: you don't really introduce the `B[H]` notation in the `BlockEncoding` docstring

mpharrigan: should probably use the words "linear combination of unitaries" somewhere in the docstring

qualtran/bloqs/block_encoding.py
```diff
@@ -336,39 +536,140 @@ def _black_box_select() -> BlackBoxSelect:
...
+    # 3x3 hubbard model U/t = 4
+    dim = 3
+    select = SelectHubbard(x_dim=dim, y_dim=dim)
+    U = 4
+    t = 1
+    prepare = PrepareHubbard(x_dim=dim, y_dim=dim, t=t, u=U)
+    N = dim * dim * 2
+    qlambda = 2 * N * t + (N * U) // 2
+    lcu_block_bloq = LCUBlockEncoding(select=select, prepare=prepare, alpha=qlambda, epsilon=0.0)
+    return lcu_block_bloq
```
mpharrigan, line 551: lol can we ommit `_bloq` suffix here and below? black box block bloq is admittedly very funny but maybe not in a helpful way :)

mpharrigan (approved):
It's admirable that the user can either use the raw select and prepare subroutines and get all the registers, but do we actually want this? I.e. why can't we always go via the "black-box" interfaces

qualtran/bloqs/block_encoding.py
fdmalone: Can you say it 5 times quickly? I knew I pushed it too far this time, I will retire black_box_block_bloq, a career highlight.

fdmalone:
> It's admirable that the user can either use the raw select and prepare subroutines and get all the registers, but do we actually want this? I.e. why can't we always go via the "black-box" interfaces

Yeah, originally I had an attrs converter which would just convert it to the abstract black box, which is *probably* the right thing to do, but for the chemistry bloqs I'll need yet another black box for the big block encodings, so thought it better to do things incrementally. (and maybe add the AutoPartition first to handle this). All these wrappers would have to be structured etc.. so all this is to say is yes, I think the end point would be use the black boxes, but maybe wait a few PRs.

I also didn't want to open a can of worms here about how we want to write these algorithms, but I think now (or shortly) we can generate the more abstract linear T style diagrams of like a walk operator and QPE etc.

mpharrigan:
Thanks for the additional color! Sounds good

qualtran/bloqs/block_encoding.py
fdmalone: Done.

qualtran/bloqs/block_encoding.py
fdmalone: Done.

qualtran/bloqs/block_encoding.py
fdmalone: Done.

qualtran/bloqs/block_encoding.py
fdmalone: RIP

qualtran/bloqs/block_encoding.py
matthagan15: Quick question: since the first kind chebyshev polynomials are just QSP/Qubitization procedures with all the phase angles set to 0 why not just reuse that? I would think that would make a nicer decomposition.

fdmalone: Probably didn't exist in the library at the time. Can you open an issue or add a comment to #984

Issue #957, opened 2024-05-15
General costs framework may not scale for costs (like `depth`) that depend upon the structure of decomposition. 


tanujkhattar:
See https://github.com/quantumlib/Qualtran/pull/913#discussion_r1594825762 for some relevant discussion on the potential shortcomings of the current method and a proposed alternative. 

We will encounter this once we add more cost keys, especially for costs where the computation of the cost would depend upon the structure of the composite bloq obtained via  `bloq.decompose_bloq()` and not just on the counts of unique callees that can be obtained via `bloq.call_graph()`. 

An example of such a cost is circuit `depth`. We can close this issue once we add such a cost to Qualtran and show that it scales for large instances of bloqs people often care about (eg: 5000 bit Adder which currently takes ~2 min to decompose into a composite bloq)

PR #969, opened 2024-05-20
[Costs] Qubit Counts

mpharrigan:
This introduces a `CostKey` to count qubits. Part of #899. 

image


qualtran/resource_counting/__init__.py
```diff
@@ -31,4 +31,6 @@
 
 from ._costing import GeneralizerT, get_cost_value, get_cost_cache, query_costs, CostKey, CostValT
 
+from ._qubit_counts import QubitCount
```
fdmalone, line 34: why is it a private module?

fdmalone (approved):
LGTM. I'll need to play around with it some more to check the chemistry algorithms.

qualtran/resource_counting/_qubit_counts.py
```diff
@@ -0,0 +1,132 @@
...
+            return min_bloq_size
+
+        # Compute the number of qubits ("width") from the bloq's decomposition. We forward
+        # the `get_callee_cost` function so this can recurse into subbloqs.
+        try:
+            cbloq = bloq.decompose_bloq()
+            logger.info("Computing %s for %s from its decomposition", self, bloq)
+            return _cbloq_max_width(cbloq._binst_graph, get_callee_cost)
+        except (DecomposeNotImplementedError, DecomposeTypeError):
+            pass
```
tanujkhattar, line 117: This is a good example of a cost which doesn't scale for large bloqs, as I had raised a concern for in your previous PR (cc https://github.com/quantumlib/Qualtran/issues/957)

Do we have a plan to make this scalable? Do we plan to do `isinstance` based dispatch for special bloqs as you had proposed in https://github.com/quantumlib/Qualtran/pull/913#discussion_r1601989838 ? 

image

tanujkhattar (changes requested):
Question about scalability, do we have any plans to address it? Is it worth discussing the design choices again and whether `my_static_costs` should be really only overridden for the leaf bloqs?

qualtran/resource_counting/_qubit_counts.py
tanujkhattar: Here is another example for the hubbard model - 

image

mpharrigan: for something like Add or QROM (which is the slow part in hubbard model) you can annotate the static qubit counts statically

tanujkhattar: I agree. And I think the right way to annotate these static costs is to express the cost as a formula of the cost of its callees. For example - the `SelectSwapQROM` can express its qubit counts in terms of `qubit_count(QROM)` and `qubit_count(SwapWithZero)` bloqs. Similarly, phase estimation can express its cost in terms of `qubit_count(U)`; instead of specifying a constant (which is not even possible for phase estimation since `U` is unknown and decomposing phase estimation can be costly in general, since you end up with a circuit where the depth is at least `2 ** size(phase_register)`)


For this to scale, it would be really nice if `my_static_costs` has access to the cache and can forward it to the `get_cost_cache` method to get the cost of its callees. Since we are designing the API right now, it'll be easier to make this change and be future proof so I'd suggest you to reconsider this proposal.

qualtran/resource_counting/__init__.py
mpharrigan: This is to force imports of the form
```
from qualtran.resource_counting import QubitCount
```

instead of

```
from qualtran.resource_counting.qubit_counts import QubitCount
```

fdmalone: the underscore isn't necessary though right? You could still do `from .qubit_counts import QubitCount` no? Or are you saying you don't want me to import from qubit_counts ever?

mpharrigan: the latter

PR #970, opened 2024-05-20
[Costs] Success probability

mpharrigan:
Add a new `CostKey` for multiplying success probabilities.  Part of #899


qualtran/bloqs/for_testing/costing.py
```diff
@@ -24,14 +24,23 @@ def _convert_callees(callees: Sequence[BloqCountT]) -> Tuple[BloqCountT, ...]:
...
 @frozen
 class CostingBloq(Bloq):
     """A bloq that lets you set the costs via attributes."""
 
     name: str
     num_qubits: int
     callees: Sequence[BloqCountT] = field(converter=_convert_callees, factory=tuple)
-    static_costs: Sequence[Tuple[CostKey, Any]] = field(converter=tuple, factory=tuple)
+    static_costs: Sequence[Tuple[CostKey, Any]] = field(
+        converter=_convert_static_costs, factory=tuple
```
fdmalone, line 42: I feel like we have at least 2 of these _tuple attrs converters floating throughout the library. Should probably consolidate at some point.

qualtran/resource_counting/_success_prob.py
```diff
@@ -0,0 +1,49 @@
...
+
+from qualtran import Bloq
+
+from ._call_graph import get_bloq_callee_counts
+from ._costing import CostKey
+
+logger = logging.getLogger(__name__)
+
+
+@frozen
```
fdmalone, line 27: Out of scope for this PR but it would be nice to have some notebook expanding on this as a concept with some examples of algorithms which have this.

qualtran/resource_counting/_success_prob.py
```diff
@@ -0,0 +1,49 @@
...
+    `my_static_costs` to provide their actual success probability.
+    """
+
+    def compute(self, bloq: 'Bloq', get_callee_cost: Callable[['Bloq'], float]) -> float:
+        tot: float = 1.0
+        callees = get_bloq_callee_counts(bloq)
+        logger.info("Computing %s for %s from %d callee(s)", self, bloq, len(callees))
+        for callee, n in callees:
+            v = get_callee_cost(callee)
+            tot *= v**n
```
fdmalone, line 42: Would we eventually need to deal with like the log of the success probability so things are additive. This looks like it runs the risk of potentially hitting numerical issues.

fdmalone (approved):
LGTM. I think multiplying lots of numbers together is potentially risky, but maybe it's not really a concern.

_why_is_everything_so_private_lately?

qualtran/resource_counting/_success_prob.py
tanujkhattar, line 42: The way this is written, would we ever need this as a separate cost? Can we just use `g, sigma = bloq.build_call_graph(pred=pred)` and then accumulate the error probability using bloq, counts in `sigma` ? 

Is there a good example that motivates this cost key that already exists in another branch / is easy to describe ?

fdmalone: I realized later that after writing this that if the success probability was so small to warrant something special it would be a dreadful quantum algorithm.

qualtran/bloqs/for_testing/costing.py
mpharrigan: it's for the type checker. In theory we could re-use converters where the types are identical, but it's way more refactorable to keep the slight amount of duplication.

The type checker isn't smart enough if you use generic types in a universal to-tupler

qualtran/resource_counting/_success_prob.py
mpharrigan: you need the contributions from inner nodes as well, not just the leafs that will show up in `sigma`. 

This was requested by someone at the IEEE meeting last year and is a good demo of a different type of cost

qualtran/resource_counting/_success_prob.py
```diff
@@ -0,0 +1,49 @@
...
+from typing import Callable
+
+from attrs import frozen
+
+from qualtran import Bloq
+
+from ._call_graph import get_bloq_callee_counts
+from ._costing import CostKey
+
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
mpharrigan: yes, I'm using logging in the cost framework. It can be very helpful to trace what's going on. I have a branch that adds some logging to the quimb simulation as well. By default, it won't print anything. you have to set up a logger e.g. with `logging.basicConfig(level=logging.INFO)`

Issue #607, opened 2024-01-27
Support overflows in classical simulation


fpapa250:
For now we can assert that the bitsize is less than the classical data type, but we should add in some functionality that protects against overflows behind the scenes.

i.e. `Add()`

```
def on_classical_vals(
        self, a: 'ClassicalValT', b: 'ClassicalValT'
    ) -> Dict[str, 'ClassicalValT']:
        assert self.bitsize <= 64  # TODO: be smarter
        # TODO: account for signed integer addition
        # https://github.com/quantumlib/Qualtran/issues/606
        return {
            'a': a,
            'b': np.uint64(a) + np.uint64(b),
        }
```

mpharrigan:
I think the appropriate fix here is to return b = (a + b) % bitsize.  Since if you were to execute this on a quantum computer, that's what it would do. It wouldn't trip an error handler

mpharrigan:
Note that we'd need C-style modulo and _not_ python style modulo

PR #971, opened 2024-05-20
Handle overflow in addition plus k classical function

dstrain115:
- Handle bitwise overflow in addition plus k.
- All other arithmetic functions either handle overflow correctly or have enough bits in the output to avoid overflow.

Fixes: #607


mpharrigan (approved):
Thanks!

Issue #929, opened 2024-05-07
Hubbard model bloq docs


mpharrigan:
hubbard_model.ipynb which documents SelectHubbard and PrepareHubbard needs to be ported to the jupyter "autodoc" system for documenting bloqs.

PR #974, opened 2024-05-20
Hubbard model nits

fdmalone:
This was bothering me when mypy complained that $t$ or $u$ (not $\mu$ as $u$ was called as a variable) were floats (which they are in general). Also uses $u$ as in the paper which for some reason doesn't use the capital $U$ hubbard U.


fdmalone:
Ok I ended up fixing the hubbard model notebook too to use the autogen feature but appended the resource estimation to the end.

qualtran/serialization/resolver_dict.py
```diff
@@ -248,6 +247,7 @@
     "qualtran.bloqs.data_loading.qrom.QROM": qualtran.bloqs.data_loading.qrom.QROM,
     "qualtran.bloqs.data_loading.select_swap_qrom.SelectSwapQROM": qualtran.bloqs.data_loading.select_swap_qrom.SelectSwapQROM,
     "qualtran.bloqs.mod_arithmetic.CModAddK": qualtran.bloqs.mod_arithmetic.CModAddK,
+    "qualtran.bloqs.mod_arithmetic.mod_addition.ModAddK": qualtran.bloqs.mod_arithmetic.mod_addition.ModAddK,
```
tanujkhattar, line 250: Is this intended as part of the PR ?

tanujkhattar (approved):

qualtran/serialization/resolver_dict.py
fdmalone: PrepareHubbard wouldn't serialized without it, I think the module name was slightly off before.

fdmalone: In particular, ModAddK moved to mod_arithmetic from arithmetic.

mpharrigan:
fixes #929

PR #981, opened 2024-05-21
Show number of controls in string representation of controlled bloq

tanujkhattar:
This is useful to visualize and analyze bloq decomposition diagrams where the decomposition may have multiple controlled-bloqs of the same subbloq but with different number of controls 

For example, 

image

is much more helpful than 

image

### DIFF
--- a/qualtran/_infra/controlled.py
+++ b/qualtran/_infra/controlled.py
@@ -462,12 +462,12 @@ class Controlled(GateWithRegisters):
 
     def pretty_name(self) -> str:
         num_ctrls = self.ctrl_spec.num_qubits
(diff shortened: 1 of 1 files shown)


fdmalone:
Does it work with symbolics?

tanujkhattar:
It works but doesn't look great right now

image


I can change it to `C[num_controls][subbloq]` and that should look nicer for symbolic expression. WDYT?

tanujkhattar:
Changed, this is how it looks now 

image

fdmalone:
LGTM

fdmalone (approved):

mpharrigan (approved):

PR #996, opened 2024-05-28
Remove t_complexity from comparison bloqs

dstrain115:
- These bloqs already had call graphs.
- Also fix them to pick up the number of cliffords from the other bloqs.

### DIFF
--- a/qualtran/bloqs/arithmetic/comparison.py
+++ b/qualtran/bloqs/arithmetic/comparison.py
@@ -50,6 +50,7 @@ from qualtran._infra.quantum_graph import Soquet
 from qualtran.bloqs.basic_gates import CNOT, TGate, XGate
 from qualtran.bloqs.mcmt.and_bloq import And, MultiAnd
 from qualtran.bloqs.mcmt.multi_control_multi_target_pauli import MultiControlX
+from qualtran.bloqs.util_bloqs import ArbitraryClifford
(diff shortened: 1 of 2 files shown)


qualtran/bloqs/arithmetic/comparison.py
```diff
@@ -884,11 +870,8 @@ def wire_symbol(self, reg: Optional[Register], idx: Tuple[int, ...] = tuple()) -
         raise ValueError(f'Unknown register symbol {reg.name}')
 
     def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
-        # TODO Determine precise clifford count and/or ignore.
-        # See: https://github.com/quantumlib/Qualtran/issues/219
-        # See: https://github.com/quantumlib/Qualtran/issues/217
         tc = t_complexity(LessThanConstant(self.bitsize, less_than_val=self.val))
-        return {(TGate(), tc.t)}
+        return {(TGate(), tc.t), (ArbitraryClifford(n=1), tc.clifford)}
```
anurudhp, line 874: shouldn't this be just `return {(LessThanConstant(self.bitsize, less_than_val=self.val), 1)}`?

dstrain115: Oh, I didn't know you could do that, but it makes sense.  I'll change it!

tanujkhattar (approved):

PR #998, opened 2024-05-28
[WIP] Use block encoding bloqs for all chemistry hamiltonians

fdmalone:
I might need to split this but given #997 I thought it better to open now to avoid any conflicts:

1. Adds Identity bloq.
2. Makes SF and DF be block_encodings.
3. Replace Reflection (which was actually incorrect) with ReflectionUsingPrepare (needed to add a wrapper bloq to make a PrepareOracle from the Identity matrix (also necessary for block encodings $|G\rangle$ state))
4. [Separate PR] Update walk operator interface.

cc @tanujkhattar WDYT


tanujkhattar:
Seems like the right thing to do! Also +1 to break into smaller PRs if its not too much work. 

Do you want me to leave a detailed review now or are you still working on it? Also, I'm starting to work on a PR that organises `utils.py -> bookkeeping/` ; maybe its best to resolve merge conflicts after thats merged? (I'll aim to send the PR in the next 1 hour)

fdmalone:
You don't need to review now, just saw the organizational PR and didn't want a conflict with reflection. I didn't do anything special when moving the files though.

fdmalone:
Yeah I can wait to resolve conflicts which should be minor. Should we reorganize reflection first too?

tanujkhattar:
> Should we reorganize reflection first too?

Yes, although I'm not sure where to best put the "reflection" bloq -- my suggestion is to create a new `bloqs/reflections/` and then just move the `reflection.py` and `reflection_using_prepare.py` to `reflections/` -- Although eventually we'd only want `reflection_using_prepare.py`. 

My suggestion is purely based on the aesthetics that we have `bloqs/rotations/` and it'd be nice to have `bloqs/reflections/` :)

Reflection is used commonly in many applications; so it doesn't feel right it to put it under any specific one like `qubitization/`, `block_encoding/` or `amp_amp/` (which doesn't exist yet)

fdmalone:
I can do this. I don't need octopus right, that's only for splitting up files?

tanujkhattar:
> I can do this. I don't need octopus right, that's only for splitting up files?

Yes, that's right.

fdmalone:
This was completed in chunks

Issue #979, opened 2024-05-21
Organize remaining bloqs in bloqs


fdmalone:
- [ ] block_endoding, qubitization_walk_operator and reflection, select_and_prepare -> something to do with qubitization OR block _encoding should live by itself, reflection will probably be deprecated #966 
- [ ] on_each.ipynb -> basic_gates
- [ ] hubbard_model -> chemistry?
- [ ] util_bloqs -> utilities (@mpharrigan rename utils?)

mpharrigan:
utils should be "bookkeeping"

fdmalone:
chemistry should maybe be further organized as chemistry/qubitization and chemistry/trotter

PR #1007, opened 2024-05-29
Move `power` bloq to `basic_gates`

tanujkhattar:
Part of a larger effort to organize bloqs. Also fixes a drive-by bug in `execute_notebook` test of `reflections` (not sure why this is not being catched by the CI)

Fixes https://github.com/quantumlib/Qualtran/issues/979


fdmalone (approved):

mpharrigan:
the notebook unit tests aren't executed by the pytest job. the notebooks are executed in the notebook job which doesn't use pytest

Issue #966, opened 2024-05-20
Refactor / replace Reflection bloq


fdmalone:
The Reflection (#510) bloq is mostly a convenience bloq for reflection around zero to make nicer diagrams (the open / closed multi-bit circles which often appear). BUT we already have ReflectionUsingPrepare which could implement Reflection using a PrepareOracle which just prepares $|0\rangle_a$ ($P = I_a$ in this case).

The issue is to potentially remove the Reflection bloq and check everything works as expected + make sure the diagrams are still nice, and probably increase docs to make the common Reflect around zero more visible.

PR #1002, opened 2024-05-28
Add identity bloq

fdmalone:
It's helpful to have this when (for example) building a PrepareOracle for ReflectionUsingPrepare to achieve a reflection about zero ($P=I$).

Part of #998

Issue #978, opened 2024-05-21
Bloq modules missing autodocs


fdmalone:
# Basic gates
- [ ] [y_gate.py](https://github.com/quantumlib/Qualtran/tree/main/qualtran/bloqs/./basic_gates/y_gate.py)
- [ ] [on_each.py](https://github.com/quantumlib/Qualtran/tree/main/qualtran/bloqs/./basic_gates/on_each.py)
# Factoring
- [ ] [mod_sub.py](https://github.com/quantumlib/Qualtran/tree/main/qualtran/bloqs/./factoring/mod_sub.py)
- [ ] [_ecc_shims.py](https://github.com/quantumlib/Qualtran/tree/main/qualtran/bloqs/./factoring/ecc/_ecc_shims.py)
- [ ] [ec_point.py](https://github.com/quantumlib/Qualtran/tree/main/qualtran/bloqs/./factoring/ecc/ec_point.py)
- [x] [reflection_using_prepare.py](https://github.com/quantumlib/Qualtran/tree/main/qualtran/bloqs/./reflection_using_prepare.py)

# Mod arithmetic
- [ ] [_shims.py](https://github.com/quantumlib/Qualtran/tree/main/qualtran/bloqs/./mod_arithmetic/_shims.py)
# Multiplexers
- [ ] [unary_iteration_bloq.py](https://github.com/quantumlib/Qualtran/tree/main/qualtran/bloqs/./multiplexers/unary_iteration_bloq.py)
- [ ] [selected_majorana_fermion.py](https://github.com/quantumlib/Qualtran/tree/main/qualtran/bloqs/./multiplexers/selected_majorana_fermion.py)
# Arithmetic
- [ ] [_shims.py](https://github.com/quantumlib/Qualtran/tree/main/qualtran/bloqs/./arithmetic/_shims.py)
- [ ] [hamming_weight.py](https://github.com/quantumlib/Qualtran/tree/main/qualtran/bloqs/./arithmetic/hamming_weight.py)
# Chemistry
- [ ] [black_boxes.py](https://github.com/quantumlib/Qualtran/tree/main/qualtran/bloqs/./chemistry/black_boxes.py)
- [ ] [chem_tutorials.py](https://github.com/quantumlib/Qualtran/tree/main/qualtran/bloqs/./chemistry/chem_tutorials.py)
[...]

PR #1008, opened 2024-05-29
Deprecate reflection bloq in favour of ReflectionUsingPrepare

fdmalone:
Pulled from #998 and builds on #1002, removes the Reflection bloq (which was actually incorrect) in favour of ReflectionUsingPrepare. There's an open issue in that, some reflections in the chemistry bloqs are multiply controlled, but ReflectionUsingPrepare only allows a single control.

I added an intermediate identity prepare oracle as input for ReflectionUsingPrepare, which will also be needed for the $|G\rangle$ state (bloq) for block encodings.


fdmalone:
Fixes #966

fdmalone:
Related #978 (added ReflectionUsingPrepare auto doc.)

qualtran/bloqs/reflections/reflection_using_prepare.py
```diff
@@ -126,3 +153,30 @@ def _circuit_diagram_info_(self, args: cirq.CircuitDiagramInfoArgs) -> cirq.Circ
         wire_symbols = ['@' if self.control_val else '@(0)'] * total_bits(self.control_registers)
         wire_symbols += ['R_L'] * total_bits(self.selection_registers)
         return cirq.CircuitDiagramInfo(wire_symbols=wire_symbols)
+
+
+@bloq_example(generalizer=ignore_split_join)
```
tanujkhattar, line 158: Can you also add a `build_call_graph` so we support symbolics?

tanujkhattar (approved):

qualtran/bloqs/reflections/reflection_using_prepare.py
fdmalone: I'm also thinking we should update the wire_symbols to potentially special case to the nice open / closed circle diagram if the reflection is a reflect around zero, but that's not super high priority.

tanujkhattar:
`MultiControlPauli` supports `cvs` so it shouldn't be very hard to add support for `cvs` to `ReflectionUsingPrepare` since we only need to control the `MultiControlPauli` and the prepare /  prepare^{\dagger} simply passes through. Do you want to do this right as part of this PR itself?

fdmalone:
@tanujkhattar Is there a simple way to allow for multiple controls on the reflection (just updating control_val -> control_vals?) I wasn't sure of the significance of the control register which looks to be used for phasing if specified rather than just adding performing the control.

fdmalone:
Yeah I'll do it now.

tanujkhattar:
Basically, we could always just allocate a new qubit for a "phase target" and do `MultiControlPauli(controls=reflection_controls + selection_controls, target=phase_target)`. The optimization to save an allocation when the control qubit is specified is really just a convenience method.

fdmalone:
SGTM

qualtran/bloqs/chemistry/df/double_factorization.py
```diff
@@ -426,8 +428,14 @@ def build_composite_bloq(
             sys=sys,
         )
         # The last ctrl is the 'target' register for the MCP gate.
-        succ_l, l_ne_zero, p, spin = bb.add(
-            Reflection((1, 1, n_n, 1), (1, 1, 0, 0)), reg0=succ_l, reg1=l_ne_zero, reg2=p, reg3=spin
+        # TODO: missing control
```
mpharrigan, line 431: what does this todo mean? can you make it more specific

qualtran/bloqs/chemistry/resource_estimation.ipynb
mpharrigan, line 1: if it's not too difficult, can you revert the (non-)changes to this file

qualtran/bloqs/chemistry/df/double_factorization.py
fdmalone: See above comment from Tanuj, it will go away once I update ReflectionUsingPrepare control_values to accept more than 1 control

qualtran/bloqs/chemistry/sf/single_factorization_test.py
```diff
@@ -120,8 +120,9 @@ def test_compare_cost_to_openfermion():
         + QR2(num_aux + 1, nprime, bp)[-1]
         + QI2(num_aux + 1, nprime)[-1]
     )
+    delta_relf_missing_ctrl = 1
```
mpharrigan, line 123: relf?

fdmalone: relf_on_the_shelf will also disappear once the control issue is fixed.

mpharrigan: does it mean reflection? wouldn't that be refl?

fdmalone: it's a typo yes

fdmalone:
> `MultiControlPauli` supports `cvs` so it shouldn't be very hard to add support for `cvs` to `ReflectionUsingPrepare` since we only need to control the `MultiControlPauli` and the prepare / prepare^{\dagger} simply passes through. Do you want to do this right as part of this PR itself?

ehhh, this is getting complicated because ReflectionUsingPrepare is a SpecializedSingleQubitControlledGate. I wasn't following this too closely, is it much work to adapt this or should we merge and follow up.

tanujkhattar:
I think it's fine doing it as a follow up. We can merge this address the multi control in a follow up

qualtran/bloqs/chemistry/sf/single_factorization_test.py
fdmalone: Done.

qualtran/bloqs/chemistry/df/double_factorization.py
fdmalone: Linked to an issue: https://github.com/quantumlib/Qualtran/issues/1022

qualtran/bloqs/chemistry/resource_estimation.ipynb
fdmalone: Rebased and removed

PR #1009, opened 2024-05-29
Don't run pylint on deleted files.

fdmalone:
I noticed in #1008 that ./check/pylint-changed-files would choke because I deleted some files. This PR adds a flag to filter out deleted files.


pavoljuhas (approved):
LGTM.  Please also fix missing quotes reported by shellcheck
```
$ shellcheck -f gcc check/pylint-changed-files 
check/pylint-changed-files:49:24: note: Double quote to prevent globbing and word splitting. [SC2086]
check/pylint-changed-files:50:23: note: Double quote to prevent globbing and word splitting. [SC2086]
```

mpharrigan:
@pavoljuhas is this bug fixed in the cirq version of this script? can we make sure the two scripts don't diverge too much

fdmalone:
> LGTM. Please also fix missing quotes reported by shellcheck
> 
> ```
> $ shellcheck -f gcc check/pylint-changed-files 
> check/pylint-changed-files:49:24: note: Double quote to prevent globbing and word splitting. [SC2086]
> check/pylint-changed-files:50:23: note: Double quote to prevent globbing and word splitting. [SC2086]
> ```

Done. Also just learned shellcheck existed!

pavoljuhas:
> @pavoljuhas is this bug fixed in the cirq version of this script? can we make sure the two scripts don't diverge too much

Not yet.  This can wait after the cirq release.  (https://github.com/quantumlib/Cirq/issues/6619)

PR #1015, opened 2024-05-29
Remove `_t_complexity_` on `MultiAnd` and support symbolic `cvs`

tanujkhattar:
Part of a larger effort to better support symbolic resource estimation across bloqs. I will update the `cvs` parameter on other bloqs in follow-up PRs.

### DIFF
--- a/qualtran/bloqs/mcmt/and_bloq.py
+++ b/qualtran/bloqs/mcmt/and_bloq.py
@@ -22,7 +22,6 @@ The `Toffoli` bloq is similar to the `And` bloq. Toffoli will flip the target bi
 to the and of its control registers. `And` will output the result into a fresh register.
 """
 import itertools
-import numbers
 from functools import cached_property
 from typing import Any, Dict, Iterable, Iterator, Optional, Set, Tuple, TYPE_CHECKING, Union
 
@@ -59,7 +58,7 @@ from qualtran.resource_counting.generalizers import (
     ignore_cliffords,
 )
 from qualtran.simulation.classical_sim import ClassicalValT
-from qualtran.symbolics import is_symbolic, Length, SymbolicInt
+from qualtran.symbolics import HasLength, is_symbolic, SymbolicInt
 
 if TYPE_CHECKING:
     import quimb.tensor as qtn
@@ -239,13 +238,14 @@ _AND_DOC = BloqDocSpec(
 )
 
 
-def _to_tuple(
-    x: Union[SymbolicInt, Iterable[SymbolicInt]]
-) -> Union[Length, Tuple[SymbolicInt, ...]]:
-    if isinstance(x, sympy.Expr):
-        return Length(x)
-    if isinstance(x, (numbers.Integral, int)):
-        return (1,) * x
+def _to_tuple_or_has_length(
+    x: Union[HasLength, Iterable[SymbolicInt]]
+) -> Union[HasLength, Tuple[SymbolicInt, ...]]:
+    if isinstance(x, HasLength):
+        if is_symbolic(x.n):
+            return x
+        else:
+            return (1,) * x.n
     return tuple(x)
 
 
@@ -256,7 +256,7 @@ class MultiAnd(Bloq):
     Args:
         cvs: A tuple of control variable settings. Each entry specifies whether that
             control line is a "positive" control (`cv[i]=1`) or a "negative" control `0`.
-            If a (symbolic) integer is passed, assumes the control values to be all 1's.
+            If a HasLength object is passed, assumes the control values to be all 1's.
 
 
     Registers:
@@ -265,20 +265,20 @@ class MultiAnd(Bloq):
         target [right]: The output bit.
     """
 
-    cvs: Union[Length, Tuple[SymbolicInt, ...]] = field(converter=_to_tuple)
+    cvs: Union[HasLength, Tuple[SymbolicInt, ...]] = field(converter=_to_tuple_or_has_length)
 
     @cvs.validator
     def _validate_cvs(self, field, val):
         if not is_symbolic(val) and len(val) < 3:
-            raise ValueError("MultiAnd must have at least 3 control values `cvs`.")
+            raise ValueError("MultiAnd must cvshave at least 3 control values `cvs`.")
 
     @property
-    def bitsize(self) -> SymbolicInt:
-        return self.cvs.bitsize if isinstance(self.cvs, Length) else len(self.cvs)
+    def n_ctrls(self) -> SymbolicInt:
+        return self.cvs.n if isinstance(self.cvs, HasLength) else len(self.cvs)
 
     @property
-    def control_values(self) -> Tuple[SymbolicInt, ...]:
-        if isinstance(self.cvs, Length):
+    def concrete_cvs(self) -> Tuple[SymbolicInt, ...]:
+        if isinstance(self.cvs, HasLength):
             raise ValueError(f"{self.cvs} is symbolic")
         return self.cvs
 
@@ -286,8 +286,8 @@ class MultiAnd(Bloq):
     def signature(self) -> Signature:
         return Signature(
             [
-                Register('ctrl', QBit(), shape=(self.bitsize,)),
-                Register('junk', QBit(), shape=(self.bitsize - 2,), side=Side.RIGHT),
+                Register('ctrl', QBit(), shape=(self.n_ctrls,)),
+                Register('junk', QBit(), shape=(self.n_ctrls - 2,), side=Side.RIGHT),
                 Register('target', QBit(), side=Side.RIGHT),
             ]
         )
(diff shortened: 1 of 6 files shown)


qualtran/bloqs/mcmt/and_bloq.py
```diff
@@ -260,26 +256,38 @@ class MultiAnd(Bloq):
     Args:
         cvs: A tuple of control variable settings. Each entry specifies whether that
             control line is a "positive" control (`cv[i]=1`) or a "negative" control `0`.
+            If a (symbolic) integer is passed, assumes the control values to be all 1's.
```
fdmalone, line 259: Is this subject to change / worth thinking about? I guess I haven't really thought about what symbolic controls mean.

tanujkhattar: Its easy to argue for the case where the control spec is simply all 1s; and thats mostly the case I want to update everywhere we currently expect `cvs: Tuple[int, ...]` 

In future, we'd ideally like these bloqs to be able to take any `cvs: CtrlSpec` and it would be more nuanced to think about what a symbolic controlled spec would mean at that point. But for now, I just want to remove blockers for enabling symbolic resource estimation for higher level bloqs like state preparation via alias sampling / reflection etc.

mpharrigan: I'm not super thrilled about having one attribute named `cvs` that can represent _either_ a list of control values _or_ the number of controls. 

```python
def __init__(self, *, cvs=None, n_ctrl=None):
  if cvs is not Nont and n_ctrl is not None: raise ValueError()
  self._cvs = cvs
  self._n_ctrl = n_ctrl

  @property
  def cvs():
    ...
  @property
  def n_ctrl():
    ..
```

qualtran/bloqs/mcmt/and_bloq.py
```diff
@@ -260,26 +256,38 @@ class MultiAnd(Bloq):
...
 
     @cvs.validator
     def _validate_cvs(self, field, val):
-        if len(val) < 3:
+        if not is_symbolic(val) and len(val) < 3:
             raise ValueError("MultiAnd must have at least 3 control values `cvs`.")
 
+    @property
+    def bitsize(self) -> SymbolicInt:
+        return self.cvs.bitsize if isinstance(self.cvs, Length) else len(self.cvs)
```
mpharrigan, line 277: n_ctrls

qualtran/bloqs/mcmt/and_bloq.py
```diff
@@ -260,26 +256,38 @@ class MultiAnd(Bloq):
...
+
+    @property
+    def control_values(self) -> Tuple[SymbolicInt, ...]:
+        if isinstance(self.cvs, Length):
+            raise ValueError(f"{self.cvs} is symbolic")
+        return self.cvs
```
mpharrigan, line 283: cvs

qualtran/bloqs/mcmt/and_bloq.py
tanujkhattar: Is there a way to do it without a custom init with frozendataclasses?  I can also do what I did for `QROM`; i.e. rename `cvs` -> `cvs_or_n_ctrl: Union[]` and then expose `cvs` and `n_ctrl` as properties.

qualtran/symbolics/types.py
```diff
@@ -46,6 +46,19 @@ def is_symbolic(self):
         return True
 
 
+@frozen
+class Length:
+    """Symbolic value for an object that has a length."""
+
+    bitsize: SymbolicInt
```
mpharrigan, line 53: this doesn't have to represent a bitwidth, right? it can be the length of anything. Perhaps `n`.

qualtran/symbolics/types.py
```diff
@@ -46,6 +46,19 @@ def is_symbolic(self):
         return True
 
 
+@frozen
+class Length:
```
mpharrigan, line 50: is this a subclass of `Shaped`?

qualtran/symbolics/types.py
```diff
@@ -46,6 +46,19 @@ def is_symbolic(self):
...
+    """Symbolic value for an object that has a length."""
```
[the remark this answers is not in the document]
tanujkhattar: Will change to `HasLength`

qualtran/bloqs/mcmt/and_bloq.py
tanujkhattar: @mpharrigan  LMK if this approach sounds good and I'll make the changes.

mpharrigan: I think you can do this with attrs by prepending an underscore to the field name and setting `frozen(kwargs=True)`

tanujkhattar: Then it would look something like this:

```
@frozen
class MultiAnd(Bloq):
    _cvs: Optional[Tuple[SymbolicInt, ...]] = field(
        converter=_to_tuple_or_none, kw_only=True, default=None
    )
    _n_ctrls: Optional[HasLength] = field(
        converter=_to_has_length_or_none, kw_only=True, default=None
    )

        assert (self._cvs is None) ^ (self._n_ctrls is None)

```

We'll have to add this logic everywhere we use `cvs: Tuple[int, ...]` right now. I'm find doing this, but there's definitely more going on in this approach than having a single `Union` which must not be None.

tanujkhattar: Also, the main reason for defining a new `Length` object is so that we can set `cvs` to be of type `Length` and `len(cvs)` still works. 

The above approach where we maintain both `n_ctrls` and `cvs` independently everywhere wouldn't require introducing `Length` since `n_ctrls` can just be `Symbolic`. 

At other places, we typically have a union of `Shaped` and numpy arrays so that `object.shape` works irrespective of the type of `object`. I think following a similar pattern here would keep things consistent. 

Tl;Dr - let's maintain `cvs: Union[HasLength, Tuple[int, ...]]`  or `cvs_or_n_ctrls: Union[HasLength, Tuple[int, ...]])` ? WDYT?

mpharrigan: oh oh oh when I first read it as `Union[Sequence, Length]` I was like, these are two different things. But `HasLength` ... yeah just one field named `cvs` is good

qualtran/bloqs/mcmt/and_bloq.py
mpharrigan: I still think this should be `n_ctrls` or similar (rather than bitsize)

qualtran/bloqs/mcmt/and_bloq.py
mpharrigan: I still think this should be meaningfully disambiguated. `and_bloq.cvs` vs `and_bloq.control_values` ... 

maybe `concrete_cvs`? idk

mpharrigan: we have `is_symbolic` in the symbolic utils. What's the opposite of symbolic? we should pick a word for it and use that word consistently throughout the library

qualtran/bloqs/mcmt/and_bloq.py
mpharrigan: can you update the docstring as well. instead of "if a symbolic integer is passed" say "if a HasLength object"

qualtran/bloqs/mcmt/and_bloq.py
tanujkhattar: I can use whatever you suggest -- I trust your expertise on the English language more than mine :)

qualtran/bloqs/mcmt/and_bloq.py
```diff
@@ -249,7 +239,13 @@ def _and_bloq() -> And:
 )
 
 
-def _to_tuple(x: Iterable[Union[int, sympy.Expr]]) -> Sequence[Union[int, sympy.Expr]]:
+def _to_tuple(
+    x: Union[SymbolicInt, Iterable[SymbolicInt]]
+) -> Union[Length, Tuple[SymbolicInt, ...]]:
+    if isinstance(x, sympy.Expr):
+        return Length(x)
```
mpharrigan, line 246: wait, shouldn't we make users pass a `HasLength` object explicitly? This is how we do `Shaped` if I'm not mistaken

tanujkhattar: Okay that's right, I'll change it.

qualtran/bloqs/mcmt/and_bloq.py
mpharrigan: let's go with "concrete" https://en.wiktionary.org/wiki/concrete

mpharrigan (approved):
lgtm with the changes brought up in the comments

tanujkhattar:
Addressed all the comments. Also realized that its a bit annoying to have `HasLength` as a drop-in replacement where `len(cvs)` just returns a symbolic value because of the way python treats the `__len__` dunder method. Explained in the docstring and added a workaround. Merging now

anurudhp:
> Also realized that its a bit annoying to have `HasLength` as a drop-in replacement where `len(cvs)` just returns a symbolic value because of the way python treats the `__len__` dunder method. 

@tanujkhattar fwiw, I faced the same issue when implementing `Shaped`, my workaround was to define `slen`:
https://github.com/quantumlib/Qualtran/blob/5aabc66c37bd4c8e8b330a896b98178109c5bd34/qualtran/symbolics/math_funcs.py#L102-L105

We should extend this by adding  `HasLength` to the input type here.

mpharrigan:
and/or make HasLength be a subclass of Shaped

PR #1018, opened 2024-05-29
Global call graph test

mpharrigan:
This PR has two commits

First, I add a dev_tool module for making one big call graph. This can be a useful starting point for querying other properties and idiosyncrasies of the bloq library; but here I use it to add a test that all bloq examples have a valid call graph. 

Then, I make two minor fixes revealed by running the "all call graph" test.


mpharrigan:
I brought this up to speed. @tanujkhattar if you could PTAL I think this would be a decent `slow` test; and it's generally useful to have a script example for making a unified call graph

tanujkhattar (approved):

Issue #1011, opened 2024-05-29
qubits are never freed in many cirq-style decompositions of arithmetic gates


tanujkhattar:
All calls to `context.qubit_manager.qalloc` should be followed by calls to `context.qubit_manager.qfree`. Many arithmetic gate decompositions do not free the allocated qubits. This is bad and should be fixed. 

- [ ] There's no `qfree` call in entire comparison file - https://github.com/quantumlib/Qualtran/blob/main/qualtran/bloqs/arithmetic/comparison.py

PR #1020, opened 2024-05-30
free ancillas in comparison gates

NoureldinYosri:
fixes https://github.com/quantumlib/Qualtran/issues/1011


mpharrigan:
why was this possible? can we add a test that would have been failing?

NoureldinYosri:
@mpharrigan I added a test to ensure we free the qubits

tanujkhattar (approved):

mpharrigan:
> I added a test to ensure we free the qubits

Thanks!

Issue #1019, opened 2024-05-30
Rough edges when swapping QROM for SelectSwapQROM in StatePreparationViaRotation


wjhuggins:
I was trying to see what would happen to the resource counts if I replaced QROM with SelectSwapQROM in StatePreparationViaRotation and I ran into an issue. A drop in replacement doesn't work. It fails in a way that is captured in this MWE:

```
import numpy as np

from qualtran.bloqs.data_loading.qrom import QROM
from qualtran.bloqs.data_loading.select_swap_qrom import SelectSwapQROM

data = np.array([1])
selection_bitsize = 0
target_bitsize = 2
control_bitsize = 1

qrom = QROM(
    [data],
    selection_bitsizes=(selection_bitsize,),
    target_bitsizes=(target_bitsize,),
    num_controls=control_bitsize,
)

select_swap_qrom = SelectSwapQROM(
    [data],
    selection_bitsizes=(selection_bitsize,),
    target_bitsizes=(target_bitsize,),
    num_controls=control_bitsize,
)
```

fdmalone:
This looks like a case where a more sensible error should be raised, does SelectSwapQROM make sense with a length 1 array (I think the answer is it should handle it gracefully).

PR #1025, opened 2024-05-30
Fix SelectSwapQROM error when selection bitsize is 0

tanujkhattar:
Fixes https://github.com/quantumlib/Qualtran/issues/1019

Also fixes a bug in computation of `find_optimal_log_block_size` which was being called with `selection_bitsize` instead of `target_bitsize`.

### DIFF
--- a/qualtran/bloqs/chemistry/thc/thc.ipynb
+++ b/qualtran/bloqs/chemistry/thc/thc.ipynb
@@ -433,7 +433,7 @@
     "    print(f\"{k+':':20s} qualtran = {binned_counts[k]:5d} vs paper cost = {v:5d}.\")\n",
     "\n",
     "print(f\"Total cost = {sum(v for v in binned_counts.values())}\")\n",
-    "assert binned_counts['data_loading'] == 432"
+    "assert binned_counts['data_loading'] == 304"
    ]
   },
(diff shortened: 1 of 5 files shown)


qualtran/bloqs/chemistry/thc/thc.ipynb
```diff
@@ -433,7 +433,7 @@
     "    print(f\"{k+':':20s} qualtran = {binned_counts[k]:5d} vs paper cost = {v:5d}.\")\n",
     "\n",
     "print(f\"Total cost = {sum(v for v in binned_counts.values())}\")\n",
-    "assert binned_counts['data_loading'] == 432"
+    "assert binned_counts['data_loading'] == 304"
```
fdmalone, line 436: do you know why this changed?

fdmalone (approved):

qualtran/bloqs/chemistry/thc/thc.ipynb
tanujkhattar: Because `find_optimal_log_block_size` is now called with the correct `target_bitsize` -- there was a bug where I was calling it with `selection_bitsize` instead so the block size was not optimal.

PR #1029, opened 2024-05-30
Create a cost model for physical resources and a static classemthod to turn bloqs into AlgorithmSummary

NoureldinYosri:


qualtran/resource_counting/_algorithm_summary_counts.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+@frozen
+class AlgorithmSummaryCounts(CostKey[AlgorithmSummaryDict]):
+    """A cost that represents the variables that affect physical resource estimation.
+
+    The result of computing this cost is a mapping from arguments of AlgorithmSummary to values
+    with two exceptions:
+        - `algorithm_qubits` are not counted and should be computed using QubitCount cost.
+        - `measurements` are not counted since qualtran doesn't measurement as a bloq.
```
mpharrigan, line 39: "doesn't measurement"

qualtran/resource_counting/_algorithm_summary_counts.py
```diff
@@ -0,0 +1,86 @@
...
+if TYPE_CHECKING:
+    from qualtran import Bloq
+
+AlgorithmSummaryDict = Dict[str, int]
+
+logger = logging.getLogger(__name__)
+
+
+@frozen
+class AlgorithmSummaryCounts(CostKey[AlgorithmSummaryDict]):
```
mpharrigan, line 33: Isn't this mostly redundant with the existing `QECGatesCost` cost key which uses the `GateCounts` data class? 

The `GateCounts` numbers are all factorized, so you should be able to add up the ones that need adding up and create an `AlgorithmsSummary`, no? If not: what are the logical incompatibilities?

NoureldinYosri: rotations ... the QECGatesCost ignores rotations (i.e. gives a GateCounts of all zeros)

mpharrigan: can we add that then

qualtran/resource_counting/_algorithm_summary_counts.py
```diff
@@ -0,0 +1,86 @@
...
+        if isinstance(bloq, And):
+            return {'toffoli_gates': int(not bloq.uncompute)}
+
+        # CSwaps aka Fredkin
+        if isinstance(bloq, TwoBitCSwap):
+            return {'toffoli_gates': 1}
+
+        callees = get_bloq_callee_counts(bloq)
+        if len(callees) == 0:
+            # If not Clifford or T of Toffoli and has no callees: assume a rotation.
```
mpharrigan, line 70: this seems like a very dangerous assumption

NoureldinYosri: the alternative returns all zeros which is the current behaviour

qualtran/resource_counting/_algorithm_summary_counts.py
NoureldinYosri: done

qualtran/surface_code/algorithm_summary.py
```diff
@@ -110,3 +117,15 @@ def to_magic_count(
                 * rotation_model.rotation_cost(error_budget / self.rotation_gates)
             )
         return ret
+
+    @staticmethod
+    def from_bloq(bloq: 'qualtran.Bloq') -> 'AlgorithmSummary':
```
mpharrigan, line 122: nit: usually we don't put `qualtran.` in the type annotations

mpharrigan (approved):
please document the depth parameter and add an explicit list of rotations that we check for (like how we do cliffords)

Issue #811, opened 2024-03-21
Consolidate bit iteration


fdmalone:
Currently we have https://github.com/quantumlib/Qualtran/blob/main/qualtran/cirq_interop/bit_tools.py and https://github.com/quantumlib/Qualtran/blob/main/qualtran/simulation/classical_sim.py which somewhat replicate functionality. These should be consolidated and possibly updated to support > 64 bit uints?

mpharrigan:
Following #1041, classical_sim bit tools still needs to be ported

PR #1032, opened 2024-05-31
[WIP] replace `bit_tools.iter_bits` with `classical_sim.ints_to_bits`

charlesyuan314:
This is just a first attempt at #811 to gather feedback and to see what tests fail.

Known issues:
- `iter_bits` still exists even though it is entirely redundant.
- `iter_bits` no longer supports sign.
- `iter_bits` no longer raises `ValueError` for too-wide inputs.

### DIFF
--- a/qualtran/cirq_interop/bit_tools.py
+++ b/qualtran/cirq_interop/bit_tools.py
@@ -26,12 +26,11 @@ def iter_bits(val: int, width: int) -> Iterator[int]:
     Args:
         val: The integer value. Its bitsize must fit within `width`
         width: The number of output bits.
-        signed: If True, the most significant bit represents the sign of
-            the number (ones complement) which is 1 if val < 0 else 0.
     Raises:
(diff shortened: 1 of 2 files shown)


charlesyuan314:
@mpharrigan @tanujkhattar @fdmalone Do you think this is a sane approach?

P.S. Looks like I can't directly request a PR review without being a member of the `quantumlib` org.

fdmalone:
I think this makes sense. The bigger migration might be to just replace iter_bits with the corresponding to_bits methods from the register data types(e.g. https://github.com/quantumlib/Qualtran/blob/main/qualtran/_infra/data_types.py#L215) which will handle the signed-ness etc. But maybe out of scope for this PR.

charlesyuan314:
Closing as I'm currently working on a PR to replace the `bit_tools` functions with `QDType` implementations as @mpharrigan suggested.

PR #1017, opened 2024-05-29
Symbolic bloq counts for `StatePreparationViaAliasSampling`

tanujkhattar:
image

image

Also added symbolic anlaysis for state prep via rotations

image

I'll pull out incremental changes into smaller PRs so they are easier to review.

PR #1034, opened 2024-05-31
Add build_call_graph and symbolic resource estimation to MultiControlPauli

tanujkhattar:
Pulled out from https://github.com/quantumlib/Qualtran/pull/1017

### DIFF
--- a/qualtran/bloqs/mcmt/multi_control_multi_target_pauli.py
+++ b/qualtran/bloqs/mcmt/multi_control_multi_target_pauli.py
@@ -13,7 +13,7 @@
 #  limitations under the License.
 
 from functools import cached_property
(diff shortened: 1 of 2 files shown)


fdmalone (approved):
you need to update the notebooks

tanujkhattar:
I don't include the symbolic example in the notebook because we still can't access the signature (and therefore call `show_bloq()` etc.) on the symbolic bloq since the `shape` is symbolic. We still need more work for all these bloqs to either 

(a) Support symbolic `shape` when instantiating a register (less likely)
(b) Refactor the bloqs so that the control register is of type `QAny(symbolic_size)` with `shape=()`. (more likely)

fdmalone:
Ok!

mpharrigan:
Can you open an issue to track the outstanding symbolic ctrl shaped register problem

PR #1037, opened 2024-05-31
Add `build_call_graph` and symbolic resource estimation to `StatePreparationViaRotation`

tanujkhattar:
Pulled out from https://github.com/quantumlib/Qualtran/pull/1017


qualtran/bloqs/state_preparation/state_preparation_via_rotation_test.py
```diff
@@ -36,6 +38,28 @@ def test_state_prep_via_rotation(bloq_autotester):
...
 
+def test_state_prep_via_rotation_symb():
+    bloq = _state_prep_via_rotation_symb.make()
+    L, phase = bloq.n_coeff, bloq.phase_bitsize
+    expected_t_count_expr = 16 * L + 8 * phase - 32
+    assert isinstance(expected_t_count_expr, sympy.Expr)
+    assert bloq.t_complexity().t == expected_t_count_expr
+
+    # Compare bloq counts via expression to actual bloq counts and make sure they
+    # are "close enough"
```
fdmalone, line 49: What does close enough mean here? What's the discrepancy from?

qualtran/bloqs/state_preparation/state_preparation_via_rotation.py
```diff
@@ -324,36 +390,53 @@ def signature(self) -> Signature:
...
-        soqs = bb.add_d(qrom, **soqs)
+        soqs = bb.add_d(self.qrom_bloq.adjoint(), **soqs)
         soqs["phase_gradient"] = phase_grad
         bb.free(cast(Soquet, soqs.pop("target0_")))
         return soqs
 
+    def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
+        return {(self.qrom_bloq, 1), (self.qrom_bloq.adjoint(), 1), (self.add_into_phase_grad, 1)}
+
+    def __repr__(self):
```
fdmalone, line 431: Why do we need a `__repr__` here?

fdmalone (approved):

qualtran/bloqs/state_preparation/state_preparation_via_rotation.py
tanujkhattar: Nicer for debugging. We should probably add a `__repr__` everywhere we have classes that store `data: NDArray` as attributes (like qroms)

qualtran/bloqs/state_preparation/state_preparation_via_rotation_test.py
tanujkhattar: The prepare amplitudes calls PRGA for  0 + 1 + 2 + 3 + 4 + 5 ... + n - 1 bitsize (where `n` is size of selection register) and corresponding `rom_values` of length `2 ** 0 + 2 ** 1 + 2 ** 2 + ...  + 2 ** (n - 1)`

When `n` is symbolic, we can't simulate this in the `build_call_graph` so I instead return a single call to PRGA with bitsize `n` and `rom_values` of size `2^n`. The constant factor of the dominant cost of `QROM` scales as `4 * L` and sum of `2 ** 0 + 2 ** 1 + 2 ** 2 + ...  + 2 ** (n - 1)` is `2^n` so dominant cost matches. But the smaller costs like addition into phase gradient register scale with selection bitsize; and so `1 + 2 + 3 + ... + n - 1  >   n` -- this is where the discrepancy comes from.

PR #1040, opened 2024-05-31
Installation docs

mpharrigan:
Document the installation procedure


docs/index.rst
```diff
@@ -51,10 +54,103 @@ and high-level quantum programs. :ref:`Read more... <bloqs_library>`
...
+~~~~~~
+
+Qualtran is a Python package. If you already have a working Python environment,
+
+.. code-block:: sh
+
+    pip install qualtran
+
+will install the library and its dependencies. Qualtran uses `Graphviz <https://graphviz.org/>`_
+to draw diagrams, which can't be installed using the Python package manager ``pip``. Follow
```
fdmalone, line 79: Is it worth just reproducing these here (or adding a little script)

fdmalone: Eh it's just apt-get nvm

docs/index.rst
```diff
@@ -51,10 +54,103 @@ and high-level quantum programs. :ref:`Read more... <bloqs_library>`
...
+
+You should be able to import ``qualtran`` from a Python command prompt.
+
+.. code-block:: python
+
+    import qualtran
+
+If this is successful,
+you can move on to learning how to `write bloqs <./_infra/Bloqs-Tutorial.html>`_ or investigate
+the :ref:`bloqs library<bloqs_library>`.
```
fdmalone, line 156: This link doesn't render?

fdmalone (approved):

docs/index.rst
fdmalone: I was being slow. Thought it was a README

docs/index.rst
mpharrigan: It's platform dependent, so I'd rather just point them to the actual docs.

docs/index.rst
mpharrigan: this will be on the front page of the readthedocs "below the fold". It's restructured text which is the weirdest markup syntax but it's the native sphinx form

PR #1041, opened 2024-05-31
Remove `bit_tools` and replace with `QDType` features

### LINKED ISSUE #811: Consolidate bit iteration

charlesyuan314:
Replace references to `iter_bits`, `iter_bits_twos_complement`, and `iter_bits_fixed_point` from `qualtran.cirq_interop.bit_tools` as they are largely duplicated by functionality in `QUInt`, `QInt`, and `QFxp` respectively. 

Move functionality of `float_as_fixed_width_int` to `QFxp.to_fixed_width_int`.

Modify `QFxp.to_bits` to enable the exactly-representable check to be bypassed and to enable overriding the twos-complement representation of negative binary fractions used by `fxpmath`. Previous users of `iter_bits_fixed_point` instead require sign-magnitude representation.

Remove `bit_tools` and move its tests to `data_types`.

Towards #811. Updates to users of `classical_sim.ints_to_bits` should follow in another PR.


charlesyuan314:
@mpharrigan PTAL.

As far as I can tell, `classical_sim.ints_to_bits` and `classical_sim.bits_to_ints` are currently only used in relation to classical simulation. They also use `np.uint64` for what I assume to be performance reasons. So I haven't touched those at this point.

mpharrigan (approved):

qualtran/bloqs/mean_estimation/complex_phase_oracle_test.py
```diff
@@ -70,7 +69,9 @@ def test_phase_oracle(bitsize: int, arctan_bitsize: int):
     prepared_state = state_vector.sum(axis=1)
     for x in range(2**bitsize):
         output_val = -2 * np.arctan(x, dtype=np.double) / np.pi
-        output_bits = [*bit_tools.iter_bits_fixed_point(np.abs(output_val), arctan_bitsize)]
+        output_bits = QFxp(arctan_bitsize, arctan_bitsize).to_bits(
+            np.abs(output_val), require_exact=False
+        )
```
mpharrigan, line 74: This got way slower and now this is one of the slowest unit tests

Issue #905, opened 2024-04-29
Remove _t_complexity_ overrides where `build_call_graph` exists. 


mpharrigan:
The t_complexity protocol can quickly use the call graph protocol to derive accurate t counts from subbloqs. This handles meta-bloqs like `Adjoint` correctly. 

For bloqs where `build_call_graph` is implemented and correct -- the `_t_complexity_` override should be removed. If "leaf values" are provided by the `_t_complexity_` override, they should be moved to a unit test. 

This is of extreme importance for the places where the `_t_complexity_(self, adjoint: bool)` method has been hacked to accept an `adjoint` parameter.

PR #1044, opened 2024-06-03
Remove _t_complexity_ if build_call_graph exists

dstrain115:
- If call graph exists, then the t complexity is redundant. These t_complexities are moved to tests to ensure that the t_complexity is as expected.

- Only issue is LPRSInterimPrep no longer supports symbolics for t_complexity.

Fixes: #905


qualtran/bloqs/phase_estimation/lp_resource_state_test.py
```diff
@@ -38,11 +38,6 @@ def test_lp_resource_state_auto(bloq_autotester):
     bloq_autotester(_lp_resource_state_small)
 
 
-def test_lp_resource_state_symb():
-    bloq = _lp_resource_state_symbolic.make()
-    assert bloq.t_complexity().t == 4 * bloq.bitsize
-
-
```
anurudhp, line 45: perhaps we should keep this test marked `@pytest.mark.xfail`? (so that we don't forget to fix symbolics for LPRS.)

dstrain115: SGTM, Done.

tanujkhattar: It is not ideal to have the symbolis not working -- maybe it's best to wait till we can add a CRz  Bloq and use that so we can continue to support symbolics. 

I don't think there's any rush to remove t_complexity; especially by breaking symbolic resource counting which is a lot more useful.

qualtran/bloqs/phase_estimation/lp_resource_state.py
```diff
@@ -94,9 +93,6 @@ def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
             }
         return ret
 
-    def _t_complexity_(self) -> 'TComplexity':
-        return TComplexity(rotations=self.bitsize + 1, clifford=2 + self.bitsize)
```
anurudhp, line 98: @tanujkhattar isn't this TComplexity expression wrong? we use `self.bitsize` CRz gates but `TComplexity.rotations` is only for single qubit rotations, right?

dstrain115: Yes, this is totally wrong.  Even if we revert the change to fix symbolics, this expression still needs to change.

dstrain115: Reverted the _t_complexity_ removal, and added a comment.  Let me know if I misrepresented the factor in the comment.

qualtran/bloqs/phase_estimation/lp_resource_state_test.py
dstrain115: Reverted this and reenabled symbolics.

qualtran/bloqs/phase_estimation/lp_resource_state.py
```diff
@@ -95,7 +95,10 @@ def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
         return ret
 
     def _t_complexity_(self) -> 'TComplexity':
-        return TComplexity(rotations=self.bitsize + 1, clifford=2 + self.bitsize)
+        # Uses self.bitsize controlled-Rz rotations which decomposes into
+        # 2 single-qubit rotations and 3 cliffords
```
anurudhp, line 99: source?

qualtran/bloqs/phase_estimation/lp_resource_state.py
anurudhp: maybe we should make this change in a different PR? because it's a bug-fix for the tcomplexity expression, unlike the other changes that remove `_t_complexity_`

mpharrigan: I don't think that's necessary

mpharrigan (approved):
In the interest of moving things forward, and because this leaves everything strictly in as-good or a better place than it was before: let's merge. Interested parties can do the CRz implementation and fix in a following PR

PR #1049, opened 2024-06-04
BugFix in `SelectSwapQROM` when multiple datasets are specified

tanujkhattar:
Two bugfixes:

1. Use `self.use_dirty_ancilla` in `_value_equality_values_` -- This is important to make sure `bloq.t_complexity()` returns different values when `bloq` is initialized with `use_dirty_ancilla=True` vs `use_dirty_ancilla=False`. Right now, they can return the same value because the t-complexity is cached
2. `qrom_batched_target[0]` can return a numpy array because the target shapes can be multi-dimensional. This is especially true when input data is multi-dimensional (i.e. `data[N][M] --> bached_qrom_data[N / k1][M / k2]`). We need the `target_0_0_..._0`'th register and not the `target_0`th register; which can be multidimensional. 

The added tests fail on main but pass with this PR.

### DIFF
--- a/qualtran/bloqs/data_loading/select_swap_qrom_test.py
+++ b/qualtran/bloqs/data_loading/select_swap_qrom_test.py
@@ -18,7 +18,13 @@ import pytest
 
 from qualtran._infra.data_types import BoundedQUInt
 from qualtran._infra.gate_with_registers import get_named_qubits, split_qubits
-from qualtran.bloqs.data_loading.select_swap_qrom import find_optimal_log_block_size, SelectSwapQROM
+from qualtran.bloqs.data_loading import QROM
+from qualtran.bloqs.data_loading.select_swap_qrom import (
+    _qroam_multi_data,
+    _qroam_multi_dim,
+    find_optimal_log_block_size,
+    SelectSwapQROM,
+)
 from qualtran.cirq_interop.bit_tools import iter_bits
 from qualtran.cirq_interop.t_complexity_protocol import t_complexity, TComplexity
 from qualtran.cirq_interop.testing import assert_circuit_inp_out_cirqsim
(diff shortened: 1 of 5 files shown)


qualtran/bloqs/data_loading/select_swap_qrom_test.py
```diff
@@ -158,3 +164,17 @@ def test_qroam_many_registers():
         log_block_sizes=(log_block_size,),
     )
     qrom.call_graph()
+
+
+@pytest.mark.parametrize('bloq_example', [_qroam_multi_data, _qroam_multi_dim])
```
fdmalone, line 169: Why isn't this just a regular old bloq_example auto tester? Won't it run more tests?

fdmalone (approved):

fdmalone:
> We need the `target_0_0_..._0`'th register and not the `target_0`th register; which can be multidimensional. 

I was a little confused by the underscores here given our weird target register convention, you just mean you want the actual zero'th target register not the zero'th row (in the 2D example)

qualtran/bloqs/data_loading/select_swap_qrom_test.py
tanujkhattar: Changed to `auto tester`. Found and fixed another bug in `nth_operation_callgraph` of `QROM`; which was not updated to support the recently added `target_shapes` feature required for `SelectSwapQROM`.

tanujkhattar:
> I was a little confused by the underscores here given our weird target register convention, you just mean you want the actual zero'th target register not the zero'th row (in the 2D example)

Yup

mpharrigan:
Why are we using value_equality_values at all?

tanujkhattar:
> Why are we using value_equality_values at all?

We store tuples of numpy arrays, which don't support value equality by default.

tanujkhattar:
@fdmalone @mpharrigan Removed the use of `_value_equality_values_` with the fancy attrs functionality Fionn mentioned above. This is ready to go now.

Issue #1052, opened 2024-06-05
Computing qubit costs for QROM is very slow


tanujkhattar:
I need the following code to run in 3-4 seconds; not 10+ minutes on a Google colab instance.

```python
%%time
from qualtran.resource_counting import get_cost_value, QubitCount

data = [*range(2**12)]
log_block_sizes = [*range(1, 9)]

qroam_clean = SelectSwapQROM.build_from_data(data, use_dirty_ancilla=False)
qroam_dirty = SelectSwapQROM.build_from_data(data, use_dirty_ancilla=True)

costs_cache = {}
qroam_t_costs_clean, qroam_t_costs_dirty = [], []
qroam_qubit_costs_clean, qroam_qubit_costs_dirty = [], []
for b in log_block_sizes:
  print(f'{b=}')
  qroam_clean_curr = qroam_clean.with_log_block_sizes(log_block_sizes=[b])
  qroam_dirty_curr = qroam_dirty.with_log_block_sizes(log_block_sizes=[b])
  qroam_t_costs_clean.append(qroam_clean_curr.t_complexity().t)
  qroam_t_costs_dirty.append(qroam_dirty_curr.t_complexity().t)
  qroam_qubit_costs_clean.append(get_cost_value(qroam_clean_curr, QubitCount(), costs_cache=costs_cache))
  qroam_qubit_costs_dirty.append(get_cost_value(qroam_dirty_curr, QubitCount(), costs_cache=costs_cache))


qrom_t_cost = QROM.build_from_data(data).t_complexity().t
qrom_qubit_cost = get_cost_value(QROM.build_from_data(data),  QubitCount(), costs_cache=costs_cache)
```

Hers is the timing from my current run -

<details>

<img width="1061" alt="image" src="https://github.com/quantumlib/Qualtran/assets/7863287/436ffaa8-2316-4594-9f76-ab7f626565e9">

</details>

I want to generate a plot like this

<img width="1381" alt="image" src="https://github.com/quantumlib/Qualtran/assets/7863287/634385c4-f4c6-4fb7-9d3c-d658202f3359">

tanujkhattar:
On current main

<img width="1189" alt="image" src="https://github.com/quantumlib/Qualtran/assets/7863287/d2edc8f3-43a7-4f45-b013-fe225865681c">


[...]

PR #1054, opened 2024-06-05
Performance improvements in cirq interop

tanujkhattar:
Low hanging fruits towards https://github.com/quantumlib/Qualtran/issues/1052

Before:

image

After:

image

### DIFF
--- a/qualtran/_infra/registers.py
+++ b/qualtran/_infra/registers.py
@@ -16,6 +16,7 @@
 import enum
 import itertools
 from collections import defaultdict
+from functools import cached_property
 from typing import cast, Dict, Iterable, Iterator, List, overload, Tuple, Union
 
 import attrs
@@ -99,6 +100,10 @@ class Register:
 
         This is the product of bitsize and each of the dimensions in `shape`.
         """
+        return self._total_bits
+
+    @cached_property
+    def _total_bits(self) -> int:
         return self.bitsize * int(np.prod(self.shape))
 
     def adjoint(self) -> 'Register':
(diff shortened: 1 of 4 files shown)


qualtran/_infra/registers.py
```diff
@@ -99,6 +100,10 @@ def total_bits(self) -> int:
 
         This is the product of bitsize and each of the dimensions in `shape`.
         """
+        return self._total_bits
+
+    @cached_property
```
mpharrigan, line 105: just use cached method on the original method?

qualtran/bloqs/data_loading/qrom.py
```diff
@@ -157,7 +157,7 @@ def decompose_zero_selection(
                     ctrl=ctrl, junk=junk, target=and_target
                 )
             yield multi_controlled_and
-            yield self._load_nth_data(zero_indx, lambda q: CNOT().on(and_target, q), **target_regs)
+            yield self._load_nth_data(zero_indx, lambda q: cirq.CNOT(and_target, q), **target_regs)
```
mpharrigan, line 160: use qualtran `CNOT` so that it can immediately unwrap from CirqGateAsBloq without doing any logic?

tanujkhattar: Using qualtran `CNOT` is slower because `CNOT` in Qualtran is not a `GateWithRegisters`. So the cirq construction wraps it in a `BloqAsCirqGate`; which ends up being slow. See the first profiling picture in my linked comment.

qualtran/_infra/registers.py
tanujkhattar: `functools` doesn't seem to have a `cached_method` and I didn't want to add another dependency. using just `cache` would require computing hash of the class itself which would again be slow.

PR #1057, opened 2024-06-07
Optimize gate count of Subtract to n-1 Toffolis

NoureldinYosri:
This uses the bit-magic `a-b = ~(~a + b)`


mpharrigan:
> This uses the bit-magic a-b = ~(~a + b)

would be useful to include in the description

qualtran/bloqs/arithmetic/subtraction.py
```diff
@@ -61,7 +63,9 @@ class Subtract(Bloq):
         b: A b_dtype.bitsize-sized input/output register (register b above).
     """
 
-    a_dtype: Union[QInt, QUInt, QMontgomeryUInt] = field()
+    a_dtype: Union[QInt, QUInt, QMontgomeryUInt] = field(
+        converter=lambda k: QUInt(k) if isinstance(k, int) else k
+    )
```
mpharrigan, line 68: let's keep this explicit for now. (if we were to do this we'd want it on both types)

qualtran/bloqs/arithmetic/subtraction_test.py
```diff
@@ -21,15 +21,20 @@
 from qualtran.resource_counting.generalizers import ignore_split_join
 
 
-def test_subtract_bloq_decomposition():
-    gate = Subtract(QInt(3), QInt(5))
+@pytest.mark.parametrize(
+    ['a_bits', 'b_bits'], [(a, b) for a in range(1, 6) for b in range(a, 6) if a + b <= 10]
+)
+def test_subtract_bloq_decomposition(a_bits, b_bits):
```
mpharrigan, line 27: can you add a test that uses the classical simulation protocol

```python
cbloq = subtract.decompose_bloq()

for in_vals:
  ref = subtract.call_classically(in_vals)
  comp = cbloq.call_classically(in_vals)
  assert ref == comp
```

mpharrigan (dismissed):
docstring!

mpharrigan:
@NoureldinYosri can you make the requested changes

NoureldinYosri:
@mpharrigan done, I was getting test failures when doing the classical simulation test because classical simulation tools couldn't handle negative values and both Add and Subtract bloqs had faulty classical action definitions for the case of signed inputs... I fixed all of these issues

qualtran/bloqs/arithmetic/subtraction_test.py
NoureldinYosri: done

qualtran/bloqs/arithmetic/subtraction.py
NoureldinYosri: done

NoureldinYosri:
@mpharrigan ptal

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -155,8 +155,11 @@ def on_classical_vals(
     ) -> Dict[str, 'ClassicalValT']:
         unsigned = isinstance(self.a_dtype, (QUInt, QMontgomeryUInt))
         b_bitsize = self.b_dtype.bitsize
-        N = 2**b_bitsize if unsigned else 2 ** (b_bitsize - 1)
-        return {'a': a, 'b': int(math.fmod(a + b, N))}
+        N = 2**b_bitsize
+        if unsigned:
+            return {'a': a, 'b': int((a + b) % N)}
+        hN = N >> 1
+        return {'a': a, 'b': int(a + b + hN) % N - hN}
```
mpharrigan, line 162: h is half? can you provide a line comment explaining the logic here and consider using a more meaningful name than `hN`. Was the previous behavior using `math.fmod` wrong? can you add a test that would have failed with the old logic?

qualtran/bloqs/arithmetic/subtraction.py
```diff
@@ -45,10 +47,9 @@
 class Subtract(Bloq):
     r"""An n-bit subtraction gate.
 
-    Implements $U|a\rangle|b\rangle \rightarrow |a\rangle|a-b\rangle$ using $4n - 4 T$ gates.
+    Implements $U|a\rangle|b\rangle \rightarrow |a\rangle|a-b\rangle$ using $4n - 4$ T gates.
 
-    This construction uses `XGate` and `AddK` to compute the twos-compliment of `b` before
-    doing a standard `Add`.
+    This construction uses the relation `a - b = ~(~a + b)` to turn the operation into addition.
```
mpharrigan, line 52: are there any quantum references where this construction is used?

mpharrigan:
What happens if a_bitsize is equal to b_bitsize? 

This is a good example of a bloq that _should_ support symbolic decomposition; at least for sizeof(a) == sizeof(b)

qualtran/bloqs/arithmetic/subtraction.py
NoureldinYosri: @tanujkhattar you mentioned that this trick is used in some phase gradient papers, right?

tanujkhattar: From https://arxiv.org/pdf/2310.03011#page=174.23

image

anurudhp:
@NoureldinYosri do take a look at #1161 and #1162, they would help simplify this decomposition

qualtran/bloqs/arithmetic/subtraction.py
```diff
@@ -112,8 +123,18 @@ def on_classical_vals(
...
         unsigned = isinstance(self.a_dtype, (QUInt, QMontgomeryUInt))
         b_bitsize = self.b_dtype.bitsize
-        N = 2**b_bitsize if unsigned else 2 ** (b_bitsize - 1)
-        return {'a': a, 'b': int(math.fmod(a - b, N))}
+        N = 2**b_bitsize
+        if unsigned:
+            return {'a': a, 'b': int((a - b) % N)}
+        # Subtraction of signed integers can result in overflow. In most classical programming languages (e.g. C++)
+        # what happens when an overflow happens is left as an implementation detail for compiler designers.
+        # However for quantum subtraction the operation shoudl be unitary and that means that the unitary of
```
mpharrigan, line 131: should

NoureldinYosri: fixed

NoureldinYosri:
> can you address @anurudhp 's comment and maybe open an issue to track

will do when the other PR is merged

PR #1060, opened 2024-06-10
Added documentation, notebook, and build_call_graph for QFTTextBook

matthagan15:
Additional documentation and notebooks for QFT per https://github.com/quantumlib/Qualtran/issues/978 , also added a `build_call_graph` for QFTTextBook.


qualtran/bloqs/qft/qft_phase_gradient.py
```diff
@@ -18,13 +18,13 @@
...
 
-from qualtran import GateWithRegisters, QFxp, QUInt, Signature
+from qualtran import bloq_example, BloqDocSpec, GateWithRegisters, QFxp, QUInt, Signature
 from qualtran.bloqs.arithmetic.multiplication import PlusEqualProduct
 
 
 @attrs.frozen
 class QFTPhaseGradient(GateWithRegisters):
-    r"""QFT implemented using phase gradient trick. Uses O(n**2) T-gates for an n-bit register.
+    r"""QFT implemented using coherent addition into a phase gradient register as opposed to controlled rotations. Uses O(n**2) T-gates for an n-bit register.
```
mpharrigan, line 27: Thanks for the improvements. Can you re-form to get closer to the ideal format of a short, one-line summary, newline, additional information. i.e. 

```python
class QFT:
  """QFT using coherent addition into a phase gradient.

  This is in opposition to controlled rotations. It uses $n^2$ T gates, etc, etc
  """

File: qualtran/bloqs/qft/qft_text_book.py [hunk 2]
```diff
@@ -11,22 +11,44 @@
...
+from qualtran.resource_counting import BloqCountT, SympySymbolAllocator
 from qualtran.symbolics import SymbolicInt
+from qualtran.symbolics.types import is_symbolic
 
 
 @attrs.frozen
 class QFTTextBook(GateWithRegisters):
-    r"""Textbook version of QFT using $\frac{N (N + 1)}{2}$ controlled rotations.
+    r"""Quantum Fourier Transform as presented in Chapter 5.1 of Nielsen and
+    Chuang
```
mpharrigan, line 36: can this fit on one line? we do 100 chars

qualtran/bloqs/qft/qft_text_book.py
```diff
@@ -11,22 +11,44 @@
...
 
+    Performs the standard QFT on a register of `bitsize` qubits utilizing
+    `bitsize` Hadamards and `bitsize` * (`bitsize` - 1) / 2 controlled Z
+    rotations, along with a reversal of qubit ordering specified via
+    `with_reverse` which defaults to `True`. `bitsize` can be provided numerically or symbolically. More specific QFT implementations
+    can be found:
```
mpharrigan, line 42: wrap this text so it looks normal in the python files

qualtran/bloqs/qft/qft_text_book.py
```diff
@@ -56,3 +81,22 @@ def decompose_from_registers(
...
+
+    def build_call_graph(self, ssa: SympySymbolAllocator) -> Set['BloqCountT']:
+        ret = {
+            (Hadamard(), self.bitsize),
+            (PhaseGradientUnitary(self.bitsize - 1, exponent=0.5, is_controlled=True)),
+        }
+
+        if self.with_reverse:
+            ret |= {(TwoBitSwap(), self.bitsize // 2)}
+        return ret
```
mpharrigan, line 93: 👍

mpharrigan (commented):
couple minor things

qualtran/bloqs/qft/qft_text_book.py
```diff
@@ -56,3 +81,22 @@ def decompose_from_registers(
         if self.with_reverse:
             for i in range(self.bitsize // 2):
                 yield cirq.SWAP(q[i], q[-i - 1])
+
+    def build_call_graph(self, ssa: SympySymbolAllocator) -> Set['BloqCountT']:
+        ret = {
+            (Hadamard(), self.bitsize),
+            (PhaseGradientUnitary(self.bitsize - 1, exponent=0.5, is_controlled=True)),
```
mpharrigan, line 88: this needs to be tuple of `(bloq, n)`. This is missing `n`.

mpharrigan:
You can use `check/format-incremental --apply` `check/pylint` and `check/mypy` and `check/pytest` to see and fix the ci checks locally

qualtran/bloqs/qft/qft_text_book.py
matthagan15: whoops yeah I got carried away trying to figure it out from "PhaseGradientUnitary". The current implementation uses PhaseGradientUnitary, which I don't think is a controlled rotation, but I think the number of those bloqs should be `bitsize` - 1?

qualtran/bloqs/qft/qft_text_book.py
matthagan15: fixed

qualtran/bloqs/qft/qft_text_book.py
matthagan15: fixed

qualtran/bloqs/qft/qft_phase_gradient.py
matthagan15: fixed

matthagan15:
I had run the formatter locally, not sure why it passed on my machine but not the CI? Also I tried running each of the other checks locally but each one failed for a different reason.

qualtran/bloqs/qft/qft_text_book.py
```diff
@@ -48,6 +70,9 @@ def decompose_from_registers(
         self, *, context: cirq.DecompositionContext, q: NDArray[cirq.Qid]  # type: ignore[type-var]
     ) -> Iterator[cirq.OP_TREE]:
         yield cirq.H(q[0])
+        if self.bitsize == 1:
+            yield cirq.H(q[0])
+            return
```
tanujkhattar, line 75: Why do we need this ?

qualtran/bloqs/qft/qft_text_book.py
```diff
@@ -56,3 +81,25 @@ def decompose_from_registers(
         if self.with_reverse:
             for i in range(self.bitsize // 2):
                 yield cirq.SWAP(q[i], q[-i - 1])
+
+    def build_call_graph(self, ssa: SympySymbolAllocator) -> Set['BloqCountT']:
+        ret = {
+            (Hadamard(), self.bitsize),
+            (
+                PhaseGradientUnitary(self.bitsize - 1, exponent=0.5, is_controlled=True),
+                self.bitsize - 1,
```
tanujkhattar, line 90: This would be incorrect when the bloq is not parameterized. If `self.bitsize` is parameterized, we can return a single bloq `self.bitsize-1` times, but if thats not the case we should return `self.bitsize-1` different bloqs, each with count 1. 

We've followed this approach in other bloqs, like `LPRSInterimPrep.build_call_graph`. For the non-symbolic case we should aim the `build_call_graph` to return accurate counts and we can make some approximations for the symbolic case where its harder to express such sequence of bloqs.

tanujkhattar (changes requested):
Left a couple of blocking comments for changes in `QFTTextBook`

qualtran/bloqs/qft/qft_text_book.py
matthagan15: good point, I was mostly following the `qft_phase_gradient` pattern and figured the explicit early return wouldnt hurt. I will remove it though since its unnecessary.

qualtran/bloqs/qft/qft_text_book.py
matthagan15: this is very helpful, the non-symbolic case makes sense I will go ahead and update that but I'm not sure what kind of approximation you mean for the symbolic case. Would a worst case count be sufficient?

matthagan15:
Just updated this PR with CI fixes, let me know if anything else needs to be addressed!

qualtran/bloqs/qft/qft_text_book.py
mpharrigan: I think "approximation" here just means you don't know the exact parameters/attributes for the callee bloqs

mpharrigan:
@tanujkhattar are your review comments addressed?

tanujkhattar:
Looking at it again now

qualtran/bloqs/qft/qft_phase_gradient.py
```diff
@@ -18,24 +18,23 @@
...
 
 
 @attrs.frozen
 class QFTPhaseGradient(GateWithRegisters):
-    r"""QFT implemented using phase gradient trick. Uses O(n**2) T-gates for an n-bit register.
-
-    Given an n-bit phase gradient state $|\phi\rangle$ prepared as
+    r"""QFT implemented using coherent addition into a phase gradient register
 
+     A variant of the Quantum Fourier Transform (QFT) that utilizes an additional register provided in a phase gradient state to switch controlled rotations to coherent additions. Given an ancilla register prepared in the state
```
tanujkhattar, line 29: Each individual line should be <= 100 characters long. We should probably start enforcing this as part of the pylint (cc @mpharrigan) but this has been part of our style guide. Can you please add some line breaks to make sure we don't have lines with > 100 characters in the docstrings? This gets taken care of for code automatically as part of the auto formatting.

qualtran/bloqs/qft/qft_phase_gradient.py
```diff
@@ -18,24 +18,23 @@
...
     $$
-        |\phi\rangle = \frac{1}{\sqrt{2^{n}}} \sum_{k=0}^{2^{n} - 1} \omega_{n}^{-k} |k\rangle
+        \frac{1}{\sqrt{2^{n}}} \sum_{k=0}^{2^{n} - 1} \omega_{n}^{-k} |k\rangle,
     $$
+    then coherent addition from the system into the ancilla applies the same phase that the controlled rotation in textbook QFT does. This reduces the number of T-gates to O(`bitsize`^2) and requires no additional arbitrary rotations beyond the one time ancilla preparation cost.
```
tanujkhattar, line 33: nit: "O(`bitsize`^2)" should probably be $\mathcal{O}(n^2)$ where `n` is the `bitsize`.

qualtran/bloqs/qft/qft_phase_gradient.py
```diff
@@ -18,24 +18,23 @@
...
 
-    Phase gradient rotations can be synthesized via additions into the phase gradient register.
-    This leads to significant reductions in T/Toffoli complexity and requires 0 arbitrary
-    rotations (given a one-time cost to prepare the gradient register). See the linked reference
-    for more details.
+    The size of the ancilla register is important, if the ancilla has less
+    qubits than the system register then the accuracy of the QFT applied will be
+    reduced. This implementation assumes an ancilla with `bitsize` qubits. See `ApproximateQFT` for an implementation or the linked reference for details.
```
tanujkhattar, line 37: As we start going through revamping the docstrings, lets make it a habit to include a "Costs" section that specifies formulas for 
- Qubit cost
- T / Toffoli cost
- Total gate cost

Can you please add this section to the two bloqs refactored in this PR?

qualtran/bloqs/qft/qft_text_book.py
```diff
@@ -12,21 +12,40 @@
...
 
 
 @attrs.frozen
 class QFTTextBook(GateWithRegisters):
-    r"""Textbook version of QFT using $\frac{N (N + 1)}{2}$ controlled rotations.
+    r"""Standard Quantum Fourier Transform from Nielsen and Chuang
 
+    Performs the QFT on a register of `bitsize` qubits utilizing
+    `bitsize` Hadamards and `bitsize` * (`bitsize` - 1) / 2 controlled Z
+    rotations, along with a reversal of qubit ordering specified via
```
tanujkhattar, line 36: Should we just use shorter symbols like `n`, `m` etc. in the docstring description, especially because it makes it easier to express formulas for costs in terms of these symbols?

qualtran/bloqs/qft/qft_text_book.py
```diff
@@ -12,21 +12,40 @@
...
+from qualtran.symbolics.types import is_symbolic
 
 
 @attrs.frozen
 class QFTTextBook(GateWithRegisters):
-    r"""Textbook version of QFT using $\frac{N (N + 1)}{2}$ controlled rotations.
+    r"""Standard Quantum Fourier Transform from Nielsen and Chuang
 
+    Performs the QFT on a register of `bitsize` qubits utilizing
+    `bitsize` Hadamards and `bitsize` * (`bitsize` - 1) / 2 controlled Z
```
tanujkhattar, line 35: nit: Use math expression instead of "`bitsize` * (`bitsize` - 1) / 2 "

qualtran/bloqs/qft/qft_text_book.py
```diff
@@ -12,21 +12,40 @@
...
+    rotations, along with a reversal of qubit ordering specified via
+    `with_reverse` which defaults to `True`. `bitsize` can be provided numerically or symbolically.
+    More specific QFT implementations can be found:
+    - `ApproximateQFT` does not apply as small phases as standard QFT and relies on a specified rotation accuracy cutoff.
```
tanujkhattar, line 39: "as small phases as" sounds bit awkward. Maybe refactor to something like "ignores small angle rotations"

qualtran/bloqs/qft/qft_text_book.py
```diff
@@ -56,3 +75,28 @@ def decompose_from_registers(
...
+                ret |= {(PhaseGradientUnitary(i, exponent=0.5, is_controlled=True), 1)}
+        if self.with_reverse:
+            ret |= {(TwoBitSwap(), self.bitsize // 2)}
+        return ret
+
+
+@bloq_example
+def _qft_text_book() -> QFTTextBook:
+    qft_text_book = QFTTextBook(3)
+    return qft_text_book
```
tanujkhattar, line 99: Please also add a symbolic bloq example and include it in the generated notebook. Make sure we are able to gate costs for the symbolic example (which should work now) 

Add tests in `qualtran/bloqs/qft/qft_text_book_test.py` that verify the gate costs for symbolic and non-symbolic example and that gate costs from decomposition match the gate costs from the bloq for the non-symbolic case

qualtran/bloqs/qft/qft_phase_gradient.py
```diff
@@ -75,3 +74,14 @@ def decompose_from_registers(
         if self.with_reverse:
             for i in range(self.bitsize // 2):
                 yield cirq.SWAP(q[i], q[-i - 1])
+
+
+@bloq_example
+def _qft_phase_gradient_small() -> QFTPhaseGradient:
+    qft_phase_gradient_small = QFTPhaseGradient(3)
+    return qft_phase_gradient_small
+
```
tanujkhattar, line 83: Ditto: Add symbolic example and add bloq_autotester tests for both the examples.

tanujkhattar (commented):
Left another round of comments. Looking good overall

qualtran/bloqs/qft/qft_phase_gradient.py
mpharrigan: an admirable goal, but let's open an issue and do this in a follow up to be clear on what's expected

qualtran/bloqs/qft/qft_text_book.py
matthagan15: fixed

qualtran/bloqs/rotations/phase_gradient.py
```diff
@@ -94,6 +111,10 @@ def decompose_from_registers(
         ctrl = quregs.get('ctrl', ())
         gate = CZPowGate if self.is_controlled else ZPowGate
         for i, q in enumerate(quregs['phase_grad']):
+            print(
+                "[decompose_from_registers] gate: ",
+                gate(exponent=self.exponent / 2**i, eps=self.eps / self.bitsize),
+            )
```
tanujkhattar, line 117: Please remove debug print

qualtran/bloqs/rotations/phase_gradient.py
```diff
@@ -67,6 +69,21 @@ class PhaseGradientUnitary(GateWithRegisters):
...
+        exponent: $t$ in the above expression for $\omega_{n, t}$, a multiplicative factor
+            for the phases applied to each state. Defaults to 1.0.
+        is_controlled: `bool` which determines if the unitary is controlled via a `ctrl` register.
+        eps: The precision for the underlying `ZPowGates`.
+
+    Costs:
+        qubits: 0 ancilla qubits are allocated.
+        T-gates: Only uses 1 T gate explicitly but does rely on more costly Z rotations.
+        rotations: Uses $n$ rotations of varying angles. The lowest accuracy rotation is
+            a single T gate and the most accurate rotation has an angle of $1/2^n$.
```
tanujkhattar, line 86: Replace `accuracy` with `rotation angle` ?

tanujkhattar (approved):
LGTM % some nits

qualtran/bloqs/rotations/phase_gradient.py
matthagan15: 🤦‍♂️ sorry about that

qualtran/bloqs/rotations/phase_gradient.py
matthagan15: made the sentence symmetric so it makes more sense.

qualtran/bloqs/rotations/phase_gradient.py
```diff
@@ -67,6 +69,21 @@ class PhaseGradientUnitary(GateWithRegisters):
...
         phase_grad: A THRU register which the phase gradient is applied to.
+        (optional) ctrl: A THRU register which specifies the control for this gate. Must have
+            `is_controlled` set to `True` to use this register.
+
+    Arguments:
+        bitsize: The number of qubits of the register being acted on
+        exponent: $t$ in the above expression for $\omega_{n, t}$, a multiplicative factor
+            for the phases applied to each state. Defaults to 1.0.
+        is_controlled: `bool` which determines if the unitary is controlled via a `ctrl` register.
+        eps: The precision for the underlying `ZPowGates`.
```
tanujkhattar, line 80: Actually, the `eps` passed here is for the state that's being prepared. The `eps` passed to the underlying `ZPowGate` is `eps / self.bitsize` because you apply `self.bitsize` rotations and you want to bound the total error to be less than `eps`.

matthagan15: true, updated.

qualtran/bloqs/rotations/phase_gradient.ipynb
```diff
@@ -54,7 +54,20 @@
...
     "#### Registers\n",
-    " - `phase_grad`: A THRU register which the phase gradient is applied to. \n",
+    " - `phase_grad`: A THRU register which the phase gradient is applied to.\n",
+    " - `(optional) ctrl`: A THRU register which specifies the control for this gate. Must have `is_controlled` set to `True` to use this register. \n",
+    "\n",
+    "#### Parameters\n",
+    " - `bitsize`: The number of qubits of the register being acted on\n",
+    " - `exponent`: $t$ in the above expression for $\\omega_{n, t}$, a multiplicative factor for the phases applied to each state. Defaults to 1.0.\n",
+    " - `is_controlled`: `bool` which determines if the unitary is controlled via a `ctrl` register.\n",
+    " - `eps`: The precision for the underlying `ZPowGates`. \n",
```
tanujkhattar, line 64: I think this file needs to be regenerated

PR #1061, opened 2024-06-10
Changes to `.flatten`

mpharrigan:
- Default predicate to flatten everything
 - If you can't flatten a bloq, don't complain; just don't do anything
 - Documentation improvements


fdmalone (approved):
I had forgotten flatten existed!

mpharrigan:
it's about to play a pivotal role in new tensor simulation

PR #1066, opened 2024-06-11
Fix CNOT.get_ctrl_system transposition error

### LINKED ISSUE #1065: Bug in `CNOT.get_ctrl_system`
There is a transposition error on the two control bits resulting in very wrong circuits if you decompose a `Controlled(...)` that contain cnots. 

The fix is simple. I'm investigating why I didn't catch this immediately

mpharrigan:
Fixes #1065 

`get_ctrl_system` is responsible for tracking which are the new soquets introduced to control the bloq and which are the existing ones. Since CNOT already has a control bit, we weave in the new control bit into the two bits in the `ctrl` register in `Toffoli`. Seemingly, at the beginning of the method I wanted to plug the new control bit into the second entry in the ctrl register but at the end of the method I thought I used the first entry in the ctrl register.

This is not an invalid thing to do for a general `CompositeBloq` but it is wrong for wiring up a Controlled(CNOT()). I added a unit test. The first assertion always passed. The second two previously failed. One needs to exercise `Controlled([cbloq containing CNOT]).decompose_bloq()` to see the bad behavior.


qualtran/bloqs/basic_gates/cnot.py
```diff
@@ -117,7 +117,7 @@ def add_controlled(
                 (c1, c2), target = bb.add(
                     bloq, ctrl=np.array([in_soqs['ctrl'], c2]), target=in_soqs['target']
                 )
-                return (c1,), (c2, target)
```
fdmalone, line 120: It's confusing that the ctrl is ordered with the new control "below" the original cnot control, or am I being stupid?

fdmalone (approved):

qualtran/bloqs/basic_gates/cnot.py
mpharrigan: It shouldn't matter as long as it's consistent -- which it wasn't. but now it is!

fdmalone: does the diagram look correct, like if it was off-on-trg, vs on-off-trg

mpharrigan: I changed it so that the "new" control bit is the first one. In either case, the diagram is topologically correct. Whether it shows up in the order you expect I couldn't say

Issue #873, opened 2024-04-15
More Leaf Bloqs


mpharrigan:
One of the fundamental principles of qualtran (and programming languages in general) is composing basic operations to make composite operations that expresses a more complex behavior. Users can call `Bloq.decompose_bloq()` to turn a bloq into its composite parts. Bloq-authors can override `Bloq.build_composite_bloq` to define a bloq in terms of its decomposition.

Above, I snuck in the terms "basic operations" and "[more] complex behavior". Very often, the 'direction' of decomposition is clear: usually our subbloqs operate on fewer qubits, correspond to fewer (quantum) clock cycles to execute on a reasonable target architecture, or are more general-purpose and appear in more than one decomposition. 

As we get closer to the bottom of the decomposition hierarchy, uncertainty about the target architecture can cause problems when writing a decomposition. Do we need to compile our `And` bloq to T or do we have a CCZ factory and we should keep them as toffolis? When compiling a toffoli to T gates, do we want to minimize depth or minimize qubits? Questions like this come up throughout the stack (i.e. what sort of adder do I want to use); but these issues become much more acute at the bottom of the stack.

There's clearly a bit of artistry here: but in a compiling/resource estimation paper: if you're swapping out e.g. different adders that goes into the "compiling" part, but if you're switching your Toffolis, that goes in the resource estimation part. I propose

 1. Making more bloqs "leaf" bloqs that raise `DecomposeTypeError` if you try to decompose them. Specifically: I want to remove decompositions from `Toffoli`, `And`, `TwoBitCSwap` and most $\le 2$ qubit gates. These gates should be defined by their tensors/name rather than a decomposition. 
    - [x] And #1346 
    - [x] Toffoli https://github.com/quantumlib/Qualtran/issues/1386
    - [x] TwoBitCSwap https://github.com/quantumlib/Qualtran/issues/1387
    - [ ] other things in `QECGatesCost` `GateCount`: clifford, rotation, measurement 
 3. Make sure the resource estimation framework can handle different "architecture" decision. For example: the default gate counts will have to return counts for all leaf bloqs. The cost model should handle multiplying the toffoli count by 4 or 7 when using a T factory. Helper functions should be in place. Edit: ✅  this is done
 4. The constructions for e.g. using 4 T gates to do an And gate are really useful! These should be kept. Maybe on bespoke methods like `decompose_to_t()` etc.

mpharrigan:
@tanujkhattar see the above plan for what I was discussing today

[...]

PR #1070, opened 2024-06-13
Tensor Simulation protocol 2: keep indices factorized

### LINKED ISSUE #793: Quimb tensor contraction is significantly slower than cirqs state vector simulation

mpharrigan:
Fixes #793 

### The problem

In Qualtran, we try to keep everything at as-high-a-level of abstraction as possible and recurse "one layer at a time" as necessary. This is better for most of the protocols (resource counting, classical simulation, diagrams). When doing a full quantum/tensor simulation we need everything to be as factored as possible for optimal performance. This is because the tensor contraction engine can find good ways of contracting the small tensors that can potentially avoid making any huge intermediate tensors.

The problem is with splits and joins. When we flatten out a high-level circuit into its constituent parts; there are a plurality of these bookkeeping operations. The cost protocols and to-cirq protocols can ignore these or handle them elegantly. But the tensor simulation engine uses a third party package: quimb which would require special support for these fancy identity operations. Currently (prior to this PR), we mapped each soquet to a quimb index. This makes a lot of conceptual sense; and indeed quimb can support indices of any size (i.e. bitsize > 1 wires). But something like a `Split` bloq would have to have one index of size `2**n` which would completely block any intelligent tensor contraction ordering if e'er one was encountered. 

### Rejected (for now) solution

One solution would have been to flatten everything into a circuit and run some sort of graph-rewriting algorithm to fuse or otherwise minimize split/join pairs. This might be desirable for other reasons; but would have been complicated to get working right.

### The solution in this PR

The solution in this PR is to take a NISQ-style, per-qubit, factorized-indices description of our tensor networks. Each tensor network index used to be a `Soquet`. Now it is a `(Connection, int)` pair where the integer `j` says which qubit of the `reg.dtype.num_qubits`-sized register this ndarray dimension is. 

I've taken the liberty of identifying quimb indices with `Connection` instead of `Soquet`, which fixes a bit of a conceptual mistake from the original implementation. 

The new Split tensor method should be ellucidating

```python
        eye = np.eye(2)
        incoming = incoming['reg']
        outgoing = cast(NDArray, outgoing['reg'])
        return [
            qtn.Tensor(data=eye, inds=[(outgoing[i], 0), (incoming, i)], tags=[str(self)])
            for i in range(self.dtype.num_qubits)
        ]
```

where it was once one `Tensor` with 1 large index coming in and n one-bit indices going out, it is now `n` 1-bit identity matrices. 

If your `Bloq` only has `QBit()` inputs and outputs, the extra index will always be `0`, e.g.

```python
        return [
            qtn.Tensor(
                data=_PAULIX, inds=[(outgoing['q'], 0), (incoming['q'], 0)], tags=[str(self)]
            )
        ]
```

### Flattening

To get good tensor performance, you ought to flatten your circuit. `Bloq.tensor_contract()` will now flatten the circuit, turn it to a tensor network, and contract it. 

It used to contract each subbloq recursively, which is not performant. There used to be a function called `flatten_for_tensor_contraction` that would use the correct predicate to flatten to the correct level. No one really knew about this or used it. After #1061 , the behavior now is to flatten as much as possible. All the bloqs in the flattened representation must have explicit tensors.

**Notable consequence:** If your bloq has both a decomposition and a `my_tensors` method, the `my_tensors` method will not be used by default. You'd have to use the `qualtran.simulation.tensor` functions yourself if you don't want to flatten all the way before embarking on tensor contraction.

The general advice is to _not_ implement `my_tensors` and rather define a decomposition if you want to simulate your bloq. Leaf bloqs need to implement `my_tensors` but the rate that they get added to the library is likely slow. 

### Done and todo

 - [x] tests pass
 - [x] #793 performance is fast
 - [x] demo showing high-bitsize one-amplitude tensor contraction
 - [x] Are there other states I should be benchmarking?
 - [x] Compare test durations before and after
 - [x] rename `cbloq_to_quimb_2` back to `cbloq_to_quimb`
 - [ ] syntactic sugar for getting statevector assuming starting in the zero state -- for follow-up


mpharrigan:
This is ready for review. I split it into two commit -- one is changes to the protocol; the other is moving all the bloqs+ to the new protocol

mpharrigan:
@tanujkhattar can you take a look when you have a chance

tanujkhattar:
Some high level comments about the approach 

> Notable consequence: If your bloq has both a decomposition and a my_tensors method, the my_tensors method will not be used by default. You'd have to use the qualtran.simulation.tensor functions yourself if you don't want to flatten all the way before embarking on tensor contraction.

Have you contrasted this with the strategy where we try to flatten till the point we hit the first node that has its `my_tensors` defined? Is there clear evidence that this strategy of flattening everything is better? Should we open an issue to track this comparison for future? 


> The general advice is to not implement my_tensors and rather define a decomposition if you want to simulate your bloq. Leaf bloqs need to implement my_tensors but the rate that they get added to the library is likely slow.

One use case for defining `my_tensors` is to verify correctness of the decomposition - do something like `bloq.tensor_contract() == bloq.decompose_bloq().tensor_contract()`. This is similar to how we use `bloq.call_classically() == cbloq.call_classicall()`  to verify decomposition via classical actions. The fact that these two statements imply different things for two similar protocols (classical and quantum simulation) would be confusion at best. Is there anyway to achieve this kind of verification with this new strategy? Should we add a predicate / flag to `bloq.tensor_contract` to control the flattening behavior so that, by default, we flatten only till we hit a bloq that has `my_tensors` defined and users can specify a `bloq.tensor_contract(flatten=True)` OR call `bloq.decompose_bloq().flatten().tensor_contract()` (which is still fairly straightforward and single line) to achieve the "flatten before contraction" ? 


> The general advice is to not implement my_tensors and rather define a decomposition if you want to simulate your bloq.

Directly implementing the decomposition is probably a higher bar. The way I imagine the workflow to be is the for many bloqs, people would first implement `my_tensors`

qualtran/bloqs/basic_gates/global_phase.py
```diff
@@ -49,6 +64,33 @@ def decompose_bloq(self) -> 'CompositeBloq':
...
+
+    def get_ctrl_system(
+        self, ctrl_spec: Optional['CtrlSpec'] = None
+    ) -> Tuple['Bloq', 'AddControlledT']:
+
+        # Default logic for more than one control.
+        if not (ctrl_spec is None or ctrl_spec == CtrlSpec()):
+            return super().get_ctrl_system(ctrl_spec=ctrl_spec)
+
+        exponent: float = np.real_if_close(np.log(self.coefficient) / (np.pi * 1.0j)).item()  # type: ignore
```
tanujkhattar, line 82: `self.coefficient` can be symbolic, in which case this line would throw an error instead of returning a symbolic `ZPowGate`. 

While this is a useful change, Why do we need to make this change as part of this PR? Is it because of the cirq-interop changes in the `Controlled` bloq? Can we pull these out in a separate PR?

qualtran/bloqs/basic_gates/su2_rotation.py
```diff
@@ -106,28 +106,18 @@ def from_matrix(mat: NDArray[np.complex_]) -> 'SU2RotationGate':
...
+                inds=[(outgoing['q'], 0), (incoming['q'], 0)],
+                tags=[str(self)],
             )
-        )
-
-    def _unitary_(self):
-        if self.is_symbolic():
-            return None
-        return self.rotation_matrix
+        ]
```
tanujkhattar, line 120: Ditto: Looks related to cirq-interop changes in `Controlled` gate that can be pulled out in a separate PR.

qualtran/bloqs/bookkeeping/join.py
```diff
@@ -81,27 +81,18 @@ def adjoint(self) -> 'Bloq':
...
-                tags=['Join', tag],
-            )
-        )
+        eye = np.eye(2)
+        incoming = cast('NDArray', incoming['reg'])
+        outgoing = outgoing['reg']
+        return [
+            qtn.Tensor(data=eye, inds=[(outgoing, i), (incoming[i], 0)], tags=[str(self)])
+            for i in range(self.dtype.num_qubits)
+        ]
```
tanujkhattar, line 95: When implementing the `my_tensors` method, a pattern like this looks super confusing. The indices now have two types of indexing going on - indexing into a numpy array if `self.reg.shape` to get the corresponding `Connection` object and then an implicit indexing into an implicit array of size `self.bitsize` by specifying a tuple `(cxn, i)`

I think this calls for adding a new type (equivalent to a qubit) so that the `ConnectionT` for a register `reg` would now be a numpy array of shape `reg.shape + (reg.bitsize,)`. This has the following advantages:
- It gets rid of an implicit vs explicit indexing. The implicit indexing is confusing at best.
- The concept aligns with the concept of a qubit - we just say that each entry in the incoming and outgoing dictionary corresponds to a qubit. 
-  This kind of `full shape` is already used as a concept in cirq style API of `decompose_from_registers` and would so it'll unify the tensor contraction API nicely with that API. 


Another alternative would be to keep `ConnectionT` a numpy array of shape `reg.shape` but override the `__getitem__` method on `Connection` object so `cxn[i]` returns a new object which is essentially a dataclass / tuple storing (cxn, i) and can be used as a tensor index (like `TensorIndex(cxn, i)`)

In general, it would be much nicer if the call sites look like `inds=[outgoing[i], incoming[i][0]]` instead of what it is right now. 

What do you think?

qualtran/cirq_interop/_cirq_to_bloq.py
```diff
@@ -239,6 +229,39 @@ def _add_my_tensors_from_gate(
     tn.add(qtn.Tensor(data=unitary, inds=outgoing_list + incoming_list, tags=[short_name, tag]))
 
 
+def _my_tensors_from_gate(
```
tanujkhattar, line 232: Looks like the `_add_my_tensors_from_gate` from gate is not removed but we've added ` _my_tensors_from_gate` -- is there a need to keep both?

tanujkhattar (changes requested):
Left a round of comments and questions.

qualtran/_infra/composite_bloq.py
```diff
@@ -538,7 +546,39 @@ def _cxn_to_soq_dict(
     return soqdict
 
 
-def _get_dangling_soquets(signature: Signature, right=True) -> Dict[str, SoquetT]:
+def _cxn_to_cxn_dict(
```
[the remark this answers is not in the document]
mpharrigan: the return value is keyed by the register name. The function derives it from the `cxns` argument, primarily. This (admitadly silly) name was chosen to keep parity with the above helper function `_cxn_to_soq_dict`. I can rename that one as well, but probably something like
 
 - `cxns_to_soq_dict`
 - `cxns_to_cxn_dict` 

and/or do you want me to change the docstring? It says "keyed" by register name

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -125,28 +121,6 @@ def dtype(self):
     def signature(self):
         return Signature([Register("a", self.a_dtype), Register("b", self.b_dtype)])
 
-    def add_my_tensors(
```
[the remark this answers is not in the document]
mpharrigan: wouldn't that be better in a unit test?

qualtran/cirq_interop/_cirq_to_bloq.py
mpharrigan: must have missed the removal, thanks!

qualtran/simulation/tensor/_dense.py
```diff
@@ -12,55 +12,136 @@
...
+    """
+
+    inds: List[Tuple[Connection, int]] = []
+    for reg in signature.rights():
+        for idx in reg.all_idxs():
+            for j in range(reg.dtype.num_qubits):
+                if idx:
+                    inds.append((outgoing[reg.name][idx], j))  # type: ignore[index]
+                else:
+                    inds.append((outgoing[reg.name], j))  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
mpharrigan: can you be more specific

qualtran/bloqs/arithmetic/addition.py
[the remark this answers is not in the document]
tanujkhattar: Maybe, but I don't see a need to remove it from bloqs that already have it defined. 

Have you benchmarked what happens if you specify the tensors explicitly in `my_tensors` and stop decomposing if a bloq has the tensor defined; vs if you flatten out everything? Now that we have an index for every qubit (unlike the situation before where had larger indices for soquets and there was a lot of splits and joins happening); I don't see why flattening out to the lowest level should be faster than stopping at a point where you have the tensor data available already. 

If its not a significant speedup in performance; I think keeping consistency with the classical simulation workflow would be very nice.

qualtran/simulation/tensor/_dense.py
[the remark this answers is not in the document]
tanujkhattar: Why do we need the `if` / `else` block ? Probably something to do with numpy indexing but I am forgetting the details so it'll be good to add a note.

mpharrigan:
> Have you contrasted this with the strategy where we try to flatten till the point we hit the first node that has its my_tensors defined? Is there clear evidence that this strategy of flattening everything is better? Should we open an issue to track this comparison for future?

This could be an interesting thing to explore. It would be an easy change by passing a predicate to the flattening call. With the following considerations

 - It's kindof hard to figure out if a bloq overrides `my_tensors`. We could do the thing where we check if the method's qualname starts with `"Bloq."` or not
 - "better" is a loaded term. For arbitrary tensor network topologies, it's better to keep things factorized in the sense that one can be smarter about contraction ordering. If you have a low number of qubits and a deep circuit within a decomposition you can very easily spend more time making the tensor network and finding a contraction ordering than just keeping things more dense to begin with

qualtran/bloqs/arithmetic/addition.py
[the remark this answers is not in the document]
mpharrigan: the example of simulating the addition of 2x32 bit integers wouldn't work.

I agree that with most other protocols we keep things as high-level as possible. It's always better to let python handle the addition of two numbers than to emulate the CPU ourselves. The tensor simulation protocol is fundamentally different due to its inherent exponential scaling

mpharrigan:
> Directly implementing the decomposition is probably a higher bar. The way I imagine the workflow to be is the for many bloqs, people would first implement my_tensors during initial verification and testing and then implement a decomposition if that becomes a bottleneck; especially for a lot of the classical action style bloqs like arithmetic, data loading etc. where action and tensor data is significantly easier to specify.

For classical-action-style bloqs I agree. It's one of my priorities to implement the tensors-from-classical-action helper.

There are some bloqs that are naturally "tensor first" like leaf bloqs or state prep maybe. Those you can write tensors for. For bloqs higher-up in the call graph, I'm not sure there are many cases where it's more natural to express the tensor than the decomposition

tanujkhattar:
> This could be an interesting thing to explore. It would be an easy change by passing a predicate to the flattening call. With the following considerations

From a users perspective, if the default behavior we choose is to decompose only till the point where a bloq has its tensors defined; then a user can do:

```python
> bloq.tensor_contract() : Use the tensor data at the highest level possible
> bloq.as_composite_bloq().flatten().tensor_contract() : Use the tensor data at the lowest level possible
``` 

Both the workflows are well supported with a single line of code. If we choose the default to go as deep as possible; the bloq authors need to:

1. Write a custom predicate, which is potentially error prone for a user to get right, to determine "go as low as possible"
2. Use the `cbloq_to_quimb` and `quimb_to_dense` APIs which are non-standard and harder to discover. 



With these considerations in mind, I'll suggest its better to change the default behavior and support both the workflows.

qualtran/simulation/tensor/_dense.py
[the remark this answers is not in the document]
mpharrigan: yes, this is the normal thing we have to do to handle both an obj or an array of obj if the register has a non-`()` shape. Would it be clearer if I changed the if condition to `if reg.shape` or `if reg.shape != ()`?

mpharrigan:
> One use case for defining my_tensors is to verify correctness of the decomposition - do something like bloq.tensor_contract() == bloq.decompose_bloq().tensor_contract(). This is similar to how we use bloq.call_classically() == cbloq.call_classicall() to verify decomposition via classical actions. The fact that these two statements imply different things for two similar protocols (classical and quantum simulation) would be confusing at best. Is there anyway to achieve this kind of verification with this new strategy?

A good unit test has the computed quantity validated against a reference value that we're pretty sure is correct. Once again: this means that we'd need non-leaf bloqs where the tensor data is somehow a natural expression of the bloq. This is the case for the classical action bloqs, but really -- the test should just use the classical action tests!

As part of this PR, I went through every bloq that had defined tensors. There were two cases where the bloq had both tensors and a decomposition (excepting some that will be leaf bloqs #873 )

 - `Add`: classical action; should just use that for testing. The existing tensors are wrong for negative values. The unfactorized form prevents efficient contraction of high-bit low-depth circuits
 - `IntState`: classical action; really just a test of the `int_to_bits` function. The unfactorized form _really_ prevents efficient contraction since this is a bunch of kets

There are no tests of the form `bloq.tensor_contract() == bloq.flatten().tensor_contract()` or cirq equivalent.

qualtran/bloqs/basic_gates/global_phase.py
mpharrigan: The new `Controlled.my_tensors` method does not work with GlobalPhase since its "tensor" is just a number (with the rest of the cirq changes reverted)

mpharrigan: This is because I'm using `subbloq.tensor_contract()` which returns a number instead of an array. Since this is the public-facing function, I don't want to change its behavior. Instead, I've put a check in `Controlled.my_tensors` to handle the case where the subbloq's tensor is just a number

qualtran/_infra/composite_bloq.py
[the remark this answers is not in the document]
mpharrigan: made these changes

qualtran/_infra/controlled.py
```diff
@@ -419,36 +409,35 @@ def add_my_tensors(
...
-                cirq.ControlledGate(self.subbloq, control_values=self.ctrl_spec.to_cirq_cv())
-            )
-        if all(reg.side == Side.THRU for reg in self.subbloq.signature):
-            # subbloq has only THRU registers, so the tensor contraction corresponds
-            # to a unitary matrix.
-            return self.tensor_contract()
-        # Unable to determine the unitary effect.
-        return NotImplemented
+        n = self.signature.n_qubits()
+        return self._tensor_data().reshape(2**n, 2**n)
```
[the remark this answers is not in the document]
mpharrigan: reverted

qualtran/bloqs/basic_gates/global_phase.py
mpharrigan: reverted and did the special case described above

qualtran/bloqs/basic_gates/su2_rotation.py
mpharrigan: reverted

qualtran/bloqs/bookkeeping/join.py
mpharrigan: There is three levels of indexing: by register name, optionally within a "shaped" register, and then the sub-register bit-indexing. 

I don't think it's worth adding an extra dimension to all the arrays where the last column is just the numbers 0..n. 

`(incoming['ctrl'], 0)` is not meaningfully more confusing than `incoming['ctrl'][0]` where indeed the latter syntax just maps to `TensorIndex(incoming['ctrl'], 0)` anyways

qualtran/simulation/tensor/_dense.py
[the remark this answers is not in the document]
mpharrigan: added a comment

tanujkhattar:
> IntState: classical action; really just a test of the int_to_bits function. The unfactorized form really prevents efficient contraction since this is a bunch of kets

Wouldn't the `my_tensors` implementation of `IntState` now just return a list of tensors instead of one big tensor? The unfactorized form vs factorized form problem here is now solved by changing the way we represent the indices and not necessarily by removing the implementation of `my_tensors` and flattening first ? 

> Both the workflows are well supported with a single line of code.. With these considerations in mind, I'll suggest its better to change the default behavior and support both the workflows.

Following my comment above, it's better to support two things than one. Please consider making this change.

qualtran/_infra/composite_bloq.py
```diff
@@ -66,6 +66,11 @@
 canonicalize and return `SoquetT`.
 """
 
+_ConnectionType = TypeVar('_ConnectionType', bound=np.generic)
+
+ConnectionT = Union[Connection, NDArray[_ConnectionType]]
```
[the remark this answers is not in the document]
dstrain115: NDArray is typed so that anything in an NDArray has to be a np.generic.  So, if you have NDArray[X], then X has to be a np.generic.  Otherwise, mypy generates a type error.  This issue happens when you jam something that's not a numpy type into a numpy array.

I don't know anything about _ConnectionType.

mpharrigan:
[d850862](https://github.com/quantumlib/Qualtran/pull/1070/commits/d85086233da09c84b89cbfa491139b5642b3c342) restores the flattening logic with a boolean flag. By default, we'll flatten as far as possible. This can be changed if there's ever an example where this is not desired

qualtran/cirq_interop/_cirq_to_bloq.py
```diff
@@ -204,39 +193,37 @@ def _wire_symbol_from_gate(
...
-    unitary = tensor_data_from_unitary_and_signature(cirq.unitary(gate), signature)
-    incoming_list = [
-        *itertools.chain.from_iterable(
-            [np.array(incoming[reg.name]).flatten() for reg in signature.lefts()]
+
+    if any(reg.side != Side.THRU for reg in signature):
+        raise ValueError(
+            f"CirqGateAsBloq.my_tensors is only supported for "
+            f"gates with thru registers. Found {gate}."
         )
```
tanujkhattar, line 216: This restriction was not present in the original `_add_my_tensors_from_gate` method; so it'll be nice to not add this restriction as part of this large scale refactoring. The `CirqGateAsBloq` specific errors should be raised in the implementation of `CirqGateAsBloq`; not in the implementation of ` _my_tensors_from_gate`. 

The original implementation works by first assuming that all registers are THRU registers and then extracting tensor data from the correct subspace whenever a register is not a THRU register. This is done by `tensor_data_from_unitary_and_signature` -- the method is still implemented but not used anymore here. 

Can you please revert the restricting changes to this method make sure the more general behavior is maintained? This method is called not just from `CirqGateAsBloq` but also from `GateWithRegisters`, that may have directional registers.

tanujkhattar (commented):
LGTM % one final comment about reverting the behavior of `my_tensors_from_gate` and support directional registers as was done previously.

qualtran/cirq_interop/_cirq_to_bloq.py
mpharrigan: this should really have a test if it's behavior we're relying on. I'll revert the changes, though

tanujkhattar (approved):

Issue #1073, opened 2024-06-13
CompositeBloq in a call graph


mpharrigan:
Right now, the call graph can contain `CompositeBloq`s as nodes. 

 - composite bloqs can be quite large and quite expensive to hash and check equality, which are operations we assume are very cheap when building and manipulating a call graph.
 - It's strange to have a composite bloq inside a call graph -- that means someone has written a decomposition that has a composite bloq within it. Usually our decompositions are made out of other, concrete bloqs. If you think of bloqs as functions and compositebloq as "anonymous functions" aka lambdas, maybe this should be a supported use case but it's still strange.
 - It makes sense to me to have a compositebloq as the root node of a call graph. This is important for `CompositeBloq` to follow the `Bloq` interface and you can imagine using this to create a union of call graphs of bloqs by wrapping them in `CompositeBloq` first. There are three ways to support this
    - status quo: put the composite bloq as the root node. Hash it hopefully a limited number of times. Cache the hash of a composite bloq
    - use a sentinel value like `None` as the root node for this special case
    - Return a networkx graph that has `n` root nodes; one for each bloq instance in the composite bloq

anurudhp:
> It's strange to have a composite bloq inside a call graph

Are there any existing bloqs where this happens? or are you just saying in general?

mpharrigan:
I don't think we have any right now in the library. Prior to #1074, a code path would be triggered where a CompositeBloq would be the root node of the call graph

mpharrigan:
A very similar thing is causing the confusion in #1097 where the user expects to be able to use a `CompositeBloq` as a root node of a recursive dag. In that, we can either

 - have all the `CostKey` support `CompositeBloq`; pay the cost of caching composite bloqs 
 - only allow `CompositeBloq` at the root call; special case it and omit it from the cache

mpharrigan:
In https://github.com/quantumlib/Qualtran/issues/1097#issuecomment-2200882072, @tanujkhattar argues for full support of `CompositeBloq` including in decompositions.

My thoughts: Pro:

 - This is "morally" right and can theoretically be well-supported with the cached hash strategy alluded to in the comment
 - I think of `CompositeBloq` analogous to anonymous functions ("lambdas"). These are broadly useful in classical programming even though we don't use them in the existing bloq library yet

Considerations:
 - It's always easier to disallow something and enable it later than vice versa
 - We need more testing of examples where there is indeed a `CompositeBloq` in the call graph

Issue #845, opened 2024-04-02
`bloq.call_graph()` should preserve ordering of successors across calls / `build_call_graph` should return a Dict instead of Set


tanujkhattar:
Every time you call `bloq.call_graph()` you can get a different `nx.DiGraph` where the ordering of successors of a node is different. This happens because `build_call_graph` returns a `Set` instead of a `Dict` so iterating over a set to get the call graph destroys ordering. 

I encountered this when writing tests for https://github.com/quantumlib/Qualtran/pull/732 where the output of `get_flame_graph_data` is different across calls because the DFS ordering of nodes in `bloq.call_graph` is different. 

@mpharrigan What's the reason to destroy ordering? Is this by design? Can we preserve ordering? I agree that most properties of the call graph analysis shouldn't depend upon the ordering but I think it's still nice to preserve the order across calls.

mpharrigan:
Are sets not ordered in Python? either a list of 2-tuples or a dict would be fine. I think the point of the "set" was to encode that order doesn't matter. Perhaps a more important property is that each bloq appears only once, which isn't captured by set. 

It's possible that networkx will shuffle things later anyways, so if you want determinism I suggest throwing in some `sorted`s.

tanujkhattar:
> Are sets not ordered in Python?

Nope. Dicts are, but sets aren't. I vote for a dict instead of a list of 2-tuples.

> Perhaps a more important property is that each bloq appears only once, which isn't captured by set.

That's true, and a dict would capture that as well. 

> It's possible that networkx will shuffle things later anyways

I think it's less likely if we just traverse the graph and not modify it. But we can deal with any potential indeterminism due to networkx when we encounter it. AFAIK, this shouldn't be an issue.

anurudhp:
@mpharrigan should the bloq author ensure that each unique bloq in the `BloqCountT` occurs only once? i.e. is something like this valid:
```py
counts = {}
if condition_1: counts.add((XGate(), 1))
if condition_2: counts.add((XGate(), 2))
return counts
```

This luckily works as long as the frequencies are different, but fails if they are equal.

fdmalone:
^^^ I've been caught out by this before

mpharrigan:
@anurudhp I think that's what this issue is about. If we change it to use dictionaries, you'd have to do something like

```python
counts = defaultdict(lambda: 0)
if condition_1: counts[XGate()] += 1
if condition_2: counts[XGate()] += 2
return counts
```

[...]

PR #1074, opened 2024-06-13
More direct bloq_counts()

mpharrigan:
Get bloq counts directly without going via the call graph. This can massively speed up `qlt_testing.assert_equivalent_bloq_counts`, which will necessarily call `CompositeBloq.bloq_counts` which will involve lots of hashing of the composite bloq xref #1073 .

This also has prompted me to finally deal with what happens if you have two of the same bloq in your list of BloqCountT, perhaps returned by `build_call_graph`. Indeed, bloq_counts returns a dictionary so we have to add those keys. This was also necessary to make sure that if two things get generalized to the same result, their counts get added into one entry. Dealing with this is a pre-requisite for #845

### DIFF
--- a/qualtran/_infra/bloq.py
+++ b/qualtran/_infra/bloq.py
@@ -360,8 +360,9 @@ class Bloq(metaclass=abc.ABCMeta):
         Returns:
             A dictionary mapping subbloq to the number of times they appear in the decomposition.
         """
-        _, sigma = self.call_graph(generalizer=generalizer, max_depth=1)
-        return sigma
+        from qualtran.resource_counting import get_bloq_callee_counts
+
+        return dict(get_bloq_callee_counts(self, generalizer=generalizer))
 
     def get_ctrl_system(
         self, ctrl_spec: Optional['CtrlSpec'] = None
--- a/qualtran/resource_counting/_call_graph.py
+++ b/qualtran/resource_counting/_call_graph.py
@@ -16,7 +16,7 @@
 
 import collections.abc
 from collections import defaultdict
-from typing import Callable, Dict, List, Optional, Sequence, Set, Tuple, Union
+from typing import Callable, cast, Dict, List, Optional, Sequence, Set, Tuple, Union
 
 import networkx as nx
 import sympy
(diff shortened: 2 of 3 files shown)


fdmalone (approved):

qualtran/_infra/bloq.py
```diff
@@ -360,8 +360,9 @@ def bloq_counts(
         Returns:
             A dictionary mapping subbloq to the number of times they appear in the decomposition.
         """
-        _, sigma = self.call_graph(generalizer=generalizer, max_depth=1)
-        return sigma
+        from qualtran.resource_counting import get_bloq_callee_counts
+
+        return dict(get_bloq_callee_counts(self, generalizer=generalizer))
```
tanujkhattar, line 365: We should change the return type to a `Mapping` instead of a dictionary so we can also return things like a `Counter`, which is more natural in this case.

tanujkhattar (commented):
To make sure I understand, we still plan to change the return type of `build_call_graph` to a `Mapping[Bloq, Union[int, sympy.Expr]]` to address https://github.com/quantumlib/Qualtran/issues/845 but that'll be done in follow-up PRs, correct?

qualtran/resource_counting/_call_graph.py
```diff
@@ -76,19 +76,19 @@ def _generalize_callees(
...
-        callee_counts.append((generalized_callee, n))
-    return callee_counts
+        callee_counts[generalized_callee] += n
+    return list(callee_counts.items())
 
 
 def get_bloq_callee_counts(
     bloq: 'Bloq',
-    generalizer: Optional['GeneralizerT'] = None,
+    generalizer: Optional[Union['GeneralizerT', Sequence['GeneralizerT']]] = None,
```
tanujkhattar, line 91: Should we take the opportunity to change `get_bloq_callee_counts` and ` _generalize_callees` to a `Mapping` / `Dict` as well? The conversion from Dict -> List -> Dict seems unnecessary.

qualtran/_infra/bloq.py
mpharrigan: I thought the general advice is to be specific with your return types. Here, it's actually a dictionary

qualtran/resource_counting/_call_graph.py
mpharrigan: let's do that all in the same pr

Issue #1080, opened 2024-06-17
287 slow tests aren't slow


mpharrigan:
Currently, we have 372 unit tests marked slow. On my macbook, 287 of them take less than 1s to run

<details>

time | test
----|----
0.04s call | qualtran/bloqs/arithmetic/hamming_weight_test.py::test_hamming_weight_compute[3]
0.54s call | qualtran/bloqs/arithmetic/hamming_weight_test.py::test_hamming_weight_compute[4]
0.15s call | qualtran/bloqs/chemistry/sparse/sparse_test.py::test_sparse_costs_against_openfermion[12-4]
0.30s call | qualtran/bloqs/chemistry/sparse/sparse_test.py::test_sparse_costs_against_openfermion[16-3]
0.09s call | qualtran/bloqs/chemistry/sparse/sparse_test.py::test_sparse_costs_against_openfermion[8-3]
0.57s call | qualtran/bloqs/data_loading/qrom_test.py::test_qrom_1d_full[0-data0]
0.32s call | qualtran/bloqs/data_loading/qrom_test.py::test_qrom_1d_full[0-data1]
0.73s call | qualtran/bloqs/data_loading/qrom_test.py::test_qrom_1d_full[1-data1]
0.34s call | qualtran/bloqs/data_loading/qrom_test.py::test_qrom_multi_dim_full[0-data1]
0.54s call | qualtran/bloqs/data_loading/qrom_test.py::test_qrom_multi_dim_full[1-data1]
0.11s call | qualtran/bloqs/data_loading/select_swap_qrom_test.py::test_select_swap_qrom[0-data0]
0.08s call | qualtran/bloqs/data_loading/select_swap_qrom_test.py::test_select_swap_qrom[0-data1]
0.02s call | qualtran/bloqs/data_loading/select_swap_qrom_test.py::test_select_swap_qrom[0-data2]
0.46s call | qualtran/bloqs/data_loading/select_swap_qrom_test.py::test_select_swap_qrom[1-data0]
0.82s call | qualtran/bloqs/data_loading/select_swap_qrom_test.py::test_select_swap_qrom[1-data1]
0.12s call | qualtran/bloqs/data_loading/select_swap_qrom_test.py::test_select_swap_qrom[None-data0]
0.08s call | qualtran/bloqs/data_loading/select_swap_qrom_test.py::test_select_swap_qrom[None-data1]
[...]

PR #1081, opened 2024-06-17
Unmark some slow tests that aren't actually slow

mpharrigan:
Low-hanging fruit from #1080

### DIFF
--- a/qualtran/bloqs/data_loading/select_swap_qrom_test.py
+++ b/qualtran/bloqs/data_loading/select_swap_qrom_test.py
@@ -31,7 +31,6 @@ from qualtran.resource_counting.t_counts_from_sigma import t_counts_from_sigma
 from qualtran.testing import assert_valid_bloq_decomposition
 
 
-@pytest.mark.slow
 @pytest.mark.parametrize(
     "data,block_size",
     [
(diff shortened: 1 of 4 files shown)


qualtran/bloqs/data_loading/select_swap_qrom_test.py
```diff
@@ -31,15 +31,14 @@
...
-@pytest.mark.slow
 @pytest.mark.parametrize(
     "data,block_size",
     [
         pytest.param(
             data,
             block_size,
             id=f"{block_size}-data{didx}",
-            marks=pytest.mark.slow if block_size == 2 and didx == 0 else (),
+            marks=pytest.mark.slow if block_size == 1 and didx == 1 else (),
```
tanujkhattar, line 41: This is probably also not slow ?

tanujkhattar (approved):

qualtran/bloqs/data_loading/select_swap_qrom_test.py
mpharrigan: It's >1s on my machine, but less than my personal 2s threshold for marking... but the other configurations should provide enough coverage for the `-m 'not slow'` executions of the unit tests.

qualtran/bloqs/hamiltonian_simulation/hamiltonian_simulation_by_gqsp_test.py
```diff
@@ -83,10 +82,9 @@ def verify_hamiltonian_simulation_by_gqsp(
     assert_matrices_almost_equal(expected_top_left, actual_top_left, atol=1e-4)
 
 
-@pytest.mark.slow
 @pytest.mark.parametrize("select_bitsize", [1])
 @pytest.mark.parametrize("target_bitsize", [1, 2])
-@pytest.mark.parametrize("t", [2, 3, 5])
+@pytest.mark.parametrize("t", [2, 3, pytest.param(5, marks=pytest.mark.slow)])
```
anurudhp, line 87: oh nice! didn't know this feature

mpharrigan: me neither

that's also why I left in the skip from the above test that tanuj noted -- it's a nice example of how to skip particular configurations :)

Issue #951, opened 2024-05-13
Unclear how to use symbolics


mstechly:
I tried to create some bloqs with symbolic costs and it's unclear how exactly to do that.

Here's the numeric version:

```
from qualtran.bloqs.arithmetic.comparison import LessThanEqual
bloq = LessThanEqual(10, 15)
bloq.t_complexity()
```

This works well and returns: `TComplexity(t=96, clifford=533, rotations=0)`

However if I try the following:

```
from qualtran.bloqs.arithmetic.comparison import LessThanEqual
import sympy
a, b = sympy.symbols("a b")
bloq = LessThanEqual(a, b)
bloq.t_complexity()
```

I'm getting: `TypeError: cannot determine truth value of Relational`

Perhaps things are working properly and I'm just using them in a wrong way, but I also wanted to give you some feedback on how it looks like from a user's perspective :)  From reading the docs and tests I wasn't able to figure out how to make it work – I didn't see tests for symbolics for this bloq.

mpharrigan:
Yes, only some bloqs support symbolic parameters but there's no way of (reliably) knowing which. Some fixes that we need

 - [ ] write a general docs page about symbolics support and how it's up to the bloq to make it work or not
 - [ ] update type annotations and arg docstrings to make it clear which parameters can be symbolic
 - [ ] make more things support symbolic parameters

mstechly:
Can you tell me which bloqs do support it, just so I have some examples to work with for now :) ?

tanujkhattar:
qrom supports it after https://github.com/quantumlib/Qualtran/pull/945

mstechly:
Thanks @tanujkhattar, now this works properly!

```
import sympy
N, M, b1, b2, c = sympy.symbols('N M b1 b2 c')
qrom_symb = QROM.build_from_bitsize((N, M), (b1, b2), num_controls=c)
qrom_symb.t_complexity()
```

`TComplexity(t=4*M*N + 4*c - 8, clifford=M*N*b1*b2 + 13*M*N + 13*c - 26, rotations=0)` 
:) 

However,  while I can decompose a numeric bloq, I can't decompose a symbolic one:

```
import sympy
N, M, b1, b2, c = sympy.symbols('N M b1 b2 c')
qrom_symb = QROM.build_from_bitsize((N, M), (b1, b2), num_controls=c)
qrom_symb.decompose_bloq()
```

```
[...]

PR #1084, opened 2024-06-18
Support symbolic decomposition for state prep via alias sampling

tanujkhattar:
Fixes https://github.com/quantumlib/Qualtran/issues/951

In general, the cirq-interop decomposition assumes access to individual bits. So anytime we would do `bb.split` / `bb.join` in the decomposition, its fine to use the cirq style decomposition as well but if we are not doing any splits or joins (example: for higher level bloqs like state prep alias sampling) its useful to migrate the cirq style decomposition to bloq style decomposition so we can support decomposing bloqs with symbolic register sizes. 

More such migrations would follow as we discover bloqs which currently have cirq-style `decompose_from_registers` implemented but can benefit from a symbolic decomposition. An alternate strategy would be to update the cirq-interop code to support symbolics wherever we can.


qualtran/bloqs/state_preparation/state_preparation_alias_sampling.py
```diff
@@ -209,21 +209,31 @@ def qrom_bloq(self) -> QROM:
...
-            return
-        selection, less_than_equal = quregs['selection'], quregs['less_than_equal']
-        sigma_mu, alt, keep = quregs['sigma_mu'], quregs['alt'], quregs['keep']
-        yield cirq.H.on_each(*sigma_mu)
-        yield self.qrom_bloq.on_registers(selection=selection, target0_=alt, target1_=keep)
-        yield LessThanEqual(self.mu, self.mu).on(*keep, *sigma_mu, *less_than_equal)
-        yield CSwap.make_on(ctrl=less_than_equal, x=alt, y=selection)
+            return soqs
+        selection, less_than_equal = soqs['selection'], soqs['less_than_equal']
+        sigma_mu, alt, keep = soqs['sigma_mu'], soqs['alt'], soqs['keep']
```
mpharrigan, line 219: please do this unpacking in the method signature

tanujkhattar: It'll be annoying to create the return dictionary twice when we want to return early. When there are these many parameters, its more convenient to work with the dictionary directly ?

qualtran/bloqs/state_preparation/state_preparation_alias_sampling_test.py
```diff
@@ -53,6 +53,8 @@ def test_state_prep_alias_sampling_symb():
     # Symbolic T-counts
     symb_t_counts = int(expected_t_count_expr.subs({L: N, sympy.Symbol(r"\epsilon"): epsilon}))
     np.testing.assert_allclose(concrete_t_counts, symb_t_counts, rtol=1e-4)
+    # Ensure the symbolic bloq can decomposes into a composite bloq
+    _ = bloq.decompose_bloq()
```
mpharrigan, line 57: `qlt_testing.assert_valid_bloq_decomposition`

tanujkhattar: done

qualtran/bloqs/state_preparation/state_preparation_alias_sampling.py
mpharrigan: oh I see

qualtran/bloqs/state_preparation/state_preparation_alias_sampling_test.py
```diff
@@ -53,6 +53,8 @@ def test_state_prep_alias_sampling_symb():
     # Symbolic T-counts
     symb_t_counts = int(expected_t_count_expr.subs({L: N, sympy.Symbol(r"\epsilon"): epsilon}))
     np.testing.assert_allclose(concrete_t_counts, symb_t_counts, rtol=1e-4)
+    # Ensure the symbolic bloq can decomposes into a composite bloq
+    assert_valid_bloq_decomposition(bloq)
```
anurudhp, line 57: also call `bloq_autotester` for the symbolic example?

tanujkhattar:
Something is going wrong with the cirq interop again due to which cirq simulation tests for reflection using prepare are failing. This is probably related to the few other places we've seen this happen recently. I'll have to spend more time to figure out exactly whats going wrong.

qualtran/bloqs/state_preparation/state_preparation_alias_sampling.py
mpharrigan: do we need to support the mu = 0 case? what is mu? can you update the docstring to the standard format with an Args: section and a Registers: section. I think the existing Signature section is out-of-date

tanujkhattar: Yes, `self.mu` case needs to be supported. `mu` is a parameter which depends on the precision and can be 0 for when precision is low (eg: probability_epsilon=0.5)

The docstring does seem to be updated and the signature of the bloq hasn't changed since it was written. I'll polish it a bit once I fix the failing tests.

mpharrigan: what do you mean? there is no args section and the register names don't match the register names

anurudhp:
Possible cause: `GreedyQubitManager`, even with `maximize_reuse=True`, does not reuse qubits in a bloq style decomposition. For example see the outputs below (for a decomp. with two chained allocate/free pairs):

Script
```py
from typing import Iterator

import cirq
from attrs import frozen
from qualtran import Bloq, BloqBuilder, GateWithRegisters, Signature, SoquetT
from qualtran.bloqs.basic_gates import CNOT
from qualtran.cirq_interop.testing import GateHelper


@frozen
class MultiAlloc(Bloq):
    @property
        return Signature.build(q=1)

    def build_composite_bloq(
        self, bb: "BloqBuilder", q: "SoquetT"
    ) -> dict[str, "SoquetT"]:
        for _ in range(2):
            a = bb.allocate(1)
            a, q = bb.add(CNOT(), ctrl=a, target=q)
            bb.free(a)

        return {"q": q}


class MultiAllocGWR(GateWithRegisters):
    @property
        return Signature.build(q=1)

    def decompose_from_registers(
    ) -> Iterator[cirq.OP_TREE]:
        (q,) = quregs["q"]
        for _ in range(2):
            (a,) = context.qubit_manager.qalloc(1)
            yield CNOT().on(a, q)
            context.qubit_manager.qfree([a])


def visualize(bloq: Bloq | GateWithRegisters):
    print(bloq)
    gate = GateHelper(
        gate=bloq,
        context=cirq.DecompositionContext(
            cirq.GreedyQubitManager(prefix="_greedy", maximize_reuse=True)
        ),
    )
    print(gate.circuit)
    print()
    print(gate.decomposed_circuit)
    print()


if __name__ == "__main__":
    visualize(MultiAlloc())
    visualize(MultiAllocGWR())
```

output
```
MultiAlloc
q: ───q───

_greedy_0: ───alloc────@───free─────────────────────────
                       │
_greedy_1: ───alloc────┼────────────────────@───free────
                       │                    │
q: ───────────Y^-0.5───@───Y^0.5───Y^-0.5───@───Y^0.5───

MultiAllocGWR
q: ───MultiAllocGWR───

_greedy_0: ───@───@───
              │   │
q: ───────────X───X───
```

tanujkhattar:
Yup, this is known. See https://github.com/quantumlib/Qualtran/pull/963#issue-2301621495

qualtran/bloqs/state_preparation/state_preparation_alias_sampling.py
tanujkhattar: updated the docstring

tanujkhattar:
@mpharrigan This is ready for a review

qualtran/bloqs/state_preparation/state_preparation_alias_sampling.py
```diff
@@ -235,21 +233,31 @@ def qrom_bloq(self) -> QROM:
...
-        *,
-        context: cirq.DecompositionContext,
-        **quregs: NDArray[cirq.Qid],  # type:ignore[type-var]
-    ) -> Iterator[cirq.OP_TREE]:
-        yield PrepareUniformSuperposition(self.n_coeff).on(*quregs['selection'])
+    def build_composite_bloq(self, bb, **soqs):
+        soqs['selection'] = bb.add(
+            PrepareUniformSuperposition(self.n_coeff), target=soqs['selection']
+        )
         if self.mu == 0:
```
anurudhp, line 240: After the fix in #1092 I think `mu` would never be `0`. So the decomp can be simplified by unpacking the registers in the method signature

qualtran/bloqs/state_preparation/state_preparation_alias_sampling_test.py
```diff
@@ -89,6 +89,8 @@ def test_state_prep_alias_sampling_symb():
         )
     )
     np.testing.assert_allclose(concrete_t_counts, symb_t_counts, rtol=5e-4)
+    # Ensure the symbolic bloq can decomposes into a composite bloq
+    assert_valid_bloq_decomposition(bloq)
```
anurudhp, line 93: bloq autotest missing

qualtran/bloqs/state_preparation/state_preparation_alias_sampling.py
anurudhp: Should also probably assert `mu > 0` in post init

anurudhp (approved):
lgtm % minor fixes

mpharrigan:
@tanujkhattar do you want to merge this

qualtran/bloqs/state_preparation/state_preparation_alias_sampling_test.py
tanujkhattar: done

qualtran/bloqs/state_preparation/state_preparation_alias_sampling.py
tanujkhattar: done

tanujkhattar:
I've addressed the nits and also added an assertion to in post init to verify that we only do state prep over 1D register. Opened https://github.com/quantumlib/Qualtran/issues/1280 to track improvements

PR #1085, opened 2024-06-18
Move QSP/GQSP helper code into `qualtran.linalg`

anurudhp:
Moved helper functions that apply more generally to polynomials. As we now have two ways (and more incoming) to compute QSP complementary polynomials, it's better to keep the generic functions separate from the specific bloqs.


tanujkhattar (approved):

qualtran/linalg/polynomial/qsp_util.py
mpharrigan, line 1: I never like "util" in a module name. `linalg/polynomial/qsp.py`? or `qsp_polynomial`? Or factor out all the testing functions into `linalg.polynomial.qsp_testing`

qualtran/linalg/testing.py
```diff
@@ -0,0 +1,20 @@
...
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+import numpy as np
+from numpy.typing import NDArray
+
+
+def assert_matrices_almost_equal(A: NDArray, B: NDArray, *, atol: float = 1e-5):
```
mpharrigan, line 18: docstring? worth mentioning that this is based on the norm which is unlike e.g. `np.testing.assert_almost_equal`

anurudhp: added docstring

qualtran/linalg/polynomial/qsp_util.py
anurudhp: renamed to `qsp_testing`, and moved one non-testing function out to `linalg/polynomial/basic.py`

PR #1086, opened 2024-06-19
Add `AutoPartition` and `BloqBuilder.add_and_partition` to fit bloqs together whose registers don't quite match

charlesyuan314:
Completes the implementation of the wrapper bloq introduced in #789 and adds an alternative interface to it as a `BloqBuilder` utility method.


charlesyuan314:
FYI @fdmalone

mpharrigan:
One of the motivations in the linked PR was to clean up the circuit diagrams by hiding the partition/unpartition pairs behind a level of decomposition. Does this still do that?

charlesyuan314:
I made `AutoPartition` be its own bloq again and added tests and documentation.

qualtran/bloqs/bookkeeping/auto_partition.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+@frozen
+class AutoPartition(Bloq):
+    """Wrap a bloq with `Partition` to fit an alternative set of input and output registers
+       such that splits / joins can be avoided in diagrams.
+
+    Args:
+        bloq: The bloq to wrap.
+        partition: A sequence of pairs specifying each register that the wrapped bloq should accept
```
mpharrigan, line 43: should this be a plural name like `paritions` or `partitionings`? If I'm understanding it correctly, each entry says how a given "outer" register gets mapped to the multiple "inner" registers, right?

qualtran/bloqs/bookkeeping/auto_partition.py
```diff
@@ -0,0 +1,116 @@
...
+            Otherwise, the output registers will follow `bloq.signature.rights()`.
+
+    Registers:
+        [user_spec]: The output registers of the wrapped bloq.
+    """
+
+    bloq: Bloq
+    partition: Sequence[Tuple[Register, Sequence[Register]]] = field(
+        converter=lambda s: tuple((r, tuple(rs)) for r, rs in s)
+    )
```
mpharrigan, line 55: would it be worth having some sort of validation that the list of first entries in this matches bloq.lefts() ?

qualtran/bloqs/bookkeeping/auto_partition.py
```diff
@@ -0,0 +1,116 @@
...
+                )
+            )
+
+    def pretty_name(self) -> str:
+        return self.bloq.pretty_name()
+
+    def build_composite_bloq(self, bb: BloqBuilder, **soqs: SoquetT) -> Dict[str, SoquetT]:
+        parts: Dict[str, Partition] = dict()
+        in_regs: Dict[str, SoquetT] = dict()
+        for out_reg, (_, bloq_regs) in zip(self.signature.lefts(), self.partition):
```
mpharrigan, line 76: is `out_reg` supposed to be equal to the first entry in the self.parition tuple, here ignored with `_`?

qualtran/bloqs/bookkeeping/auto_partition.py
```diff
@@ -0,0 +1,116 @@
...
+class AutoPartition(Bloq):
+    """Wrap a bloq with `Partition` to fit an alternative set of input and output registers
+       such that splits / joins can be avoided in diagrams.
+
+    Args:
+        bloq: The bloq to wrap.
+        partition: A sequence of pairs specifying each register that the wrapped bloq should accept
+            and the registers from `bloq.signature.lefts()` that concatenate to form it.
+        partition_output: If True, the output registers will also follow `partition`.
+            Otherwise, the output registers will follow `bloq.signature.rights()`.
```
mpharrigan, line 46: what's the motivation for this flag? I would have assumed the `False` setting would be most useful so you could fully wrap a bloq's factorized registers

qualtran/bloqs/bookkeeping/auto_partition.py
```diff
@@ -0,0 +1,116 @@
...
+                parts[soq_name].adjoint(),
+                **{reg.name: bloq_out_regs[reg.name] for reg in parts[soq_name].signature.rights()},
+            )
+        return out_regs
+
+    def build_call_graph(self, ssa: SympySymbolAllocator) -> Set[BloqCountT]:
+        return self.bloq.build_call_graph(ssa)
+
+    def _t_complexity_(self) -> TComplexity:
+        return self.bloq._t_complexity_()
```
mpharrigan, line 96: you shouldn't need this override, I think

qualtran/bloqs/bookkeeping/auto_partition.py
charlesyuan314: This flag is not important if we're only wrapping Bloqs with the same left and right registers. It could be useful in cases where we only want to insert a `Partition` on the left side but keep the original signature of the bloq on the right side. In that case, `partition_output` should be False. The default case, where both sides are wrapped, is True.

qualtran/_infra/composite_bloq_test.py
```diff
@@ -556,6 +556,25 @@ def test_t_complexity():
...
+
+    bb = BloqBuilder()
+    bloq = Controlled(Swap(3), CtrlSpec(qdtypes=QUInt(4), cvs=0b0110))
+    a = bb.add_register_from_dtype('a', QAny(7))
+    b = bb.add_register_from_dtype('b', QAny(3))
+    assert a is not None
+    assert b is not None
+    ctrl_r, x_r, y_r = bloq.signature.lefts()
+    a, b = bb.add_and_partition(
+        bloq, [(Register('a', QAny(7)), [y_r, ctrl_r]), (Register('b', QAny(3)), [x_r])], a=a, b=b
```
mpharrigan, line 571: Maybe this would cause problems, but what do you think about an API where we can use the names of the factorized registers to set up the auto-partition. 

```python
bb.add_and_partition(CSwap(), {Register('a', QAny(7)): ['y', 'ctrl']. Register('b', QAny(3): ['x']}, a=a, b=b)
```

qualtran/bloqs/bookkeeping/auto_partition.py
charlesyuan314: I think the appropriate invariant is that the concatenation of the second entries in `partition` matches `bloq.lefts()`. I'll add this validation.

qualtran/_infra/composite_bloq_test.py
charlesyuan314: I think this works. I'll change it.

qualtran/bloqs/bookkeeping/auto_partition.py
charlesyuan314: In this version of the code they are equal. I will remove the redundant `zip`.

qualtran/bloqs/bookkeeping/auto_partition.py
charlesyuan314: Yes, I will change to `partitions`.

qualtran/bloqs/bookkeeping/auto_partition.py
mpharrigan: Oh, I read the documentation and variable name backwards. I think of the "partitioned" form of registers as being the many, split-up registers and the unpartitioned form being the single register. So when I read `partition_output = True`, I assume the output will be left in the partitioned state. 

Maybe a completely different flag toggle like `left_only = False` or something like that. It's probably also useful to capture what you wrote in the above comment ^ in the docstring somewhere, i.e. that (full) wrapping only works for consistent left&right registers

mpharrigan (commented):
Very comprehensive! Thanks for adding docs as well. LGTM modulo wording fixes on the flag-for-one-sidedness

charlesyuan314:
@mpharrigan PTAL!

mpharrigan (approved):
sweeeeeeeeet

PR #1088, opened 2024-06-21
Serialize `sympy.codegen.cfunctions.log2`

anurudhp:
This is used by `qualtran.symbolics.math_funcs.log2`.


anurudhp:
@mpharrigan ptal

mpharrigan (approved):

PR #1089, opened 2024-06-21
Implement decomposition for `BitonicSort`

anurudhp:
Decomposition for sorting lists whose size is a power of 2. Partially addresses https://github.com/quantumlib/Qualtran/issues/219

Described briefly in https://www.nature.com/articles/s41534-018-0071-5, circuit based on [wiki](https://en.wikipedia.org/wiki/Bitonic_sorter#:~:text=at%20the%20bottom.-,Alternative%20representation,-%5Bedit%5D).


anurudhp:
@tanujkhattar @fdmalone ptal

qualtran/bloqs/arithmetic/comparison_test.py
```diff
@@ -230,7 +230,9 @@ def test_greater_than_manual():
     qlt_testing.assert_wire_symbols_match_expected(
         GreaterThanConstant(bitsize, 17), ['In(x)', '⨁(x > 17)']
     )
-    assert cbloq.t_complexity() == t_complexity(LessThanEqual(bitsize, bitsize))
+    assert cbloq.t_complexity() == (
+        t_complexity(LessThanEqual(bitsize, bitsize)) + TComplexity(clifford=1)
+    )
```
fdmalone, line 236: Can we add a correctness test for the circuit here / tensor contraction / classical simulation?

anurudhp: added some classical simulation tests. indeed caught a lot of bugs 😅. now works on random lists of size 2-16

fdmalone: Great, (make sure the tests don't take too long or else @mpharrigan will be on your case :))

qualtran/bloqs/arithmetic/sorting.ipynb
```diff
@@ -36,19 +36,20 @@
    },
    "source": [
     "## `Comparator`\n",
-    "Compare and potentially swaps two n-bit numbers.\n",
+    "Compare and potentially swaps two numbers in {0, ... L - 1}.\n",
```
fdmalone, line 39: Why the change from bitsize (which is common in qualtran) to integer range. Is it just more convenient?

qualtran/bloqs/arithmetic/sorting.py
```diff
@@ -11,118 +11,305 @@
...
+    """
+    L: SymbolicInt
+    k: SymbolicInt
+
+    def __attrs_post_init__(self):
+        k = self.k
+        if not is_symbolic(k):
+            assert k >= 1, "length of input lists must be positive"
+            # TODO support non-power-of-two input lengths
+            assert (k & (k - 1)) == 0, "length of input lists must be a power of 2"
```
fdmalone, line 177: cute

qualtran/bloqs/arithmetic/sorting.ipynb
anurudhp: papers seem to use "integers in range [n]" etc. often, so I thought this would be more useful. And having `BoundedQUInt` makes it much nicer to support this than earlier.

qualtran/bloqs/arithmetic/comparison_test.py
anurudhp: hahaha yep, total is ~0.5s, should be good

qualtran/bloqs/arithmetic/sorting_test.py
```diff
@@ -27,19 +30,54 @@ def test_bitonic_sort(bloq_autotester):
 
 
 def test_comparator_manual():
-    bloq = Comparator(4)
-    assert bloq.t_complexity().t == 88 - 4
-    with pytest.raises(NotImplementedError):
-        bloq.decompose_bloq()
+    bloq = Comparator(2**4)
+    assert bloq.t_complexity().t == 88 - 4 - 7 * 4
```
fdmalone, line 34: why the -7 * 4?

qualtran/bloqs/arithmetic/sorting_test.py
```diff
@@ -27,19 +30,54 @@ def test_bitonic_sort(bloq_autotester):
...
+
+    _ = bloq.t_complexity()
+
+
+def test_bitonic_sort_classical_sim():
+    L = 8
+    xs = np.array([4, 2, 7, 1])
+    bloq = BitonicSort(L, len(xs))
+    sorted_xs, _ = bloq.call_classically(xs=xs)
+    assert np.all(sorted_xs == sorted(xs))
```
fdmalone, line 67: sweet

qualtran/bloqs/arithmetic/sorting.py
```diff
@@ -11,118 +11,305 @@
...
     return cmp_symb
 
 
 _COMPARATOR_DOC = BloqDocSpec(bloq_cls=Comparator, examples=[_cmp_symb])
 
 
 @frozen
-class BitonicSort(Bloq):
-    r"""Sort k n-bit numbers.
+class ParallelComparators(Bloq):
```
fdmalone, line 105: Add unit tests for this and Bitonic merge, we may want to split this file which now has several bloqs, which sort of strays from the ~ 1 bloq / file ideal / WDYT @mpharrigan

qualtran/bloqs/arithmetic/sorting.py
```diff
@@ -11,118 +11,318 @@
...
+
+        junk = np.concatenate(all_junks)
+        assert len(junk) == self.num_comparisons
+
+        return {'result': result, 'junk': np.array(junk)}
+
+
+@frozen
+class BitonicSort(Bloq):
+    r"""Sort k numbers in the range {0, ..., L-1}.
```
tanujkhattar, line 247: Use brackets `[0, L)` or `[0, L]` instead of `{0, ..., L - 1}` for consistency. 

```suggestion
    r"""Sort k numbers in the range [0, ..., L-1].
```

qualtran/bloqs/arithmetic/sorting.py
```diff
@@ -11,118 +11,305 @@
...
 
     Args:
-        bitsize: Number of bits used to represent each integer.
+        L: Upper-bound (excluded) on the input integers.
         k: Number of integers to sort.
```
fdmalone, line 251: we're sort of using k to mean several different things here (each half, whole, list, ...), maybe a more descriptive variable name is in order.

qualtran/bloqs/arithmetic/sorting.py
```diff
@@ -11,118 +11,305 @@
...
-    bitsize: SymbolicInt
-    k: int
+    L: SymbolicInt
+    k: SymbolicInt
+
+    def __attrs_post_init__(self):
+        k = self.k
+        if not is_symbolic(k):
+            assert k >= 1, "length of input list must be positive"
+            # TODO support non-power-of-two input lengths
```
fdmalone, line 269: Open an issue and link it here.

qualtran/bloqs/arithmetic/sorting.py
```diff
@@ -11,118 +11,305 @@
...
+        k: Number of integers in each half
+    """
+    L: SymbolicInt
+    k: SymbolicInt
+
+    def __attrs_post_init__(self):
+        k = self.k
+        if not is_symbolic(k):
+            assert k >= 1, "length of input lists must be positive"
+            # TODO support non-power-of-two input lengths
```
fdmalone, line 182: Add link to issue here too.

qualtran/bloqs/arithmetic/sorting_test.py
```diff
@@ -27,19 +30,54 @@ def test_bitonic_sort(bloq_autotester):
...
+
+
```
fdmalone, line 69: your new bloqs should add autotesters and bloq-doc-specs too.

fdmalone (changes requested):
So nice! I'm glad someone finally did this.

qualtran/bloqs/arithmetic/sorting.ipynb
tanujkhattar: Is there an advantage in terms of complexity of sorting? I would say this notation is definitely more confusing. I had to look at the unit tests to figure out what you mean by `Sort k numbers in the range {0, ..., L-1}.` My first guess was maybe you want to sort the first `k` of the first `L` numbers or something. 

Saying something like `Sort K n-bit numbers in complexity `O()` complexity` is a lot cleaner. If there's no advantage in terms of complexity and everywhere we are using `n=bitsize(L - 1)` to determine the bitsize anyway, then I'd say lets change the bloqs to use `bitsize` of the numbers to be sorted as the attribute by default and we can provide a builder method that expects an `L` and computes the bitsize for you.

qualtran/bloqs/arithmetic/sorting.py
```diff
@@ -11,118 +11,318 @@
...
 _COMPARATOR_DOC = BloqDocSpec(bloq_cls=Comparator, examples=[_cmp_symb])
 
 
 @frozen
-class BitonicSort(Bloq):
-    r"""Sort k n-bit numbers.
+class ParallelComparators(Bloq):
+    """Given k numbers in [0, L), compare every pair that is `offset` apart.
+
+    Args:
```
tanujkhattar, line 108: Please more details to the docstring - do we compare and swap or we just store the result of the comparison somewhere? Are the RIGHT junk registers storing the result of the comparison? Or are they supposed to be completely ignored?

qualtran/bloqs/arithmetic/sorting.py
```diff
@@ -11,118 +11,318 @@
...
+
+
+@frozen
+class BitonicMerge(Bloq):
+    r"""Merge two sorted sequences of integers.
 
     Args:
-        bitsize: Number of bits used to represent each integer.
+        L: Upper-bound (excluded) on the input integers.
+        k: Number of integers in each half
```
tanujkhattar, line 172: Please specify that `k` must be a power of `2`. Also specify the complexity in terms of `T / Toffoli costs`, `ancilla cost` and `circuit depth`. 

Also, would it be a lot more work if we want to support merging two lists of different sizes?

qualtran/bloqs/arithmetic/sorting.py
```diff
@@ -11,118 +11,318 @@
...
+
+        return {'result': result, 'junk': np.array(junk)}
+
+
+@frozen
+class BitonicSort(Bloq):
+    r"""Sort k numbers in the range {0, ..., L-1}.
+
+    Args:
+        L: Upper-bound (excluded) on the input integers.
```
tanujkhattar, line 250: Please specify that `k` must be a power of `2`. Also specify the complexity in terms of `T / Toffoli costs`, `ancilla cost` and `circuit depth`.

qualtran/bloqs/arithmetic/sorting.py
```diff
@@ -11,118 +11,318 @@
...
+            # TODO support non-power-of-two input lengths
+            assert (k & (k - 1)) == 0, "length of input list must be a power of 2"
 
     @property
     def signature(self):
-        return Signature([Register("input", QAny(self.bitsize), shape=(self.k,))])
+        return Signature(
+            [
+                Register("xs", BoundedQUInt(self.bitsize, self.L), shape=(self.k,)),
+                Register("junk", QBit(), shape=(self.num_comparisons,), side=Side.RIGHT),
```
tanujkhattar, line 277: Why do we have `junk` RIGHT registers everywhere ? Are we saving on any uncomputation? Can you also expand the docstring to explain this?

qualtran/bloqs/arithmetic/sorting_test.py
```diff
@@ -27,19 +30,54 @@ def test_bitonic_sort(bloq_autotester):
...
-    with pytest.raises(NotImplementedError):
-        bloq.decompose_bloq()
+
+    bloq = BitonicSort(2**bitsize, k)
+    assert bloq.num_comparisons == 24
+
+    _ = bloq.t_complexity()
+
+
+def test_bitonic_sort_classical_sim():
```
tanujkhattar, line 62: Would be nice to have a test that tests sorting permutations for all `n!` classical inputs for a list of size `n`. This makes sure we are not missing any corner case that wouldn't get sorted as expected.

qualtran/bloqs/arithmetic/sorting_test.py
```diff
@@ -27,19 +30,54 @@ def test_bitonic_sort(bloq_autotester):
...
+    assert bloq.t_complexity() == TComplexity(t=15 * bitsize + 4, clifford=56 * bitsize + 24)
+
+
+@pytest.mark.parametrize("L", [5, 8, 12])
+def test_comparator_classical_sim(L: int):
+    bloq = Comparator(L)
+    for a in range(L):
+        for b in range(L):
+            res_a, res_b, anc = bloq.call_classically(a=a, b=b)
+            assert res_a <= res_b
```
tanujkhattar, line 49: Also assert that `res_a` and `res_b` are equal to `min(a, b)` and `max(a,b)` and `anc` stores the result of the comparison.

tanujkhattar (changes requested):
+1, great to have this implemented!

qualtran/bloqs/arithmetic/sorting.py
fdmalone: Should definitely be [0, L) for exclusive upper limit

qualtran/bloqs/arithmetic/sorting.py
anurudhp: So `BitonicSort` is the only end bloq that algorithms use. The rest are just helpers to make the decomposition more modular, I feel they really can't be used in general, other than as sub-steps of `BitonicSort`.

If the intermediate bloqs are more useful, we could split them up into their own files.

fdmalone: Ok, I think they should still be unit tested if they have implementations though.

qualtran/bloqs/arithmetic/sorting.ipynb
anurudhp: Actually, as the comparison bloqs only accept bitsize, this doesn't give any improvement. It does make more sense to just keep it as bitsize, (and provide a helper constructor that takes `L` ). I'll revert the changes and update the docs.

qualtran/bloqs/arithmetic/sorting.py
anurudhp: Yep will add unit tests for each

qualtran/bloqs/arithmetic/sorting.py
anurudhp: used bitsizes instead

qualtran/bloqs/arithmetic/sorting.ipynb
anurudhp: Used bitsizes everywhere.

qualtran/bloqs/arithmetic/sorting_test.py
anurudhp: added the three assertions

qualtran/bloqs/arithmetic/sorting_test.py
anurudhp: added a testing all permutations of `[1,2,3,4]`

qualtran/bloqs/arithmetic/sorting.py
anurudhp: This is the comparator circuit from the [reference](https://www.nature.com/articles/s41534-018-0071-5):
image

I wrote the signature similar to the `And` bloq where the result bit is `RIGHT`.

qualtran/bloqs/arithmetic/sorting_test.py
anurudhp: I had some reason to justify this, but can't remember anymore. I _think_ somewhere in the previous t_complexity it assumed CSwap was `t=14`, but the current impl is `t=7`

tanujkhattar: The paper assumes CSWAP is `t=4` using 1 ancilla, right? We should probably open an issue to track that though. Textbook CSWAP should be `t=7`. 


CSWAP to 1 Toffoli can be reduced as follows
image


and then the Toffoli can be implemented using 4 T either using a Relative phase toffoli or using an ancilla with AND gate.

qualtran/bloqs/arithmetic/sorting.py
tanujkhattar: I think its worth highlighting this in the docstring -- i.e. the junk stores the result of the comparisons. This would be necessary to make the sorting bloq reversible so you can apply the unitary in reverse and get back to the original sequence of unsorted elements if needed.

qualtran/bloqs/arithmetic/sorting.py
anurudhp: done

qualtran/bloqs/arithmetic/sorting.py
anurudhp: done

qualtran/bloqs/arithmetic/sorting.py
anurudhp: updated all docstrings.

qualtran/bloqs/arithmetic/sorting.py
anurudhp: updated the docstring explaining this.

qualtran/bloqs/arithmetic/sorting.py
anurudhp: It isn't too difficult, but the edge cases need to be dealt with care. I wanted to get the power-of-2 case working first, will implement the general case in a follow up.

qualtran/bloqs/arithmetic/sorting.py
anurudhp: updated to `half_length`

qualtran/bloqs/arithmetic/sorting.py
anurudhp: added simulation tests for each supporting bloq

qualtran/bloqs/arithmetic/sorting_test.py
anurudhp: added bloq_examples with autotesters and updated the bloqdocspecs

qualtran/bloqs/arithmetic/sorting.py
anurudhp: added to docstring

qualtran/bloqs/arithmetic/sorting.py
```diff
@@ -48,82 +64,331 @@ class Comparator(Bloq):
...
+class BitonicMerge(Bloq):
+    r"""Merge two sorted sequences of n-bit integers.
+
+    Given two sorted lists of length `half_length`, merges them inplace into a single
+    sorted list.
+
+    Currently only supports `half_length` equal to a power of two (#1090).
+
+    If each half has length $k$, then the merge network uses $k * (\log{k} + 1)$ comparisons
+    when $k$ is a power of 2.
```
tanujkhattar, line 200: Please also add information / $$..$$ formula to specify what the `junk` stores since it can be potentially useful depending upon the context.

qualtran/bloqs/arithmetic/sorting.py
```diff
@@ -48,82 +64,331 @@ class Comparator(Bloq):
...
+
+_COMPARATOR_DOC = BloqDocSpec(bloq_cls=Comparator, examples=[_comparator, _comparator_symb])
+
+
+@frozen
+class ParallelComparators(Bloq):
+    """Given k n-bit numbers, for each pair that is `offset` apart, compare and swap if needed to make them ordered.
+
+    For each block of `2 * offset` numbers, apply a `Comparator` between each pair that is `offset` apart.
+    Uses up to $k / 2$ Comparators.
```
tanujkhattar, line 109: Please add a short note about what the junk stores

qualtran/bloqs/arithmetic/sorting.py
```diff
@@ -48,82 +64,331 @@ class Comparator(Bloq):
...
-        for the moment. Issue #219
+    where $y_1, y_2, \ldots, y_k = \mathrm{sorted}(x_1, x_2, \ldots, x_k$, and the junk register
+    stores the result of comparisons done during the sorting. Note that the junk register will
+    be entangled with the input list register.
+
+    Currently only supports $k$ being a power of two (#1090).
+
+    The bitonic sorting network requires $\frac{k}{2} \frac{\log{k} (\log{k} + 1)}{2}$ total comparisons,
+    and has depth $\frac{\log{k} (\log{k} + 1)}{2}$, when $k$ is a power of 2. Each comparison generates
+    one ancilla qubit, so the total size of junk register equals the number of comparisons.
```
tanujkhattar, line 309: This is great!

qualtran/serialization/sympy.py
```diff
@@ -25,6 +25,8 @@ def _get_sympy_function_type(expr: sympy.Basic) -> int:
 
     This method converts a sympy function to its sympy_pb2.Function enum representation.
     """
+    import sympy.codegen.cfunctions
+
```
tanujkhattar, line 29: Why the import? We don't seem to be using it?

qualtran/serialization/sympy.py
```diff
@@ -62,6 +64,8 @@ def _get_sympy_function_from_enum(enum: int) -> Any:
     Sympy functions are represented as a sympy_pb2.Function enum. This method converts
     this int enum.
     """
+    import sympy.codegen.cfunctions
+
```
tanujkhattar, line 68: ditto

tanujkhattar (approved):
LGTM % some final comments

qualtran/serialization/sympy.py
anurudhp: I added `sympy.codegen.cfunctions.log2` in https://github.com/quantumlib/Qualtran/pull/1088, and the tests passed because `sympy_test.py` imported it. But otherwise `import sympy` itself does not load `sympy.codegen`.

fdmalone (approved):

qualtran/bloqs/arithmetic/sorting.py
anurudhp: updated

qualtran/bloqs/arithmetic/sorting.py
anurudhp: Added a brief description (similar to the one in `BitonicSort`).

Did you also want a description of the order of the ancillas?

qualtran/bloqs/arithmetic/sorting.py
```diff
@@ -48,82 +64,342 @@ class Comparator(Bloq):
...
+    For each block of `2 * offset` numbers, apply a `Comparator` between each pair that is `offset` apart.
+    For an offset of $\delta$, this requires
+
+    $$
+        \delta \left\lfloor \frac{k}{2\delta} \right\rfloor
+        + \max((k\mod 2\delta) - \delta, 0)
+    $$
+
+    The above expression is at most $k / 2$.
+    Each comparison generates one ancilla qubit, which are aggregated into the `junk` register.
```
tanujkhattar, line 117:
```suggestion
    Each comparison generates one ancilla qubit which stores the result of the comparison. The ancillas 
    are aggregated into the `junk` register.
```

qualtran/bloqs/arithmetic/sorting.py
```diff
@@ -48,82 +64,342 @@ class Comparator(Bloq):
...
+
+    For each block of `2 * offset` numbers, apply a `Comparator` between each pair that is `offset` apart.
+    For an offset of $\delta$, this requires
+
+    $$
+        \delta \left\lfloor \frac{k}{2\delta} \right\rfloor
+        + \max((k\mod 2\delta) - \delta, 0)
+    $$
+
+    The above expression is at most $k / 2$.
```
tanujkhattar, line 116: There seems to be a break after "this requires, <math expression> . The above expression".

We should finish the "this requires," line. Maybe something like
```suggestion
    number of comparisons. The above expression is at most $k / 2$.
```

qualtran/bloqs/arithmetic/sorting.py
```diff
@@ -48,82 +64,342 @@ class Comparator(Bloq):
...
+    r"""Merge two sorted sequences of n-bit integers.
+
+    Given two sorted lists of length `half_length`, merges them inplace into a single
+    sorted list.
+
+    Currently only supports `half_length` equal to a power of two (#1090).
+
+    If each half has length $k$, then the merge network uses $k (1+\log{k})$ comparisons
+    when $k$ is a power of 2.
+    Each comparison generates one ancilla qubit, which are aggregated into the `junk` register.
```
tanujkhattar, line 209: ditto: Let's specify that the ancilla qubit stores the result of comparisons

tanujkhattar (approved):

qualtran/bloqs/arithmetic/sorting.py
tanujkhattar: The order is not as important, but we should at least specify that the ancilla qubits store the results of intermediate comparisons.

qualtran/bloqs/arithmetic/sorting.py
anurudhp: oops, fixed

qualtran/bloqs/arithmetic/sorting.py
anurudhp: updated

qualtran/bloqs/arithmetic/sorting.py
anurudhp: updated

PR #1094, opened 2024-06-27
Add `Unitary` and `TensorProduct` block encodings

charlesyuan314:
Add instances of `BlockEncoding` that implement 1) a trivial block encoding of a unitary operator and 2) a tensor product of block encodings.

Extend the `BlockEncoding` interface to expose the properties `alpha`, `ancilla_bitsize`, and `epsilon` of block encodings.

With the newly designed block encoding interface, the `selection_registers`, `junk_registers`, and `target_registers` methods should now be considered deprecated and will be removed in a future PR.

### DIFF
--- a/qualtran/_infra/registers.py
+++ b/qualtran/_infra/registers.py
@@ -162,7 +162,7 @@ class Signature:
             registers: keyword arguments mapping register name to QDType. All registers
                 will be 0-dimensional and THRU.
         """
-        return cls(Register(name=k, dtype=v) for k, v in registers.items() if v.num_qubits)
+        return cls(Register(name=k, dtype=v) for k, v in registers.items())
 
     def lefts(self) -> Iterable[Register]:
         """Iterable over all registers that appear on the LEFT as input."""
--- a/qualtran/_infra/registers.py
+++ b/qualtran/_infra/registers.py
@@ -143,15 +143,18 @@ class Signature:
         self._rights = _dedupe((reg.name, reg) for reg in self._registers if reg.side & Side.RIGHT)
 
     @classmethod
-    def build(cls, **registers: Union[int, sympy.Expr]) -> 'Signature':
+    def build(cls, ignore_zero: bool = True, **registers: Union[int, sympy.Expr]) -> 'Signature':
         """Construct a Signature comprised of simple thru registers given the register bitsizes.
 
         Args:
+            ignore_zero: if True, registers with bitsize of zero will be ignored.
             registers: keyword arguments mapping register name to bitsize. All registers
                 will be 0-dimensional and THRU.
         """
         return cls(
-            Register(name=k, dtype=QBit() if v == 1 else QAny(v)) for k, v in registers.items()
+            Register(name=k, dtype=QBit() if v == 1 else QAny(v))
+            for k, v in registers.items()
+            if v or not ignore_zero
         )
 
     @classmethod
@@ -160,7 +163,7 @@ class Signature:
 
         Args:
             registers: keyword arguments mapping register name to QDType. All registers
-                will be 0-dimensional and THRU.
+                will be 0-dimensional and THRU. Registers with bitsize of zero are not ignored.
         """
         return cls(Register(name=k, dtype=v) for k, v in registers.items())
 
--- a/qualtran/_infra/registers.py
+++ b/qualtran/_infra/registers.py
@@ -151,7 +151,7 @@ class Signature:
                 will be 0-dimensional and THRU.
         """
         return cls(
-            Register(name=k, dtype=QBit() if v == 1 else QAny(v)) for k, v in registers.items() if v
+            Register(name=k, dtype=QBit() if v == 1 else QAny(v)) for k, v in registers.items()
         )
 
     @classmethod
--- a/qualtran/_infra/registers.py
+++ b/qualtran/_infra/registers.py
@@ -143,18 +143,15 @@ class Signature:
         self._rights = _dedupe((reg.name, reg) for reg in self._registers if reg.side & Side.RIGHT)
 
     @classmethod
-    def build(cls, ignore_zero: bool = True, **registers: Union[int, sympy.Expr]) -> 'Signature':
+    def build(cls, **registers: Union[int, sympy.Expr]) -> 'Signature':
         """Construct a Signature comprised of simple thru registers given the register bitsizes.
 
         Args:
-            ignore_zero: if True, registers with bitsize of zero will be ignored.
             registers: keyword arguments mapping register name to bitsize. All registers
                 will be 0-dimensional and THRU.
         """
         return cls(
-            Register(name=k, dtype=QBit() if v == 1 else QAny(v))
-            for k, v in registers.items()
-            if v or not ignore_zero
+            Register(name=k, dtype=QBit() if v == 1 else QAny(v)) for k, v in registers.items() if v
         )
 
     @classmethod
@@ -163,9 +160,9 @@ class Signature:
 
         Args:
             registers: keyword arguments mapping register name to QDType. All registers
-                will be 0-dimensional and THRU. Registers with bitsize of zero are not ignored.
+                will be 0-dimensional and THRU.
         """
-        return cls(Register(name=k, dtype=v) for k, v in registers.items())
+        return cls(Register(name=k, dtype=v) for k, v in registers.items() if v.num_qubits)
 
     def lefts(self) -> Iterable[Register]:
         """Iterable over all registers that appear on the LEFT as input."""
--- a/qualtran/_infra/registers_test.py
+++ b/qualtran/_infra/registers_test.py
@@ -130,7 +130,7 @@ def test_signature_build():
     sig1 = Signature([Register("r1", QInt(7)), Register("r2", QBit())])
     sig2 = Signature.build_from_dtypes(r1=QInt(7), r2=QBit())
     assert sig1 == sig2
-    sig1 = Signature([Register("r1", QInt(7))])
+    sig1 = Signature([Register("r1", QInt(7)), Register("r2", QAny(0))])
     sig2 = Signature.build_from_dtypes(r1=QInt(7), r2=QAny(0))
     assert sig1 == sig2
 
--- a/qualtran/_infra/registers_test.py
+++ b/qualtran/_infra/registers_test.py
@@ -139,8 +139,13 @@ def test_signature_build():
     assert sig1 == sig2
     assert sig1.n_qubits() == 7
 
-    sig1 = Signature([Register("r1", QAny(0)), Register("r2", QAny(2))])
+    sig1 = Signature([Register("r2", QAny(2))])
     sig2 = Signature.build(r1=0, r2=2)
+    assert sig1 == sig2
+    assert sig1.n_qubits() == 2
+
+    sig1 = Signature([Register("r1", QAny(0)), Register("r2", QAny(2))])
+    sig2 = Signature.build(ignore_zero=False, r1=0, r2=2)
     assert sig1 == sig2
     assert sig1.n_qubits() == 2
 
--- a/qualtran/_infra/registers_test.py
+++ b/qualtran/_infra/registers_test.py
@@ -33,17 +33,6 @@ def test_register():
     assert r == r.adjoint()
 
 
-def test_zero_register():
-    r = Register("my_reg", QAny(0))
-    assert r.name == "my_reg"
-    assert r.bitsize == 0
-    assert r.shape == tuple()
-    assert r.side == Side.THRU
-    assert r.total_bits() == 0
-
-    assert r == r.adjoint()
-
-
 def test_multidim_register():
     r = Register("my_reg", QBit(), shape=(2, 3), side=Side.RIGHT)
     idxs = list(r.all_idxs())
@@ -138,21 +127,10 @@ def test_signature_build():
     sig2 = Signature.build(r1=5, r2=2)
     assert sig1 == sig2
     assert sig1.n_qubits() == 7
-
-    sig1 = Signature([Register("r2", QAny(2))])
-    sig2 = Signature.build(r1=0, r2=2)
-    assert sig1 == sig2
-    assert sig1.n_qubits() == 2
-
-    sig1 = Signature([Register("r1", QAny(0)), Register("r2", QAny(2))])
-    sig2 = Signature.build(ignore_zero=False, r1=0, r2=2)
-    assert sig1 == sig2
-    assert sig1.n_qubits() == 2
-
     sig1 = Signature([Register("r1", QInt(7)), Register("r2", QBit())])
     sig2 = Signature.build_from_dtypes(r1=QInt(7), r2=QBit())
     assert sig1 == sig2
-    sig1 = Signature([Register("r1", QInt(7)), Register("r2", QAny(0))])
+    sig1 = Signature([Register("r1", QInt(7))])
     sig2 = Signature.build_from_dtypes(r1=QInt(7), r2=QAny(0))
     assert sig1 == sig2
 
--- a/qualtran/_infra/registers_test.py
+++ b/qualtran/_infra/registers_test.py
@@ -33,6 +33,17 @@ def test_register():
     assert r == r.adjoint()
 
 
+def test_zero_register():
+    r = Register("my_reg", QAny(0))
+    assert r.name == "my_reg"
+    assert r.bitsize == 0
+    assert r.shape == tuple()
+    assert r.side == Side.THRU
+    assert r.total_bits() == 0
+
+    assert r == r.adjoint()
+
+
 def test_multidim_register():
     r = Register("my_reg", QBit(), shape=(2, 3), side=Side.RIGHT)
     idxs = list(r.all_idxs())
@@ -127,6 +138,12 @@ def test_signature_build():
     sig2 = Signature.build(r1=5, r2=2)
     assert sig1 == sig2
     assert sig1.n_qubits() == 7
+
+    sig1 = Signature([Register("r1", QAny(0)), Register("r2", QAny(2))])
+    sig2 = Signature.build(r1=0, r2=2)
+    assert sig1 == sig2
+    assert sig1.n_qubits() == 2
+
     sig1 = Signature([Register("r1", QInt(7)), Register("r2", QBit())])
     sig2 = Signature.build_from_dtypes(r1=QInt(7), r2=QBit())
     assert sig1 == sig2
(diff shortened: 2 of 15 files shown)


qualtran/_infra/registers_test.py
charlesyuan314, line 1: TODO: add an explicit test for zero registers

qualtran/bloqs/block_encoding/tensor_product.py
```diff
@@ -0,0 +1,154 @@
...
+    @property
+    def signal_state(self) -> PrepareOracle:
+        raise NotImplementedError
+
+    def build_composite_bloq(
+        self, bb: BloqBuilder, system: SoquetT, ancilla: SoquetT, resource: SoquetT
+    ) -> Dict[str, SoquetT]:
+        transpose = lambda x: zip(*x)
+        sys_regs, anc_regs, res_regs = transpose(
+            [evolve(r, name=f"{r.name}{i}") for r in u.signature.lefts()]
```
tanujkhattar, line 115: Just iterate on `for r in u.signature` instead of `u.signature.lefts()` since all registers in `u` are THRU registers?

qualtran/bloqs/block_encoding/block_encoding_base.py
```diff
@@ -67,6 +68,31 @@ class `BlockEncoding` bloq, which expects values for $\alpha$, $\epsilon$,
...
+
+    @property
+    def num_resource(self) -> int:
+        """The number of resource qubits not counted in ancillas."""
+        raise NotImplementedError
+
+    @property
+    def epsilon(self) -> SymbolicFloat:
+        """The precision to which the block encoding is to be prepared."""
+        raise NotImplementedError
```
tanujkhattar, line 94: Make these `abc.abstractmethod` and make the base class derive from `metaclass=abc.ABCMeta` instead of raising `NotImplementedError`

qualtran/bloqs/block_encoding/tensor_product.py
```diff
@@ -0,0 +1,154 @@
...
+            anc_part.adjoint(), **{r.name: ap for r, ap in zip(anc_regs, anc_out_regs)}
+        )
+        resource = bb.add(
+            res_part.adjoint(), **{r.name: ap for r, ap in zip(res_regs, res_out_regs)}
+        )
+        return {"system": system, "ancilla": ancilla, "resource": resource}
+
+
+@bloq_example
+def _tensor_product_block_encoding() -> TensorProduct:
```
tanujkhattar, line 139: Please add an example with symbolics.

qualtran/bloqs/block_encoding/tensor_product.py
```diff
@@ -0,0 +1,154 @@
...
+    Registers:
+        system: The system register.
+        ancilla: The ancilla register.
+        resource: The resource register.
+
+    References:
+        [Quantum algorithms: A survey of applications and end-to-end complexities](https://arxiv.org/abs/2310.03011). Dalzell et al. (2023). Ch. 10.2.
+    """
+
+    U: Sequence[BlockEncoding] = field(
```
tanujkhattar, line 61: Does it make sense to change `U` to be a Union type of `HasLength` and `Tuple[BlockEncoding, ...]` so we can support a symbolic number of block encodings?

qualtran/bloqs/block_encoding/tensor_product.py
```diff
@@ -0,0 +1,154 @@
...
+
+    @property
+    def selection_registers(self) -> Tuple[Register, ...]:
+        return (self.signature.get_right("ancilla"),)
+
+    @property
+    def signal_state(self) -> PrepareOracle:
+        raise NotImplementedError
+
+    def build_composite_bloq(
```
tanujkhattar, line 110: Let's also implement a `build_call_graph`.

qualtran/bloqs/block_encoding/tensor_product.py
tanujkhattar, line 115: I think this naming convention may require a suffix underscore (`_`). -- cc @fdmalone ran into an issue where the cirq interop adds a suffix in front of qubits due to which two different qubits can end up with the same name. This is the reason why the target registers in `QROM`s have a underscore in the suffix.

qualtran/bloqs/block_encoding/tensor_product_test.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+    bloq = TensorProduct([u1, u2])
+    assert bloq.dtype == QAny(1 + 2)
+    assert bloq.alpha == 0.5 * 0.5
+    assert bloq.epsilon == 0.5 * 0.01 + 0.5 * 0.1
+    assert bloq.num_ancillas == 2 + 1
+    assert bloq.num_resource == 1 + 1
+
+
+def test_tensor_product_tensors():
```
tanujkhattar, line 55: Can you add tests that convert the bloq to a cirq circuit? For example: Construct a `cirq.Circuit(bloq_op)` and decompose it using `cirq.decompose()`.

qualtran/bloqs/block_encoding/unitary.py
```diff
@@ -0,0 +1,111 @@
...
+
+    U: Bloq
+    dtype: QDType
+    alpha: SymbolicFloat = 1
+    num_ancillas: int = 0
+    num_resource: int = 0
+    epsilon: SymbolicFloat = 0
+
+    def __attrs_post_init__(self):
+        assert self.U.signature == self.U.signature.adjoint()
```
tanujkhattar, line 64: This test is same as saying all registers in `self.U` should be THRU registers. I think writing this test explicitly would be cleaner in terms of what we want to convey

qualtran/bloqs/block_encoding/unitary.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def pretty_name(self) -> str:
+        return f"B[{self.U.pretty_name()}]"
+
+    @property
+    def target_registers(self) -> Tuple[Register, ...]:
+        return tuple(self.signature.rights())
+
+    junk_registers: Tuple[Register, ...] = ()
+    selection_registers: Tuple[Register, ...] = ()
```
tanujkhattar, line 79: What's happening here? Just override the properties and return an empty tuple instead? This pattern is super confusing.

qualtran/bloqs/block_encoding/unitary.py
```diff
@@ -0,0 +1,111 @@
...
+
+    @property
+    def signal_state(self) -> PrepareOracle:
+        raise NotImplementedError
```
tanujkhattar, line 83: Can you add a docstring for why this is NotImplemented?

qualtran/bloqs/block_encoding/unitary.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def build_composite_bloq(
```
tanujkhattar, line 85: Let's implement a `build_call_graph`, in case `self.signature` is symbolic the cirq-style protocols would fail during resource estimation.

qualtran/_infra/registers.py
```diff
@@ -162,7 +162,7 @@ def build_from_dtypes(cls, **registers: QDType) -> 'Signature':
             registers: keyword arguments mapping register name to QDType. All registers
                 will be 0-dimensional and THRU.
         """
-        return cls(Register(name=k, dtype=v) for k, v in registers.items() if v.num_qubits)
+        return cls(Register(name=k, dtype=v) for k, v in registers.items())
```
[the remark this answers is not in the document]
tanujkhattar: Okay, I get the motivation but I think we should pull out this change in a separate PR since it has potential to break existing stuff and add thorough tests.

tanujkhattar (changes requested):
Great start, left a round of comments.

qualtran/_infra/registers.py
```diff
@@ -143,15 +143,18 @@ def __init__(self, registers: Iterable[Register]):
         self._rights = _dedupe((reg.name, reg) for reg in self._registers if reg.side & Side.RIGHT)
 
     @classmethod
-    def build(cls, **registers: Union[int, sympy.Expr]) -> 'Signature':
+    def build(cls, ignore_zero: bool = True, **registers: Union[int, sympy.Expr]) -> 'Signature':
```
tanujkhattar, line 146: Adding an `ignore_zero:` flag to make tests pass for this PR would be hacky -- Let's pull out this change in a separate PR and fix it properly OR update the block encodings design so we don't have 0 sized registers. The only reason tests are not failing for block encodings with 0 bitsizes is because we don't have the right tests that would trigger a failure. 

I suggest it may be easier to continue the assumption that we won't have registers with bitsize 0 (and therefore not make this change). The block encodings implementation can do the following:

Iterate on `self.subbloq.registers` and for each register which is present, pull out the ancilla / resource part from `soqs` dict. This pattern has been used across the library at multiple places where the signature of the subbloq can be dynamic (for example: where the subbloq may or may not have a control register and this behavior is dependent upon an attribute `cvs` that the subbloq expects).

qualtran/bloqs/block_encoding/block_encoding_base.py
```diff
@@ -67,6 +68,31 @@ class `BlockEncoding` bloq, which expects values for $\alpha$, $\epsilon$,
     def pretty_name(self) -> str:
         return 'B[H]'
 
+    @property
+    def dtype(self) -> QDType:
+        """The type of the system being block encoded."""
+        raise NotImplementedError
+
+    @property
+    def alpha(self) -> SymbolicFloat:
```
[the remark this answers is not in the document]
charlesyuan314: It's always a float because it's the normalization of $$||A||$.

qualtran/bloqs/block_encoding/tensor_product.py
```diff
@@ -0,0 +1,154 @@
...
+    def junk_registers(self) -> Tuple[Register, ...]:
+        return (self.signature.get_right("resource"),)
+
+    @property
+    def selection_registers(self) -> Tuple[Register, ...]:
+        return (self.signature.get_right("ancilla"),)
+
+    @property
+    def signal_state(self) -> PrepareOracle:
+        raise NotImplementedError
```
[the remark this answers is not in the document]
charlesyuan314: Yes. I'll add documentation stating as such.

qualtran/bloqs/block_encoding/block_encoding_base.py
charlesyuan314: My concern is that existing instances of `BlockEncoding` then could not be instantiated unless we changed them now to implement these methods. Do you agree?

qualtran/_infra/registers.py
charlesyuan314: @mpharrigan As we just discussed this today, do you want to weigh in on the question of zero-sized registers?

qualtran/bloqs/block_encoding/tensor_product.py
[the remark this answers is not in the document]
tanujkhattar: You can also open an issue and assign it to yourself.

qualtran/bloqs/block_encoding/tensor_product.py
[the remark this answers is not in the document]
mpharrigan: we've been moving to just using `Sequence` as the type annotation since it's more readable than tuple with its ellipses and it raises fewer false positives from type checkers that can't deduce types by respecting the `converter` argument.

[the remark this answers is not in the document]
tanujkhattar: But that's the wrong type? Shouldn't class attributes have more specific types (similar to return values) and the constructor have generic types? Having the right type would also help us identify issues where we forget to provide a converter to convert the input to a tuple; which is important and would cause problems with hashing. 

Eg: https://mypy-play.net/?mypy=latest&python=3.12&gist=160e99b38930a8fcc5564ab7f790593b

Maybe @dstrain115 can chime in on advice for what's the best pattern.

qualtran/bloqs/block_encoding/tensor_product.py
mpharrigan: is this still an issue? ideally we wouldn't be mutilating our signatures to support compiling to cirq

tanujkhattar: I don't know if it is still an issue. We should verify and fix it before potentially creating new tech debt since we know we have encountered the issue in the past.

qualtran/bloqs/block_encoding/block_encoding_base.py
```diff
@@ -67,6 +68,31 @@ class `BlockEncoding` bloq, which expects values for $\alpha$, $\epsilon$,
     def pretty_name(self) -> str:
         return 'B[H]'
 
+    @property
+    def dtype(self) -> QDType:
+        """The type of the system being block encoded."""
+        raise NotImplementedError
```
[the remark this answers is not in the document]
charlesyuan314: Given the source of confusion, I will replace the `dtype: QDType` property with simply a `system_bitsize: int` property.

qualtran/bloqs/block_encoding/tensor_product.py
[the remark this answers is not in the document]
dstrain115: I think it is okay for a class method to be a Sequence, but then it should not be an error if there is a list or other sequence assigned to U.  (For instance, if you want to be flexible in the types that U accepts.  

If U has to be a tuple, then it should be a Tuple type.

qualtran/bloqs/block_encoding/tensor_product.py
```diff
@@ -0,0 +1,154 @@
...
+    @cached_property
+    def dtype(self) -> QDType:
+        return QAny(bitsize=sum(u.dtype.num_qubits for u in self.U))
+
+    def pretty_name(self) -> str:
+        return f"B[{'⊗'.join(u.pretty_name()[2:-1] for u in self.U)}]"
+
+    @cached_property
+    def alpha(self) -> SymbolicFloat:
+        return reduce(lambda a, b: a * b.alpha, self.U, 1.0)
```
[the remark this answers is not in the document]
charlesyuan314: I will make this change on top of #1101.

qualtran/bloqs/block_encoding/tensor_product.py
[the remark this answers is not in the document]
charlesyuan314: I changed the name to `block_encodings` and the type to `Tuple`.

qualtran/bloqs/block_encoding/tensor_product.py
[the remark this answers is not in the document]
charlesyuan314: I will open this issue at merge time.

qualtran/bloqs/block_encoding/block_encoding_base.py
charlesyuan314: I will open this issue at merge time.

qualtran/bloqs/block_encoding/tensor_product.py
charlesyuan314: I'm not really sure how that would work. We can't compute parameters such as the `alpha` of the tensor product without concretely knowing its components. Just knowing the number of components seems to not be enough.

qualtran/bloqs/block_encoding/tensor_product.py
charlesyuan314: I added the underscore suffix.

qualtran/bloqs/block_encoding/tensor_product_test.py
charlesyuan314: I'm not deeply familiar with how Cirq works. Could you point me to an example in the codebase of a test for a bloq that is done this way?

qualtran/_infra/registers.py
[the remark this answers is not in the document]
charlesyuan314: Is the scope of the tests above zero-sized registers specifically, or more generally for the behavior of the `Unitary` and `TensorProduct` bloqs?

(In other words, if we were to back out the zero-sized stuff, are they still needed? I responded below to your other comment about Cirq-style tests.)

qualtran/bloqs/block_encoding/tensor_product_test.py
mpharrigan: why do we need this test?

qualtran/bloqs/block_encoding/tensor_product.py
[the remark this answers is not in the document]
charlesyuan314: I made this change.

qualtran/_infra/registers.py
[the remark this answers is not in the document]
tanujkhattar: The tests were specifically for zero-sized registers.

qualtran/bloqs/block_encoding/tensor_product.py
```diff
@@ -0,0 +1,192 @@
...
+    def selection_registers(self) -> Tuple[Register, ...]:
+        return (self.signature.get_right("ancilla"),)
+
+    @property
+    def signal_state(self) -> PrepareOracle:
+        """This method will be implemented in the future after PrepareOracle is updated for the BlockEncoding interface."""
+        raise NotImplementedError
+
+    def build_call_graph(self, ssa: SympySymbolAllocator) -> Set[BloqCountT]:
+        return {(bloq, 1) for bloq in self.block_encodings}
```
tanujkhattar, line 115: Let's instead use a dictionary to add up the bloq counts and then convert to a set; so that if `self.block_encodings` has two entries for the same `bloq` then we have the right counts.

qualtran/bloqs/block_encoding/tensor_product_test.py
tanujkhattar: To verify cirq interop and full decomposition. There's a lot going on in these high level bloqs, like:
- Potentially zero sized registers (which were present when I added the original comment)
- `Partition`s with dynamically generated output register names. 
- The above two happening recursively


It's good to verify that the recursive composition works without any issues with cirq interop (eg: the register names with a suffix, which may or may not have been an issue).

qualtran/_infra/registers.py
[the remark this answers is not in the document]
charlesyuan314: I have removed the zero-sized registers from this PR.

qualtran/_infra/registers.py
charlesyuan314: I have removed the zero-sized registers from this PR.

qualtran/bloqs/block_encoding/tensor_product_test.py
charlesyuan314: I added a test for Cirq decomposition.

tanujkhattar (approved):
LGTM! Thanks for the changes!

Please open issues to track 
-  Migration of NotImplementedError to abc.abstractmethods. 
-  Implementation of signal_state for the newly added block encodings.

PR #1099, opened 2024-06-30
Greedy topological sort of the binst graph to minimize qubit allocations / deallocations

tanujkhattar:
This PR adds a new stable greedy topological sorting method and changes the default topological sort to use this method when iterating over nodes of a bloq instance graph. 

This is useful to 

- Improve qubit cost analysis
- Improve cirq-interop and minimize qubit allocations when specifying a GreedyQubitManager. 

Fixes the issue mentioned in https://github.com/quantumlib/Qualtran/pull/963#issuecomment-2192745758, would be useful for https://github.com/quantumlib/Qualtran/issues/1097 and can be improved by addressing https://github.com/quantumlib/Qualtran/issues/1098

PR #1100, opened 2024-07-01
Bugfix in Cirq Interop: Attempt 2

tanujkhattar:
Another attempt at https://github.com/quantumlib/Qualtran/pull/963. Opened a new PR instead of rebasing that one since a major reorganization has occurred since that attempt. 

High level updates are:
- cirq.inverse doesn't play well with qubit allocation / deallocation. Therefore, removed `decompose_from_registers` from the `Adjoint` bloq and default to using the bloq-style `decompose_bloq` API
- `InteropQubitManager` had a bug where the managed qubits could be allocated twice (eg: by the underlying greedy qubit manager)

The above two were the primary reason for https://github.com/quantumlib/Qualtran/issues/818, which is now fixed. As part of these upgrades, a few other minor improvements were done. For example:
- `SGate(is_adjoint=True).as_cirq_op()` was returning `cirq.S` instead of `cirq.S**-1`. This is now fixed. 
- `Allocate` and `Free` now have a `as_cirq_op` defined and won't show up as explicit operations in the cirq circuit.
- Comparison bloqs are made self adjoint because tests of `ReflectionUsingPrepare` rely on not decomposing `LessThanEqual` gate so the cirq simulator can make sure of the optimized implementation of _apply_unitary_ for `cirq.ArithmeticGate`'s. This gets lost if we have `Adjoint(LessThanEqual)` instead of `LessThanEqual` in the circuit. Flattening out the entire cirq circuit for `ReflectionUsingPrepare -> StatePreparationViaAliasSampling -> LessThanEqual` is catastrophic for performance. cc @mpharrigan It'll be nice to see how the new tensor network simulation perform for this specific test. It'll be nice to get rid of `construct_gate_helper_and_qubit_order` and rewrite `get_3q_uniform_dirac_notation` test with the same parameters but using tensor network simulations. 

Currently based on top of https://github.com/quantumlib/Qualtran/pull/1099

The only pending thing in this PR is that tests for `SelectSwapQROM` are still failing; and the failure is related to the one seen in https://github.com/quantumlib/Qualtran/pull/961. More investigations are needed here to figure out whats going wrong.


mpharrigan:
I think some of the selectswapqrom tests have zero-bitsize registers that have snuck in. Maybe that could be causing a problem?

tanujkhattar:
@mpharrigan The SelectSwapQROM failure was because `SGate(is_adjoint=True).as_cirq_op()` was returning `cirq.S` instead of `cirq.S**-1`  :(

All tests are now fixed and this is ready for a review. I've also added more simulation tests and I request everyone to add full state vector simulation tests as part of the test suite so we can catch such bugs earlier. Classical simulation was giving correct answer despite full state vector simulation being wrong; so one cannot replace the other from our perspective.

mpharrigan (approved):
thank you!

PR #1101, opened 2024-07-01
Generalize `symbolics.prod` to allow `SymbolicFloat` or `SymbolicInt`

charlesyuan314:
Prior to this change, `symbolics.prod` only accepted `SymbolicInt` arguments and returned `SymbolicInt`. Now, it also accepts `SymbolicFloat` to return `SymbolicFloat`.

The function also now accepts an iterator rather than being variadic, matching the interface of built-in `sum`.

This goal is achieved via a new type variable `SymbolicT`. There does not seem to be a correct approach using `@overload` instead due to limitations of mypy, which does not consider `float | Expr` and `int | Expr` to be two valid overloads for a function since `int` is automatically promoted to `float`.

### DIFF
--- a/qualtran/symbolics/math_funcs.py
+++ b/qualtran/symbolics/math_funcs.py
@@ -85,13 +85,13 @@ def smin(*args):
 # This is only used in the type signature of functions that should be generic over different
 # symbolic types, in situations where Union and @overload are not sufficient.
 # The user should not need to invoke it directly. Rather, the user can use a function that
-# takes AnySymbolic by calling it with e.g. a SymbolicInt. Correspondingly, if the type signature
-# of the function returns AnySymbolic, then this call will then return a SymbolicInt.
-AnySymbolic = TypeVar('AnySymbolic', SymbolicInt, SymbolicFloat, SymbolicComplex)
+# takes SymbolicT by calling it with e.g. a SymbolicInt. Correspondingly, if the type signature
+# of the function returns SymbolicT, then this call will then return a SymbolicInt.
+SymbolicT = TypeVar('SymbolicT', SymbolicInt, SymbolicFloat, SymbolicComplex)
 
 
-def prod(args: Iterable[AnySymbolic]) -> AnySymbolic:
-    ret: AnySymbolic = 1
+def prod(args: Iterable[SymbolicT]) -> SymbolicT:
+    ret: SymbolicT = 1
     for arg in args:
         ret = ret * arg
     return ret
--- a/qualtran/symbolics/math_funcs.py
+++ b/qualtran/symbolics/math_funcs.py
@@ -11,13 +11,12 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-from typing import cast, Iterable, overload, Sized, Tuple, Union
+from typing import cast, Iterable, overload, Sized, Tuple, TypeVar, Union
 
 import numpy as np
 import sympy
 
 from qualtran.symbolics.types import (
-    AnySymbolic,
     HasLength,
     is_symbolic,
     Shaped,
@@ -81,6 +80,14 @@ def smin(*args):
     if is_symbolic(*args):
         return sympy.Min(*args)
     return min(*args)
+
+
+# This is only used in the type signature of functions that should be generic over different
+# symbolic types, in situations where Union and @overload are not sufficient.
+# The user should not need to invoke it directly. Rather, the user can use a function that
+# takes AnySymbolic by calling it with e.g. a SymbolicInt. Correspondingly, if the type signature
+# of the function returns AnySymbolic, then this call will then return a SymbolicInt.
+AnySymbolic = TypeVar('AnySymbolic', SymbolicInt, SymbolicFloat, SymbolicComplex)
 
 
 def prod(args: Iterable[AnySymbolic]) -> AnySymbolic:
--- a/qualtran/symbolics/math_funcs.py
+++ b/qualtran/symbolics/math_funcs.py
@@ -11,12 +11,13 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-from typing import cast, overload, Sized, Tuple, Union
+from typing import cast, Iterable, overload, Sized, Tuple, Union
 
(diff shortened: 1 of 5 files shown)


qualtran/symbolics/math_funcs.py
```diff
@@ -82,8 +83,8 @@ def smin(*args):
     return min(*args)
 
 
-def prod(*args: SymbolicInt) -> SymbolicInt:
-    ret: SymbolicInt = 1
+def prod(args: Iterable[AnySymbolic]) -> AnySymbolic:
+    ret: AnySymbolic = 1
```
tanujkhattar, line 87: I think it's better to define an @overload of this method so that if all inputs are `Iterable[SymbolicInt]` then the return type should be a `SymbolicInt` and if the input is `Iterable[SymbolicFloat]` then the return type should be `SymbolicFloat`. 

Using `AnySymbolic` would mean that call sites that expect a `SymbolicInt` would now need to deal with `Union[SymbolicFloat, SymbolicInt]`  instead. I don't think `AnySymbolic` is a useful addition since `SymbolicFloat` is a superset of `SymbolicInt` so the union might as well just correspond to `SymbolicFloat`.

mpharrigan: @charlesyuan314 wrote 

> here does not seem to be a correct approach using @overload instead due to limitations of mypy, which does not consider float | Expr and int | Expr to be two valid overloads for a function since int is automatically promoted to float.

did you see this?

qualtran/symbolics/types.py
```diff
@@ -26,6 +26,9 @@
 SymbolicComplex = Union[complex, sympy.Expr]
 document(SymbolicComplex, """A complex value or a sympy expression.""")
 
+AnySymbolic = TypeVar('AnySymbolic', SymbolicInt, SymbolicFloat)
+document(AnySymbolic, """A type variable that can be either SymbolicInt or SymbolicFloat.""")
```
mpharrigan, line 30: should this union in `SymbolicComplex` as well?

mpharrigan: why `TypeVar` instead of `Union` like the others?

qualtran/symbolics/math_funcs.py
charlesyuan314: This is where I got stuck @tanujkhattar, if you have a suggestion for how to make `@overload` work:

https://mypy-play.net/?mypy=latest&python=3.12&flags=show-error-codes%2Callow-redefinition%2Ccheck-untyped-defs

I tried a couple of other things, but mypy is pretty stubborn with what it considers a valid overload.

qualtran/symbolics/types.py
charlesyuan314: Let me try to give a technical explanation here --- I hope it's not too obvious or pedantic of one.

Assuming that mypy works along the lines of other type systems, the difference is that a `TypeVar` is a form of [bounded parametric polymorphism](https://en.wikipedia.org/wiki/Parametric_polymorphism#Bounded_parametric_polymorphism) whereas a `Union` is a form of [width subtyping](https://en.wikipedia.org/wiki/Subtyping#Width_and_depth_subtyping).

What that mumbo jumbo means is that a function that accepts a `Union[int, float]` and returns a `Union[int, float]` is _one function_ whose more precise relationship between input and output (`int`s go to `int`s and `float`s go to `float`s) must be deduced through external means. 

In mypy, that external means is the `@overload` system, which unfortunately considers `Iterable[int | Expr]` as "the same or broader" parameter type as `Iterable[float | Expr]`, as shown in the playground. This behavior seems to be because `int` is implicitly convertible to `float` in Python, but I do not fully understand it. Whether it is a bug or feature I cannot tell, or whether there is yet some trick to get around the problem.

By contrast, a function that accepts `AnySymbolic` and returns `AnySymbolic` is a _family of functions_ parameterized by the type variable `AnySymbolic`. Here, we also _bound_ it to be either `SymbolicInt` or `SymbolicFloat`, but that is a totally ancillary thing. Formally, the type checker is obligated to treat `AnySymbolic` as if it were some totally opaque type `T`.

The rules of parametric polymorphism state that the type checker must check each function in this family independently, without considering any other. In particular, the fact that `int` is implicitly convertible to `float` cannot influence on the process that the type checker takes to determine that giving the function an `Iterable[int]` must cause it to output an `int`.

That is my best explanation so far of why `TypeVar` works here whereas `Union` does not.

qualtran/symbolics/math_funcs.py
charlesyuan314: > Using AnySymbolic would mean that call sites that expect a SymbolicInt would now need to deal with Union[SymbolicFloat, SymbolicInt]

To my knowledge, this isn't true. The type checker automatically instantiates `AnySymbolic` to be `SymbolicInt` when `prod` is called with a `SymbolicInt`. One way to think about the situation is that there are two versions of `prod` (call them `prod_int` and `prod_float`), each one of which is selected automatically behind the scenes without the caller needing to care about it.

> I don't think AnySymbolic is a useful addition since SymbolicFloat is a superset of SymbolicInt so the union might as well just correspond to SymbolicFloat.

`AnySymbolic` is not really intended to be used by itself as a replacement for the other symbolic types. It's just a type checking tool used in the type signature of `prod` (and possibly similar functions) to get around this specific behavior of the type checker. I'm very open to a way of making the typing work without this `TypeVar` and with `@overload` alone, but I haven't found one yet.

qualtran/symbolics/types.py
charlesyuan314: If it helps, `@overload` is analogous to function overloading in C++ whereas `TypeVar` is analogous to `template<typename T>`.

mpharrigan: thanks for the explanation!

qualtran/symbolics/math_funcs.py
mpharrigan: Perhaps we can rename AnySymbolic to `SymbolicT` which is the convention for the other TypeVars used as generic types elsewhere in the library (and in Python more generally)

qualtran/symbolics/types.py
mpharrigan: Yeah, we should name it `SymbolicT`, which is the convention for typevars elsewhere in the library and makes the distinction between the Union ones

qualtran/symbolics/math_funcs.py
charlesyuan314: Renamed.

mpharrigan (approved):

Issue #1102, opened 2024-07-01
Design discussion: Introduce zero-sized registers to support quantum interfaces


mpharrigan:
In https://github.com/quantumlib/Qualtran/pull/1094, the desire to have zero-size registers is introduced.

The motivation is to have a consistent set of register names for a "quantum interface" and just set the size to zero if that register isn't actually "used". 

I'll write my personal thoughts in a follow-on comment in this issue.

mpharrigan:
For motivation: we want a class of bloqs that follow the same "quantum interface". Namely: the same, known set of registers with fixed names.

```python
bb.add(arbitrary_bloq, select=select, system=system, ancilla=ancilla)
```

where `arbitrary_bloq` will happily accept `ancilla` even if we don't use it. I think the closest analogy in a classical programming language is to have a callable that must take all the possible arguments but a) might ignore one of them b) we can pass in `None` to the ignored argument.  It's distinct from the concept of a "default argument", which is the opposite case where we _don't_ pass in a value but a default one is substituted. 

The reason why we can't simply take in extraneous wires and just ignore them is that we want to accurately estimate the quantum resource requirements of these algorithms, so we need to encode the fact that that line isn't actually used. 

In general, I think it's helpful to ask how zero-sized registers would be handled by the protocols

 - qubit counts: easy: the zero size registers don't add anything to qubit requirements
 - bloq counts: at some point, a bloq that is called with a zero-size register will need to alter its decomposition to ignore that wire. For example: if a Toffoli is called with a zsr in one of its control lines, it should probably be counted as a CNOT (?)
 - drawing: the zero-size registers/Connections/wires would exist and probably need to be drawn. This could be unintuitive
 - classical simulation: I think the analogous classical value would be `None`? 
 - tensor simulation: I'm not sure how zero size registers would fit into this
 - composition: you can only compose zero-size registers with other zero-size registers (like any data type). It's unclear to me how a zero-size register would be handled as we recurse through multiple levels of decomposition. Would we need a special allocation and freeing to allocate these null values?

----------

We have the power to use the classical, Python layer of qualtran to dynamically set the signature of a bloq. There's a hierarchy of how dynamic a signature is; with a corresponding hierarchy in complexity. 

 1. completely fixed (e.g. CNOT)
 2. simple parameterization of bitsizes (e.g. Swap)
 3. different data types (e.g. Add signed/unsigned)
[...]

PR #1105, opened 2024-07-02
Make sure no zero-bitsize connections appear

mpharrigan:
This adds additional assertions to `qualtran.testing` to make sure none of the `Connection` objects have zero bitsize. The presence of such registers/connections likely breaks assumptions used by the rest of the library. cc #1102 

Originally, I was going to add asserts to the `QDType`s themselves, but 1) they're used in several places other than defining registers that actually appear in a composite bloq causing "false positives" in the test 2) performing those assertions at QDType instantiation makes it impossible to get around; whereas this current change will only result in a failure if you're using `qlt_testing.assert_valid_bloq_decomposition` or its friends (or the `bloq_autotester`) so this PR probably has some false negatives still remaining.


mpharrigan:
cc @NoureldinYosri found this in reviewing #1057

mpharrigan:
@tanujkhattar `test_select_swap_qrom` results in zero-sized registers for the 3rd data example of `[[1],[2],[3]]`. Can you fix? 

https://github.com/quantumlib/Qualtran/blob/e2c6b82f0315ccfbdd1ebe9f65c8f1b324fc0c05/qualtran/bloqs/data_loading/qrom_base.py#L227

mpharrigan:
`StatePreparationViaRotations`

https://github.com/quantumlib/Qualtran/blob/cf2044d9eecae76c8165d2405ee858d32986b647/qualtran/bloqs/state_preparation/state_preparation_via_rotation.py#L160

passed as selection bitsize to `PRGAViaPhaseGradient`. I don't know if this should actually be zero and the selection register needs to be omitted or if we still use a selection bit to select 1 value

@tanujkhattar can you fix

mpharrigan:
@tanujkhattar can you approve. Looks like the failures are indeed fixed :)

tanujkhattar (approved):

PR #1106, opened 2024-07-02
Add `Product` block encoding

charlesyuan314:
Add an instance of `BlockEncoding` that implements the product of constituent block encodings.


qualtran/bloqs/block_encoding/product.py
```diff
@@ -0,0 +1,264 @@
...
+    def junk_registers(self) -> Tuple[Register, ...]:
+        return (self.signature.get_right("resource"),)
+
+    @property
+    def selection_registers(self) -> Tuple[Register, ...]:
+        return (self.signature.get_right("ancilla"),)
+
+    @property
+    def signal_state(self) -> PrepareOracle:
+        """This method will be implemented in the future after PrepareOracle is updated for the BlockEncoding interface."""
```
mpharrigan, line 116: nit: this can be a line comment, and wrap to 100 chars.

If this isn't going to be immediately remedied (although I suspect it might be), it would be good practice to link to an issue as well

qualtran/symbolics/math_funcs.py
```diff
@@ -71,12 +71,34 @@ def bit_length(x: SymbolicFloat) -> SymbolicInt:
 
 
 def smax(*args):
-    if any(isinstance(arg, sympy.Basic) for arg in args):
+    if len(args) == 0:
```
mpharrigan, line 74: can you add some simple tests for these?

qualtran/bloqs/block_encoding/product.py
```diff
@@ -0,0 +1,264 @@
...
+def _product_block_encoding() -> Product:
+    from qualtran.bloqs.basic_gates import Hadamard, TGate
+    from qualtran.bloqs.block_encoding.unitary import Unitary
+
+    product_block_encoding = Product((Unitary(TGate()), Unitary(Hadamard())))
+    return product_block_encoding
+
+
+@bloq_example
+def _product_block_encoding_override() -> Product:
```
mpharrigan, line 224: what does override mean?

qualtran/bloqs/block_encoding/product.py
```diff
@@ -0,0 +1,264 @@
...
+    from qualtran.bloqs.basic_gates import Hadamard, TGate
+    from qualtran.bloqs.block_encoding.unitary import Unitary
+
+    u1 = evolve(Unitary(TGate()), alpha=0.5, ancilla_bitsize=2, resource_bitsize=1, epsilon=0.01)
```
mpharrigan, line 228: why are you using evolve instead of just providing the arguments to the `Unitary` constructor?

qualtran/bloqs/block_encoding/product.py
```diff
@@ -0,0 +1,264 @@
...
+from qualtran.bloqs.basic_gates.x_basis import XGate
+from qualtran.bloqs.block_encoding import BlockEncoding
+from qualtran.bloqs.block_encoding.lcu_select_and_prepare import PrepareOracle
+from qualtran.bloqs.bookkeeping.partition import Partition
+from qualtran.symbolics import is_symbolic, prod, smax, ssum, SymbolicFloat, SymbolicInt
+
+
+@frozen
+class Product(BlockEncoding):
+    r"""Product of a sequence of block encodings.
```
mpharrigan, line 43: can you briefly summarize the construction itself in addition to the block encoding parameter formulas

qualtran/bloqs/block_encoding/product.py
```diff
@@ -0,0 +1,264 @@
...
+
+    Builds the block encoding $B[U_1 * U_2 * \cdots * U_n]$ given block encodings $B[U_1], \ldots, B[U_n]$.
+
+    When each $B[U_i]$ is a $(\alpha_i, a_i, \epsilon_i)$-block encoding of $U_i$, we have that
+    $B[U_1 * \cdots * U_n]$ is a $(\prod_i \alpha_i, n - 1 + \max_i a_i, \sum_i \alpha_i \epsilon_i)$-block
+    encoding of $U_1 * \cdots * U_n$.
```
mpharrigan, line 49: My personal preference, feel free to ignore:

There's a lot going on inside the second parentheses. What about splitting it out a bit: where each bui is an alpphai ai epsiloni block encoding, the block encoding of the product has normalization constant $\prod alpha_i$, ancilla bitsize $n-1 max etc$, and epsilon $sum alphai epsiloni$.

mpharrigan (approved):
Some minor things to consider, but LGTM

qualtran/bloqs/block_encoding/product.py
tanujkhattar: You can consider using https://asciiflow.com/#/ to draw an ascii diagram of the circuit and paste it in the docstring.

qualtran/bloqs/block_encoding/product.py
```diff
@@ -0,0 +1,264 @@
...
+                res_bits_used += res_bits
+                if res_bits_left > 0:
+                    res_regs.append(res_after)
+                res_part = Partition(self.resource_bitsize, tuple(res_regs))
+                res_part_soqs = bb.add_d(res_part, x=soqs["resource"])
+                u_soqs["resource"] = res_part_soqs["res"]
+
+            # connect the constituent bloq
+            u_out_soqs = bb.add_d(u, **u_soqs)
+            system = u_out_soqs["system"]
```
tanujkhattar, line 176: Am  I understanding this correct that you are using the simpler construction where each of the block encoding has it's own ancilla register and they don't share any ancillas? i.e. the construction from Figure 1 instead of Figure 2 ?

image



If that is the case, can we do the partition / unpartition of the ancilla registers once outside the loop instead of doing it as part of the loop? That would clean up the decomposition quite a bit I believe.

tanujkhattar (approved):
Please add an explanation of the decomposition to the docstring and consider doing the partitions and unpartitions once outside the for-loop

qualtran/bloqs/block_encoding/product.py
```diff
@@ -0,0 +1,264 @@
...
+            if self.ancilla_bitsize > 0:
+                flag_bits_soq = cast(NDArray, anc_part_soqs["flag_bits"])
+                if anc_bits > 0:
+                    anc_used_soq = cast(Soquet, u_out_soqs["ancilla"])
+                    if i == n - 1:
+                        anc_part_soqs["anc_used"] = anc_used_soq
+                    else:
+                        # set corresponding flag if ancillas are all zero
+                        ctrl, flag_bits_soq[i] = bb.add_t(
+                            XGate().controlled(CtrlSpec(qdtypes=(anc_used_soq.reg.dtype,), cvs=0)),
```
tanujkhattar, line 195: Consider using `MultiControlPauli` instead of `XGate().controlled()`

qualtran/bloqs/block_encoding/product.py
tanujkhattar: nvm, seems like you are using decomposition from Fig 2. I missed the part where you were adding MCX gates below. 

In this case, I would do a parition of ancilla register into (a) Flag register and (b) Ancilla for individual bloq encodings outside the forloop and then add `n` bloqs for the `i` block encodings, each of them wrapped in an `AutoPartition`. The `MultiControlX` gate can use the unpartitioned ancilla directly. This would give us circuit diagrams very similar to the one drawn in the picture above.

charlesyuan314: This may require some change to `AutoPartition`, and so I will do it in a follow-up PR.

charlesyuan314:
I will do the [suggested refactor](https://github.com/quantumlib/Qualtran/pull/1106#discussion_r1671506044) to clean up the circuit diagram in a follow-up PR.

@mpharrigan I've addressed all of the other feedback.

qualtran/conftest.py
```diff
@@ -97,6 +97,9 @@ def assert_bloq_example_serializes_for_pytest(bloq_ex: BloqExample):
         'tensor_product_block_encoding',
         'tensor_product_block_encoding_override',
         'tensor_product_block_encoding_symb',
+        'product_block_encoding',
+        'product_block_encoding_override',
+        'product_block_encoding_symb',
```
tanujkhattar, line 102: Why are these not serializable ?

tanujkhattar: @charlesyuan314 I'll merge the PR and we can continue this discussion and potentially address in a follow-up PR

charlesyuan314: I believe it is because this bloq has an attribute of type `Tuple[BlockEncoding, ...]`. It results in the following at serialization time:
```
qualtran.testing.BloqCheckException: Serialization Failed:
Object arrays cannot be saved when allow_pickle=False
```

Are bloqs contained as attributes something that can be made serializable?

tanujkhattar: > Object arrays cannot be saved when allow_pickle=False

I think it's because so far we haven't had any bloq that stores a sequence of subbloqs as an attribute. 

To support serialization we'll have to update https://github.com/quantumlib/Qualtran/blob/f8d2db04b2986268fbe423f0e294b10152758736/qualtran/protos/bloq.proto#L40 to add a new ` NDArray subbloqs`  that stores an NDArray of integers corresponding to IDs of each of the subbloqs and then update https://github.com/quantumlib/Qualtran/blob/f8d2db04b2986268fbe423f0e294b10152758736/qualtran/serialization/bloq.py#L349 to check if the argument is an ndarray of bloqs, then delegate to a different method that serializes it. 

Can you please open an issue to track this?

Issue #721, opened 2024-02-29
Basic Unary Iteration / QROM


mpharrigan:
unary iteration / qrom bloqs support a lot of features (multi-dimensional select; `break`) that obscure the key points of the design. They're also written cirq-style where each target op `op[l]` is a bundle of gates instead of one bloq per index. Writing an idiomatic "basic" 1d unary iteration and/or qrom bloq would be very instructive.

mpharrigan:
A sketch of how this might look

```python
class Set(Bloq):
  datum: int
  bitsize: int

  return Signature.build(x=bitsize)

  def decompose(bb, x):
    for bit in iter_bits(self.datum):
       if bit:
         bb.Add(XGate()
 ```

```python
class ApplyL(Bloq):
  ops: List[Bloq]
  selection_bitsize: int = field(validator=len(ops) fits)
  target_bitsize: int

  return Signature.build(ctrl=1, selection=self.selection_bitsize, target=self.target_bitsize)

  def decompose(bb, ctrl, selection, target):
    selection = selection.split()
    l = 0
    for ...:
        # construct and ladder
        _, add_ctrl_op = self.ops[l].get_ctrl_system()
        s, target = add_ctrl_op(ctrl = [current selection bit], {'x': target})
        l += 1
        # unconstruct and ladder
```

```python
qrom = ApplyL([Set(d, target_bitsize) for d in data], selection_bitsize, target_bitsize)
```

tanujkhattar:
This should go in a tutorial notebook. We shouldn't have two QROM implementations.

tanujkhattar:
After `ApplyLthBloq` and `UnaryIterationGate`, I think this can be closed. The "basic" unary iteration can be accessed using the `ApplyLthBloq` interface and power users can derive from `UnaryIterationGate` for more control.

PR #1107, opened 2024-07-03
Create `ApplyLthBloq` as a simple SELECT oracle

charlesyuan314:
This bloq provides a simple way to coherently index into a list of bloqs that have identical signature and all THRU registers. Unlike the more specialized Select oracles that are currently available, it makes no additional assumptions on the behavior of these bloqs and deduces necessary register sizes automatically.

This bloq essentially implements `ApplyL` from #721. It will be used to implement linear combinations of block encodings.


qualtran/bloqs/multiplexers/apply_lth_bloq.py
```diff
@@ -0,0 +1,100 @@
...
+
+    References:
+        [Encoding Electronic Spectra in Quantum Circuits with Linear T Complexity](https://arxiv.org/abs/1805.03662).
+        Babbush et. al. (2018). Section III.A. and Figure 7.
+    """
+
+    ops: Tuple[Bloq, ...] = field(
+        converter=lambda x: x if isinstance(x, tuple) else tuple(x), validator=validators.min_len(1)
+    )
+    control: bool = False
```
tanujkhattar, line 53: Consider deriving from `SpecializedSingleQubitControlledGate` so you get the `get_ctrl_system` implementation for free; which is useful to support constructing controlled bloqs by calling `bloq.controlled()`

qualtran/bloqs/multiplexers/apply_lth_bloq.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def __attrs_post_init__(self):
+        if not all(u.signature == self.ops[0].signature for u in self.ops):
+            raise ValueError("All ops must have the same signature.")
+        if not all(r.side == Side.THRU for u in self.ops for r in u.signature):
+            raise ValueError("All ops must have only THRU registers.")
```
tanujkhattar, line 59: What's the motivation of this bloq with these restrictive constraints? Where will this be used?

qualtran/bloqs/multiplexers/apply_lth_bloq.py
```diff
@@ -0,0 +1,100 @@
...
+        [user_spec]: The output registers of the bloqs in `ops`.
+
+    References:
+        [Encoding Electronic Spectra in Quantum Circuits with Linear T Complexity](https://arxiv.org/abs/1805.03662).
+        Babbush et. al. (2018). Section III.A. and Figure 7.
+    """
+
+    ops: Tuple[Bloq, ...] = field(
+        converter=lambda x: x if isinstance(x, tuple) else tuple(x), validator=validators.min_len(1)
+    )
```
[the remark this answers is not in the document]
tanujkhattar: In fact, consider accepting a ndarray instead of a Tuple and make it a `Union[Shaped, NDArray[Bloq]]` so that users can specify N-dimensional SELECT operations.

The SELECT operations are often controlled on multiple selection registers. Expecting a list os `ops` implies that you expect users to create a single "contiguous" register and represent your LCU as a summation of a single variable. However, it's often preferred and is more optimal to keep the summation indices factorized and implement a SELECT operation multiplexed by multiple selection registers. 

For example, see Appendix A of https://arxiv.org/pdf/2011.03494

image


The unary iteration base class supports specifying multiple selection registers. You can look at implementation of `ApplyGateToLthQubit` or `SelectedMajoranaFermion` gates; both of which supports multiple selection registers

qualtran/bloqs/multiplexers/apply_lth_bloq.py
```diff
@@ -0,0 +1,100 @@
...
+        return tuple(self.ops[0].signature)
+
+    def nth_operation(
+        self,
+        context: cirq.DecompositionContext,
+        control: cirq.Qid,
+        selection: int,
+        **targets: Sequence[cirq.Qid],
+    ) -> cirq.OP_TREE:
+        return self.ops[selection].on_registers(**targets).controlled_by(control)
```
tanujkhattar, line 84: `self.ops[selection]` will give you a `Bloq`. To get a qualtran-style controlled bloq; you need to call `self.ops[selection].controlled()`. You can then apply this Bloq on qubits to get a cirq operation where; if needed; the bloq gets wrapped in `BloqAsCirqGate` if the bloq is not a `GateWithRegisters`. 

This requires you to know the name of the control registers so you can pass in the right qubits (imagine how you would do it in qualtran style). This has a limitation because we may not know what the name of the control register is; especially of the `Bloq` itself is a `Controlled` bloq. 


Using the cirq interop, we can get around this constraint because the cirq style API doesn't force you to provide qubits by keyword arguments. So you can do something like

```python
bloq = self.ops[selection]
target_qubits = merge_qubits(bloq.signature, **targets)
return bloq.controlled().on(control, *target_qubits)
```



What you have right now will return a `cirq.ControlledOperation()` where the `sub_operation` is `BloqAsCirqGate.on(*targets)` -- We want to avoid this. 

I think the fix to support this line and yet do the right thing would be to have a special `GWROperation` class that derives from `cirq.Operation` and overrides the `controlled_by` method to delegate to `Controlled` bloq; but that's out of scope for this PR.

qualtran/bloqs/multiplexers/apply_lth_bloq.py
```diff
@@ -0,0 +1,100 @@
...
+            Register(
+                'selection', BoundedQUInt(int(np.ceil(np.log2(len(self.ops)))), len(self.ops))
+            ),
+        )
+
+    @cached_property
+    def target_registers(self) -> Tuple[Register, ...]:
+        return tuple(self.ops[0].signature)
+
+    def nth_operation(
```
tanujkhattar, line 77: Please implement `nth_operation_callgraph` and / or `build_call_graph` to support symbolics. You can also consider overriding the `break_early` method if you expect this to be used in a situation where many of the subbloqs are identical and we can optimize the implementation of SELECT to apply the subbloqs for the entire range of selection values only once. You can look at the QROMBase class for an example of how this is used.

tanujkhattar (changes requested):
Left a round of comments. Curious to learn more about the motivation for this Bloq. 

I would also suggest to update the existing `ApplyGateToLthQubit` instead of adding a new `ApplyLthBloq` since this seems like a generalization of the one that exists (with a caveat that we need a wrapper which would have a signature acting on "all" qubits but would decompose into specific subsets of operations acting on a subset of target qubits) -- but I'm happy to accept this as a new Bloq as well.

qualtran/bloqs/multiplexers/apply_lth_bloq.py
charlesyuan314: This bloq will be used to implement linear combinations of block encodings given a list of those constituent block encodings and their corresponding coefficients. 

Before creating this bloq, I checked the existing implementations of Select oracles, and did not find any whose interface simply accepts a list of bloqs and coherently indexes into that list as needed for the block encoding.

I could be wrong, but the constraints here (all ops in the list have the same signature of only THRU registers) are strictly weaker than certain other existing Select oracles (e.g. `ApplyGateToLthQubit`, which additionally requires the gates to be single-qubit). 

Is there a way in which you find these constraints to be inappropriately restrictive?

charlesyuan314: As you mentioned in your main comment:

> Wouldn't the restriction that all block encodings in the LCU should have the same signature be restrictive?

Could you give an example of this condition being overly restrictive? Our quantum interface for block encodings already requires these signatures to be identical up to the sizes of `ancilla` and `resource` registers. In this case, the implementation of the `LinearCombination` block encoding is responsible for harmonizing these registers --- I've implemented that logic as part of a later PR.

The reason I didn't generalize the logic to harmonize registers and put it into this bloq is because the logic seems to be case-dependent. For example, the bits of the `ancilla` register are reused among the constituent block encodings, but it's unclear whether this should hold of arbitrary other registers in the signatures of bloqs that `ApplyLthBloq` selects from.

charlesyuan314:
> I would also suggest to update the existing `ApplyGateToLthQubit` instead of adding a new `ApplyLthBloq` since this seems like a generalization of the one that exists (with a caveat that we need a wrapper which would have a signature acting on "all" qubits but would decompose into specific subsets of operations acting on a subset of target qubits) 

I agree that something like that wrapper would be needed to implement this functionality by generalizing the existing `ApplyGateToLthQubit`. I'm happy to think about this more once we reach agreement on the nature of and restrictions on the signatures of the constituents in `ops`. Tracking in the comment thread [above](https://github.com/quantumlib/Qualtran/pull/1107#discussion_r1664947150).

qualtran/bloqs/multiplexers/apply_lth_bloq.py
charlesyuan314: Incorporated the suggested snippet above.

qualtran/bloqs/multiplexers/apply_lth_bloq.py
tanujkhattar: Suppose you want to construct a block encoding for an SK model hamiltonian. The hamiltonian is represented as

$$
H = \sum_{i < j} w_{i, j} Z_{i} Z_{j}
$$

We can rewrite the hamiltonian as a summation over a single contiguous index $k$ that goes from $0$ to $\frac{n(n-1)}{2}$ s.t.

$$
H = \sum_{k} w_{k} Z_{k, 0} Z_{k, 1}
$$

Now I have an LCU over $\frac{n(n-1)}{2}$ terms. Here each term is a `Unitary` block encoding with `U=ZZ` gate; i.e. `Unitary(U=ZZ, alpha=1, ancilla_bitsize=0, resource_bitsize=0, epsilon=0)`. The `system_bitsize` would be 2 since the unitary acts on 2 qubits. 

Now, if I do the more "intuitive" thing and simply construct an array of size $k$ where each element is a unitary block encoding as described above and pass it to your `LCUBlockEncoding` bloq to construct a block encoding for my hamiltonian, then I'll get a wrong result. This is because the current interface does not encode information about "which qubits you want to apply the i'th bloq on". To encode this information, I'll have to wrap all these unitaries in a wrapper which encodes in the information about `Z_{k, 0} Z_{k, 1}` such that each bloq acts on all the $k$ qubits instead of only two qubits and then pass this list of wrappers to the `LCUBlockEncoding` interface. 


Instead of forcing users to wrap every term of the LCU in a wrapper that acts on the entire system register; we can also provide an abstraction which accepts a `Bloq` and maybe an ndarray of indices that tell you on which subset of qubits from your system register you want to apply the bloq on to construct the block encoding. 

But I'm just curious to understand better if you've already thought about this problem and have a preferred approach one way or the other. 

> Before creating this bloq, I checked the existing implementations of Select oracles, and did not find any whose interface simply accepts a list of bloqs and coherently indexes into that list as needed for the block encoding.

This is because all existing Select oracles are for specific problem instances where we have more problem specific information about "where to apply the i'th bloq" and so we didn't need a generalized version of indexing into the i'th bloq given a list of bloqs.

qualtran/bloqs/multiplexers/apply_lth_bloq.py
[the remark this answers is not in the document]
charlesyuan314: > Consider making `ops` a `Union[HasLength, Tuple[Bloq, ...]]` to support symbolics.

One challenge with making the list symbolic and length- or shape-only here is that it's not clear how to compute the call graph if we don't know concretely what the constituent bloqs are --- the number of them doesn't seem to be enough information. We had the same dilemma with `TensorProduct` in #1094.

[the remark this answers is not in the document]
charlesyuan314: > consider accepting a ndarray instead of a Tuple and make it a `Union[Shaped, NDArray[Bloq]]` so that users can specify N-dimensional SELECT operations.

Setting aside the symbolic issue of `HasLength`/`Shaped` above, I took a stab at switching to `NDArray`. The issue is that `NDArray` is a mutable/non-hashable type, which breaks the use of bloqs as keys in sets/dicts by the call graph infrastructure:

```
qualtran/_infra/bloq.py:343: in call_graph
    return get_bloq_call_graph(self, generalizer=generalizer, keep=keep, max_depth=max_depth)
qualtran/resource_counting/_call_graph.py:241: in get_bloq_call_graph
    _build_call_graph(bloq, generalizer, ssa, keep, max_depth, g=g, depth=0)
qualtran/resource_counting/_call_graph.py:142: in _build_call_graph
    g.add_node(bloq)
../.virtualenvs/qualtran/lib/python3.11/site-packages/networkx/classes/digraph.py:460: in add_node
    if node_for_adding not in self._succ:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = ApplyLthBloq(ops=array([TGate(is_adjoint=False), Hadamard(), ZGate(), XGate()],
      dtype=object), selection_regs=None, control_val=1)

    def __hash__(self):
>           return hash((
                    -8356619234005614914,
                    self.ops,
                    self.selection_regs,
                    self.control_val,
                ))
E           TypeError: unhashable type: 'numpy.ndarray'
```

I haven't found examples of existing bloqs that have `NDArray` attributes and overcome this problem. The closest examples are bloqs that make the attribute immutable by fixing its dimension, e.g. to `Tuple[Tuple[Bloq, ...], ...]`. But that isn't what we want either.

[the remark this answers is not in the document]
charlesyuan314: One way to potentially avoid the above issue is to emulate `ApplyGateToLthQubit` and use a `Callable[..., Bloq]` rather than a `NDArray[Bloq]` to define the nth bloq. Assuming the callable is hashed via its pointer this may work, but I haven't worked out how good the resulting interface looks to the bloq user.

It seems `SelectedMajoranaFermion` does not encounter this issue because it only stores a single gate that it applies selectively.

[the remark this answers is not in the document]
tanujkhattar: Here is an example - https://github.com/quantumlib/Qualtran/blob/3315602fc762b3b94300d449aebe2a23fbddb29a/qualtran/bloqs/data_loading/qrom_base.py#L153

You define the attribute as an `attrs.field()` and use a custom function to convert an NDArray to a flattened tuple that can be used in comparisons and hashing by overriding the `eq=_ndarray_to_tuple`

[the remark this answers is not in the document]
charlesyuan314: I took a stab at using the `Callable` approach in the latest commit, but I'm not sure the code is more robust than in the one-dimensional case. Happy to receive feedback.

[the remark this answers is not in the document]
tanujkhattar: Sorry, what's the benefit of using a `Callable` instead of directly expecting an NDArray in the input and using the approach I described to make the NDArray hashable / comparable? A callable has a drawback that its not (a) comparable (b) hashable and (c) serializable. Expecting an NDArray directly would fix all these issues.

qualtran/bloqs/multiplexers/apply_lth_bloq.py
```diff
@@ -0,0 +1,127 @@
...
+    @cached_property
+    def selection_registers(self) -> Tuple[Register, ...]:
+        if self.selection_regs is None:
+            return (
+                Register(
+                    'selection',
+                    BoundedQUInt(int(ceil(log2(len(self._ops.flat)))), len(self._ops.flat)),
+                ),
+            )
+        return self.selection_regs
```
tanujkhattar, line 86: My original suggestion to expect an `NDArray` of `Bloq`s instead of a flat list was so that we can use multiple selection registers instead of a single one and have a more accurate representation of a  "nested for-loop" to index into the NDArray.

Thus, we should have `len(self._ops.shape)` different selection registers instead of a single selection register of of length `self._ops.flat`.

qualtran/bloqs/multiplexers/apply_lth_bloq.py
charlesyuan314: > Instead of forcing users to wrap every term of the LCU in a wrapper that acts on the entire system register; we can also provide an abstraction which accepts a Bloq and maybe an ndarray of indices that tell you on which subset of qubits from your system register you want to apply the bloq on to construct the block encoding.

Thanks for giving the example. It makes sense and was not a case that I had previously considered.

If it's OK with you, I'd like to add support for such unitaries that act over distinct parts of the system register in a follow-up PR after this one.

qualtran/bloqs/multiplexers/apply_lth_bloq.py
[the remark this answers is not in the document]
charlesyuan314: Thanks for the suggestion of the conversion. I'll give it a try to replace the `Callable`.

qualtran/bloqs/multiplexers/apply_lth_bloq.py
charlesyuan314: That makes sense. I'll make the change.

charlesyuan314:
I made the suggested changes to feature an `NDArray` of bloqs to iterate over.

After merging this PR, I plan to create an issue to track support by `ApplyLthBloq` for sub-bloqs acting over different subsets of the system register. Afterwards, it would be possible to consider merging the capabilities of `ApplyLthBloq` and `ApplyGateToLthQubit`.

qualtran/bloqs/multiplexers/apply_lth_bloq.py
mpharrigan: Thanks for the write-up @tanujkhattar . If I understand it correctly, you're objecting to wrapping each operation so it has a consistent set of registers. This introduces a level of indirection; but makes a very conceptually simple bloq composition. If I understand it correctly, it would additionally unblock having a symbolic decomposition of a SELECT oracle which has myriad benefits

qualtran/bloqs/multiplexers/apply_lth_bloq.py
```diff
@@ -0,0 +1,121 @@
...
+
+@frozen
+class ApplyLthBloq(UnaryIterationGate, SpecializedSingleQubitControlledGate):  # type: ignore[misc]
+    r"""A SELECT operation that executes one of a list of bloqs $U_l$ based on a quantum index:
+
+    $$
+    \mathrm{SELECT} = \sum_{l}|l \rangle \langle l| \otimes U_l
+    $$
+
+    This bloq uses the unary iteration scheme to apply `ops[selection]` controlled on an optional single-bit `control` register.
```
mpharrigan, line 38: nit: wrap to 100 chars

mpharrigan (approved):
This LGTM. @tanujkhattar do you have any additional notes?

qualtran/bloqs/multiplexers/apply_lth_bloq.py
```diff
@@ -0,0 +1,121 @@
...
+
+    Args:
+        ops: List of bloqs. Each bloq must have identical registers that are all THRU.
+        selection_regs: List of selection registers, 1D selection index inferred if not provided.
```
tanujkhattar, line 42: This seems out of sync - we infer an n-D selection index now if not provided.

qualtran/bloqs/multiplexers/apply_lth_bloq.py
```diff
@@ -0,0 +1,121 @@
...
+        selection: The indices of the bloq in `ops` to execute.
+        control: The control bit if specified above.
+        [user_spec]: The output registers of the bloqs in `ops`.
+
+    References:
+        [Encoding Electronic Spectra in Quantum Circuits with Linear T Complexity](https://arxiv.org/abs/1805.03662).
+        Babbush et. al. (2018). Section III.A. and Figure 7.
+    """
+
+    ops: NDArray[Bloq] = field(  # type: ignore[type-var]
```
tanujkhattar, line 55: Why do we need the type ignore? Due to the lambda function? Can we write the converter as a separate method to avoid type ignore?

qualtran/bloqs/multiplexers/apply_lth_bloq.py
```diff
@@ -0,0 +1,121 @@
...
+        converter=lambda x: np.array(x) if isinstance(x, Iterable) else x,
+        eq=lambda d: tuple(d.flat),
+    )
+    selection_regs: Optional[Tuple[Register, ...]] = None
+    control_val: Optional[int] = None
+
+    def __attrs_post_init__(self):
+        if cast(int, prod(self.ops.shape)) <= 1:
```
tanujkhattar, line 63: Since `self.ops` is an NDArray, we can just use `np.prod` with `self.ops.shape` and not worry about casting.

qualtran/bloqs/multiplexers/apply_lth_bloq.py
```diff
@@ -0,0 +1,121 @@
...
+                for i, k in enumerate(self.ops.shape)
+            )
+        return self.selection_regs
+
+    @cached_property
+    def target_registers(self) -> Tuple[Register, ...]:
+        return tuple(self.ops.flat[0].signature)
+
+    def nth_operation_callgraph(self, ssa: SympySymbolAllocator, **kwargs: int) -> Set[BloqCountT]:
+        return self.ops[tuple(kwargs.values())].controlled().build_call_graph(ssa)
```
tanujkhattar, line 91: We can just return `{self.ops[tuple(kwargs.values())].controlled()}`  instead of explicitly calling the `build_call_graph` of the subbloq. The call graph of `ApplyLthBloq` should simply decompose into `self.ops` and their subgraph would be inferred automatically by the call graph infrastructure.

qualtran/bloqs/multiplexers/apply_lth_bloq.py
```diff
@@ -0,0 +1,121 @@
...
+    def nth_operation(
+        self,
+        context: cirq.DecompositionContext,
+        control: cirq.Qid,
+        **kwargs: Union[int, Sequence[cirq.Qid]],
+    ) -> cirq.OP_TREE:
+        sel_regs = {r.name for r in self.selection_registers}
+        selection_indices = list(cast(int, v) for k, v in kwargs.items() if k in sel_regs)
+        tgt_regs = {r.name for r in self.target_registers}
+        targets = {k: cast(Sequence[cirq.Qid], v) for k, v in kwargs.items() if k in tgt_regs}
```
tanujkhattar, line 102:
```suggestion
        targets = {reg.name: cast(Sequence[cirq.Qid], kwargs[reg.name]) for reg in self.target_registers}
```

qualtran/bloqs/multiplexers/apply_lth_bloq.py
```diff
@@ -0,0 +1,121 @@
...
+        return self.ops[tuple(kwargs.values())].controlled().build_call_graph(ssa)
+
+    def nth_operation(
+        self,
+        context: cirq.DecompositionContext,
+        control: cirq.Qid,
+        **kwargs: Union[int, Sequence[cirq.Qid]],
+    ) -> cirq.OP_TREE:
+        sel_regs = {r.name for r in self.selection_registers}
+        selection_indices = list(cast(int, v) for k, v in kwargs.items() if k in sel_regs)
```
tanujkhattar, line 100:
```suggestion
        selection_indices = list(cast(int, kwargs[reg.name]) for reg in self.selection_registers)
```

tanujkhattar (approved):
LGTM % comments.

qualtran/bloqs/multiplexers/apply_lth_bloq.py
tanujkhattar: > If I understand it correctly, you're objecting to wrapping each operation so it has a consistent set of registers

Quoting from my comment above "But I'm just curious to understand better if you've already thought about this problem and have a preferred approach one way or the other." 

I am not objecting to anything. I was just pointing out that we will have cases where the LCU block encoding would be composed of parts which act on different parts of the system. We should either provide a wrapper or an abstraction to capture which bloq acts on which part of the subsystem. Both of these would be fine ways of doing things. 

If we use a wrapper, we also risk exposing an API that users can use incorrectly -- see my example above on following the more intuitive approach and how that leads to a bloq which doesn't raise an error on construction but is an incorrect bloq encoding.

qualtran/bloqs/multiplexers/apply_lth_bloq.py
charlesyuan314: This is due to `NDArray[Bloq]`. Strictly speaking, `numpy.typing.NDArray` requires its type argument to be a numpy dtype such as `float64`, not `Bloq`.

We already do `# type: ignore[type-var]` for `NDArray[Soquet]`, etc. throughout the code base. I'll add a comment here for clarification.

charlesyuan314:
I've addressed all of the latest round of comments.

PR #1108, opened 2024-07-03
Move `SelectOracle` and `PrepareOracle` to `qualtran/interfaces/`

charlesyuan314:
Move the definitions of `SelectOracle` and `PrepareOracle` to a new directory `qualtran/interfaces`, so as to remove the unnecessary dependency on block encodings, which become circular with new work on block encoding implementations.

Change the base class of `SelectOracle` and `PrepareOracle` to the less restrictive `Bloq` rather than `GateWithRegisters`, and update instances appropriately.


tanujkhattar:
What's the cause of the circular dependency? Can you please provide some more context? 

I don't think we should have a generic `bloqs/interfaces` -- we have interfaces for many different types of bloqs (eg: (a) unary iteration (b) QROMs) and will continue to have more (eg: arithmetic bloqs). We should place the interfaces in the directory where they belong. In this case, the SELECT and PREPARE oracles are specific for LCU based block encodings so they should be place appropriately.

tanujkhattar:
@charlesyuan314 `qualtran.bloqs.block_encodings.linear_combination.LinearCombination` can import `qualtran.bloqs.multiplexers.apply_lth_bloq.ApplyLthBloq` locally in the `decompose_bloq` method instead of globally at the top level to avoid the circular dependency. Does this solve the problem? If not, there are also other tricks we can use like the `LazyLoader` from Cirq - https://github.com/quantumlib/Cirq/blob/32d48331fecd5f075c25e8f05c4551380bcd6c45/cirq-core/cirq/_import.py#L176

Once you open a PR, we can look into the technical details of how to avoid the circular dependency but I don't think we should move the LCU Select / Prepare interfaces to a new `qualtran/bloqs/<module>`.

charlesyuan314:
Sounds good, I will try the above approaches!

PR #1110, opened 2024-07-04
Permutation Bloq

anurudhp:
Implements a simple circuit that permutes basis states in `O(N logN)`, and `O(d logN)` for a d-prefix (used to prepare d-sparse states).


qualtran/bloqs/arithmetic/addition.py
```diff
@@ -359,6 +361,63 @@ def _cvs_converter(vv):
     return tuple(int(v) for v in vv)
 
 
+@frozen
+class XorK(Bloq):
```
mpharrigan, line 365: cool! I would probably move this to its own module in e.g. `arithmetic/bitwise.py` or something. but this organization isn't blocking this pr

qualtran/bloqs/arithmetic/comparison.py
```diff
@@ -953,10 +954,61 @@ def wire_symbol(self, reg: Optional[Register], idx: Tuple[int, ...] = tuple()) -
...
+            raise DecomposeTypeError(f"cannot decompose symbolic {self}")
+
+        bits_k = x.reg.dtype.to_bits(self.val)
+
+        if self.bitsize == 1:
+            if self.val == 0:
+                x = bb.add(XGate(), q=x)
+            x, target = bb.add(CNOT(), ctrl=x, target=target)
+            if self.val == 0:
+                x = bb.add(XGate(), q=x)
```
mpharrigan, line 973: can you add a note that this is just a negative-control CNOT so that in the future, if and when we add a `cv` attribute to `CNOT` we can replace it

qualtran/bloqs/arithmetic/comparison.py
```diff
@@ -953,10 +954,61 @@ def wire_symbol(self, reg: Optional[Register], idx: Tuple[int, ...] = tuple()) -
...
+            else:
+                op = MultiAnd(HasLength(self.bitsize))
+
+            bloq_counts: dict[Bloq, int] = defaultdict(lambda: 0)
+            bloq_counts[op] += 1
+            bloq_counts[op.adjoint()] += 1
+            bloq_counts[CNOT()] += 1
+            return set(bloq_counts.items())
+
+        return super().build_call_graph(ssa)
```
mpharrigan, line 1011: nit: I generally prefer inverting this so the method isn't as indented

```python
if not self.is_symbolic():
  return super().build_call_graph

op = ...
return bloq_counts
```

That being said --- my original design for the call graph protocol was always to provide the counts you'd expect to find in e.g. a paper, which would use the `is_symbolic` path even for concrete numbers. Then you can reconcile them in e.g. the unit test by using the `generalizer` argument and built-in `generalizer` functions. Maybe I'll open an issue because I know this sort of thing has come up in other places

qualtran/bloqs/arithmetic/permutation.py
```diff
@@ -0,0 +1,248 @@
...
+@frozen
+class PermutationCycle(Bloq):
+    r"""Apply a single permutation cycle on the basis states.
+
+    Args:
+        N: the total size the permutation acts on.
+        cycle: the permutation cycle to apply.
+
+    Registers:
+        q: integer register storing a value in [0, ..., N - 1]
```
mpharrigan, line 75: lets reserve the register name "q" for a single qubit. We generally use "x" (my pref) or "a" for an arbitrary value

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -359,6 +361,63 @@ def _cvs_converter(vv):
...
+
+        soqs['x'] = bb.join(xs)
+        if ctrls is not None:
+            soqs['ctrls'] = ctrls
+        return soqs
+
+    def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
+        bit_flip_bloq = MultiControlX(cvs=self.cvs) if len(self.cvs) > 0 else XGate()
+        num_flips = self.bitsize if self.is_symbolic() else sum(self.dtype.to_bits(self.k))
+        return {(bit_flip_bloq, num_flips)}
```
tanujkhattar, line 418: We shouldn't do `n` `MCX` gates to do `n` bit flips when number of controls > 1

Instead, divide it into 3 cases:

1) Num controls == 0 : Simply do an XGate()
2) Num controls == 1 : Simply do CNOT() gates
3) Num controls > 1 : Use a temporary ancilla and do a `MultiAnd(cvs=self.cvs)`  with the temporary ancilla as the target. Then reduce to Case(2) with the temporary ancilla as control.

tanujkhattar: In fact, I'd suggest we only support Case (1) and (2) and reducing Case (3) to Case (1) is a general strategy used for all multi controlled bloq so we can add a separate Bloq that does this reduction (maybe at a later point)

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -359,6 +361,63 @@ def _cvs_converter(vv):
...
+        if self.is_symbolic():
+            raise DecomposeTypeError(f"cannot decompose symbolic {self}")
+
+        xs = bb.split(cast(Soquet, soqs.pop('x')))
+        ctrls = soqs.pop('ctrls', None)
+
+        for i, bit in enumerate(self.dtype.to_bits(self.k)):
+            if bit == 1:
+                if len(self.cvs) > 0 and ctrls is not None:
+                    ctrls, xs[i] = bb.add(MultiControlX(cvs=self.cvs), ctrls=ctrls, x=xs[i])
```
tanujkhattar, line 406: ditto: Please don't use `MultiControlX` inside the loop. See my comment below

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -359,6 +361,63 @@ def _cvs_converter(vv):
...
+        k: The classical integer value to be XOR-ed to x.
+        cvs: A tuple of control values. Each entry specifies whether that control line is a
+            "positive" control (`cv[i]=1`) or a "negative" control (`cv[i]=0`).
+
+    Registers:
+        x: A quantum register of type `self.dtype` (see above).
+        ctrls: A sequence of control qubits (only when `cvs` is non-empty).
+    """
+    dtype: QDType
+    k: SymbolicInt
```
tanujkhattar, line 379: If `k` is a `SymbolicInt` should we restrict `dtype` to be `QUInt` / `QInt` ? Looking at the implementation below, it seems like this would also work for other types when `k` is not necessarily an integer.

qualtran/bloqs/arithmetic/comparison.py
```diff
@@ -956,10 +957,62 @@ def wire_symbol(self, reg: Optional[Register], idx: Tuple[int, ...] = tuple()) -
             return TextBox(f"⨁(x = {self.val})")
         raise ValueError(f'Unknown register symbol {reg.name}')
 
+    def is_symbolic(self):
+        return is_symbolic(self.bitsize, self.val)
+
+    def build_composite_bloq(
+        self, bb: 'BloqBuilder', x: 'Soquet', target: 'Soquet'
+    ) -> Dict[str, 'SoquetT']:
```
tanujkhattar, line 965: You can shorten the decomposition significantly by simply delegating to `MultiControlPauli(cvs=bits_k, target_gate=cirq.X).on(*x, *target)`. 


This is also cleaner IMO. We will soon update the `MultiControlPauli` bloq to not use `cirq.X` and instead use `XGate()` so that shouldn't be a blocker for using the bloq here.

qualtran/bloqs/arithmetic/comparison.py
```diff
@@ -956,10 +957,62 @@ def wire_symbol(self, reg: Optional[Register], idx: Tuple[int, ...] = tuple()) -
...
+                cv = ssa.new_symbol('cv')
+                op = And(cv, cv)
+        else:
+            op = MultiAnd(HasLength(self.bitsize))
+
+        bloq_counts: dict[Bloq, int] = defaultdict(lambda: 0)
+        bloq_counts[op] += 1
+        bloq_counts[op.adjoint()] += 1
+        bloq_counts[CNOT()] += 1
+        return set(bloq_counts.items())
```
tanujkhattar, line 1015: ditto: Simply delegate to `MultiControlPauli`.

qualtran/bloqs/arithmetic/permutation.py
```diff
@@ -0,0 +1,248 @@
...
+def _convert_cycle(cycle) -> Union[tuple[int, ...], Shaped]:
+    if isinstance(cycle, Shaped):
+        return cycle
+    return tuple(cycle)
+
+
+@frozen
+class PermutationCycle(Bloq):
+    r"""Apply a single permutation cycle on the basis states.
+
```
tanujkhattar, line 69: Can you add some more details on what does it mean to "apply a permutation cycle" ? Maybe add a latex equation for the action of unitary on an input state?

qualtran/bloqs/arithmetic/permutation.py
```diff
@@ -0,0 +1,248 @@
...
+    N: SymbolicInt
+    cycle: Union[tuple[int, ...], Shaped] = field(converter=_convert_cycle)
+
+    @cached_property
+    def signature(self) -> Signature:
+        return Signature.build_from_dtypes(x=BoundedQUInt(self.bitsize, self.N))
+
+    @cached_property
+    def bitsize(self):
+        return bit_length(self.N - 1)
```
tanujkhattar, line 91: Will there ever be a case when the user wants to specify the `bitsize` instead of the default `log2(N)` ? We can make it an attribute and use this as a default value if the user doesn't specify anything; if you think that's relevant.

qualtran/bloqs/arithmetic/permutation.py
```diff
@@ -0,0 +1,248 @@
...
+        return cycles
+    return tuple(_convert_cycle(cycle) for cycle in cycles)
+
+
+@frozen
+class Permutation(Bloq):
+    """Apply a permutation of [0, N - 1] on the basis states.
+
+    Decomposes a permutation into cycles and applies them in order.
+    See :meth:`from_dense_permutation` to construct this bloq from a permutation.
```
tanujkhattar, line 154: Ditto: Please add a latex equation describing the action of the unitary on an input state.

qualtran/bloqs/arithmetic/permutation.py
```diff
@@ -0,0 +1,248 @@
...
+    @classmethod
+    def from_dense_permutation(cls, permutation: Sequence[int]):
+        N = len(permutation)
+        cycles = tuple(decompose_permutation_into_cycles(permutation))
+        return cls(N, cycles)
+
+    @classmethod
+    def from_sparse_permutation_prefix(cls, N: int, permutation_prefix: Sequence[int]):
+        cycles = tuple(decompose_sparse_prefix_permutation_into_cycles(permutation_prefix, N))
+        return cls(N, cycles)
```
tanujkhattar, line 193: Please add a docstring explaining `permutation_prefix` - copying the docstring from `decompose_sparse_prefix_permutation_into_cycles` should suffice. 

For specifying a sparse permutation, I would naturally think of specifying a dictionary instead of a sequence of prefix but this also LGTM % adding docstrings for clarity

tanujkhattar (approved):
LGTM % comments

qualtran/bloqs/arithmetic/addition.py
anurudhp: moved to `arithmetic/bitwise.py`

qualtran/bloqs/arithmetic/comparison.py
anurudhp: Inverted the method.

qualtran/bloqs/arithmetic/addition.py
anurudhp: I simplified this to only support a single `control_val`, as #1131 will handle the CtrlSpec -> QBit part.

qualtran/bloqs/arithmetic/comparison.py
anurudhp: fixed

qualtran/bloqs/arithmetic/comparison.py
anurudhp: removed this (as the decomp is trivial and works for symbolic)

qualtran/bloqs/arithmetic/permutation.py
anurudhp: added the unitary action on basis states

qualtran/bloqs/arithmetic/permutation.py
anurudhp: used a dictionary instead, actually made the code simpler!

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -490,23 +431,46 @@ def build_composite_bloq(
...
+        # Apply XGates to qubits in k where the bitstring has value 1. Apply CNOTs when the gate is
+        # controlled.
+        for i in range(self.bitsize):
+            if binary_rep[i] == 1:
+                if len(self.cvs) > 0 and ctrls is not None:
+                    ctrls, k_split[i] = bb.add(
+                        MultiControlX(cvs=self.cvs), ctrls=ctrls, x=k_split[i]
+                    )
+                else:
+                    k_split[i] = bb.add(XGate(), q=k_split[i])
```
tanujkhattar, line 450: Sorry, why are we doing this change here instead of using `XorK` ? I think this is added after my last review?

qualtran/bloqs/arithmetic/permutation.py
```diff
@@ -67,6 +67,20 @@ def _convert_cycle(cycle) -> Union[tuple[int, ...], Shaped]:
...
+        $$
+
+    for each $i \in [0, k)$, and
+
+        $$
+            U|x\rangle \mapsto |x\rangle
+        $$
+
+    and for every $x \not\in C$.
+
```
tanujkhattar, line 83: Is this incomplete?

qualtran/conftest.py
```diff
@@ -97,6 +97,12 @@ def assert_bloq_example_serializes_for_pytest(bloq_ex: BloqExample):
         'tensor_product_block_encoding',
         'tensor_product_block_encoding_properties',
         'tensor_product_block_encoding_symb',
+        'product_block_encoding',
+        'product_block_encoding_properties',
+        'product_block_encoding_symb',
+        'apply_lth_bloq',
+        'linear_combination_block_encoding',
+        'phase_block_encoding',
```
tanujkhattar, line 105: Why are the block encodings being added here as part of this PR ?

qualtran/bloqs/arithmetic/addition.py
tanujkhattar: nvm, I got confused due to "show changes since your last review"

qualtran/conftest.py
tanujkhattar: ditto, please ignore.

tanujkhattar:
LGTM!

PR #1109, opened 2024-07-04
More basic gates

mpharrigan:
I'm going to split these up into independent PRs, but

 - controlled global phase gate gives you a Z
 - Controlled Hadamard
 - Controlled Z
 - Controlled Y
 - make `cirq_gate_as_bloq` a public function

PR #1112, opened 2024-07-09
Make `cirq_gate_as_bloq` public

mpharrigan:
This is generally useful when switching between Cirq gates and bloqs that are basic gates; part of #1109

### DIFF
--- a/qualtran/cirq_interop/_cirq_to_bloq.py
+++ b/qualtran/cirq_interop/_cirq_to_bloq.py
@@ -319,7 +319,12 @@ def _gather_input_soqs(
     return qvars_in
 
 
-def _cirq_gate_to_bloq(gate: cirq.Gate) -> Bloq:
+def cirq_gate_to_bloq(gate: cirq.Gate) -> Bloq:
+    """For a given Cirq gate, return an equivalent bloq.
+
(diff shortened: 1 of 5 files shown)


qualtran/cirq_interop/_cirq_to_bloq.py
```diff
@@ -378,6 +378,11 @@ def _cirq_gate_to_bloq(gate: cirq.Gate) -> Bloq:
     if gate in CIRQ_GATE_TO_BLOQ_MAP:
         return CIRQ_GATE_TO_BLOQ_MAP[gate]
 
+    if isinstance(gate, cirq.ControlledGate):
+        return Controlled(
+            cirq_gate_to_bloq(gate.sub_gate), CtrlSpec.from_cirq_cv(gate.control_values)
+        )
+
```
tanujkhattar, line 385: Why this change?

mpharrigan: for #1115  or any case where we have an explicit bloq that takes priority over the general `Controlled` wrap.

mpharrigan:
(the other basic gates were based on top of this change, so merging this will cause those tests to pass)

tanujkhattar (approved):

PR #1113, opened 2024-07-09
Controlled GlobalPhase is ZGate

mpharrigan:
The single-bit control of a global phase is `ZGate`. 

More basic gates #1109

### DIFF
--- a/qualtran/bloqs/basic_gates/global_phase.py
+++ b/qualtran/bloqs/basic_gates/global_phase.py
@@ -12,26 +12,45 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from typing import Dict, List, TYPE_CHECKING
+from typing import Dict, Iterable, List, Optional, Sequence, Tuple, TYPE_CHECKING
 
 import attrs
 import cirq
+import numpy as np
 from attrs import frozen
 
-from qualtran import bloq_example, BloqDocSpec, ConnectionT, DecomposeTypeError
+from qualtran import (
+    AddControlledT,
+    Bloq,
+    bloq_example,
+    BloqBuilder,
+    BloqDocSpec,
+    CompositeBloq,
+    ConnectionT,
+    CtrlSpec,
+    DecomposeTypeError,
+    SoquetT,
+)
 from qualtran.cirq_interop import CirqGateAsBloqBase
 from qualtran.cirq_interop.t_complexity_protocol import TComplexity
 from qualtran.symbolics import sconj, SymbolicComplex
 
+from .rotation import ZPowGate
+
 if TYPE_CHECKING:
     import quimb.tensor as qtn
-
-    from qualtran import CompositeBloq
 
 
 @frozen
 class GlobalPhase(CirqGateAsBloqBase):
-    """Global phase operation of $z$ (where $|z| = 1$).
+    """Applies a global phase to the circuit as a whole.
+
+    The unitary effect is to multiply the state vector by the complex scalar
+    $z$ where $|z| = 1$.
+
+    The global phase of a state or circuit does not affect any observable quantity, but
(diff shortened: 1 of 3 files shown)


qualtran/bloqs/basic_gates/global_phase.py
```diff
@@ -58,6 +77,27 @@ def my_tensors(
...
+        self, ctrl_spec: Optional['CtrlSpec'] = None
+    ) -> Tuple['Bloq', 'AddControlledT']:
+
+        # Delegate to superclass logic for more than one control.
+        if not (ctrl_spec is None or ctrl_spec == CtrlSpec()):
+            return super().get_ctrl_system(ctrl_spec=ctrl_spec)
+
+        # Otherwise, it's a ZPowGate
+        exponent: float = np.real_if_close(np.log(self.coefficient) / (np.pi * 1.0j)).item()  # type: ignore
+        bloq = ZPowGate(exponent=exponent, eps=self.eps)
```
tanujkhattar, line 90: Please handle the case when `self.coefficient` is symbolic

tanujkhattar: Should we also have an override where we return a `ZGate` instead of `ZPowGate` if `exponent` is `1` ?

qualtran/bloqs/basic_gates/global_phase.py
```diff
@@ -58,6 +77,27 @@ def my_tensors(
...
+    def get_ctrl_system(
+        self, ctrl_spec: Optional['CtrlSpec'] = None
+    ) -> Tuple['Bloq', 'AddControlledT']:
+
+        # Delegate to superclass logic for more than one control.
+        if not (ctrl_spec is None or ctrl_spec == CtrlSpec()):
+            return super().get_ctrl_system(ctrl_spec=ctrl_spec)
+
+        # Otherwise, it's a ZPowGate
+        exponent: float = np.real_if_close(np.log(self.coefficient) / (np.pi * 1.0j)).item()  # type: ignore
```
anurudhp, line 89: this wouldn't work when `self.coefficient` is symbolic. Maybe we should update this to `self.exponent` to match the other rotation gates?

qualtran/bloqs/basic_gates/global_phase.py
mpharrigan: That's a good question, and I think it's part of a broader issue. How do we handle cases where you can express the same "thing" (unitary) via different bloq+attribute combinations? I think the answer is to keep things as non-magical as possible when constructing bloqs and handle overlap in the bloq analysis functions. Happy to discuss more.

For this PR in particular: since `ZPowGate(exponent=1)` is a valid bloq, I'll leave as-is. 

I'll handle symbolic coefficient

qualtran/bloqs/basic_gates/global_phase.py
mpharrigan: Thanks. 

Did you want to change the semantics to just store the exponent directly as the attribute or just change the name?

anurudhp: the former, then we can safely pass `self.exponent` to `ZPowGate`. But would require updating all call sites. Based on a preliminary scan, most of them are `1j, -1j, -1` and would simplify SU2: https://github.com/quantumlib/Qualtran/blob/3315602fc762b3b94300d449aebe2a23fbddb29a/qualtran/bloqs/basic_gates/su2_rotation.py#L131

I'm happy to take this PR if you'd like.

mpharrigan: please :)

anurudhp:
can be closed: this change was merged #1117

mpharrigan:
Thank you!

PR #1115, opened 2024-07-09
CYGate

mpharrigan:
Add the controlled Y gate 

Basic gates #1109

PR #1116, opened 2024-07-09
CZ

mpharrigan:
Add the controlled-Z CZ gate


qualtran/bloqs/basic_gates/z_basis.py
```diff
@@ -262,13 +283,67 @@ def as_cirq_op(
...
+
 
 @bloq_example
 def _zgate() -> ZGate:
     zgate = ZGate()
     return zgate
 
 
+@frozen
+class CZ(Bloq):
```
tanujkhattar, line 302: Please add `CZPow` similar to `ZPow`

mpharrigan:
I need to hook this up to the docs system, marking as draft

qualtran/bloqs/basic_gates/z_basis.py
tanujkhattar, line 302: https://github.com/quantumlib/Qualtran/pull/1115/files adds a `CYGate` whereas this PR adds a `CZ` gate. Let's keep the `CX / CY / CZ` family's name consistent -- let's either add `CPGate` or `CP` where `P` is the pauli in question.

qualtran/bloqs/basic_gates/z_basis.py
```diff
@@ -262,13 +283,67 @@ def as_cirq_op(
...
+    """Two-qubit controlled-Z gate.
+
+    Registers:
+        ctrl: One-bit control register.
+        target: One-bit target register.
+    """
+
+    @cached_property
+    def signature(self) -> 'Signature':
+        return Signature.build(q1=1, q2=1)
```
tanujkhattar, line 312: Should we just accept `Signature([Register('q', QBit(), shape=2)])` instead of a `q1` and `q2` ? This will keep things consistent with rotation bloqs deriving from `CirqGateAsBloqBase`. We can also follow the same pattern for 3 qubit gates instead of q1, q2, q3 etc.

tanujkhattar (approved):

qualtran/bloqs/basic_gates/z_basis.py
mpharrigan: I'll make the change in the other pr's to be `ctrl` and `target`. Conceivably, that could be the case here as well; but CZ is symmetric.

qualtran/bloqs/basic_gates/z_basis.py
mpharrigan: My original motivation was to use `CNOT` and `CZ` for those two gates since those are very common names. I.e. folks would expect to find a CNOT bloq and a CZ bloq through tab completion or searching the docs site. `CY` just doesn't feel as common to me, although I take the point on the consistency angle. But I really don't want to change `CNOT` to be `CX`.

PR #1122, opened 2024-07-10
Add `Unused` to `AutoPartition`

charlesyuan314:
Previously, `AutoPartition` could combine registers in the signature of the bloq being wrapped and expose the larger combined registers. This change enables it to also add new registers not known to the bloq being wrapped, which get passed through unaffected. The user can specify such a new register of `n` bits by using `Unused(n)` in place of the name of an existing register.

Also adds `BloqBuilder.add_d_and_partition` which is the same as `BloqBuilder.add_and_partition` but returns a dictionary.

This feature will be used to simplify the implementation of the `Product` block encoding and to implement the `LinearCombination` block encoding.


qualtran/_infra/composite_bloq.py
```diff
@@ -1042,6 +1046,18 @@ def add_and_partition(
...
+        self,
+        bloq: Bloq,
+        partitions: Sequence[Tuple[Register, Sequence[Union[str, 'Unused']]]],
+        left_only: bool = False,
+        **in_soqs: SoquetInT,
+    ) -> Dict[str, SoquetT]:
+        """Version of `add_and_partition` that returns a dictionary. See `BloqBuilder.add_d`."""
+        from qualtran.bloqs.bookkeeping.auto_partition import AutoPartition
+
+        return self.add_d(AutoPartition(bloq, partitions, left_only), **in_soqs)
```
mpharrigan, line 1059: I'm a little worried about continually expanding the API surface of `BloqBuilder` by having all these variants as methods, especially when it's just straightforwardly passing through each argument to existing methods. In this case, I think the one-liner `self.add_d(AutoPartition(...), ...)` is straightforward enough to use when writing decompositions directly. What do you think?

qualtran/bloqs/bookkeeping/auto_partition.py
```diff
@@ -40,6 +48,8 @@ class AutoPartition(Bloq):
         bloq: The bloq to wrap.
         partitions: A sequence of pairs specifying each register that the wrapped bloq should accept
             and the names of registers from `bloq.signature.lefts()` that concatenate to form it.
+            If the bloq being wrapped does not use a portion of the register being exposed, an
+            instance of `Unused` may be used in place of a register name from the bloq signature.
```
mpharrigan, line 52: I'd update the docstring to imply that unused takes a bitstring, i.e. put "instance of `Unused(n)` may be used..."

qualtran/_infra/composite_bloq.py
charlesyuan314: Yes, that is fair. I backed this out.

qualtran/bloqs/bookkeeping/auto_partition.py
mpharrigan: This docstring writ large is getting a little unwieldy if you don't already know what this bloq does. I've also noticed that the class docstring doesn't match the format of "short one-line summary, followed by details." My suggested text follows

```
"""Automatically partitions and unpartitions registers to match the signature of a sub-bloq

This can be used for adapting factored registers into a 'quantum interface'. Wrapping in AutoPartition
hides the partitioning splits and joins behind a level of decomposition, which can produce more
helpful circuit diagrams.

Args:
  bloq: The sub-bloq to wrap. Its register names are used as the 
    second items in the `partitions` tuples
  partitions: A sequence of pairs specifying how to partition each exterior register. The first element of 
    each pair is the `Register` that is exposed as part of AutoPartition's exterior signature. The 
    second element is a list of register names from the original subbloq, that, when combined, form the
    exposed register. If a portion of the exposed register is unused by the original subbloq, 
    the sentinel value `Unused(n)` can be used in place of a register name.

...
"""

mpharrigan: a suggestion ^ feel free to further revise

mpharrigan (commented):
Just some comments on docs and api surface. The functionality lgtm

qualtran/bloqs/bookkeeping/auto_partition.py
charlesyuan314: Thanks! I revised the docs a little further.

mpharrigan (approved):

PR #1127, opened 2024-07-11
remove `jupyter_kernel.lock` files

anurudhp:
seem to be introduced by #1060

@tanujkhattar


mpharrigan:
mea culpa. This file shows up because of my unreliable solution to get the notebooks to run in parallel

mpharrigan (approved):

PR #1128, opened 2024-07-11
Simplify `Product` block encoding decomposition

charlesyuan314:
This change refactors the `Product` block encoding, following [suggestion in discussion](https://github.com/quantumlib/Qualtran/pull/1106#discussion_r1671506044), to perform fewer partitions of the ancilla and resource registers up front and to rely on `AutoPartition` to simplify the circuit diagram of each constituent block.


mpharrigan:
I'm staring at the `product_block_encoding_properties.decompose_bloq()` diagram and I think this mostly makes sense. Why are the "ancilla" and "resource" registers still being split with different ones being fed into `B[U_a]` vs `B[U_b]`?

charlesyuan314:
With `ancilla`, we are doing just one split at the beginning into the `flag_bits` and the inner `ancilla` that is then fed to each of the constituents. The constituents use `AutoPartition` to ignore the unused bits. I have updated a comment to make it clearer that there are not separate individual `ancilla` registers.

With `resource`, there is a question of whether these bits are being returned to zero. If so, the constituents can share the resource bits in the same way as `ancilla` but without need for the flag bits. I have gone ahead with this assumption for now and changed the code to not split the `resource` register and rely only on `AutoPartition`. This cleans up the diagram.

It is possible that some block encodings will leave their `resource` state dirty and thus require flag bits here. I think properly supporting this behavior without using excessive bits could be done via a flag attribute in the `BlockEncoding` interface, analogous to the check of whether the `signal_state` is identity.

I can open an issue tracking the general fix above and add a code comment referencing it. Does that sound like a reasonable plan?

mpharrigan:
Thanks for the careful thought! Issue sounds like a good thought

mpharrigan (approved):

PR #1129, opened 2024-07-11
Add `Phase` block encoding

charlesyuan314:
Add an instance of `BlockEncoding` that applies a phase to the matrix being block encoded.

This bloq will be used to handle negative coefficients in `LinearCombination`.


qualtran/bloqs/block_encoding/negate.py
```diff
@@ -0,0 +1,151 @@
...
+from qualtran.bloqs.basic_gates import XGate, ZGate
+from qualtran.bloqs.block_encoding import BlockEncoding
+from qualtran.bloqs.block_encoding.lcu_select_and_prepare import PrepareOracle
+from qualtran.bloqs.bookkeeping.auto_partition import AutoPartition, Unused
+from qualtran.resource_counting import BloqCountT, SympySymbolAllocator
+from qualtran.symbolics import SymbolicFloat, SymbolicInt
+
+
+@frozen
+class Negate(BlockEncoding):
```
tanujkhattar, line 39: Can we write something that absorbs an arbitrary phase instead of just `-1` ?

tanujkhattar: nvm, I guess absorbing the negatives is sufficient since we are guaranteed that the LCU coefficients are real numbers and LCU can already work with arbitrary positive numbers ?

anurudhp: I was also wondering this, do LCU hamiltonians always have real coeffs?

charlesyuan314: I was pondering this. The only cases I've seen for block encodings have real coefficients. But I can change this to an arbitrary phase if we want to be more general.

charlesyuan314: Thanks @anurudhp for pointing out that `GlobalPhase` works here to enable us to apply an arbitrary phase and also save the resource qubit.

qualtran/bloqs/block_encoding/phase.py
```diff
@@ -0,0 +1,122 @@
...
+from qualtran.resource_counting import BloqCountT, SympySymbolAllocator
+from qualtran.symbolics import SymbolicFloat, SymbolicInt
+
+
+@frozen
+class Phase(BlockEncoding):
+    r"""Apply a phase to a block encoding.
+
+    Given $B[A]$ as a $(\alpha, a, \epsilon)$-block encoding of $A$, produces a
+    $(\alpha, a, \epsilon)$-block encoding $B[\exp(i\pi\phi)A]$ of $\exp(i\pi\phi)A$.
```
mpharrigan, line 33: nit / feel free to ignore: "produces a block encoding $B[x]$ of $x$." took me some mental energy to parse that $x$ was indeed the same in the two expressions and you were just being precise about the notation. Maybe just "produces a block encoding of $x$"?

mpharrigan (approved):

PR #1132, opened 2024-07-11
Make QubitizationWalkOperator expect a BlockEncoding instead of Select and Prepare

fdmalone:
A bit uneasy about the controlled aspect of this, forcing block encodings to be SpecializedSingleQubitControlGats seems wrong, or do we only expect singly controlled walk operators?


qualtran/bloqs/qubitization/qubitization_walk_operator.py
```diff
@@ -82,66 +86,67 @@ class QubitizationWalkOperator(SpecializedSingleQubitControlledGate):
...
-        return ReflectionUsingPrepare(self.prepare, control_val=self.control_val, global_phase=-1)
+        return ReflectionUsingPrepare(
+            self.block_encoding.signal_state, control_val=self.control_val, global_phase=-1
+        )
 
     @cached_property
     def sum_of_lcu_coefficients(self) -> Optional[SymbolicFloat]:
         r"""value of $\lambda$, i.e. sum of absolute values of coefficients $w_l$."""
-        return self.prepare.l1_norm_of_coeffs
+        return self.block_encoding.alpha
```
anurudhp, line 129: `QubitizationWalkOperator` is also a block-encoding of `H` right? We should perhaps inherit from `BlockEncoding`, get rid of `sum_of_lcu_coefficients`, and instead always use `walk_op.alpha` (which would be `BH.alpha`).

anurudhp:
I think this breaks the current decomp of HamiltonianSimulationByGQSP's, which uses prepare, prepare^dagger at either end (which shouldn't be neccessary anymore I think)

fdmalone:
> I think this breaks the current decomp of `HamiltonianSimulationByGQSP`, which uses prepare, prepare^dagger at either end (which shouldn't be neccessary anymore I think)

Is it not necessary? (I'm just getting back to this)

anurudhp:
> > I think this breaks the current decomp of `HamiltonianSimulationByGQSP`, which uses prepare, prepare^dagger at either end (which shouldn't be neccessary anymore I think)
> 
> Is it not necessary? (I'm just getting back to this)

So the current walk operator is: `PREP . R_0 . PREP^\dagger . SELECT`, but the upgraded one using block-encodings is `R_0 . (PREP^\dagger . SELECT . PREP)` (term in brackets is the block-enc). So the new one automatically handles the basis change as the PREP is in front instead of at the end.

fdmalone:
GQSP assumes the block encoding is of the LCU form right? If we force the QWO to be built from a block encoding we'll need to handle this.

anurudhp:
GQSP actually doesn't assume anything, it just applies the map

$$U \mapsto \begin{bmatrix} P(U) & \cdot \\\ \cdot & \cdot \end{bmatrix}$$

Only Hamiltonian simulation assumes $U$ is a walk-operator for some hamiltonian $H$. The paper does assume an LCU form, but the algorithm works for any arbitrary walk operator. This is because the PREP / PREP^\dagger are never controlled, so you can always re-group them the new way to merge into a block-enc. 

I'm not sure why they restricted the paper to a  specific form. Possibly for a simpler exposition.

fdmalone:
Some issues:

* It seemed a bit wrong (but maybe not) to make all bloq encodings GateWithRegisters just to derive from the SingleQubitControlledGate so I just duplicated the get_ctrl_system logic.
* The diagrams may look a little more cluttered as the signature of the block encoding includes junk registers whereas before only Selects signature was used (which excludes the junk registers from Prepare)
* I pushed alpha out of the constructor of the LCU block encodings and just grab this from Prepare
* To avoid GQSP issues I expose a prepare property for the LCU block encodings so that still works without any more changes.
*  Make block encoding epsilon optional as it's currently not really used anywhere

tanujkhattar:
> It seemed a bit wrong (but maybe not) to make all bloq encodings GateWithRegisters just to derive from the SingleQubitControlledGate so I just duplicated the get_ctrl_system logic.

Maybe it's worth making `SingleQubitControlledGate` not derive from `GateWithRegisters` and instead just derive from the `Bloq` class and avoid the duplication here? This change can be handled as part of a separate PR (we can just open an issue for now), especially if we already have multiple call sites that derive from `SingleQubitControlledGate` but don't necessarily need it to be a `GateWithRegisters` ?

fdmalone:
> > It seemed a bit wrong (but maybe not) to make all bloq encodings GateWithRegisters just to derive from the SingleQubitControlledGate so I just duplicated the get_ctrl_system logic.
> 
> Maybe it's worth making `SingleQubitControlledGate` not derive from `GateWithRegisters` and instead just derive from the `Bloq` class and avoid the duplication here? This change can be handled as part of a separate PR (we can just open an issue for now), especially if we already have multiple call sites that derive from `SingleQubitControlledGate` but don't necessarily need it to be a `GateWithRegisters` ?

I'll do this.

fdmalone:
@tanujkhattar I integrated those 2 PRs PTAL when you get a chance

fdmalone:
@tanujkhattar / @anurudhp PTAL, setting the ancilla to zero for LCUBlockEncoding fixed the OOM issue (back to the old diagrams)

mpharrigan:
@charlesyuan314 do you want to take a look at this

qualtran/bloqs/qubitization/qubitization_walk_operator.py
```diff
@@ -22,13 +22,13 @@
 \mathrm{PREPARE}|0\rangle = |\mathcal{L}\rangle \\
 (\langle \mathcal{L} | \otimes \mathbb{1}) \mathrm{SELECT} (|\mathcal{L} \rangle \otimes \mathbb{1}) = H / \lambda
 $$.
-
+sum_of_l1
```
charlesyuan314, line 25: Should this be here?

qualtran/bloqs/qubitization/qubitization_walk_operator.py
```diff
@@ -84,66 +89,71 @@ class QubitizationWalkOperator(GateWithRegisters, SpecializedSingleQubitControll
         Babbush et. al. (2018). Figure 1.
     """
 
-    select: SelectOracle
-    prepare: PrepareOracle
+    block_encoding: BlockEncoding
```
charlesyuan314, line 92: One problem here is that the `selection_registers`, `target_registers`, and `junk_registers` methods have been removed from `BlockEncoding` by #1246.

They're still present in `LCUBlockEncoding` and `LCUBlockEncodingZeroState`. Though not optimal, would it make sense for this attribute to be made a `Union[LCUBlockEncoding, LCUBlockEncodingZeroState]` for now so that the attributes can still be used?

qualtran/bloqs/block_encoding/block_encoding_base.py
```diff
@@ -93,6 +93,11 @@ def epsilon(self) -> SymbolicFloat:
         """The precision to which the block encoding is to be prepared."""
         raise NotImplementedError
 
+    @property
```
charlesyuan314, line 96: See my comment in `QubitizationWalkOperator` --- given the removal of the other `_registers` methods from the `BlockEncoding` interface, could we add `control_registers` only to `LCUBlockEncoding` and `LCUBlockEncodingZeroState`?

We can then revisit how to add control registers to the block encoding quantum interface.

qualtran/bloqs/qubitization/qubitization_walk_operator.py
fdmalone: Whoops! Removed.

qualtran/bloqs/qubitization/qubitization_walk_operator.py
fdmalone: Sure.

qualtran/bloqs/block_encoding/block_encoding_base.py
fdmalone: Done.

charlesyuan314:
Latest changes look good from a block encoding perspective!

qualtran/bloqs/block_encoding/lcu_block_encoding.py
```diff
@@ -66,8 +78,6 @@ class LCUBlockEncoding(BlockEncoding):
     Args:
         alpha: The normalization constant upper bounding the spectral norm of
             the Hamiltonian. Often called lambda.
```
anurudhp, line 80: remove this?

qualtran/bloqs/block_encoding/lcu_block_encoding.py
```diff
@@ -108,15 +121,30 @@ def selection_registers(self) -> Tuple[Register, ...]:
 
     @cached_property
     def junk_registers(self) -> Tuple[Register, ...]:
-        return self.prepare.junk_registers
+        return ()
```
anurudhp, line 124: why this change?

qualtran/bloqs/block_encoding/lcu_block_encoding.py
```diff
@@ -108,15 +121,30 @@ def selection_registers(self) -> Tuple[Register, ...]:
...
     def target_registers(self) -> Tuple[Register, ...]:
         return self.select.target_registers
 
+    @property
+    def alpha(self) -> SymbolicFloat:
+        return self.prepare.l1_norm_of_coeffs
+
+    @cached_property
+    def epsilon(self) -> SymbolicFloat:
+        return 0.0
```
anurudhp, line 136: link issue #1247?

qualtran/bloqs/block_encoding/lcu_block_encoding.py
```diff
@@ -210,13 +253,28 @@ def junk_registers(self) -> Tuple[Register, ...]:
...
```
anurudhp, line 262: link issue #1247?

qualtran/bloqs/block_encoding/lcu_block_encoding.py
```diff
@@ -127,9 +155,25 @@ def build_composite_bloq(self, bb: 'BloqBuilder', **soqs: SoquetT) -> Dict[str,
...
 
+    def wire_symbol(self, reg: Optional[Register], idx: Tuple[int, ...] = tuple()) -> 'WireSymbol':
+        if reg is None:
+            return Text('')
+        if reg.name == 'control':
+            return Circle(filled=bool(self.control_val))
+        else:
+            return TextBox('B[H]')
+
+    def get_single_qubit_controlled_bloq(self, control_val: int) -> 'LCUBlockEncoding':
```
anurudhp, line 166: this should probably raise an error if `self.control_val` is not `None`.

qualtran/bloqs/block_encoding/lcu_block_encoding.py
```diff
@@ -227,6 +285,22 @@ def _extract_soqs(bloq: Bloq) -> Dict[str, 'SoquetT']:
...
 
+    def wire_symbol(self, reg: Optional[Register], idx: Tuple[int, ...] = tuple()) -> 'WireSymbol':
+        if reg is None:
+            return Text('')
+        if reg.name == 'control':
+            return Circle(filled=bool(self.control_val))
+        else:
+            return TextBox('B[H]')
+
+    def get_single_qubit_controlled_bloq(self, control_val: int) -> 'LCUBlockEncodingZeroState':
```
anurudhp, line 296: this should probably raise an error if `self.control_val` is already not `None`.

qualtran/bloqs/qubitization/qubitization_walk_operator.py
```diff
@@ -153,6 +161,15 @@ def _circuit_diagram_info_(self, args: cirq.CircuitDiagramInfoArgs) -> cirq.Circ
     def adjoint(self) -> 'QubitizationWalkOperator':
         return attrs.evolve(self, uncompute=not self.uncompute)
 
+    @cached_property
+    def prepare(self) -> Union[PrepareOracle, BlackBoxPrepare]:
+        """Get the Prepare bloq if appropriate from the block encoding."""
```
anurudhp, line 166: could you add a TODO that says this should use `self.signal_state`?

qualtran/bloqs/chemistry/pbc/first_quantization/select_and_prepare.py
```diff
@@ -270,6 +274,10 @@ def selection_registers(self) -> Tuple[Register, ...]:
     def junk_registers(self) -> Tuple[Register, ...]:
         return (Register("succ_nu", QBit()), Register("plus_t", QBit()))
 
+    @property
+    def l1_norm_coeffs(self) -> Optional[SymbolicFloat]:
```
anurudhp, line 278: no `Optional` here? Perhaps raise a `ValueError` if `self.sum_of_l1_coeffs` is None?

anurudhp (approved):
lgtm % nits

qualtran/bloqs/block_encoding/lcu_block_encoding.py
fdmalone: Done.

qualtran/bloqs/block_encoding/lcu_block_encoding.py
fdmalone: Done.

qualtran/bloqs/block_encoding/lcu_block_encoding.py
fdmalone: Done.

qualtran/bloqs/block_encoding/lcu_block_encoding.py
fdmalone: Done.

qualtran/bloqs/qubitization/qubitization_walk_operator.py
fdmalone: Done.

fdmalone: Eh, need to make the walk operator a block encoding, some issues with controls I'll leave for follow up.

fdmalone: https://github.com/quantumlib/Qualtran/issues/1266

qualtran/bloqs/chemistry/pbc/first_quantization/select_and_prepare.py
fdmalone: Done.

fdmalone: Done.

qualtran/bloqs/block_encoding/lcu_block_encoding.py
fdmalone: This was to get around the OOM issues (This is just block encoding select which doesn't have junk registers, now the circuit widths are the same as before (no ancilla appearing.)

PR #1133, opened 2024-07-12
Add `LinearCombination` block encoding

charlesyuan314:
Add an instance of `BlockEncoding` that builds a block encoding of the linear combination of its
constituent block encodings, given a list of coefficients.


qualtran/bloqs/block_encoding/linear_combination.py
```diff
@@ -0,0 +1,351 @@
...
+    These oracles will be automatically instantiated if not specified by the user.
+
+    Args:
+        _block_encodings: A sequence of block encodings.
+        _lambd: Corresponding coefficients.
+        lambd_bits: Number of bits needed to represent coefficients precisely.
+        _prepare: If specified, oracle preparing $\sum_i \sqrt{|\lambda_i|} \ket{i}$
+            (state should be normalized and can have junk).
+        _select: If specified, oracle taking
+            $\ket{i}\ket{\psi} \mapsto \text{sgn}(\lambda_i) \ket{i} U_i\ket{\psi}$.
```
mpharrigan, line 67: `\ket` won't render properly in the jupyter notebooks. Mathjax lets you define commands, so you can define it if you want; but it has to be proximal to the usage

qualtran/bloqs/block_encoding/linear_combination.py
```diff
@@ -0,0 +1,351 @@
...
+class LinearCombination(BlockEncoding):
+    r"""Linear combination of a sequence of block encodings.
+
+    Builds the block encoding $B[\lambda_1 U_1 + \lambda_2 U_2 + \cdots + \lambda_n U_n]$ given
+    block encodings $B[U_1], \ldots, B[U_n]$ and coefficients $\lambda_i \in \mathbb{R}$.
+
+    When each $B[U_i]$ is a $(\alpha_i, a_i, \epsilon_i)$-block encoding of $U_i$, we have that
+    $B[\lambda_1 U_1 + \cdots + \lambda_n U_n]$ is a $(\sum_i \lvert\lambda_i\rvert\alpha_i, \lceil
+    \log_2 n \rceil + \max_i a_i, (\sum_i \lvert\lambda_i\rvert)\max_i \epsilon_i)$-block encoding
+    of $\lambda_1 U_1 + \cdots + \lambda_n U_n$.
```
mpharrigan: personal preference (feel free to ignore): I prefer de-densifying the explanation of the block encoding attributes of the combination; i.e. "with normalization constant blah, ancilla count y, and error sum max eps"

qualtran/bloqs/block_encoding/linear_combination.py
```diff
@@ -0,0 +1,351 @@
...
+
+
+@frozen
+class LinearCombination(BlockEncoding):
+    r"""Linear combination of a sequence of block encodings.
+
+    Builds the block encoding $B[\lambda_1 U_1 + \lambda_2 U_2 + \cdots + \lambda_n U_n]$ given
+    block encodings $B[U_1], \ldots, B[U_n]$ and coefficients $\lambda_i \in \mathbb{R}$.
+
+    When each $B[U_i]$ is a $(\alpha_i, a_i, \epsilon_i)$-block encoding of $U_i$, we have that
```
charlesyuan314, line 51: @mpharrigan I plan to reformat the long math here like you suggested in the past.

qualtran/bloqs/block_encoding/linear_combination.py
```diff
@@ -0,0 +1,351 @@
...
+    $B[\lambda_1 U_1 + \cdots + \lambda_n U_n]$ is a $(\sum_i \lvert\lambda_i\rvert\alpha_i, \lceil
+    \log_2 n \rceil + \max_i a_i, (\sum_i \lvert\lambda_i\rvert)\max_i \epsilon_i)$-block encoding
+    of $\lambda_1 U_1 + \cdots + \lambda_n U_n$.
+
+    Under the hood, this bloq uses LCU Prepare and Select oracles to build the block encoding.
+    These oracles will be automatically instantiated if not specified by the user.
+
+    Args:
```
mpharrigan, line 60: the way `attrs` data classes work is that if you define the attribute with a preceeding underscore, it will be "private" and users shouldn't be accessing that attribute on the class. But in the `__init__` constructor, the leading underscore is stripped from the argument name. When we document these constructor arguments (which is what we're doing here), lets remove the preceding underscore

qualtran/bloqs/block_encoding/linear_combination.py
```diff
@@ -0,0 +1,351 @@
...
+    @cached_property
+    def block_encodings(self):
+        """Appropriately negated constituent block encodings."""
+        return tuple(
+            Phase(be, phi=1, eps=0) if l < 0 else be
+            for l, be in zip(self._lambd, self._block_encodings)
+        )
+
+    @cached_property
+    def lambd(self):
```
mpharrigan, line 117: most of the time we want these bloq classes to be simple containers of data. An implicit promise is usually that

```python
my_x = ...
MyClass(x=my_x).x == my_x
```

I understand here that there are _input_ block encodings and coefficients and the _useful_ block encodings and coefficients that are subtly different. I'd ask that the names of these properties are different than the names of the input arguments

qualtran/bloqs/block_encoding/linear_combination.py
```diff
@@ -0,0 +1,351 @@
...
+            raise ValueError(
+                "If given, select oracle must have block encoding `system` register as target."
+            )
+
+    @cached_property
+    def block_encodings(self):
+        """Appropriately negated constituent block encodings."""
+        return tuple(
+            Phase(be, phi=1, eps=0) if l < 0 else be
+            for l, be in zip(self._lambd, self._block_encodings)
```
mpharrigan, line 113: is this right? we ignore `be` if its coefficient is less than zero?

qualtran/bloqs/block_encoding/linear_combination.py
charlesyuan314: Agreed, I made this change just now.

qualtran/bloqs/block_encoding/linear_combination.py
charlesyuan314: I'll rename them.

qualtran/bloqs/block_encoding/linear_combination.py
charlesyuan314: Sorry, what do you mean here? If the coefficient is less than zero we use `Phase(be, phi=1, eps=0)` which negates `be`, otherwise we use `be` unmodified.

charlesyuan314:
The use case would be in case the user has more optimized oracles on hand. I think this is something @tanujkhattar mentioned as being useful.

qualtran/bloqs/block_encoding/linear_combination.py
mpharrigan: doh I didn't see that `be` is being passed to `Phase`. I thought you were just emitting a `GlobalPhase` operation and ignoring that entry

mpharrigan (approved):

PR #1134, opened 2024-07-12
cleanup redundant code in `MultiControlledX`, use `MultiControlledPauli`

### LINKED ISSUE #1131: Add support for `MultiControlledU` via `C-U` and `MultiAnd`
- [x] Generalize `MultiControlledPauli` to accept any Bloq `U` with a single-bit control.
- [x] Replace `MultiControlledX` with `MultiControlledU(XGate())`
- [ ] Return this as the default in `Bloq.controlled()`?
- [x] Document: When the user needs $\text{MC-U}^\dagger \cdot V \cdot \text{MC-U}$, they can use `MultiAnd` to compute the controls, and only uncompute at the end (instead of twice when using `MultiControlledU`)

Related: #914

anurudhp:
partially addresses #1131

### DIFF
--- a/qualtran/bloqs/arithmetic/addition.py
+++ b/qualtran/bloqs/arithmetic/addition.py
@@ -444,7 +444,7 @@ class AddK(Bloq):
             if binary_rep[i] == 1:
                 if len(self.cvs) > 0 and ctrls is not None:
                     ctrls, k_split[i] = bb.add(
-                        MultiControlX(cvs=self.cvs), ctrls=ctrls, x=k_split[i]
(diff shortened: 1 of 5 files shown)


mpharrigan:
In the linked issue, it's proposed to beef up MultiControlledPauli into a MultiControlledU and (then?) do the migration. How much extra effort would it be to do it in this direction? It's a shame to get rid of a readable, qualtran-native decomposition.

anurudhp:
That should be easy, I can do that first

anurudhp:
handled in #1155

Issue #1072, opened 2024-06-13
mean_estimation.complex_phase_oracle_test.test_phase_oracle is very slow because of Fxp


mpharrigan:
This is the change: https://github.com/quantumlib/Qualtran/pull/1041/files#r1638968757

Somehow, this has turned the parameterized unit test from a 1 second to a 21 second ordeal on my machine. 

Is it worth resuscitating our own fixed-point-bits function? or is there a better incantation to the Fxp library to get it to go faster? 

Perhaps related: https://github.com/quantumlib/Qualtran/issues/643#issuecomment-2065491533

fdmalone:
I never dug into what was the cause of the slowness, whether it was Fxp or whether it was something else.

mpharrigan:
@anurudhp do you know if your change fixed this

anurudhp:
Unfortunately not, we are still using `fxpmath.Fxp` to convert input floats to bits, which seems to be very slow.

PR #1137, opened 2024-07-14
Support arbitrary `CtrlSpec` to single qubit control

anurudhp:
towards #1131 

- Introduces `CtrlSpecAnd` which is a generalized And/MultiAnd for an arbitrary `CtrlSpec`.

To follow:
- Fallback for `Controlled.decompose_bloq` to use this, followed by single-controlled subbloq, and uncompute.

### DIFF
--- a/qualtran/bloqs/bookkeeping/partition.py
+++ b/qualtran/bloqs/bookkeeping/partition.py
@@ -16,6 +16,7 @@ from typing import Dict, List, Tuple, TYPE_CHECKING
 
 import numpy as np
 from attrs import evolve, field, frozen, validators
+from numpy.typing import NDArray
 
 from qualtran import (
     bloq_example,
@@ -137,7 +138,7 @@ class Partition(_BookkeepingBloq):
             start += size
         return out_vals
 
-    def _classical_unpartition(self, **vals: 'ClassicalValT'):
+    def _classical_unpartition_to_bits(self, **vals: 'ClassicalValT') -> NDArray[np.uint8]:
         out_vals = []
         for reg in self.regs:
             reg_val = vals[reg.name]
@@ -145,14 +146,16 @@ class Partition(_BookkeepingBloq):
                 out_vals.append(ints_to_bits(reg_val.ravel(), reg.bitsize).ravel())
             else:
                 out_vals.append(ints_to_bits(reg_val, reg.bitsize)[0])
-        big_int = np.concatenate(out_vals)
-        return {'x': bits_to_ints(big_int)[0]}
+        big_int_bits = np.concatenate(out_vals)
+        return big_int_bits
 
     def on_classical_vals(self, **vals: 'ClassicalValT') -> Dict[str, 'ClassicalValT']:
         if self.partition:
             return self._classical_partition(vals['x'])
         else:
-            return self._classical_unpartition(**vals)
+            big_int_bits = self._classical_unpartition_to_bits(**vals)
+            (big_int,) = bits_to_ints(big_int_bits)
+            return {'x': big_int}
 
     def wire_symbol(self, reg: Register, idx: Tuple[int, ...] = tuple()) -> 'WireSymbol':
         if reg is None:
--- a/qualtran/bloqs/bookkeeping/partition.py
+++ b/qualtran/bloqs/bookkeeping/partition.py
@@ -16,7 +16,6 @@ from typing import Dict, List, Tuple, TYPE_CHECKING
 
 import numpy as np
 from attrs import evolve, field, frozen, validators
-from numpy.typing import NDArray
 
 from qualtran import (
     bloq_example,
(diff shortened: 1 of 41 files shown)


qualtran/bloqs/mcmt/ctrl_spec_activation_test.py
```diff
@@ -0,0 +1,36 @@
...
+    _ctrl_on_nd_bits,
+    CtrlSpecActivation,
+)
+
+
+@pytest.mark.parametrize(
+    "example", [_ctrl_on_bits, _ctrl_on_nd_bits, _ctrl_on_int, _ctrl_on_multiple_values]
+)
+def test_examples(bloq_autotester, example):
+    bloq_autotester(example)
```
tanujkhattar, line 30: Please add correctness tests via classical / tensor simulation.

qualtran/bloqs/mcmt/ctrl_spec_activation.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+@frozen
+class CtrlSpecActivation(Bloq):
+    """Computes a single qubit which is `1` iff the CtrlSpec is satisfied.
+
+    This reduces an arbitrary control specification to a single qubit, which can be used
+    to then control a bloq. Therefore, a bloq author is only required to implement a
+    single-controlled version of their bloq, and can be generalized to arbitrary controls.
+
```
tanujkhattar, line 49: This has a caveat that the current implementation works only for the case when the `CtrlSpec` is a logical `AND` of all individual control register clauses and each control register clause is restricted to be an `AND` with a single constant. 

Users can derive `CtrlSpec` to provide more general control spec and `activation_function_dtypes` and `is_active` methods were designed to support this. Although the docstring of `CtrlSpec` explicitly states

```
CtrlSpec uses logical AND among all control register clauses. If you need a different boolean
    function, open a GitHub issue.
```

For this PR, we should

a) Document this restriction in the docstring. 
b) Maybe rename to `CtrlSpecAND` (or something similar) which is more indicative of this restriction and we explicitly state that we don't support constructing a circuit for arbitrary boolean functions that a user may encode as part of the `CtrlSpec`.

tanujkhattar (commented):
Since we are decomposing a `CtrlSpec` to an `AND` gate, I'd suggest renaming to `CtrlSpecAND`. See my detailed comment above. 

Looking good otherwise.

qualtran/bloqs/mcmt/ctrl_spec_activation_test.py
anurudhp: added classical simulation test.

mpharrigan:
If we go with `CtrlSpecAnd` I suggest following the capitalization of existing `And`

qualtran/bloqs/mcmt/ctrl_spec_activation.py
anurudhp: Updated the bloq name and the docstring explaining this restriction

qualtran/bloqs/bookkeeping/partition.py
```diff
@@ -137,22 +138,24 @@ def _classical_partition(self, x: 'ClassicalValT') -> Dict[str, 'ClassicalValT']
...
 
-    def _classical_unpartition(self, **vals: 'ClassicalValT'):
-        out_vals = []
+    def _classical_unpartition_to_bits(self, **vals: 'ClassicalValT') -> NDArray[np.uint]:
+        out_vals: list[list[int]] = []
         for reg in self.regs:
             reg_val = vals[reg.name]
             if isinstance(reg_val, np.ndarray):
-                out_vals.append(ints_to_bits(reg_val.ravel(), reg.bitsize).ravel())
+                out_vals.extend(reg.dtype.to_bits(val) for val in reg_val.ravel())
```
tanujkhattar, line 146: Why this change? Also, why this change only in the `_classical_unpartition` and not in `_classical_partition` ? Was the previous logic wrong for certain cases? Can you please add tests which pass with the new version and fail with the original one?

tanujkhattar (approved):
LGTM % comments about partition

qualtran/bloqs/bookkeeping/partition.py
anurudhp: I'm using this to convert the `cvs` to bits in `CtrlSpecAnd` which allows any dtype. But the current classical simulation protocol only supports integer values in `ClassicalValT`, so seems like it worked either way.

anurudhp: Also `bits_to_ints` / `ints_to_bits` are faster than manually calling `to_bits` / `from_bits` on each element. (similar to https://github.com/quantumlib/Qualtran/issues/1072). So only changed the part I'm using for now.

tanujkhattar: Sorry, I don't understand. What was the part which was failing earlier and now gets fixed? I presume even now, `CtrlSpecAnd` would support only `QUInt`, `QInt` and `QAny` types (or, types for which `bits_to_ints` works correctly) since the classical bits need to be partitioned correctly (right now, partitioning is still based on `bits_to_ints` but unpartitioning uses `dtype.to_bits()`) ?

anurudhp: `CtrlSpecAnd` also works for `QFxp` as it's just bitwise equality

anurudhp: I'm using this only for extracting the classical control bits in `CtrlSpecAnd` (i.e. `_flat_cvs`), so I never merge these bits back to get the constants. 

But this would break the classical simulator, but as the classical sim only supports ints now, we don't have any tests that use Fxp.

qualtran/bloqs/mcmt/ctrl_spec_and_test.py
```diff
@@ -0,0 +1,53 @@
...
+
+@pytest.mark.parametrize("ctrl_spec", [CtrlSpec(), CtrlSpec(cvs=0), CtrlSpec(cvs=[0])])
+def test_raises_for_single_qubit_controls(ctrl_spec: CtrlSpec):
+    with pytest.raises(ValueError):
+        _ = CtrlSpecAnd(ctrl_spec)
+
+
+@pytest.mark.parametrize(
+    "ctrl_spec",
+    [CtrlSpec(QUInt(4), cvs=4), CtrlSpec((QUInt(4), QUInt(4)), cvs=(np.array(3), np.array(5)))],
```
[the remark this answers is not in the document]
anurudhp: will add a tensor test for it, as `get_classical_truth_table` uses classical sim which doesn't support Fxp yet.

qualtran/bloqs/bookkeeping/partition.py
tanujkhattar: I think for now, it's better to say that `CtrlSpecAnd` does not support `QFxp`, because in reality it does not. The decomposition would be incorrect because `Partition` and `Unpartition` doesn't support `QFxp`. So it's fine to fail early at `_flat_cvs` instead of having a wrong decomposition. 

The alternative would be to compute the `_flat_cvs` without invoking `Partition` so that the computation of `_flat_cvs` works for all data types and update `Partition` bloq for `_classical_unpartition` and `_classical_partition` whenever we are ready to do so. 

The current logic leaves the state of things in a complicatedly broken way where one direction works (maybe) but the other doesn't.

anurudhp: I reverted these changes. For now CtrlSpecAnd only supports integer types (checked in post_init). 

Will add more support after the classical sim for QFxp is merged.

qualtran/bloqs/mcmt/ctrl_spec_and_test.py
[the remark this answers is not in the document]
anurudhp: not supported in this PR

tanujkhattar (approved):

Issue #643, opened 2024-02-07
Support data types in classical simulation


mpharrigan:
Add dtype specific validation to the various steps. Maybe beef up the test fixtures as well. cc @fdmalone

fdmalone:
From #651 we need updates for:

1. QFxp
2. BoundedQInt
3. QIntOnesComplement

mpharrigan:
Is this done?

fdmalone:
No, I need to profile to see why Fxp is so slow.

Issue #606, opened 2024-01-27
Refactor `classical_sim.py` bitwise functions to allow for signed values


fpapa250:
`bits_to_ints()` and `ints_to_bits()` only work for conversions of unsigned integers. This limits the classical simulation correctness for bloqs using `Join()` because its `on_classical_vals()` implementation uses these functions. Therefore when we get the classical simulation of a bloq that uses `Join()` it cannot represent signed integers so we can't have tests for classical simulation of signed integer representations.

Currently this affects the signed implementation of `Add()` and doesn't allow us to classically simulate `SimpleAddConstant()`. Once this functionality has been changed we need to add tests for the classical simulation of the signed version of `SimpleAddConstant()`.

fdmalone:
in general I think we want a type_to_bits and bits_to_type method which will appropriately dispatch based on the type of the register. For fixed point we can probably write a little python implementation to define operations given some limited precision.

fpapa250:
For tracking purposes: Also an issue for comparison of negative integers in the `GreaterThan()` bloq.

PR #1142, opened 2024-07-15
Vectorized variants of `to_bits` and `from_bits`

### LINKED ISSUE #811: Consolidate bit iteration

anurudhp:
#811 
#606 
#643 

- Add new vectorized variants `to_bits_array` and `from_bits_array`
- Port the `ints_to_bits` and `bits_to_ints` functions from `classical_sim` to `QUInt.to_bits_array` and `QUInt.from_bits_array` 

more discussion: https://github.com/quantumlib/Qualtran/pull/1137#discussion_r1678347938


anurudhp:
ptal! I haven't removed `ints_to_bits` and `bits_to_ints` yet, as a sanity check that the old tests pass. 

I'm debugging some Fxp issue right now, after that I'll move the above tests and remove the old functions.

qualtran/_infra/data_types.py
```diff
@@ -528,6 +581,13 @@ def from_bits(self, bits: Sequence[int]) -> Fxp:
         fxp_bin = "0b" + bits_bin[: -self.num_frac] + "." + bits_bin[-self.num_frac :]
         return Fxp(fxp_bin, dtype=self.fxp_dtype_str)
 
+    def from_bits_array(self, bits_array: NDArray[np.uint8]):
+        assert isinstance(self.bitsize, int), "cannot convert to bits for symbolic bitsize"
+        # TODO figure out why `np.vectorize` is not working here
```
mpharrigan, line 586: open an issue and link? Do you have any theories? as I understand it: `np.vectorize` just does a python for loop under-the-hood

qualtran/_infra/data_types.py
```diff
@@ -77,10 +77,30 @@ def get_classical_domain(self) -> Iterable[Any]:
...
+    def to_bits_array(self, x_array: NDArray[Any]) -> NDArray[np.uint8]:
+        """Yields an NDArray of bits corresponding to binary representations of the input elements.
+
+        Often, converting an array can be performed faster than converting each element individually.
+        This operation accepts any NDArray of values, and the output array satisfies
+        `output_shape = input_shape + (self.bitsize,)`.
+        """
+        return np.vectorize(
+            lambda x: np.asarray(self.to_bits(x), dtype=np.uint8), signature='()->(n)'
+        )(x_array)
```
mpharrigan, line 89: I'm sure you know this, but as far as I understand it `np.vectorize` will use a python for-loop under-the-hood and you don't get any special performance improvements by using it. You get the correct api and broadcasting behavior, however. 

why is the signature argument needed?

mpharrigan (approved):
LGTM from an API perspective

qualtran/_infra/data_types.py
anurudhp: Without the signature, it tries to pack each output as a single entry in the array, and fails when we return a vector that needs to be treated as an additional dimension

qualtran/_infra/data_types.py
anurudhp: I think it's something to do with how Fxp interacts with numpy. Fxp has some inbuilt support to operate over NDArrays, so perhaps mixing the order up causes issues. I didn't investigate more though.

qualtran/bloqs/rotations/phase_gradient.py
```diff
@@ -302,15 +313,23 @@ def scaled_val(self, x: int) -> int:
...
-        return out['x'], out['phase_grad']
+        phase_grad_out = (phase_grad + self.sign * self.scaled_val(x)) % (2**self.phase_bitsize)
+        return x, phase_grad_out
+
+    def on_classical_vals(self, **kwargs) -> Dict[str, Union['ClassicalValT', Fxp]]:
+        x_fxp, phase_grad_fxp = kwargs['x'], kwargs['phase_grad']
 
-    def on_classical_vals(self, **kwargs) -> Dict[str, 'ClassicalValT']:
-        x, phase_grad = kwargs['x'], kwargs['phase_grad']
+        x, phase_grad = _extract_raw_int_from_fxp(x_fxp), _extract_raw_int_from_fxp(phase_grad_fxp)
```
tanujkhattar, line 332: The `self.scaled_val(x)` logic was written because `on_classical_vals` expected an `int` instead of `Fxp`. Now that `on_classical_vals` expects an `Fxp` object directly, we can avoid converting `Fxp -> int -> scaled_int -> scaled_fxp` and directly do `Fxp -> Scaled Fxp` here. 

Maybe rename `self.scaled_val -> self._scaled_val_int` and call from `self.apply` and add a `self.scaled_val` that expects an Fxp object for `x` and returns the scaled value as 

```python
    def scaled_val(self, x: Fxp) -> Fxp:
        """Computes `phase_grad + x` using fixed point arithmetic."""
        return x.like(_fxp(0, self.phase_bitsize)) >> self.right_shift
```

qualtran/bloqs/rotations/phase_gradient.py
```diff
@@ -222,6 +222,17 @@ def _phase_gradient_state() -> PhaseGradientState:
...
 
+def _extract_raw_int_from_fxp(val: Fxp) -> int:
+    """extracts the bits of the Fxp as a binary number (ignoring the dot)"""
+    if isinstance(val, (int, np.integer)):
+        return int(val)
+    return int(val.bin(), 2)
+
+
+def _phase_int_to_fxp(val: int, frac_bitsize: int) -> Fxp:
+    return Fxp(val, raw=True, n_word=frac_bitsize, n_frac=frac_bitsize, signed=False)
```
tanujkhattar, line 233: We should probably not need these two methods after the refactoring suggested below.

qualtran/_infra/data_types.py
tanujkhattar: An `Fxp` object can wrap a numpy array -- so to get a ND collection of Fxp objects, you construct a `Fxp(numpy_array_of_int_or_float_values)` instead of `np.array([Fxp(x) for x in array_of_int_or_float_values])`

See https://github.com/francof2a/fxpmath?tab=readme-ov-file#arithmetic for more details

qualtran/bloqs/rotations/phase_gradient.py
```diff
@@ -520,12 +547,16 @@ def apply(
...
-    def on_classical_vals(self, x: int, phase_grad: int) -> Dict[str, 'ClassicalValT']:
         phase_grad_out = (phase_grad + self.scaled_val(x)) % 2**self.phase_bitsize
-        return {'x': x, 'phase_grad': phase_grad_out}
+        return x, phase_grad_out
+
+    def on_classical_vals(self, x: Fxp, phase_grad: Fxp) -> Dict[str, Fxp]:
+        x_ = _extract_raw_int_from_fxp(x)
+        phase_grad_ = _extract_raw_int_from_fxp(phase_grad)
+
+        phase_grad_out = (phase_grad_ + self.scaled_val(x_)) % 2**self.phase_bitsize
```
tanujkhattar, line 557: Again, we should probably just use 

```python
scaled_x = _mul_via_repeated_add(x, self.gamma_fxp, self.phase_dtype.bitsize)
phase_grad_out = phase_grad + np.sign(self.gamma) * scaled_x
```

tanujkhattar (approved):
LGTM % updates to phase gadient bloqs to avoid `fxp -> int -> scaled_int -> scaled_fxp` conversions and directly do fxp arithmetic

qualtran/bloqs/rotations/phase_gradient.py
anurudhp: refactored `on_classical_vals` and removed these

qualtran/bloqs/rotations/phase_gradient.py
anurudhp: cleaned up

qualtran/bloqs/rotations/phase_gradient.py
```diff
@@ -293,36 +293,49 @@ def with_registers(self, *new_registers: Union[int, Sequence[int]]):
...
+    def on_classical_vals(self, **kwargs) -> Dict[str, Union['ClassicalValT', Fxp]]:
+        x: Fxp = kwargs['x']
+        phase_grad: Fxp = kwargs['phase_grad']
+
+        if self.controlled_by is None or self.controlled_by == kwargs['ctrl']:
+            # widen appropriately so that right shifting does not drop necessary bits
+            x = x.like(
+                QFxp(x.n_int + phase_grad.n_frac, phase_grad.n_frac, phase_grad.signed)._fxp_dtype
+            )
+            scaled_x = x >> self.right_shift
```
tanujkhattar, line 329: nit: Since `self.x_dtype` is `QFxp(self.x_bitsize, self.x_bitsize, signed=False)`, `x.n_int` is always 0 and so we can simply use `x.like(_fxp(self.phase_bitsize))` to do the resizing before shifting. 
```suggestion
            scaled_x = x.like(_fxp(self.phase_bitsize)) >> self.right_shift
```

qualtran/bloqs/rotations/phase_gradient.py
```diff
@@ -520,21 +522,32 @@ def apply(
...
+    def on_classical_vals(self, x: Fxp, phase_grad: Fxp) -> Dict[str, Fxp]:
+        if is_symbolic(self.phase_bitsize):
+            raise ValueError(f"Cannot classically simulate with symbolic {self.phase_bitsize=}")
+        if is_symbolic(self.gamma):
+            raise ValueError(f"Cannot classically simulate with symbolic {self.gamma=}")
 
-    def on_classical_vals(self, x: int, phase_grad: int) -> Dict[str, 'ClassicalValT']:
-        phase_grad_out = (phase_grad + self.scaled_val(x)) % 2**self.phase_bitsize
+        scaled_x = _mul_via_repeated_add(
+            x.like(phase_grad), gamma_fxp=self.abs_gamma_fxp, out=self.phase_bitsize
```
tanujkhattar, line 535: I think it's wrong to do the `x.like(phase_grad)` truncation here. `x` can be bigger than phase_grad and in this case, we should not discard the extra bits in `x`. The `_mul_via_repeated_add` will add shifted substrings of `x` into `phase_grad` but if we discard the extra bits here then the shifted substrings would just be `0` and lead to a larger approximation error. 

Ideally, a test should be failing at this point

qualtran/bloqs/rotations/phase_gradient_test.py
```diff
@@ -136,36 +180,51 @@ def test_add_into_phase_grad_controlled(controlled: int):
...
     assert len(basis_map) == len(set(basis_map.values()))
     circuit = cirq.Circuit(bloq.on(*cirq.LineQubit.range(num_bits)))
     np.testing.assert_allclose(circuit.unitary(), expected_unitary, atol=1e-8)
 
 
+_ADD_SCALED_VAL_INTO_PHASE_REG_EXAMPLES: list[AddScaledValIntoPhaseReg] = [
+    AddScaledValIntoPhaseReg.from_bitsize(4, 7, 0.123, 6),
+    AddScaledValIntoPhaseReg.from_bitsize(2, 8, 1.3868682, 8),
+    AddScaledValIntoPhaseReg.from_bitsize(4, 9, -9.0949456, 5),
+    AddScaledValIntoPhaseReg.from_bitsize(6, 4, 2.5, 2),
```
tanujkhattar, line 194: Maybe try adding a test for `AddScaledValIntoPhaseReg.from_bitsize(8, 3, 7.5, 1)` and see if that fails the classical simulation test due to `x.like()` truncation above?

qualtran/bloqs/rotations/phase_gradient.py
anurudhp: I was planning to remove `_fxp` and only use `QFxp` methods to construct Fxps so that there's only one way to construct them (easier to maintain). 

I'll use `x.like(self.phase_dtype._fxp_dtype)` here instead.

qualtran/bloqs/rotations/phase_gradient.py
anurudhp: Ah you're right. All the current tests have `x.n_int = 0`. I'll add a test to catch this and fix it

tanujkhattar: The issue is not necessarily because of `n_int`; but also because the fractional bitsize can be larger than `phase_grad` and if gamma > 1; then you'd be doing a left shift and adding into phase_grad; so you'll add bits 5, 6 7 into a phase grad register of size 3 if left shift due to gamma is 4

qualtran/bloqs/rotations/phase_gradient.py
tanujkhattar: The `_fxp` helper here sets useful properties on the `Fxp` config object (like `shifting='trunc'`) which are important to guarantee correctness of the multiply via add logic. This is hard to enforce when constructing `Fxp` objects via `QFxp`; since this is relevant in the context of this arithmetic bloq; please use the `_fxp` helper here instead of `self.phase_dtype._fxp_dtype` (which would be wrong the way it's implemented right now; unless we change the default options on `QFxp._fxp_dtype`

anurudhp: I actually fixed the `_fxp_dtype` to match the `_fxp` implementation, as we'd almost always want that.

qualtran/bloqs/rotations/phase_gradient.py
```diff
@@ -365,37 +366,30 @@ def _add_into_phase_grad() -> AddIntoPhaseGrad:
...
     res = _fxp(0, out)
+    x_fxp = x_fxp.copy()
+    x_fxp.resize(n_int=x_fxp.n_int, n_frac=max(out, x_fxp.n_frac))
     for i, bit in enumerate(gamma_fxp.bin()):
         if bit == '0':
             continue
         shift = gamma_fxp.n_int - i - 1
         # Left/Right shift by `shift` bits.
-        res += x_fxp << shift if shift > 0 else x_fxp >> abs(shift)
+        # Issue: Fxp << and >> does not preserve config, so we use `* 2**shift` instead.
```
tanujkhattar, line 385: Do you have a minimal example or link to documentation? Can you file an issue at the `fxpmath` repo? This seems wrong.

tanujkhattar (changes requested):
Blocking this because the scope of the PR has grown a lot since last LGTM. Let's pull out vectorization changes and merge them first and then deal with Fxp upgrades and nuances later

anurudhp:
Implemented in #1215

PR #1143, opened 2024-07-16
Create `SparseMatrix` block encoding

charlesyuan314:
Add an instance of `BlockEncoding` that builds a block encoding of a sparse-access matrix given
oracles for the indices of non-zero entries in each row / column and an oracle for its entries.

For now, we test its behavior using oracles for a matrix with the same entry in every position.
Oracles for diagonal and tridiagonal matrices are planned next.

### DIFF
--- a/qualtran/bloqs/block_encoding/sparse_matrix.py
+++ b/qualtran/bloqs/block_encoding/sparse_matrix.py
@@ -12,6 +12,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import abc
 from functools import cached_property
 from typing import cast, Dict, Tuple
 
@@ -49,10 +50,6 @@ class RowColumnOracle(Bloq):
     Here, $r(i, \ell)$ and $c(j, \ell)$ give the $\ell$-th nonzero entry in the $i$-th row
     and $j$-th column of the matrix, respectively.
 
-    Args:
-        n: The number of bits used to represent an index.
-        num_nonzero: The number of nonzero entries in each row or column.
-
     Registers:
         l: As input, index specifying the `l`-th non-zero entry to find in row / column `i`.
            As output, position of the `l`-th non-zero entry in row / column `i`.
@@ -63,8 +60,15 @@ class RowColumnOracle(Bloq):
         https://arxiv.org/pdf/2201.08309). Lin Lin (2022). Ch. 6.5.
     """
 
-    n: SymbolicInt
-    num_nonzero: SymbolicInt
+    @property
+    @abc.abstractmethod
+    def n(self) -> SymbolicInt:
+        """The number of bits used to represent an index."""
+
+    @property
+    @abc.abstractmethod
+    def num_nonzero(self) -> SymbolicInt:
+        """The number of nonzero entries in each row or column."""
 
     def __attrs_post_init__(self):
         if self.num_nonzero > 2**self.n:
@@ -82,9 +86,6 @@ class EntryOracle(Bloq):
     In the reference, this is the interface of
     $$O_A : \ket{0}\ket{i}\ket{j} \mapsto (A_{ij}\ket{0} + \sqrt{1 - |A_{ij}|^2}\ket{i}\ket{j}).$$
 
-    Args:
-        n: The number of bits used to represent an index of the matrix.
-
     Registers:
         q: The flag qubit that is rotated proportionally to the value of the entry.
         i: The row index.
@@ -95,7 +96,10 @@ class EntryOracle(Bloq):
         https://arxiv.org/pdf/2201.08309). Lin Lin (2022). Ch. 6.5.
     """
 
-    n: SymbolicInt
+    @property
+    @abc.abstractmethod
+    def n(self) -> SymbolicInt:
+        """The number of bits used to represent an index."""
 
     @cached_property
     def signature(self) -> Signature:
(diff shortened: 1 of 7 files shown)


charlesyuan314:
@mpharrigan PTAL!

qualtran/bloqs/block_encoding/sparse_matrix.py
```diff
@@ -0,0 +1,270 @@
...
+        i: The row / column index.
+
+    References:
+        [Lecture Notes on Quantum Algorithms for Scientific Computation](
+        https://arxiv.org/pdf/2201.08309). Lin Lin (2022). Ch. 6.5.
+    """
+
+    @property
+    @abc.abstractmethod
+    def n(self) -> SymbolicInt:
```
anurudhp, line 65: maybe something more descriptive, like `index_bitsize`?

qualtran/bloqs/block_encoding/sparse_matrix.py
```diff
@@ -0,0 +1,270 @@
...
+    def num_nonzero(self) -> SymbolicInt:
+        """The number of nonzero entries in each row or column."""
+
+    def __attrs_post_init__(self):
+        if self.num_nonzero > 2**self.n:
+            raise ValueError("Cannot have more than 2 ** n non-zero elements")
+
+    @cached_property
+    def signature(self) -> Signature:
+        return Signature.build_from_dtypes(l=QAny(self.n), i=QAny(self.n))
```
anurudhp, line 79: could use more specific dtypes? Perhaps `QUInt`, or `BoundedQUInt` similar to bloqs like `UnaryIteration`.

anurudhp (approved):
lgtm! % nits

qualtran/bloqs/block_encoding/sparse_matrix.py
```diff
@@ -0,0 +1,279 @@
...
+    and $j$-th column of the matrix, respectively.
+
+    Registers:
+        l: As input, index specifying the `l`-th non-zero entry to find in row / column `i`.
+           As output, position of the `l`-th non-zero entry in row / column `i`.
+        i: The row / column index.
+
+    References:
+        [Lecture Notes on Quantum Algorithms for Scientific Computation](
+        https://arxiv.org/pdf/2201.08309). Lin Lin (2022). Ch. 6.5.
```
mpharrigan, line 62: extreme nit: can you link to the arxiv landing page rather than the pdf. And you can keep the link on one line (url and title) even if it goes too long

qualtran/bloqs/block_encoding/sparse_matrix.py
```diff
@@ -0,0 +1,279 @@
...
+from qualtran.bloqs.block_encoding import BlockEncoding
+from qualtran.bloqs.block_encoding.lcu_select_and_prepare import PrepareOracle
+from qualtran.bloqs.state_preparation.prepare_uniform_superposition import (
+    PrepareUniformSuperposition,
+)
+from qualtran.symbolics import SymbolicFloat, SymbolicInt
+
+
+@frozen
+class RowColumnOracle(Bloq):
```
mpharrigan, line 46: Granted I'm not an expert on exactly why, but If this is supposed to be an abstract class -- which I think it is since it has abstract methods -- I think you're supposed to add `(Bloq, metaclass=abc.ABCMeta)` to the inheritance

qualtran/bloqs/block_encoding/sparse_matrix_test.py
```diff
@@ -0,0 +1,56 @@
...
+    assert bloq.ancilla_bitsize == 2 + 1
+    assert bloq.resource_bitsize == 0
+
+
+def test_sparse_matrix_tensors():
+    bloq = _sparse_matrix_block_encoding()
+    alpha = bloq.alpha
+    bb = BloqBuilder()
+    system = bb.add_register("system", 2)
+    ancilla = cast(Soquet, bb.add(IntState(0, 3)))
```
mpharrigan, line 49: do you actually need this `cast`?

qualtran/bloqs/block_encoding/sparse_matrix_test.py
```diff
@@ -0,0 +1,56 @@
...
+    system, ancilla = bb.add_t(bloq, system=system, ancilla=ancilla)
```
mpharrigan, line 50: you can still use `add` here (which will return either a single value or a tuple depending on the bloq signature (this is how python function calls work))

mpharrigan (approved):
just a couple style things looks pretty nice 👍

qualtran/bloqs/block_encoding/sparse_matrix_test.py
charlesyuan314: If I'm correct, you're expressing the same preference here as you did in https://github.com/quantumlib/Qualtran/pull/1158#discussion_r1683582059.

Is it a strong stylistic preference in Qualtran to use `add` over `add_t`? I didn't see this preference expressed in the documentation of `BloqBuilder::add_t`.

qualtran/bloqs/block_encoding/sparse_matrix_test.py
charlesyuan314: It is for making my IDE type checker (Pylance) happier. Mypy is more forgiving here.

qualtran/bloqs/block_encoding/sparse_matrix.py
charlesyuan314: Appreciated!

qualtran/bloqs/block_encoding/sparse_matrix.py
charlesyuan314: Actually, if we want to really get precise, inheriting from `Bloq` was sufficient here. A child of a class with `metaclass=abc.ABCMeta` gets the same benefit as its parent.

```
>>> class A(metaclass=abc.ABCMeta):
...     pass
... 
>>> class B(A):
...     @abc.abstractmethod
...     def f(): ...
... 
>>> B()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Can't instantiate abstract class B with abstract method f
```

`Bloq` already declares the `metaclass` here. I guess inheriting from `abc.ABC` makes it clear how this benefit would pass down to child classes.

PR #1145, opened 2024-07-17
`ControlledAddOrSubtract` Bloq

anurudhp:
Has half the T-cost of a controlled addition gate. Useful to perform an Rz rotation via addition into phase gradient, at half the cost of a ZPow.


qualtran/bloqs/arithmetic/controlled_add_or_subtract.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @cached_property
+    def signature(self) -> Signature:
+        return Signature.build_from_dtypes(ctrl=QBit(), a=self.a_dtype, b=self.b_dtype)
+
+    def build_composite_bloq(
+        self, bb: 'BloqBuilder', ctrl: 'Soquet', a: 'Soquet', b: 'Soquet'
+    ) -> dict[str, 'SoquetT']:
+        ctrl = bb.add(XGate(), q=ctrl)
+        ctrl, b = bb.add(OnEach(self.b_dtype.num_qubits, XGate()).controlled(), ctrl=ctrl, q=b)
```
tanujkhattar, line 65: Should this be updated to use the `Negate` bloq ?

anurudhp: No, `Negate` computes the two's complement negative. But flipping every bit of `x` gives `-x - 1` in two's complement form.

qualtran/bloqs/arithmetic/controlled_add_or_subtract.py
```diff
@@ -0,0 +1,100 @@
...
+from qualtran.bloqs.arithmetic import Add
+from qualtran.bloqs.basic_gates import OnEach, XGate
+
+if TYPE_CHECKING:
+    from qualtran import BloqBuilder, Soquet, SoquetT
+
+
+@frozen
+class ControlledAddOrSubtract(Bloq):
+    """Transforms |1>|a>|b> to |1>|a>|b + a> and |0>|a>|b> to |0>|a>|b - a>
```
mpharrigan, line 29: consider moving the ket notation to the details and using a more human-friendly first line

> Performs either addition or subtraction depending on the control bit

anurudhp: updated docstring

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -530,8 +530,8 @@ def get_ctrl_system(
         def _add_ctrled(
             bb: 'BloqBuilder', ctrl_soqs: Sequence['SoquetT'], in_soqs: Dict[str, 'SoquetT']
         ) -> Tuple[Iterable['SoquetT'], Iterable['SoquetT']]:
-            ctrl, x = bb.add_t(bloq, ctrls=ctrl_soqs[0], **in_soqs)
-            return (ctrl,), (x,)
+            ctrls, x = bb.add_t(bloq, ctrls=np.asarray(ctrl_soqs), **in_soqs)
+            return np.asarray(ctrls).tolist(), (x,)
```
mpharrigan, line 534: was this a bug? is there a test that would have caught it? 

actually: what's going on here? the conditions above exclude any situation in where there is more than one `SoquetT` being passed in. For this bloq, that one `SoquetT` is an array of soquets

qualtran/bloqs/arithmetic/controlled_add_or_subtract.py
```diff
@@ -0,0 +1,125 @@
...
+
+    $$
+        |1\rangle |a\rangle |b\rangle \mapsto |1\rangle |a\rangle |b + a\rangle \\
+        |0\rangle |a\rangle |b\rangle \mapsto |0\rangle |a\rangle |b - a\rangle
+    $$
+
+    Given two numbers `a`, `b` and a control bit `ctrl`, this bloq computes:
+    - the sum `b + a` when `ctrl=1`,
+    - the difference `b - a` when `ctrl=0`,
+    and stores it in the second register (`b`).
```
mpharrigan, line 42: markdown nit: this line gets rendered as part of the second bullet point. You'd need a newline to separate it

qualtran/bloqs/arithmetic/controlled_add_or_subtract.py
```diff
@@ -0,0 +1,125 @@
...
+
+    n = sympy.Symbol("n")
+    ctrl_add_or_sub_signed_symb = ControlledAddOrSubtract(QInt(n), QInt(n))
+    return ctrl_add_or_sub_signed_symb
+
+
+_CONTROLLED_ADD_OR_SUBTRACT_DOC = BloqDocSpec(
+    bloq_cls=ControlledAddOrSubtract,
+    import_line='from qualtran.bloqs.arithmetic.controlled_add_or_subtract import ControlledAddOrSubtract',
+    examples=[_ctrl_add_or_sub_unsigned, _ctrl_add_or_sub_signed, _ctrl_add_or_sub_signed_symb],
```
mpharrigan, line 124: the first example listed will be used for the "call graph", which should be the symbolic one if it supports symbolics which it does

mpharrigan:
Some nits but LGTM, but I'll let @tanujkhattar take a final look

qualtran/bloqs/arithmetic/controlled_add_or_subtract.py
```diff
@@ -0,0 +1,125 @@
...
+        |0\rangle |a\rangle |b\rangle \mapsto |0\rangle |a\rangle |b - a\rangle
+    $$
+
+    Given two numbers `a`, `b` and a control bit `ctrl`, this bloq computes:
+    - the sum `b + a` when `ctrl=1`,
+    - the difference `b - a` when `ctrl=0`,
+    and stores it in the second register (`b`).
+
+    This uses a controlled `Negate` followed by an uncontrolled `Add`,
+    which has half the T-cost of a controlled `Add`.
```
tanujkhattar, line 45: This is probably stale because the bloq uses controlled `BitwiseNot` instead of `Negate` ?

tanujkhattar (approved):
LGTM % nits

qualtran/bloqs/arithmetic/controlled_add_or_subtract.py
anurudhp: yep, fixed!

qualtran/bloqs/arithmetic/controlled_add_or_subtract.py
anurudhp: ah makes sense, put the symbolic one in front

qualtran/bloqs/arithmetic/controlled_add_or_subtract.py
anurudhp: oops, thanks! added a newline

qualtran/bloqs/arithmetic/addition.py
anurudhp: It fails on `test_tensor`:
```
            for li in reg.all_idxs():
>               idxed_soq = in_soq[li]
E               IndexError: too many indices for array: array is 0-dimensional, but 1 were indexed
```

PR #1149, opened 2024-07-17
Add `Xor` bloq

charlesyuan314:
Add a bloq to copy a register in computational basis via CNOTs that can be used to simplify diagrams.


qualtran/bloqs/mcmt/multi_control_multi_target_pauli.py
```diff
@@ -372,3 +373,59 @@ def build_composite_bloq(
 
     def pretty_name(self) -> str:
         return f'C^{len(self.cvs)}-NOT'
+
+
+@frozen
+class Copy(Bloq):
```
tanujkhattar, line 379: Can we call it `Xor` and place it in `qualtran/arithmetic/bitwise` ? I would want to avoid a bloq called `Copy` since one of the very common mistakes beginners make is to assume we can "copy" data for free and do not handle the entanglement properly.

qualtran/bloqs/arithmetic/bitwise.py
```diff
@@ -118,3 +122,64 @@ def _cxork() -> XorK:
     cxork = XorK(QUInt(8), 0b01010111).controlled()
     assert isinstance(cxork, XorK)
     return cxork
+
+
+@frozen
+class Xor(Bloq):
```
tanujkhattar, line 128: Can we follow the same signature as `XorK` for consistency? This would mean:

a) Supporting singly controlled version by deriving from `SpecializedSingleQubitControlledGate` and have an attribute `control_val: Optional[int] = None`.
b) Use `dtype: QDType` instead of `bitsize: SymbolicInt`
c) Rename registers to `x` and `y` s.t. you xor `x` into the `y` register. This is mainly so the controlled version has. a register called `ctrl`. We also often use `x` / `y` or `a`/`b` in arithmetic bloqs to denote source / target registers. 

Among the above 3 suggestions, I think maybe (a) is redundant if we just fallback to using `Controlled(Xor())`, but (b) and (c) should be incorporated irrespective of consistency with `XorK`.

qualtran/bloqs/arithmetic/bitwise.py
```diff
@@ -118,3 +123,73 @@ def _cxork() -> XorK:
...
+        ys = bb.split(y)
+
+        for i in range(len(xs)):
+            xs[i], ys[i] = bb.add_t(CNOT(), ctrl=xs[i], target=ys[i])
+
+        return {'x': bb.join(xs), 'y': bb.join(ys)}
+
+    def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
+        return {(CNOT(), self.dtype.num_qubits)}
+
```
tanujkhattar, line 172: You need to use a `Toffoli()` instead of a `CNOT` whenever `self.control_val` is not None. So this decomposition would be incorrect.  

I think I take my suggestion back -- for a simple bloq like this, it's probably best to just let the `Controlled()` bloq handle the controlled version and it's cost instead of explicitly defining the controlled bloq. The advantages of having a custom controlled Bloq, as I see it, is mainly 
a) More control over diagrams
b) Explicit control over decomposition and costings, though in this case that shouldn't matter.

@anurudhp Do you remember why you explicitly defined the controlled version for `XorK` in https://github.com/quantumlib/Qualtran/pull/1110/files?

qualtran/bloqs/arithmetic/bitwise.py
```diff
@@ -118,3 +123,73 @@ def _cxork() -> XorK:
...
+        if not isinstance(self.dtype.num_qubits, int):
+            raise DecomposeTypeError("`dtype.num_qubits` must be a concrete value.")
+
+        xs = bb.split(x)
+        ys = bb.split(y)
+
+        for i in range(len(xs)):
+            xs[i], ys[i] = bb.add_t(CNOT(), ctrl=xs[i], target=ys[i])
+
+        return {'x': bb.join(xs), 'y': bb.join(ys)}
```
tanujkhattar, line 168: I think you need to specify the `dtype` to `bb.join()` or else it'll join to `QAny` and the output soquet dtype would be different from expected dtype. I'm surprised this doesn't raise an error already.

qualtran/bloqs/arithmetic/bitwise.py
```diff
@@ -125,56 +126,65 @@ def _cxork() -> XorK:
...
+        self, x: 'ClassicalValT', y: 'ClassicalValT'
     ) -> Dict[str, 'ClassicalValT']:
-        return {'ctrl': ctrl, 'x': ctrl}
+        return {'x': x, 'y': x}
 
 
 @bloq_example
 def _xor() -> Xor:
-    xor = Xor(4)
+    xor = Xor(QAny(4))
```
tanujkhattar, line 181: Please add a test with a dtype other than `QAny`; this is probably why the `bb.join()` bug didn't raise an error.

qualtran/bloqs/arithmetic/bitwise.py
```diff
@@ -125,56 +126,65 @@ def _cxork() -> XorK:
...
     def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
-        return {(CNOT(), self.bitsize)}
+        return {(CNOT(), self.dtype.num_qubits)}
 
     def on_classical_vals(
-        self, ctrl: 'ClassicalValT', x: 'ClassicalValT'
+        self, x: 'ClassicalValT', y: 'ClassicalValT'
     ) -> Dict[str, 'ClassicalValT']:
-        return {'ctrl': ctrl, 'x': ctrl}
+        return {'x': x, 'y': x}
```
tanujkhattar, line 176: The classical return type should be `y: x ^ y` instead of `y: x`

Returning `y: x` would be a non-unitary operation in general since you'd map every arbitrary state $|x\rangle|y\rangle \rightarrow |x\rangle|x\rangle$ -- This is precisely the mistake that happens when thinking about a bloq that can "Copy" an input register.  Xoring would yield a unitary action of $|x\rangle|y\rangle \rightarrow |x\rangle|x \oplus y\rangle$

qualtran/bloqs/arithmetic/bitwise_test.py
```diff
@@ -36,3 +40,28 @@ def test_xork_classical_sim():
...
+def test_xor_symb(bloq_autotester):
+    bloq_autotester(_xor_symb)
+
+
+def test_xor_call():
+    bloq = _xor()
+    x, y = bloq.call_classically(x=7, y=0)
+    assert x == 7 and y == 7
+    x, y = bloq.decompose_bloq().call_classically(x=7, y=0)
+    assert x == 7 and y == 7
```
tanujkhattar, line 58: Please add classical action tests when `y != 0`  so the xoring behavior is verified. Probably best to iterate over all input values and verify the classical action.

qualtran/bloqs/arithmetic/bitwise.py
charlesyuan314: Falling back to `Controlled()`.

anurudhp: Actually not sure, I think I had it initially to match `AddK`, but forgot to remove it later.

anurudhp: I think it's best to avoid specialized controls unless we want to explicitly specify the diagrams for `Xor` and `XorK`. I'll cleanup the `XorK` bloq.

PR #1151, opened 2024-07-17
Add `InvertRealNumber` bloq

charlesyuan314:
Cost out the arithmetic operation of inverting a real number >= 1. Uses Newton-Raphson iteration as specified by [this reference](https://arxiv.org/pdf/1511.08253).

This bloq can be combined with inverse square root to compute the square root of a number.

### DIFF
--- a/qualtran/bloqs/arithmetic/multiplication.py
+++ b/qualtran/bloqs/arithmetic/multiplication.py
@@ -30,6 +30,7 @@ from qualtran import (
     Side,
     Signature,
 )
+from qualtran.bloqs.arithmetic.subtraction import Subtract
 from qualtran.bloqs.basic_gates import CNOT, TGate, Toffoli, XGate
 from qualtran.bloqs.mcmt import MultiControlPauli
 from qualtran.symbolics import HasLength, smax, SymbolicInt
@@ -539,9 +540,6 @@ class InvertRealNumber(Bloq):
         return "1/a"
 
     def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
-        # import here to avoid circular dependency on qualtran.bloqs.arithmetic
-        from qualtran.bloqs.arithmetic.subtraction import Subtract
-
         # initial approximation: Figure 4
         num_int = self.bitsize - self.num_frac
         # Newton-Raphson: Eq. (1)
--- a/qualtran/bloqs/arithmetic/multiplication.py
+++ b/qualtran/bloqs/arithmetic/multiplication.py
@@ -30,9 +30,9 @@ from qualtran import (
     Side,
     Signature,
 )
-from qualtran.bloqs.basic_gates import TGate, Toffoli
-from qualtran.symbolics import smax
-from qualtran.symbolics.types import SymbolicInt
+from qualtran.bloqs.basic_gates import CNOT, TGate, Toffoli, XGate
+from qualtran.bloqs.mcmt import MultiControlPauli
+from qualtran.symbolics import HasLength, smax, SymbolicInt
 
 if TYPE_CHECKING:
     import quimb.tensor as qtn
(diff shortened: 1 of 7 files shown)


qualtran/bloqs/arithmetic/multiplication.py
```diff
@@ -495,3 +495,81 @@ def _square_real_number() -> SquareRealNumber:
...
+    Args:
+        bitsize: Number of bits used to represent the number.
+        num_frac: Number of fraction bits in the number.
+
+    Registers:
+        a: `bitsize`-sized input register.
+        result: `bitsize`-sized output register.
+
+        References:
+        [Quantum Algorithms and Circuits for Scientific Computing](
```
fdmalone, line 519: The paper link should be a single line: https://github.com/quantumlib/Qualtran/blob/7f3e92959b497084f2bf9d033326490423154557/qualtran/Autodoc.ipynb

qualtran/bloqs/arithmetic/multiplication.py
```diff
@@ -495,3 +495,81 @@ def _square_real_number() -> SquareRealNumber:
...
+    """
+
+    bitsize: int
+    num_frac: int
+
+    def __attrs_post_init__(self):
+        if (
+            isinstance(self.num_frac, int)
+            and isinstance(self.bitsize, int)
+            and self.num_frac >= self.bitsize
```
fdmalone, line 530: num_frac > bitsize is a separate error right (I feel like this should be checked in __attrs_post_init__ of QFxp itself.). Also bitsize and num_frac are typed as ints in the constructor so I'm not sure what the first two checks are checking. I think you just want to check that num_frac != self.bitsize (as then it would be purely fractional and < 1) or am I getting confused? The number of integer bits is bitsize - num_frac.

qualtran/bloqs/arithmetic/multiplication.py
```diff
@@ -495,3 +495,81 @@ def _square_real_number() -> SquareRealNumber:
...
+                Register("result", QFxp(self.bitsize, self.num_frac)),
+            ]
+        )
+
+    def pretty_name(self) -> str:
+        return "1/a"
+
+    def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
+        # import here to avoid circular dependency on qualtran.bloqs.arithmetic
+        from qualtran.bloqs.arithmetic.subtraction import Subtract
```
fdmalone, line 548: This is suspicious...

fdmalone: (As in signals some sort of larger problem (not really related to this PR)

qualtran/bloqs/arithmetic/multiplication_test.py
```diff
@@ -166,6 +171,27 @@ def test_plus_equal_product():
     assert t_complexity(bloq) == TComplexity(t=8 * max(a_bit, b_bit) ** 2)
```
fdmalone, line 173: Add the auto tester.

fdmalone (changes requested):
Some small questions / comments. Looking good!

qualtran/bloqs/arithmetic/multiplication.py
charlesyuan314: Thanks for the catch, this part is outdated and I will fix.

qualtran/bloqs/arithmetic/multiplication.py
charlesyuan314: You were right to be suspicious. I found the better fix which was to qualify some imports in `subtraction.py`. Now we don't need the workaround.

charlesyuan314:
@fdmalone with the changes, do you think this is ready to merge?

fdmalone (approved):

PR #1152, opened 2024-07-17
Fix `XorK` call graph, remove control_val

anurudhp:


qualtran/bloqs/arithmetic/bitwise.py
```diff
@@ -50,33 +47,23 @@ def _cvs_converter(vv):
 
 
 @frozen
-class XorK(SpecializedSingleQubitControlledGate):
+class XorK(Bloq):
```
tanujkhattar, line 50: Do we need the `_cvs_converter` defined above? It doesn't seem to be used in this file

tanujkhattar, line 50: While we are at it, let's also override the `wire_symbol` method for both the bloqs in this file and return `TextBox(f"⊕ {K}")` for the `XorK` and `TextBox(f"x⊕y")` for `Xor` for the target registers

tanujkhattar (approved):
LGTM % comments

qualtran/bloqs/arithmetic/bitwise.py
anurudhp: added wire symbols and a circuit diagram unit test

anurudhp: oops, removed

PR #1158, opened 2024-07-18
Add `SubtractFrom` to subtract from a register in place

charlesyuan314:
Add a new version of `Subtract` that instead of taking $|a\rangle|b\rangle \mapsto |a\rangle|a-b\rangle$, takes it to $|a\rangle|b-a\rangle$. This is essentially equivalent to the statement `b -= a`.

The implementation uses an additional `Negate`, which could likely be improved, but this is a first approach. For now, it only supports `a` and `b` register of equal dtype.

The reason we need this for `Subtract` but not for `Add` is because addition is commutative and we could always swap the operands, but we can't for subtraction.

### DIFF
--- a/qualtran/bloqs/arithmetic/subtraction.py
+++ b/qualtran/bloqs/arithmetic/subtraction.py
@@ -171,7 +171,7 @@ _SUB_DOC = BloqDocSpec(
 
 @frozen
 class SubtractFrom(Bloq):
-    """A version of `Subtract` that stores the input in the register being subtracted from.
+    """A version of `Subtract` that stores the output in the register being subtracted from.
 
     Implements $U|a\rangle|b\rangle \rightarrow |a - b\rangle|b\rangle$.
 
@@ -179,8 +179,8 @@ class SubtractFrom(Bloq):
         dtype: Quantum datatype used to represent the integers a, b, and a - b.
 
     Registers:
-        a: A dtype.bitsize-sized input register (register a above).
-        b: A dtype.bitsize-sized input/output register (register b above).
+        a: A dtype.bitsize-sized input/output register (register a above).
+        b: A dtype.bitsize-sized input register (register b above).
     """
 
     dtype: Union[QInt, QUInt, QMontgomeryUInt] = field()
--- a/qualtran/bloqs/arithmetic/subtraction.py
+++ b/qualtran/bloqs/arithmetic/subtraction.py
@@ -171,16 +171,17 @@ _SUB_DOC = BloqDocSpec(
 
 @frozen
 class SubtractFrom(Bloq):
-    """A version of `Subtract` that stores the output in the register being subtracted from.
+    """A version of `Subtract` that leaves behind the value being subtracted.
 
-    Implements $U|a\rangle|b\rangle \rightarrow |a - b\rangle|b\rangle$.
+    Implements $U|a\rangle|b\rangle \rightarrow |a\rangle|b - a\rangle$, essentially equivalent to
+    the statement `b -= a`.
 
     Args:
         dtype: Quantum datatype used to represent the integers a, b, and a - b.
 
     Registers:
-        a: A dtype.bitsize-sized input/output register (register a above).
-        b: A dtype.bitsize-sized input register (register b above).
+        a: A dtype.bitsize-sized input register (register a above).
+        b: A dtype.bitsize-sized input/output register (register b above).
     """
 
     dtype: Union[QInt, QUInt, QMontgomeryUInt] = field()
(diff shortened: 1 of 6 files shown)


qualtran/bloqs/arithmetic/subtraction.py
```diff
@@ -167,3 +167,93 @@ def _sub_diff_size_regs() -> Subtract:
...
+        return {
+            (Negate(self.dtype), 1),
+            (Negate(self.dtype).adjoint(), 1),
+            (Add(self.dtype_as_unsigned, self.dtype_as_unsigned), 1),
+        }
+
+    def build_composite_bloq(self, bb: 'BloqBuilder', a: Soquet, b: Soquet) -> Dict[str, 'SoquetT']:
+        neg = Negate(self.dtype)
+        b = bb.add(neg, x=b)  # a, -b
+        b, a = bb.add_t(Add(self.dtype_as_unsigned, self.dtype_as_unsigned), a=b, b=a)  # a - b, -b
```
mpharrigan, line 233: you can still use `bb.add()` in these cases

mpharrigan:
Makes sense to me. @tanujkhattar do you have any qualms with this approach? also cc @NoureldinYosri for your inventory of arithmetic bloqs

qualtran/bloqs/arithmetic/subtraction.py
```diff
@@ -167,3 +167,93 @@ def _sub_diff_size_regs() -> Subtract:
...
+
+    Implements $U|a\rangle|b\rangle \rightarrow |a - b\rangle|b\rangle$.
+
+    Args:
+        dtype: Quantum datatype used to represent the integers a, b, and a - b.
+
+    Registers:
+        a: A dtype.bitsize-sized input/output register (register a above).
+        b: A dtype.bitsize-sized input register (register b above).
+    """
```
tanujkhattar, line 184: IMO, it would be nice if we have two bloqs for subtract:

1. $|a\rangle |b\rangle \rightarrow |a\rangle |a - b\rangle$
2. $|a\rangle |b\rangle \rightarrow |a\rangle |b - a\rangle$


i.e. in both the cases we preserve the convention that subtraction happens in the second register (`b`) and the first register (`a`) is unmodified. The two cases where you'd want to modify the source instead of target can be reduced by changing the order in which you call the bloq at the call site. 

The approach LGTM otherwise.

qualtran/bloqs/arithmetic/subtraction.py
charlesyuan314: Yes. My IDE type checker (Pylance) prefers that I use `add_t` though for the slightly finer grained information.

qualtran/bloqs/arithmetic/subtraction.py
charlesyuan314: That seems fine. I changed the signature of this bloq to 2.

qualtran/bloqs/arithmetic/subtraction.py
```diff
@@ -167,3 +167,84 @@ def _sub_diff_size_regs() -> Subtract:
...
+
+@frozen
+class SubtractFrom(Bloq):
+    """A version of `Subtract` that leaves behind the value being subtracted.
+
+    Implements $U|a\rangle|b\rangle \rightarrow |a\rangle|b - a\rangle$, essentially equivalent to
+    the statement `b -= a`.
+
+    Args:
+        dtype: Quantum datatype used to represent the integers a, b, and a - b.
```
anurudhp, line 180:
```suggestion
        dtype: Quantum datatype used to represent the integers a, b, and b - a.
```

qualtran/bloqs/arithmetic/subtraction.py
```diff
@@ -167,3 +167,84 @@ def _sub_diff_size_regs() -> Subtract:
 _SUB_DOC = BloqDocSpec(
     bloq_cls=Subtract, examples=[_sub_symb, _sub_small, _sub_large, _sub_diff_size_regs]
 )
+
+
+@frozen
+class SubtractFrom(Bloq):
+    """A version of `Subtract` that leaves behind the value being subtracted.
```
anurudhp, line 174: This phrasing seems a bit confusing to me, perhaps you can state it more concretely?

qualtran/bloqs/arithmetic/subtraction.py
```diff
@@ -167,3 +167,84 @@ def _sub_diff_size_regs() -> Subtract:
...
+            return directional_text_box('b - a', side=reg.side)
+        else:
+            raise ValueError()
+
+    def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
+        return {(Negate(self.dtype), 1), (Subtract(self.dtype, self.dtype), 1)}
+
+    def build_composite_bloq(self, bb: 'BloqBuilder', a: Soquet, b: Soquet) -> Dict[str, 'SoquetT']:
+        a, b = bb.add_t(Subtract(self.dtype, self.dtype), a=a, b=b)  # a, a - b
+        b = bb.add(Negate(self.dtype), x=b)  # a, b - a
```
anurudhp, line 225: This effectively reduces to `Negate(b); Add(a, b); Negate(b)` right? I think it would be cheaper to use ones-complement (`BitwiseNot`) instead of negate, because `Negate(a) = BitwiseNot(a); AddK(a, 1)` and we can avoid the `AddK` which requires Toffolis.
```
BitwiseNot(b) # a, -1 - b
Add(a, b) # a, a - 1 - b
BitwiseNot(b) # a, -1 - (a - 1 - b) = a, -a + b
```

But this isn't blocking this PR. I was going to add a `BitwiseNot` gate, I can update this decomposition then.

anurudhp (approved):

qualtran/bloqs/arithmetic/subtraction.py
charlesyuan314: Sounds great, let's switch to `BitwiseNot` in the next PR.

PR #1159, opened 2024-07-18
Add `typing_extensions>=4.10.0` dep

charlesyuan314:
Make sure the version of the package is up to date to support #1156.


tanujkhattar:
Please also run `dev_tools/requirements/re-pip-compile.sh` to recompile the output files

charlesyuan314:
Ran the script.

dev_tools/requirements/envs/format.env.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.10
+# This file is autogenerated by pip-compile with Python 3.11
```
tanujkhattar, line 2: I think the recommendation is to run the script using Python 3.10 (cc @mpharrigan). There's a docker version which you can try running if you are on a linux machine.

charlesyuan314:
Re-ran using `re-pip-compile-in-docker.sh`.

mpharrigan:
Hey @charlesyuan314 thanks for finding the typing_extensions version. Based on the changelog, this is only from Feb 2024, so I think it's worth putting explicitly in runtime.txt 

It looks like mypy is getting bumped and causing problems, and I don't want to mire this PR in solving unrelated devops issues; so I'm going to push changes to this PR to revert the `re-pip-compile-in-docker.sh` changes which I can handle in a subsequent pr

mpharrigan (approved):

mpharrigan:
I thought this was the mypy version being bumped causing all the new mypy issues, but with local testing the mypy bump only causes one minor issue. The typing_extensions version isn't actually bumped by the re-render.

If anyone has any ideas what could be causing the plethora of new type checking errors please don't hesitate to throw them out

mpharrigan:
it's sympy ...

PR #1160, opened 2024-07-19
Disallow zero sized registers in QROMs and PRGAViaPhaseGradient

tanujkhattar:
Unblocks https://github.com/quantumlib/Qualtran/pull/1105


mpharrigan (approved):
sweeeeeeet

PR #1161, opened 2024-07-19
Add `BitwiseNot`

anurudhp:


charlesyuan314 (approved):

NoureldinYosri:
why do we need this when we have [XorK](https://github.com/quantumlib/Qualtran/blob/22b3d5e2062610e908484e46d04771cc25aba56c/qualtran/bloqs/arithmetic/bitwise.py#L46)?

anurudhp:
> why do we need this when we have [XorK](https://github.com/quantumlib/Qualtran/blob/22b3d5e2062610e908484e46d04771cc25aba56c/qualtran/bloqs/arithmetic/bitwise.py#L46)?

Just a convenient bloq to use instead of the less readable `XorK(dtype, -1)` or `XorK(dtype, 2**bitsize-1)`. And also avoids the detour of computing `bits(k)` etc.

NoureldinYosri:
> Just a convenient bloq to use instead of the less readable


```py3
class XorK:
    ...

    @staticmethod
    def bitwise_not(dtype):
        return ...
```

tanujkhattar:
I support having an additional Bloq instead of delagating to `XorK`. It would also be more readable in diagrams. Classical programming languages also have both `Xor` and `-` operators defined, even though one could always write `a ^ 2**size - 1` instead of `~a`.

PR #1162, opened 2024-07-19
Add `SignExtend` for two's complement sign extension

anurudhp:
Useful for a lot of arithmetic ops which allow different bitsizes. [wiki](https://en.wikipedia.org/wiki/Sign_extension)


NoureldinYosri (changes requested):
I was going to create this bloq as well ... this is where I stand https://github.com/quantumlib/Qualtran/pull/1163

you can incorporate my code into yours and then we can merge yours 

---

I don't know of a way to do resizing for QFxp that would make sense. I think this belongs in bookkeeping not arithmetic + this bloq needs to implement classical action ... I added all of this to my PR which you can copy

qualtran/bloqs/arithmetic/sign_extension.py
```diff
@@ -0,0 +1,125 @@
...
+@bloq_example
+def _sign_extend() -> SignExtend:
+    from qualtran import QInt
+
+    sign_extend = SignExtend(QInt(8), QInt(16))
+    return sign_extend
+
+
+@bloq_example
+def _sign_extend_fxp() -> SignExtend:
```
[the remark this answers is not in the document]
anurudhp: It works the same because `x: Fxp(n, n_frac)` is just an unsigned `x*2**n_frac: UInt(n)`. And our Fxp uses 2's complement. 

The only thing is I assert the frac sizes are the same. To extend only the fractional side we'd just add 0s at the end instead, which isn't supported by this bloq for now.

qualtran/bloqs/arithmetic/sign_extension.py
```diff
@@ -0,0 +1,125 @@
...
+    def signature(self) -> 'Signature':
+        return Signature(
+            [
+                Register('x', self.inp_dtype, side=Side.LEFT),
+                Register('y', self.out_dtype, side=Side.RIGHT),
+            ]
+        )
+
+    @cached_property
+    def extend_bitsize(self):
```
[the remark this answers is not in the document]
anurudhp: oops, will fix

qualtran/bloqs/arithmetic/sign_extension.py
```diff
@@ -0,0 +1,125 @@
...
+        return self.out_dtype.num_qubits - self.inp_dtype.num_qubits
+
+    def build_composite_bloq(self, bb: 'BloqBuilder', x: 'Soquet') -> dict[str, 'SoquetT']:
+        extend_ys = bb.allocate(self.extend_bitsize)
+        xs = bb.split(x)
+
+        xs[0], extend_ys = bb.add(
```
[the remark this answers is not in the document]
anurudhp: Oh right, yeah will assert that in attrs post init.

anurudhp:
> I think this belongs in bookkeeping not arithmetic + this bloq needs to implement classical action

My understanding is that bookkeeping bloqs are only for building the correct bloq graph, but do not actually perform any quantum action. But this bloq actually performs a non-trivial action, so it makes more sense to keep it in arithmetic.

mpharrigan:
>My understanding is that bookkeeping bloqs are only for building the correct bloq graph, but do not actually perform any quantum action. But this bloq actually performs a non-trivial action, so it makes more sense to keep it in arithmetic.

This is a reasonable set of logic to apply to the organization, thanks

qualtran/bloqs/arithmetic/sign_extension.py
```diff
@@ -0,0 +1,125 @@
...
+    Soquet,
+    SoquetT,
+)
+from qualtran.bloqs.mcmt import MultiTargetCNOT
+from qualtran.resource_counting import BloqCountT, SympySymbolAllocator
+from qualtran.symbolics import is_symbolic
+
+
+@frozen
+class SignExtend(Bloq):
```
tanujkhattar, line 38: Since we need to support the adjoint, should the bloq just be called `Resize` ? It's better to explicitly support resizing down and add tests than expecting `Adjoint(SignExtend())` to do the right thing? 

Also, if we are adding a `Resize` we can explain in docstring the behavior of resizing a signed vs unsigned type and probably just support resizing both instead of supporting resizing only the signed types.

mpharrigan: Up to the consensus, but my 2 cents would be 1) resize definitely sounds more like a bookkeeping operation and hides the fact that there's a tiny bit of logic to respect the sign bit 2) the linked wiki page is called "sign extend"

tanujkhattar: Ops, I missed the Wikipedia page. SignExtend LGTM then -- let's just support both signed and unsigned types and document the behavior for Adjoint / add a new Bloq for SignTruncate (Truncate seems to be used as the opposite of Extend at some places)? 

https://jarrettbillingsley.github.io/teaching/classes/cs0447/slides/03%20-%20Signed%20Integers,%20Extension,%20Truncation,%20and%20Addition.pptx

NoureldinYosri: > let's just support both signed and unsigned types

+1

---

my decomposition supports both directions https://github.com/quantumlib/Qualtran/pull/1163

qualtran/bloqs/arithmetic/sign_extension.py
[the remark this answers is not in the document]
anurudhp: added `SignTruncate` for the adjoint

qualtran/bloqs/arithmetic/sign_extension.py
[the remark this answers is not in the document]
anurudhp: added

qualtran/bloqs/arithmetic/sign_extension.py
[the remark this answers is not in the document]
anurudhp: Restricted to `QInt | QIntOnesComp`

qualtran/bloqs/arithmetic/sign_extension.py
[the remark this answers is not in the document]
anurudhp: Removed for now, will add examples later.

qualtran/bloqs/arithmetic/sign_extension.py
anurudhp: added `SignTruncate` for the adjoint. This PR supports signed values only.

In a follow-up, we can add unsigned support using a separate bloq (say `ZeroExtend` to stick to convention) and a support bloq `ResizeInt` which delegates to the correct extension or truncation bloq based on the dtypes

NoureldinYosri: with sign extend and zero extend we will have a pattern of checking dtype before using the appropiate bloq

```py3
if isinstance(self.dtype, ...): bb.add(SignExtend...)
else: bb.add(ZeroExtend...)
```

how about `ArithmeticResize` for a bloq that contains both logics?

tanujkhattar: I think it's fine to merge this PR and add the logic of zero extend either in the same bloq, or a new bloq.

qualtran/bloqs/arithmetic/conversions/sign_extension.py
```diff
@@ -0,0 +1,193 @@
...
+        x = bb.join(xs)
+
+        return {'y': x}
+
+    def build_call_graph(self, ssa: 'SympySymbolAllocator') -> set['BloqCountT']:
+        return {(MultiTargetCNOT(self.truncate_bitsize), 1)}
+
+    def on_classical_vals(self, x: 'ClassicalValT') -> dict[str, 'ClassicalValT']:
+        bits = self.inp_dtype.to_bits(int(x))
+        y = self.out_dtype.from_bits(bits[self.truncate_bitsize :])
```
tanujkhattar, line 181: I think we should assert here that all the dropped bits are 0 always (after xoring the MSB; or assert that all `bits` are equal to `y[0]`) -- as a sanity check that the adjoint is called on the correct input.

qualtran/bloqs/arithmetic/conversions/ones_complement_to_twos_complement.py
```diff
@@ -0,0 +1,66 @@
...
+            [
+                Register('x', QIntOnesComp(self.bitsize), side=Side.LEFT),
+                Register('y', QInt(self.bitsize), side=Side.RIGHT),
+            ]
+        )
+
+    def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
+        # Take the sign qubit as a control and cnot the remaining qubits, then
+        # add it to the remaining n-1 bits.
+        return {(Toffoli(), (self.bitsize - 2))}
```
tanujkhattar, line 57: Add classical action, my_tensors and simulation tests? Or open an issue and add it in a follow up if needed?

tanujkhattar: Oh I see this is just a move; nvm

tanujkhattar (approved):
LGTM - Please open an issue to document any future improvements needed to `SignExtend` (like supporting unsigned types)

qualtran/bloqs/arithmetic/conversions/sign_extension.py
anurudhp: added an assertion, and a unit test to see that it raises correctly

NoureldinYosri (approved):

PR #1166, opened 2024-07-19
Add `ExplicitEntryOracle` to block encode a matrix

charlesyuan314:
This bloq enables us to block encode a square matrix given its entries as a classical `NDArray`, when used with the `SparseMatrix` block encoding and the `FullRowColumnOracle`.

Caveats: this is mostly for prototyping and testing, as real algorithms should compute their entries more efficiently. Under the hood, this bloq depends on QROM and quantum variable rotations to a rather high degree of precision.


qualtran/bloqs/block_encoding/sparse_matrix.py
```diff
@@ -254,10 +265,74 @@ class UniformEntryOracle(EntryOracle):
     def build_composite_bloq(
         self, bb: BloqBuilder, q: Soquet, **soqs: SoquetT
     ) -> Dict[str, SoquetT]:
-        soqs["q"] = cast(Soquet, bb.add(Rx(2 * np.arccos(self.entry)), q=q))
+        soqs["q"] = cast(Soquet, bb.add(Ry(2 * np.arccos(self.entry)), q=q))
```
anurudhp, line 268: whoops! How did the previous unit tests not catch this?

charlesyuan314: This isn't a functional aspect of the block encoding. Rx will just induce a phase on the subspace where the ancilla is not $\ket{0}$ without affecting the matrix being encoded.

The unit tests only test the functional aspect that on the subspace where the ancilla is $\ket{0}$, the correct matrix is encoded. That happens regardless of whether we use Rx or Ry.

The code in the previous PR used Rx because [this reference](https://arxiv.org/abs/2302.10949) uses it. I wanted to switch it to Ry to be consistent with the [other reference](https://arxiv.org/abs/2201.08309), but either is correct in principle.

charlesyuan314: Added a code comment to make this clear.

anurudhp: Ah nice, thanks for the explanation!

mpharrigan (approved):
nice!

PR #1169, opened 2024-07-22
Add random tests for `Product` block encoding

charlesyuan314:
These tests increase confidence in the correctness of the implementation. They are pretty fast to run.

### DIFF
--- a/qualtran/bloqs/block_encoding/product_test.py
+++ b/qualtran/bloqs/block_encoding/product_test.py
@@ -12,7 +12,6 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-import random
 from typing import cast
 
 import cirq
--- a/qualtran/bloqs/block_encoding/product_test.py
+++ b/qualtran/bloqs/block_encoding/product_test.py
@@ -12,6 +12,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import random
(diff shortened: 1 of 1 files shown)


qualtran/bloqs/block_encoding/product_test.py
```diff
@@ -150,3 +152,28 @@ def test_product_cirq():
     initial_state = [0, 0, 0]
     final_state = [1, 0, 0]
     assert_circuit_inp_out_cirqsim(circuit, qubits, initial_state, final_state)
+
+
+random.seed(1234)
```
anurudhp, line 157: Is it safe to randomize globally? iirc, pytest runs multiple tests in the same process, so they'd share the random states. Maybe safer to pass an rng?

qualtran/bloqs/block_encoding/product_test.py
```diff
@@ -150,3 +152,28 @@ def test_product_cirq():
...
+
+
+def gen_test():
+    n = random.randint(3, 6)
+    bitsize = random.randint(1, 3)
+    gates = [MatrixGate.random(bitsize, random_state=1234) for _ in range(n)]
```
anurudhp, line 163: The seed here is fixed, so all the matrices will be the same, right?

anurudhp (approved):
lgtm % nits about randomness

PR #1171, opened 2024-07-22
Add script to check whether autogenerating notebooks is idempotent

charlesyuan314:
This script can be useful in a manual or CI test that a PR has appropriately updated the notebooks associated with a code change.

Running `check/autogenerate-notebooks` will either return 0 if `dev_tools/autogenerate-bloqs-notebooks-v2.py` leaves the Git working directory clean, or return 1 and print a message like the following:

```
...
[hamiltonian_simulation_by_gqsp] Replacing HamiltonianSimulationByGQSP.call_graph.py cell.
M qualtran/bloqs/block_encoding/block_encoding.ipynb
```

### DIFF
--- a/check/autogenerate-notebooks
+++ b/check/autogenerate-notebooks
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+
+################################################################################
+# Autogenerates the notebooks and reports whether there were any changes.
+#
+# Usage:
+#     check/autogenerate-notebooks
+#
+################################################################################
+
+python dev_tools/autogenerate-bloqs-notebooks-v2.py || exit $?
+
+if output=$(git status --porcelain) && [ -z "$output" ]; then
+  # Working directory clean
+  exit 0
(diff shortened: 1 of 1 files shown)


mpharrigan:
would this catch the situation where a new notebook is added to the autogenerate script but not `git add`-ed ?

charlesyuan314:
I think so. To test, I deleted the notebook, committed, ran the script, and got return code 1:
```
?? qualtran/bloqs/block_encoding/block_encoding.ipynb
```
Now we're in the situation where we have the new notebook but it wasn't `git add`-ed. Running the script again still gives return code 1 and the same output.

The invocation of `git status` here should find any modified or untracked files.

mpharrigan:
got it. This will be good for the CI, but the exit status may be a "false positive" for messy people like me that (deliberately) keep untracked files around

mpharrigan (approved):

PR #1173, opened 2024-07-22
Fix classical simulation when selection bitsize is 0 for QROM

tanujkhattar:
This got missed from https://github.com/quantumlib/Qualtran/pull/1160


mpharrigan (approved):
Thanks!

PR #1174, opened 2024-07-22
Update classical action of addition gates and fix classical action bug in Join

NoureldinYosri:


qualtran/simulation/classical_sim.py
```diff
@@ -265,3 +265,20 @@ def format_classical_truth_table(
...
     ]
     return '\n'.join([heading] + entries)
+
+
+def signed_addition(a: int, b: int, N: int, is_signed: bool) -> int:
+    """Performs addition mod N of (un)signed in a reversible way.
+
+    Addition of signed integers can result in overflow. In most classical programming languages (e.g. C++)
+    what happens when an overflow happens is left as an implementation detail for compiler designers.
+    However for quantum subtraction the operation should be unitary and that means that the unitary of
```
mpharrigan, line 275: wrap lines

qualtran/simulation/classical_sim.py
```diff
@@ -265,3 +265,20 @@ def format_classical_truth_table(
         for invals, outvals in truth_table
     ]
     return '\n'.join([heading] + entries)
+
+
+def signed_addition(a: int, b: int, N: int, is_signed: bool) -> int:
+    """Performs addition mod N of (un)signed in a reversible way.
```
mpharrigan, line 271: addition of signed? recommend: addition of two integers

qualtran/simulation/classical_sim_test.py
```diff
@@ -168,6 +170,35 @@ def test_apply_classical_cbloq():
...
+    [
+        (x, y, n_bits)
+        for n_bits in range(2, 5)
+        for x, y in itertools.product(range(-(2 ** (n_bits - 1)), 2 ** (n_bits - 1)), repeat=2)
+    ],
+)
+def test_signed_addition_signed(x, y, n_bits):
+    half_n = 1 << (n_bits - 1)
+    # Addition of signed ints `x` and `y`` is a cyclic rotation of the interval [-2^(n-1), 2^n) by `y`.
+    R = [*range(-(2 ** (n_bits - 1)), 2 ** (n_bits - 1))]
```
mpharrigan, line 196: is there a good reason why we're not using pep-8 variable names here?

qualtran/simulation/classical_sim.py
```diff
@@ -265,3 +265,20 @@ def format_classical_truth_table(
         for invals, outvals in truth_table
     ]
     return '\n'.join([heading] + entries)
+
+
+def signed_addition(a: int, b: int, N: int, is_signed: bool) -> int:
```
mpharrigan, line 270: Args section to document the arguments; I'd suggest renaming `N` to `mod`, and I'd suggest changing `mod` and `is_signed` to be keyword-only arguments (by putting `..., *, mod, is_signed)`. Baring that, I would strongly recommend _using_ keyword args when calling this function to make the flags obvious

qualtran/bloqs/bookkeeping/join.py
```diff
@@ -95,7 +94,7 @@ def my_tensors(
         ]
 
     def on_classical_vals(self, reg: 'NDArray[np.uint]') -> Dict[str, int]:
-        return {'reg': bits_to_ints(reg)[0]}
+        return {'reg': self.dtype.from_bits(reg.tolist())}
```
mpharrigan, line 97: why is `tolist` required?

mpharrigan (commented):
Thanks! Some style things and there's a failing test but then I think this is a nice improvement!

qualtran/bloqs/bookkeeping/join.py
NoureldinYosri: mypy

mpharrigan: it doesn't recognize `ndarray` as `Sequence`?

NoureldinYosri: expects Sequence[int] but got ndarray[np.int64]

qualtran/simulation/classical_sim.py
NoureldinYosri: ptal

qualtran/simulation/classical_sim.py
NoureldinYosri: done ... though I prefer to pass `num_bits` rather than `N` or `mod`

qualtran/simulation/classical_sim_test.py
NoureldinYosri: renamed

NoureldinYosri:
@mpharrigan I modified my change to the classical action of Join for QFxp to retain the old behaviour ... the behaviour for QFxp is fixed in https://github.com/quantumlib/Qualtran/pull/1142

mpharrigan (approved):

PR #1175, opened 2024-07-22
Generalize full matrix block encoding to top-left non-zero matrices

charlesyuan314:
- Generalize `FullRowColumnOracle` to `TopLeftRowColumnOracle` to block encode matrices with non-zero top left block. The index computation is the same, so we just need to let `num_nonzero` be overridden.
- Adapt `SparseMatrix` block encoding to when `num_nonzero` is not equal to `2**system_bitsize`.


qualtran/bloqs/block_encoding/sparse_matrix.py
```diff
@@ -239,14 +250,31 @@ def build_composite_bloq(
...
+
+    @_num_nonzero.default
+    def _num_nonzero_default(self):
+        # Default to a matrix with all entries non-zero
+        return 2**self.system_bitsize
 
     @cached_property
     def num_nonzero(self) -> SymbolicInt:
-        return 2**self.system_bitsize
+        return self._num_nonzero
```
anurudhp, line 277: I think you can just use `num_nonzero` as an attribute. As the class is frozen, it can't be modified.

charlesyuan314: That's true in principle, but pylint doesn't accept having `num_nonzero` be an attribute here in order to override the `@property @abstractmethod` named `num_nonzero` in the parent class:
```
qualtran/bloqs/block_encoding/sparse_matrix_test.py:85:17: E0110: Abstract class 'TopLeftRowColumnOracle' with abstract methods instantiated (abstract-class-instantiated)
```

Hence, I had to redundantly declare the property with a different name from the attribute.

anurudhp: Ah didn't realize it was an abstract property in the parent class

PR #1176, opened 2024-07-22
Improve static type hints for symbolic math

charlesyuan314:
- Add `@overload` type signatures to `math_funcs` to specify that non-symbolic inputs yield non-symbolic outputs.
- Switch from `isinstance` to `is_symbolic`.


mpharrigan (approved):
Thanks!

PR #1177, opened 2024-07-23
Add `SymmetricBanded` matrix block encoding

charlesyuan314:
Add row and column oracles to block encode a symmetric [banded matrix](https://en.wikipedia.org/wiki/Band_matrix). The symmetry here refers to the pattern of non-zero entries, not necessarily the entries themselves, which are determined separately by the entry oracle.


qualtran/bloqs/block_encoding/sparse_matrix.py
```diff
@@ -281,6 +284,65 @@ def build_composite_bloq(self, bb: BloqBuilder, **soqs: SoquetT) -> Dict[str, So
...
+        return {
+            (Add(QUInt(self.system_bitsize), QUInt(self.system_bitsize)), 1),
+            (AddK(self.system_bitsize, -self.bandsize, signed=True), 1),
+        }
+
+    def build_composite_bloq(self, bb: BloqBuilder, l: SoquetT, i: SoquetT) -> Dict[str, SoquetT]:
+        if is_symbolic(self.system_bitsize) or is_symbolic(self.bandsize):
+            raise DecomposeTypeError(f"Cannot decompose symbolic {self=}")
+
+        i, l = bb.add_t(Add(QUInt(self.system_bitsize), QUInt(self.system_bitsize)), a=i, b=l)
```
mpharrigan, line 340: The idiomatic style we're going for is to use `add` in non-programatic cases and no `cast`

qualtran/bloqs/block_encoding/sparse_matrix_test.py
```diff
@@ -125,5 +131,142 @@ def test_top_left_matrix():
...
+
+
+def test_symmetric_banded_row_column_oracle():
+    n = 3
+    bloq = SymmetricBandedRowColumnOracle(n, bandsize=1)
+
+    def test_entry(l, i):
+        bb = BloqBuilder()
+        l_soq = cast(Soquet, bb.add(IntState(l, n)))
+        i_soq = cast(Soquet, bb.add(IntState(i, n)))
```
mpharrigan, line 182: lets get rid of all of these `cast`s. They add noise to the code

qualtran/bloqs/block_encoding/sparse_matrix.py
```diff
@@ -281,6 +284,65 @@ def build_composite_bloq(self, bb: BloqBuilder, **soqs: SoquetT) -> Dict[str, So
...
+            or is_symbolic(self.num_nonzero)
+        ):
+            raise DecomposeTypeError(f"Cannot call symbolic {self=} classically")
+
+        assert not isinstance(l, np.ndarray) and not isinstance(i, np.ndarray)
+        if l >= self.num_nonzero:
+            raise IndexError("l out of bounds")
+        return ((l + i - self.bandsize) % (2**self.system_bitsize), i)
+
+    def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set[BloqCountT]:
```
mpharrigan, line 330: add a test that calls 

```python
import qualtran.testing as qlt_testing

qlt_testing.assert_equivalent_bloq_counts(bloq)
```

which compares a decomposition against the call graph

mpharrigan (approved):
A couple of things, but lgtm after fixing them

qualtran/bloqs/block_encoding/sparse_matrix.py
```diff
@@ -281,6 +284,65 @@ def build_composite_bloq(self, bb: BloqBuilder, **soqs: SoquetT) -> Dict[str, So
...
+        if 2**self.system_bitsize < 2 * self.bandsize:
+            raise ValueError(
+                f"bandsize={self.bandsize} too large for system_bitsize={self.system_bitsize}"
+            )
+
+    def call_classically(self, l: ClassicalValT, i: ClassicalValT) -> Tuple[ClassicalValT, ...]:
+        if (
+            is_symbolic(self.bandsize)
+            or is_symbolic(self.system_bitsize)
+            or is_symbolic(self.num_nonzero)
```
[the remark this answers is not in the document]
charlesyuan314: You're right that it would still work. Is it expected that we would ever `call_classically` a bloq with symbolic attributes but concrete `ClassicalValT`?

PR #1179, opened 2024-07-23
Make SpecializedSingleQubitControlledGate a bloq

fdmalone:
There's no particular reason this should have been a GWR so keep it as a bloq.

I decided to do this now before #1132.


qualtran/_infra/single_qubit_controlled.py
```diff
@@ -0,0 +1,84 @@
...
+
+from qualtran._infra.bloq import Bloq
+from qualtran._infra.controlled import CtrlSpec
+from qualtran._infra.registers import Register
+
+if TYPE_CHECKING:
+    from qualtran import AddControlledT, BloqBuilder, SoquetT
+
+
+class SpecializedSingleQubitControlledGate(Bloq):
```
tanujkhattar, line 28: Should we replace the `Gate` suffix with a `Bloq` suffix? cc @mpharrigan for suggestions on naming the API

fdmalone: Although Bloq(Bloq) is a bit redundant, SpecializedSingleQubitControlledMixin?

fdmalone: I chose -Mixin

qualtran/_infra/single_qubit_controlled.py
```diff
@@ -0,0 +1,84 @@
...
+
+from qualtran._infra.bloq import Bloq
+from qualtran._infra.controlled import CtrlSpec
+from qualtran._infra.registers import Register
+
+if TYPE_CHECKING:
+    from qualtran import AddControlledT, BloqBuilder, SoquetT
+
+
+class SpecializedSingleQubitControlledMixin(Bloq):
```
tanujkhattar, line 28: Strictly speaking, a `Mixin` would not derive from the `Bloq` class to avoid multiple inheritance when the children derive from both `GWR` and `SpecializedSingleQubitControlledMixin` and inherit `Bloq` from both the parents.

tanujkhattar (approved):
LGTM % one comment

qualtran/_infra/single_qubit_controlled.py
fdmalone: Good point. Maybe Extension / Helper?

fdmalone: Decorator

tanujkhattar: Sorry for the late reply -- Maybe just make it an actual `Mixin`,  call it a `Mixin` and don't derive from `Bloq` class ?

The call sites would look like

`MyGate(GateWithRegisters, SpecializedSingleQubitControlledMixin)` or `MyBloq(Bloq, SpecializedSingleQubitControlledMixin)` ?

PR #1180, opened 2024-07-23
Add `ToFxp` to convert a `QUInt` to a `QFxp`

charlesyuan314:
This bloq enables us to safely convert between `QUInt` and `QFxp` of compatible size and to divide integers by powers of two in the process.

If the bitsize of the output is equal to the input, then `Cast` could be used as an unsafe alternative. This bloq is a little safer and supports using a larger register to store the divided the value.

### DIFF
--- a/qualtran/bloqs/arithmetic/conversions.py
+++ b/qualtran/bloqs/arithmetic/conversions.py
@@ -15,23 +15,29 @@
 from functools import cached_property
 from typing import Dict, Optional, Set, Tuple, TYPE_CHECKING
 
-from attrs import frozen
+from attrs import field, frozen
 
 from qualtran import (
     Bloq,
     bloq_example,
+    BloqBuilder,
     BloqDocSpec,
+    DecomposeTypeError,
+    QFxp,
     QInt,
     QIntOnesComp,
     QUInt,
     Register,
     Side,
     Signature,
+    Soquet,
+    SoquetT,
 )
-from qualtran.bloqs.basic_gates import Toffoli
+from qualtran.bloqs.basic_gates import CNOT, Toffoli
 from qualtran.cirq_interop.t_complexity_protocol import TComplexity
 from qualtran.drawing import WireSymbol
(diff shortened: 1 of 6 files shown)


qualtran/bloqs/arithmetic/conversions.py
```diff
@@ -150,3 +156,77 @@ def _signed_to_twos() -> SignedIntegerToTwosComplement:
...
+
+        a_bits = bb.split(a)
+        b_bits = bb.split(b)
+        offset = self.b_dtype.num_frac - self.divide_by
+        for i in range(len(a_bits)):
+            a_bits[i], b_bits[offset + i] = bb.add_t(
+                CNOT(), ctrl=a_bits[i], target=b_bits[offset + i]
+            )
+
+        return {"a": bb.join(a_bits), "b": bb.join(b_bits)}
```
tanujkhattar, line 213: I think the `ToFxp` bloq should really just be a bookkeeping bloq and we shouldn't do any operations (like CNOT) to implement it. The reason is that to "Convert a register of type QUInt to QFxp" we really  just need to reinterpret where the dot is in the binary representation of the `QUInt`; and this is determined by the `divide_by` parameter. You are not changing the binary string representation of any input (maybe concatenating the input with a bunch of 0's on the left if bitsize of output is greater than input; but I think we should use the `SignExtend` bloq to do the resizing instead of doing it here). 

The current behavior has the unitary $\sum_{a}|a\rangle|b\rangle \rightarrow |a\rangle |b + a\rangle$. Here $a$ is the `QUInt` register and $b$ is the `QFxp` register. Instead what we want is to keep $a$ as it is but just reinterpret it as a `QFxp` instead of a `QUInt`.

tanujkhattar (changes requested):

qualtran/bloqs/arithmetic/conversions.py
fdmalone: Cast + Pad maybe?

PR #1184, opened 2024-07-24
Only do one level of serialization check

### LINKED ISSUE #953: Serialization check -- only do one level
The current automatic check for serialization is too slow to operate as a unit test or as part of the `bloq_report_card` notebook. We've been sprinkling in some skips but top-level bloqs will always be slow and always be sneaking in.

The check should just do one level of decomposition and serialization. The serializability of grand-callees of a bloq are not under the purview of the check/unit test for that bloq.

charlesyuan314:
Fixes #953.

Before, `qubitization_qpe_hubbard_model_large` took >5 min* to check.
After, it takes 23.3s to check.

*Run was cut off after 5 min.


mpharrigan:
Sorry, I'm a bit confused and I apologize if this is actually a can of worms: but does this imply that the prior behavior was doing one decomposition (i.e. not unlimited)? 
 - why is this the default instead of doing all of them?
 - Why is it so slow to do one level of decomposition? 

If my understanding is correct I don't think we actually want to reduce the depth to zero. Why does it still take 23s to serialize just the arguments? Does the hubbard model bloq have a bloq as a class attribute? Yes: It goes QubitizationQPE -> QubitzationWalkOperator -> {SelectOracle, PrepareOracle}. I'm still mildly confused why it still takes 20s to serialize those 3 stubs (i.e. it won't try to decompose any of the bloqs that are present as attributes)

cc @tanujkhattar who may have more context here

mpharrigan:
oh, it also has `LPResourceState` and `QFTTextBox` as attributes

charlesyuan314:
> does this imply that the prior behavior was doing one decomposition (i.e. not unlimited)?

Yes, I think that's right.

mpharrigan:
Ok some more color: `bloqs_to_proto` only uses the `max_depth` argument when building the library of bloqs to consider serializing; and after it does that (which includes all the bloqs that appear as attributes) it will decompose each bloq it found

mpharrigan:
If we consider two types of serialized bloqs: Fully-serialized and Stub bloqs where the former has `decomposition`, `bloq_counts` filled out and the latter just has `bloq_id` `name` `registers` and `args`:

Then, I would want the `max_depth` argument to control how deep in the call graph to get fully-serialized bloqs. Just to be clear: a max-depth of 1 would mean the root bloq would be fully serialized and its callees and attributes-that-happen-to-be-bloqs would all be stubs

I suspect this may have been the intention, but adding the bloqs-that-are-attributes to the `bloq_to_idx` table short-circuited the max depth logic

mpharrigan:
Or it could be correctly skipping serializing the decomposition when it finds that the callees of a stub bloq aren't in the `bloq_to_idx` table (via `KeyError`) but only after doing the expensive step of decomposition

tanujkhattar:
I agree the behaviour for max_depth should change. 



Testing decomposition all the way down has been useful to catch bugs in the past, so I think there's value in having a test which decomposes all the way down -- But I'm fine pulling out that logic as a separate test and not as part of seriallzation

PR #1183, opened 2024-07-24
[surface code] Organization

mpharrigan:
This is based off of #1157. See the one new commit for just the changes in this PR until that one gets merged and I rebase.

 - Use the names "Beverland et al" and "Gidney Fowler" everywhere
 - Split the CCZ2T _factory_ from the functions associated with the _model_ 
 - rename to "QECScheme" 
 - Remove global variables, which weren't pep8 names and the names would have been too many capital letters. Port them to consistently-named factory methods on their respective classes. 
 - Relative imports in `__init__.py`

PR #1185, opened 2024-07-24
[surface code] PhysicalCostModel

mpharrigan:
The final piece that both 1) includes circa two off-the-shelf models and 2) lets you mix-and-match components 

Following #1183


mpharrigan:
@NoureldinYosri ptal

NoureldinYosri (approved):

PR #1187, opened 2024-07-24
Ignore zero contributions in bloq classification.

fdmalone:


mpharrigan (approved):

mpharrigan:
add a unit test?

fdmalone:
I added a test that would have failed previously.

PR #1191, opened 2024-07-25
Refactor phase gradient bloq dtypes

anurudhp:
splitting #1142, part 1.

- add `dtype` properties to all bloqs and use them in signatures
- split unit tests into smaller components


mpharrigan (approved):
this lgtm. the library changes seem minimal and the test changes -- splitting into more specific unit tests is always appreciated

qualtran/bloqs/rotations/phase_gradient_test.py
```diff
@@ -77,95 +77,122 @@ def test_phase_gradient_gate(n: int, exponent, controlled):
...
             d = {'x': x, 'phase_grad': phase_grad}
             c1 = bloq.on_classical_vals(**d)
             c2 = cbloq.on_classical_vals(**d)
             assert c1 == c2, f'{d=}, {c1=}, {c2=}'
+
+
+@pytest.mark.parametrize(
+    'bloq',
+    [
+        pytest.param(bloq, marks=pytest.mark.slow if bloq.num_qubits() > 12 else ())
```
mpharrigan, line 182: bloq2 is still getting executed on my machine (7 seconds) despite it looking like it should be 4+9=13 qubits

mpharrigan: consider using bloq.signature.n_qubits, which is explicit about how it handles left/right registers

anurudhp: About bloq2, my guess would be that because `AddIntoPhaseGrad` doesn't define a decomposition, but only a classical action, and cirq.unitary computes the gate much faster for classical gates.

PR #1194, opened 2024-07-25
QREF integration

mstechly:
I added `qref_interop` module to Qualtran.
- Allows to convert Bloqs into QREF objects
- Includes demo of how to use it with Bartiq

There's a couple of issues with this PR:

- module is called `qref_interop`, but it's actually not "interop" but "one-way conversion" at this moment. Should we rename it or leave it as is? 
- I've written tests using `get_bloq_examples` from `dev_tools`. However, this is outside from Qualtran package, so not a way to go. I like the functionality though, so perhaps it's worth moving it to the Qualtran package?
- Failing tests are either skipping (failure on qualtran side) or xfailing (failure on QREF side). I found some recurring issues with certain bloqs, will post them in separate issue/issues. Since this is very much beta, I didn't want those tests to fail yet. Not sure if you agree with this approach.
- I'm not sure if the notebook demo is compelling enough, but decided I'll share this version and get your feedback first.
- As we discussed I tried analyzing call graph and getting some information from there, but we'd need to add some features in QREF to actually make it useful, so it's not even included in this PR.

CC @mpharrigan @tanujkhattar


mpharrigan:
Awesome! This will be great to have in the repository! Thanks so much for opening a PR. I suspect we may have a couple nits before merging, but this looks like a great start. 

We'll have to add qref to the dependencies. For better or for worse, we've been putting our interop dependencies alongside the ordinary runtime dependencies (maybe we'll factor these out into optional dependencies in the future) in `dev_tools/requirements/deps/runtime.txt`.

qualtran/qref_interop/_bloq_to_qref.py
```diff
@@ -0,0 +1,311 @@
...
+    When importing BloqInstance we derive name from Bloq's default name and
+    instance ID to prevent duplication of names between siblings.
+
+    See `import_from_qualtran` and `_bloq_instance_name` for more info.
+    """
+    return bloq_to_routine(instance.bloq, name=_bloq_instance_name(instance))
+
+
+def _names_and_dir_from_register(reg: Register) -> Iterable[tuple[str, str]]:
+    """Yield names and directions of Bartiq Ports corresponding to QualtranRegister.
```
tanujkhattar, line 164:
```suggestion
    """Yield names and directions of Bartiq Ports corresponding to Qualtran Register.
```

qualtran/qref_interop/_bloq_to_qref.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+def _expand_name_if_needed(reg_name, shape) -> Iterable[str]:
+    """Given a register name, expand it into sequence of names if it has nontrivial shape.
+
+    Examples:
+        "reg", () -> "reg"
+        "reg", (3,) -> "reg_0", "reg_1", "reg_2"
+    """
+    return (reg_name,) if shape == () else (f"{reg_name}_{i}" for i in range(shape[0]))
```
tanujkhattar, line 183: Should this be expanded recursively? We can also have registers with n-dimensional shapes. For example

```
"reg", (2, 3) -> "reg_0_0", "reg_0_1", "reg_0_2", "reg_1_0", "reg_1_1", "reg_1_2", 
```

qualtran/qref_interop/_bloq_to_qref.py
```diff
@@ -0,0 +1,311 @@
...
+        bloq: Bloq for which information should be extracted.
+        name: Optional override for name. Setting it to None will cause some reasonable
+            name (depending on bloq type) to be inferred. The typical use of non-trivial
+            name is when a better name is known from the parent Bloq.
+
+    Returns:
+        A dictionary that can be unpacked into arguments of RoutineV1 initializer.
+    """
+    ports = [port for reg in bloq.signature for port in _ports_from_register(reg)]
+    local_variables = {}
```
tanujkhattar, line 62: Can you add a comment explaining what's the use case for `local_variables` ? It seems like it may be used when the string representation of the symbolic size of a qualtran register does not satisfy the `isIdentifier()` condition which is needed for any string to be a valid port size in QREF ?

qualtran/qref_interop/_bloq_to_qref.py
```diff
@@ -0,0 +1,311 @@
...
+            if bloq.uncompute:
+                name += "_uncompute"
+        except AttributeError:
+            pass
+        try:
+            if bloq.is_adjoint:
+                name += "_adjoint"
+        except AttributeError:
+            pass
+    input_params = _extract_input_params(bloq, ports, local_variables)
```
tanujkhattar, line 81: Should `_extract_input_params` just deal with `bloq` and `ports` and the `local_variables` should be subtracted in this function because it's specific to the mapping of new port sizes to old port sizes done here?

qualtran/qref_interop/_bloq_to_qref.py
```diff
@@ -0,0 +1,311 @@
...
+            pass
+    return [str(symbol) for symbol in input_params]
+
+
+def _extract_input_params_from_ports(ports) -> list[str]:
+    input_params = set()
+    for port in ports:
+        input_params = input_params | sympy.sympify(port.size).free_symbols
+
+    return [str(param) for param in input_params]
```
tanujkhattar, line 311: Can you please add docstrings to these two methods?

qualtran/qref_interop/_bloq_to_qref.py
```diff
@@ -0,0 +1,311 @@
...
+    return resources
+
+
+def _ensure_primitive_type(value: Any) -> Union[int, float, str, None]:
+    """Ensure given value is of primitive type (e.g. is not a sympy expression)."""
+    return value if value is None or isinstance(value, (int, float, str)) else str(value)
+
+
+def _extract_input_params(bloq: Bloq, ports: list[PortV1], local_variables) -> list[str]:
+    """Extracts input_params from bloq's t_complexity and port sizes."""
```
tanujkhattar, line 287: When I first read `input params`, I think of the attributes that can be used to instantiate a bloq. But in this case, it looks like the parameters are symbolic parameters that either appear in the bloq's signature or in the bloq's t_complexity formulas. 

I think it's word adding a brief explanation of what qref considers to be an input parameters and how we are extracting them from Bloqs.

qualtran/qref_interop/_bloq_to_qref.py
```diff
@@ -0,0 +1,311 @@
...
+
+    The relative name is always computed wrt. the parent RoutineV1.
+    This function correctly recognizes the fact, that in any connection, the input parent
+    ports serve as outputs for the connection (and vice versa).
+    """
+    if len(soquet.idx) > 1:
+        raise NotImplementedError(
+            "Soquets referencing more than one index in composite register are not yet supported. "
+            f"The error was caused by the following soquet: {soquet}."
+        )
```
tanujkhattar, line 243: This answers my question about recursively expanding registers with `len(shape) > 1` -- I think it's worth raising a `NotImplementedError` in `_expand_name_if_needed` method as well.

qualtran/qref_interop/_bloq_to_qref.py
```diff
@@ -0,0 +1,311 @@
...
+                "value": _ensure_primitive_type(getattr(t_complexity, name)),
+                "type": "additive",
+            }
+        )
+    return resources
+
+
+def _ensure_primitive_type(value: Any) -> Union[int, float, str, None]:
+    """Ensure given value is of primitive type (e.g. is not a sympy expression)."""
+    return value if value is None or isinstance(value, (int, float, str)) else str(value)
```
tanujkhattar, line 283: If the goal is to check whether the value is a symbolic or not; you can consider using the `is_symbolic` method in `qualtran.symbolics` -- it may be more consistent since `isinstance` check would fail for types like `numpy.int64`.

qualtran/qref_interop/_bloq_to_qref.py
```diff
@@ -0,0 +1,311 @@
...
+        obj: object to be imported. Can be either Bloq or BloqInstance.
+        name: optional name override. This can be useful e.g. if you are converting
+            CompositeBloq (which do not store meaningful names in Qualtran) and
+            know some good name for it.
+
+    Return:
+        A Bartiq object corresponding to the source Qualtran object. For both Bloqs
+        and BloqInstances the returned object is of type RoutineV1.
+
+    """
```
tanujkhattar, line 125: If I understand correctly, as part of the conversion; we are not doing any bloq decompositions and the expectation is that the user should flatten out a bloq that has a decomposition into a `CompositeBloq` at an appropriate level and then convert the `CompositeBloq` to qref's `SchemaV1`; the conversion would not have any further information about the structure of the subbloqs (eg: info about decomposing a bloq that appears in the composite bloq but is not decomposed). Is this understanding correct?

qualtran/qref_interop/bartiq_demo.ipynb
```diff
@@ -0,0 +1,2214 @@
...
+    "## Writing custom USP implementation\n",
+    "\n",
+    "So far so good, but we used Bartiq to derive the same information we already had in Qualtran.\n",
+    "\n",
+    "There are two limitations of Qualtran implementation, with which Bartiq can help us, let's see what these are.\n",
+    "\n",
+    "### Numeric implementation\n",
+    "\n",
+    "The example above is purely numeric. We can get the symbolic cost of the whole `StatePreparationAliasSampling` (see snippet below) but we can't decompose it using `decompose_bloq()`.\n",
+    "\n",
```
tanujkhattar, line 1746: This will be fixed once https://github.com/quantumlib/Qualtran/pull/1084 is merged; so symbolic decomposition would work as expected for the given bloq example

qualtran/qref_interop/bartiq_demo.ipynb
```diff
@@ -0,0 +1,2214 @@
...
+   "execution_count": 17,
+   "id": "72f0b802-a3cd-4e43-86da-ad1bbe0fe39b",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/Users/mstechly/Documents/code/2024-06-26-qualtran-conversion/bartiq/src/bartiq/compilation/_compile.py:117: UserWarning: Found the following issues with the provided routine after the compilation has finished: [\"Input param N found in subroutine: CompositeBloq.PrepareUniformSuperposition_0, which is not among top level params: {'PrepareUniformSuperposition_0.N'}.\"]\n",
+      "  warnings.warn(\n"
```
tanujkhattar, line 1988: We usually clear the output of all notebooks before committing them. The notebooks are run before generating the docs so that the docs generated on https://qualtran.readthedocs.io/en/latest/index.html have the expected output. 

Can you please clear all output from the notebook (assuming the notebooks runs after qualtran runtime dependencies are updated)

qualtran/qref_interop/_bloq_to_qref.py
```diff
@@ -0,0 +1,311 @@
+import networkx as nx
```
tanujkhattar, line 1: nit: I think it's worth splitting up this file into smaller files with methods grouped based on the function they are performing. But happy to merge it as a single file for now and leave the reorganization as a future improvement.

qualtran/qref_interop/test_bloq_to_qref.py
tanujkhattar, line 1: In qualtran, we follow the nomenclature of making the test file `file_name_test.py` for source file `file_name.py`.
Please rename the file to `qualtran/qref_interop/_bloq_to_qref_test.py`

tanujkhattar (changes requested):
Thanks for opening the PR! The conversion already looks great! 

I've left a round of comments and suggestions.

tanujkhattar:
Also, I think it's worth adding `qref` to `dev_tools/requirements/deps/runtime.txt` in a separate PR and then run `dev_tools/requirements/re-pip-compile-in-docker.sh` to recompile the `.env.txt` files.

qualtran/qref_interop/_bloq_to_qref.py
mstechly: Resolved in another discussion.

qualtran/qref_interop/_bloq_to_qref.py
mstechly: After some thoughts I decided this logic should actually go to `bartiq`, will update `bartiq` and then remove it from here.

qualtran/qref_interop/_bloq_to_qref.py
mstechly: That's a fair point, I might not be extracting it correctly. I'll see if I can get the info I need from `signature`.

qualtran/qref_interop/_bloq_to_qref.py
mstechly: Done!

qualtran/qref_interop/test_bloq_to_qref.py
mstechly: Good point, done!

mstechly:
@tanujkhattar I'll do as you suggested with the separate PR for the dependencies once this PR is clean and I make any necessary changes in qref/bartiq (there's at least one right now).

qualtran/qref_interop/_bloq_to_qref.py
mstechly: Actually, it would require some major changes in bartiq and if done like this it should work, so I'll leave it as it is.
Added a comment.

mstechly: An extra question – is there a way to get all the symbols used in a particular `bloq`? 
In most cases the symbols taken from `signature` are enough, but sometimes (e.g. in `PermutationCycle`), I don't know how to retrieve the information that it requires having `L` defined, other than parsing the `t_complexity` of it.

qualtran/qref_interop/_bloq_to_qref_test.py
```diff
@@ -0,0 +1,261 @@
+from dev_tools.qualtran_dev_tools.bloq_finder import get_bloq_examples
```
tanujkhattar, line 1: @mpharrigan This is not part of the qualtran package; so we should probably not use it in tests to avoid adding a dependency.

mpharrigan: Yeah, @mstechly called it out in the original PR description. For the tests, I'd recommend picking a few key examples to test the qref interop functionality and then later we can add 

 - a qref conversion check to the "bloq report card" which is run ad-hoc
 - a qref conversion check to the `bloq_autotester` pytest fixture, which will run against most bloq examples

mstechly: Ok, I've chosen some examples (open to using other examples) and changed the code.

qualtran/qref_interop/bartiq_demo.ipynb
mstechly: Ok, will fix it then :)

mstechly (commented):
I think I have addressed all the feedback, let me know what you think.

mstechly:
@tanujkhattar @mpharrigan – I've signed the CLA and ran all the checks locally to make sure they're passing.

Apart from the outstanding discussion, some other comments:
- I'm importing private methods in `_bloq_to_qref_test.py` like this: `from qualtran.bloqs.arithmetic.addition import _add_oop_large`. If there's a better way to import it, please let me know.
- As @mpharrigan noticed, the version of QREF and Bartiq I suggested were incompatible – I made a new PR to update versions in Qualtran: https://github.com/quantumlib/Qualtran/pull/1284
- Until this PR is merged, the notebooks check might be failing, but I hope others will work. 

Hopefully we can close it soon 😅

mpharrigan:
> I'm importing private methods in _bloq_to_qref_test.py like this: from qualtran.bloqs.arithmetic.addition import _add_oop_large. If there's a better way to import it, please let me know.

This is fine, especially from a test file; and because this is a bloq example, so it's not "private" in the traditional sense.

I'll merge the version bump PR, take a final look over this, and we should be able to merge imminently!

mpharrigan (approved):
looks great! Really nice notebook

qualtran/qref_interop/_bloq_to_qref.py
mpharrigan: We don't have an interface that does exactly this. We often try to treat symbols and concrete values as similarly as possible in the code (this was our design decision; it has its plus-es and minus-es). In practice, this means symbols get passed in as constructor arguments (aka bloq attributes). But they can be arbitrarily hidden: 

```python
    N, L = sympy.symbols("N L", positive=True, integer=True)
    cycle = Shaped((L,))
    permutation_cycle_symb = PermutationCycle(N, cycle)
```

the `L` symbol is passed as a constructor argument, but contained within a `Shaped` object (which we use to support cases where the concrete value is an array but we only need the shape to do e.g. resource counts).

Since everything is an `attrs` dataclass, you can extract the bloq attributes with (pseudocode)

```python
fields = attrs.fields(bloq)
vals = {f.name: getattrs(bloq, f.name) for f in fields}
```

so you can see that the vals dictionary contains `{'N': sympy.Symbol('N')` (easy), and `'cycle': Shaped(sympy.Symbol('L'))}` (you have to poke at it to see that it has a symbol named L

qualtran/qref_interop/_bloq_to_qref.py
mpharrigan: sounds good, I think this is fine as long as we're clear that a `RoutineV1` is analogous to one `CompositeBloq`

PR #1196, opened 2024-07-26
Reorganize Block Encoding notebooks into separate section

charlesyuan314:
Separate the docs that were formally under one page "Block Encoding" into its own sidebar section.

The content has not changed. It will be updated at a later point and a top-level Concept for block encodings will also be added then.

Screenshot 2024-07-26 15 43 12


mpharrigan (approved):
this is not related to the current pr, but I noticed during review: you can remove the `import_line='...'` from most (all) of the bloqdocspecs since it is equivalent to the default. At some point, I may make it so you're only allowed to specify that field if the default doesn't suffice, which would highlight situations that are maybe organized strangely

PR #1197, opened 2024-07-26
Fix dtypes in IntVector and PlusEqualsProduct

anurudhp:
(part of #1142)

### DIFF
--- a/qualtran/bloqs/basic_gates/z_basis.py
+++ b/qualtran/bloqs/basic_gates/z_basis.py
@@ -33,6 +33,7 @@ from qualtran import (
     DecomposeTypeError,
     QAny,
     QBit,
+    QDType,
     Register,
     Side,
     Signature,
@@ -42,7 +43,6 @@ from qualtran import (
 from qualtran.bloqs.bookkeeping import ArbitraryClifford
 from qualtran.cirq_interop.t_complexity_protocol import TComplexity
 from qualtran.drawing import Circle, directional_text_box, Text, TextBox, WireSymbol
-from qualtran.simulation.classical_sim import ints_to_bits
 
 if TYPE_CHECKING:
     import cirq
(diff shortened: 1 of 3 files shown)


qualtran/bloqs/basic_gates/z_basis.py
```diff
@@ -384,12 +384,16 @@ def check(self, attribute, val):
         if val >= 2**self.bitsize:
             raise ValueError(f"`val` is too big for bitsize {self.bitsize}")
 
+    @cached_property
+    def dtype(self) -> QDType:
+        if self.bitsize == 1:
+            return QBit()
+        return QAny(self.bitsize)
```
tanujkhattar, line 391: Is this change supposed to be part of this PR? Looks unrelated.

tanujkhattar (approved):
I'm surprised this does not raise an error because the call site in the decomposition of `QFTPhaseGradient` is passing in a register of type `QFxp` and would need to be updated to include a cast; at the very least. 


It's maybe best to open an issue to track the changes needed at the call site and merge this change so we don't go down the rabbit hole. This is a positive improvement so LGTM

qualtran/bloqs/basic_gates/z_basis.py
anurudhp: I just pulled out the dtype to avoid repeating the code (in signature and build_composite_bloq)

anurudhp:
> I'm surprised this does not raise an error because the call site in the decomposition of `QFTPhaseGradient` is passing in a register of type `QFxp` and would need to be updated to include a cast; at the very least.

I think it's because it uses a cirq style decomp which may not be respecting the dtypes?
https://github.com/quantumlib/Qualtran/blob/174b5493b36a14af3bf0b2c004411386ef972a61/qualtran/bloqs/qft/qft_phase_gradient.py#L82-L97

tanujkhattar:
> I think it's because it uses a cirq style decomp which may not be respecting the dtypes?

I'd be very surprised if that's the root cause. 

@fdmalone Are data type verification checks turned on by default? Do you know what may be going on here?

fdmalone:
@tanujkhattar The type checking will only raise an error if things are grossly wrong. By default it's set to LOOSE (which allows a limited set of conversions without a cast from wholly fractional / wholly integer fxps to uints, is that the case here?)
https://github.com/quantumlib/Qualtran/blob/main/qualtran/_infra/data_types.py#L728

PR #1198, opened 2024-07-27
Fix `AutoPartition` indexing bug with multiple `Unused`

charlesyuan314:
Fix a bug in `AutoPartition` where having `Unused` assigned to the same position in two different exposed registers would map them to the same internal `_unused` register and break decomposition.


mpharrigan (approved):
thanks for the fix

PR #1199, opened 2024-07-27
Add vectorized `from_bits` and `to_bits`

anurudhp:
Adds `from_bits_array` and `to_bits_array`, and uses these functions in classical_sim's `ints_to_bits` and `bits_to_ints` (which will be removed in a follow up PR).

part of #1142

### DIFF
--- a/qualtran/_infra/data_types.py
+++ b/qualtran/_infra/data_types.py
@@ -353,6 +353,10 @@ class QUInt(QDType):
         if is_symbolic(self.bitsize):
             raise ValueError(f"Cannot compute bits for symbolic {self.bitsize=}")
 
+        if self.bitsize > 64:
+            # use the default vectorized `to_bits`
+            return super().to_bits_array(x_array)
+
         w = int(self.bitsize)
         x = np.atleast_1d(x_array)
         if not np.issubdtype(x.dtype, np.uint):
--- a/qualtran/_infra/data_types.py
+++ b/qualtran/_infra/data_types.py
@@ -77,9 +77,29 @@ class QDType(metaclass=abc.ABCMeta):
     def to_bits(self, x) -> List[int]:
         """Yields individual bits corresponding to binary representation of x"""
 
+    def to_bits_array(self, x_array: NDArray[Any]) -> NDArray[np.uint8]:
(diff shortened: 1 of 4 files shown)


qualtran/_infra/data_types.py
```diff
@@ -528,6 +592,13 @@ def from_bits(self, bits: Sequence[int]) -> Fxp:
         fxp_bin = "0b" + bits_bin[: -self.num_frac] + "." + bits_bin[-self.num_frac :]
         return Fxp(fxp_bin, dtype=self.fxp_dtype_str)
 
+    def from_bits_array(self, bits_array: NDArray[np.uint8]):
+        assert isinstance(self.bitsize, int), "cannot convert to bits for symbolic bitsize"
+        # TODO figure out why `np.vectorize` is not working here
```
tanujkhattar, line 597: Add a link to an open issue with the TODO?

tanujkhattar (approved):

qualtran/_infra/data_types.py
anurudhp: I'm planning to remove the `fxpmath` dependency in classical sim for now, in a follow-up PR, so this might become irrelevant. I can perhaps open an issue later when we decide to actually use fxpmath as our floating point library.

tanujkhattar: You want to remove all code that uses `fxpmath` and remove the `QFxp` type ?

anurudhp: No no, the `QFxp` type will stay, but will use integer values (i.e. binary repr of the fxp) for the classical simulator, instead of `Fxp`.

And other uses of `fxpmath` in the code will not be touched, e.g. the current phase gradient `scaled_val` etc.

tanujkhattar: So whenc calling `bloq.call_classically(phase_grad=_fxp_value_)`; the `_fxp_value_` would be an int I believe ?

anurudhp: Yep, like it is right now. I'll add a helper method to `QFxp` to convert a float/Fxp value into the correct int value for passing to the classical sim.

PR #1201, opened 2024-07-27
Fix symbolic expression in `bit_length`

anurudhp:
The previous symbolic expression was `ceil(log2(x))` which is incorrect (e.g. `n=4`). This refactor uses a single correct expression for both symbolic and non-symbolic cases.

- Added a test to verify the new expression matches.
- This also simplifies some awkward looking symbolic terms, e.g. `ceil(log2(2**n - 1))` to `n` (when `n` is a sympy symbol declared with `positive=True, integer=True`)

### DIFF
--- a/qualtran/bloqs/block_encoding/sparse_matrix_test.py
+++ b/qualtran/bloqs/block_encoding/sparse_matrix_test.py
@@ -33,7 +33,6 @@ from qualtran.bloqs.block_encoding.sparse_matrix import (
     UniformEntryOracle,
 )
 from qualtran.resource_counting.generalizers import ignore_split_join
-from qualtran.symbolics import ceil, log2
 
 
 def test_sparse_matrix(bloq_autotester):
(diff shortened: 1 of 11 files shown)


qualtran/bloqs/block_encoding/sparse_matrix_test.py
```diff
@@ -83,8 +82,8 @@ def test_call_graph():
 
     bloq = _sparse_matrix_symb_block_encoding()
     _, sigma = bloq.call_graph(generalizer=ignore_split_join)
-    n = sympy.Symbol('n')
-    assert sigma[Hadamard()] == 6 * ceil(log2(2**n - 1))
+    n = sympy.Symbol('n', integer=True, positive=True)
+    assert sigma[Hadamard()] == 6 * n
```
anurudhp, line 86: one example of a natural simplification

anurudhp, line 86: one example of a natural simplification

tanujkhattar (approved):
Nice!

PR #1202, opened 2024-07-27
Avoid `ArbitraryClifford` in call graphs when exact gates are known

anurudhp:


tanujkhattar (approved):
This is an improvement, thanks!

PR #1204, opened 2024-07-28
QFxp: use integer values for classical sim instead of `fxpmath.Fxp`

anurudhp:
- Use raw integer values for simulating QFxp
- Move `fxpmath` based functionality in QFxp to private functions

To pass inputs to the classical simulator, one should use `QFxp.to_fixed_width_int` to convert a float to the correct int for simulation. 

(part of #1142)

In a follow-up I will remove the classical_sim.py `ints_to_bits` and `bits_to_ints` and update all bloqs to use qdtype features.


qualtran/_infra/data_types.py
```diff
@@ -533,34 +534,107 @@ class QFxp(QDType):
...
+                raise ValueError("bitsize must be >= num_frac.")
+
     @property
     def num_qubits(self):
         return self.bitsize
 
     @property
     def num_int(self) -> SymbolicInt:
-        return self.bitsize - self.num_frac - int(self.signed)
+        return self.bitsize - self.num_frac
```
charlesyuan314, line 559: This doesn't seem consistent with the docstring anymore?
> In total there are `bitsize = (n_sign + num_int + num_frac)` bits used

fdmalone:
What's the motivation here? Is it faster?

anurudhp:
> What's the motivation here? Is it faster?

Yeah, Fxp is some quite slow, about 2-5x. And `fxpmath` is quite unstable, it does not correctly preserve behaviour for truncation etc. I ran into a lot of issues in #1142.

The current simulator uses ints for QFxp as well (as it always uses `ints_to_bits` and `bits_to_ints`), so this PR does not change that behaviour. It just deprecates the older functions, and always uses the QDType interfaces. And when we finalize which library to use for fixed points later, we can just update the code in QFxp without touching the rest of the codebase.

fdmalone:
Can / should we remove the fxpmath dependency?

qualtran/_infra/data_types.py
anurudhp: thanks!

mpharrigan (commented):
I'm a little worried about this change, although I understand the frustration with Fxp. One of the primary motivations of the classical simulation protocol is to write classical programs that look like classical programs. This is super useful for a) unit testing, where one needs to inspect one side of the assertion and deem it obviously correct and b) looking at what a bloq does. 

Let me think some more about this

mpharrigan:
Can you describe the motivations, consequences, and potential alternatives for this change?

anurudhp:
### Motivation

Currently, the library mostly uses `ints_to_bits` and `bits_to_ints` from `classical_sim.py` to convert classical values (in Split, Join, Partition etc.), and `QDType.from_bits` and `.to_bits` to generate values in the classical simulator. This behaviour is divergent for `QFxp` as the former uses a fixed-width-int representation and the latter uses `fxpmath.Fxp`.

All existing bloqs assume the input values are integers for QFxp registers, e.g. Phase Gradient:
https://github.com/quantumlib/Qualtran/blob/15cb66190e5382b155aa4d2bdcdc0ec6b1b92c45/qualtran/bloqs/rotations/phase_gradient.py#L312-L325

### This PR
- Consolidates the conversions, removing the classical sim functions, and always using `QDType.from_bits` and `.to_bits`.
- Use fixed-witdth-int representations always as `fxpmath` is unstable at the moment.

### Consequences
- End users must use `QFxp.to_fixed_width_int(actual_floating_point_value)` to pass the correct input values to `bloq.call_classically`.

### Potential Alternatives
- We would need a stable fixed-point library, or write our own light-weight wrappers around floats.
- We also need to explicitly update classical calls for bloqs using QFxp (like `Add` etc) to work with QFxp. Right now they work only for ints, and coincidentally the behaviour of the integer representations for Fxp matches this.

mpharrigan:
Thanks for the write-up!

> All existing bloqs assume the input values are integers for QFxp registers, e.g. Phase Gradient:

If this is true, then this removes some of my concern about the current PR as a near-term solution to the problem(s).

Can you add some text to the classical sim docs and classical sim docstrings that explains the behavior? Do we want to say what the integers are? The to_fixed_width_int says "re-interpret the bits as an int" but how are the bits laid out in Fxp? 

A more robust approach could be to create a lightweight wrapper around _ints_ that at least makes the user call the classical functions correctly rather than having to convert to int and back

```python
class FxpVal:
  _as_int: int
  total_bits: int
  frac_bits: int

  def __add__(self, other):
    return self._as_int + other._as_int

  # etc
```

qualtran/_infra/data_types.py
```diff
@@ -521,46 +521,120 @@ class QFxp(QDType):
...
+
+    def from_bits(self, bits: Sequence[int]):
+        return self._int_qdtype.from_bits(bits)
+
+    def assert_valid_classical_val(self, val: int, debug_str: str = 'val'):
+        self._int_qdtype.assert_valid_classical_val(val, debug_str)
+
+    def to_fixed_width_int(self, x: Union[float, Fxp], *, require_exact: bool = False) -> int:
+        """Returns the interpretation of the binary representation of `x` as an integer."""
+        bits = self._fxp_to_bits(x, require_exact, require_exact)
```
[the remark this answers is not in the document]
anurudhp: oops! this was a typo. Fixed and used a kwarg to be safe.

qualtran/_infra/data_types.py
```diff
@@ -521,46 +521,120 @@ class QFxp(QDType):
...
-    then $\pi$ \approx 3.140625 = 11.001001, where the . represents the decimal place.
+    real number is signed we store negative values in two's complement form. The first
+    bit can therefore be treated as the sign bit in such cases (0 for +, 1 for -).
+    In total there are `bitsize = (num_int + num_frac)` bits used to represent the number.
+    E.g. Using `(bitsize = 8, num_frac = 6, signed = False)` then
+    $\pi$ \approx 3.140625 = 11.001001, where the . represents the decimal place.
 
     We can specify a fixed point real number by the tuple bitsize, num_frac and
-    signed, with num_int determined as `(bitsize - num_frac - n_sign)`.
+    signed, with num_int determined as `(bitsize - num_frac)`.
```
mpharrigan, line 531: There's some docstrings and comments hiding around the code. Can you add one or two short paragraphs to the class docstring for QFxp to describe the relationship to the classical simulation protocol and how we'll use these "raw ints" in the classical simulation protocol.

Comments don't show up in the docs and may be missed, and docstrings on private methods may be missed. You can also add a line to the docstrings of the public methods that says something like 'see the class docstring for details on the classical simulation format'

qualtran/_infra/data_types.py
```diff
@@ -521,46 +521,120 @@ class QFxp(QDType):
...
+    def _int_qdtype(self) -> Union[QUInt, QInt]:
+        """The corresponding integer type used to represent raw values of this type.
+
+        This raw integer value is used in the classical simulator to represent values
+        of QFxp registers.
+
+        For example, QFxp(6, 2) has 2 int bits and 4 frac bits, and the corresponding
+        int type is QUInt(6). So a true classical value of `10.0011` will have a raw
+        integer representation of `100011`.
+        """
```
mpharrigan, line 574: might be worth moving the bulk of this discussion to the class docstring and then just saying "the dtype in accordance with the format described in the class docstring"

qualtran/_infra/data_types.py
```diff
@@ -521,46 +521,120 @@ class QFxp(QDType):
...
+        return QInt(self.bitsize) if self.signed else QUInt(self.bitsize)
+
+    def get_classical_domain(self) -> Iterable[int]:
+        yield from self._int_qdtype.get_classical_domain()
```
mpharrigan, line 578: this needs a docstring, since it would be surprising to an unsuspecting user. If you bulk out the class docstring, it can be a one-liner that refers to the complete description in the class docstring.

qualtran/_infra/data_types.py
```diff
@@ -521,46 +521,120 @@ class QFxp(QDType):
...
+
+    def to_bits(self, x) -> List[int]:
+        return self._int_qdtype.to_bits(x)
+
+    def from_bits(self, bits: Sequence[int]):
+        return self._int_qdtype.from_bits(bits)
+
+    def assert_valid_classical_val(self, val: int, debug_str: str = 'val'):
+        self._int_qdtype.assert_valid_classical_val(val, debug_str)
```
mpharrigan, line 587: short docstring

qualtran/_infra/data_types.py
```diff
@@ -521,46 +521,120 @@ class QFxp(QDType):
...
+
+    def to_fixed_width_int(self, x: Union[float, Fxp], *, require_exact: bool = False) -> int:
+        """Returns the interpretation of the binary representation of `x` as an integer."""
```
mpharrigan, line 590: Refer to a description of the format. Do we want users to _know_ what the classical format is? Yes, probably.

qualtran/_infra/data_types.py
```diff
@@ -521,46 +521,120 @@ class QFxp(QDType):
...
+        if self.signed:
+            return f'QFxp({self.bitsize}, {self.num_frac}, True)'
+        else:
+            return f'QFxp({self.bitsize}, {self.num_frac})'
+
+    # Experimental `fxpmath.Fxp` support.
+    # This support is currently experimental, and does not hook into the classical
+    # simulator protocol. Once the library choice for fixed-point classical real
+    # values is finalized, the code will be updated to use the new functionality
+    # instead of delegating to raw integer values (see above).
```
mpharrigan, line 604: this can migrate to the docstring of a public method. Probably `fxp_dtype_template`

qualtran/_infra/data_types.py
```diff
@@ -521,46 +521,120 @@ class QFxp(QDType):
...
 
     @property
-    def _fxp_dtype(self) -> Fxp:
-        return Fxp(None, dtype=self.fxp_dtype_str)
+    def fxp_dtype_template(self) -> Fxp:
```
mpharrigan, line 607: why are we using the word "template" here? because it doesn't actually have a value associated with it?

```
"""Prepare an empty `fxpmath.Fxp` value container for experimental fixed point support.

This constructs a `Fxp` object with no value. To assign the returned object a fixed point value, 
use ... [idk, exaplain how to use it].

Fxp support is experimental and doesn't hook into the classical simulator protocol etc etc

This corresponds to the Fxp constructor arguments:
 - op sizing, etc etc
"""
```

mpharrigan: I'd also make this a method instead of a property

mpharrigan (commented):
preliminary lgtm, but some documentation suggestions

qualtran/_infra/data_types.py
anurudhp: I used the term from the fxpmath readme: https://github.com/francof2a/fxpmath#:~:text=It%20is%20a%20good%20idea%20create%20Fxp%20objects%20like%20template%3A

This is used to type-cast Fxp values, instead of being an empty container, hence a property.

qualtran/_infra/data_types.py
anurudhp: done

qualtran/_infra/data_types.py
anurudhp: I've updated the docstring to explain how to use this function

qualtran/_infra/data_types.py
anurudhp: Updated the class docstring

qualtran/_infra/data_types.py
anurudhp: Moved to class docstring

qualtran/_infra/data_types.py
anurudhp: added

anurudhp:
@mpharrigan ptal! I've updated the docstrings, and added a brief explanation in the `classical_sim.ipynb` notebook about how to use QFxp correctly.

qualtran/_infra/data_types.py
```diff
@@ -530,6 +530,23 @@ class QFxp(QDType):
     We can specify a fixed point real number by the tuple bitsize, num_frac and
     signed, with num_int determined as `(bitsize - num_frac)`.
 
+
+    Classical Simulation:
```
mpharrigan, line 534: doc nit: These "numpydoc-style" headings have semantic meaning. Otherwise, use markdown. Here, I'd suggest using a `<h3>`

```
### Classical simulation

To hook into the classical simulator...
```

qualtran/_infra/data_types.py
```diff
@@ -563,56 +580,109 @@ def is_symbolic(self) -> bool:
...
     def fxp_dtype_template(self) -> Fxp:
         """A template of the `Fxp` data type for classical values.
 
-        -   op_sizing='same' and const_op_sizing='same' ensure that the returned object is not resized
-            to a bigger fixed point number when doing operations with other Fxp objects.
-        -   shifting='trunc' ensures that when shifting the Fxp integer to left / right; the digits are
-            truncated and no rounding occurs
-        -   overflow='wrap' ensures that when performing operations where result overflows, the overflowed
-            digits are simply discarded.
+        Usage:
```
mpharrigan, line 665: section formatting -- but you probably don't need this as a section. the docstring is always about usage :)

qualtran/_infra/data_types.py
```diff
@@ -563,56 +580,109 @@ def is_symbolic(self) -> bool:
...
+
+            To construct an `Fxp` with this config, one can use:
+            `Fxp(float_value, like=QFxp(...).fxp_dtype_template)`,
+            or given an existing value `some_fxp_value: Fxp`:
+            `some_fxp_value.like(QFxp(...).fxp_dtype_template)`.
+
+        The following Fxp configuration is used:
+            - op_sizing='same' and const_op_sizing='same' ensure that the returned
```
mpharrigan, line 673: do markdown bullets `[space]-[space]` (no indent)

qualtran/_infra/data_types.py
```diff
@@ -563,56 +580,109 @@ def is_symbolic(self) -> bool:
...
+              object is not resized to a bigger fixed point number when doing
+              operations with other Fxp objects.
+            - shifting='trunc' ensures that when shifting the Fxp integer to
+              left / right; the digits are truncated and no rounding occurs
+            - overflow='wrap' ensures that when performing operations where result
+              overflows, the overflowed digits are simply discarded.
+
+        Notes:
```
mpharrigan, line 681: this doesn't really need to be delineated into a section

qualtran/_infra/data_types.py
mpharrigan: you're never going to want to plumb through any of the options?

mpharrigan (approved):
lgtm after docstring formatting fixes. you can use the scripts in dev_tools to preview how the api reference docs get rendered

qualtran/_infra/data_types.py
anurudhp: fixed

qualtran/_infra/data_types.py
anurudhp: removed

qualtran/_infra/data_types.py
anurudhp: thanks! fixed

PR #1206, opened 2024-07-29
Move Trigonometric bloqs to common location

anurudhp:
- new folder `arithmetic/trigonometric`
- Move `ArcTan` from `mean_estimation/`


anurudhp:
@tanujkhattar do you remember the reference for the TComplexity of arctan? https://github.com/quantumlib/Qualtran/blob/e1bf3940ad5cc9368f4c5c9948f3305c822544d6/qualtran/bloqs/mean_estimation/arctan.py#L60-L62

tanujkhattar:
@anurudhp Thanks for looking into this. The TComplexity of arctan is a gross under estimation and I'm the culprit for it :)

It comes from https://arxiv.org/pdf/2208.07544 which states the complexity to compute arctan; in bigO notation, should be $\mathcal{O}(n\log{n})$ and I dropped the $\log{n}$ term when implementing the "approximate" TComplexity here. 

However, the argument for big-O complexity in the linked references essentially follows from reducing arctan to a bunch of constant multiplications (eg: as done in https://stackoverflow.com/a/23097989) and then using the complexity to multiply two numbers as $\mathcal{O}(n\log{n})$ from https://projecteuclid.org/journals/annals-of-mathematics/volume-193/issue-2/Integer-multiplication-in-time-Onmathrmlog-n/10.4007/annals.2021.193.2.4.short; which is an important theoretical result but requires very large numbers for the nlogn scaling to kick in. The best quantum implementations of multiplication circuits is probably https://arxiv.org/abs/2403.18006 which scales as $n^{1.3}$

Tl;Dr - We should probably update the approximate T-count to $\mathcal{O}(n^2)$, which is also the scaling we use for `PlusEqualProduct` and arctan reduces to a constant number of `PlusEqualProduct`.

anurudhp:
> Tl;Dr - We should probably update the approximate T-count to O(n^2) which is also the scaling we use for `PlusEqualProduct` and arctan reduces to a constant number of `PlusEqualProduct`.

Would it still be a constant number of `PlusEqualProduct`? Would it not depend on the target bitsize?

PR #1209, opened 2024-07-30
Make param `ctrl_spec` of `get_ctrl_system` mandatory

anurudhp:
Makes custom override code much cleaner as the user doesn't have to check for None.

For example, a custom single controlled version would look like:
```py
    if ctrl_spec != CtrlSpec(): # default fallback
        return super().get_ctrl_system(ctrl_spec=ctrl_spec)
    
    ... # custom single controlled version.
```

instead of
```py
def get_ctrl_system(
    self, ctrl_spec: Optional['CtrlSpec'] = None
) -> Tuple['Bloq', 'AddControlledT']:
    if not (ctrl_spec is None or ctrl_spec == CtrlSpec()):
        return super().get_ctrl_system(ctrl_spec=ctrl_spec)

    ... # custom
```

This does not affect bloq users, who can still use `bloq.controlled()`. It only makes it simpler for bloq authors to provide custom versions.


tanujkhattar (commented):
LGTM from my side -- The call site doesn't change with it; so users should still be able to do `bloq.controlled()` which is important. 

cc @mpharrigan to see if there are any concerns and approve.

mpharrigan (approved):
lgtm, @anurudhp pinged me about this offline before :)

PR #1210, opened 2024-07-30
Seed chemistry examples

### LINKED ISSUE #1170: `thc.ipynb` assertion fails occasionally
This notebook has started to fail occasionally as of a few weeks ago.

Example log [here](https://github.com/quantumlib/Qualtran/actions/runs/10044825760/job/27760744975?pr=1169).

fdmalone:
Resolves #1170.

I couldn't immediately figure out why this is happening in practice as the T count of the QROAM bloq should just depend on the length of the data and the target bitsizes / which are both identical for these two examples in the new unit tests... but there does seem to be a saving of 2 Ands / Toffolis which is probably worth investigating more. In this case there are no swaps so the cost is just plain old qroms.


qualtran/bloqs/chemistry/thc/prepare_test.py
```diff
@@ -36,12 +39,32 @@ def test_thc_prepare(bloq_autotester):
...
+        num_spat: The number of spatial orbitals.
+        seed: If not None then seed the rng with this value. Otherwise seed is
+            not set explcitly in this function.
+
+    Retuns:
+        t_l: The eigenvalues of the one-body Hamiltonian.
+        zeta: The central tensor (matrix) of size num_mu x num_mu.
+    """
+    if seed is not None:
+        np.random.seed(seed)
```
mpharrigan, line 56: can you make a new np.random.RandomState instead of (re-)seeding the global generator? And the constructor will do the correct thing for `None` values so you don't need the conditional

fdmalone:
> Yeah I'm curious how this can change the t count. Could you reproduce different t counts with different random seeds or is this still conjecture?

Yes, the unit tests uses two different seeds which captures the flakiness. I'll have another look.

qualtran/bloqs/chemistry/thc/prepare_test.py
fdmalone: TIL

fdmalone:
Phew, our QROM bloq is too smart. The data had some repeated pattern accidentally which caused variable spaced qrom to kick in! I wonder if we should flag that this happens to avoid confusion. I'm not sure if this is just a recent thing when the QROM bloq interfaces were unified. Maybe variable space qrom wasn't enabled for QROAM until then @tanujkhattar ?

mpharrigan:
Thanks for investigating and for opening an issue!

mpharrigan (approved):

tanujkhattar:
>  I'm not sure if this is just a recent thing when the QROM bloq interfaces were unified. Maybe variable space qrom wasn't enabled for QROAM until then @tanujkhattar ?

I don't think this is affected by the unification of QROM interfaces. This has been this way since the variable spaced QROM optimization was implemented. 

I think earlier we were using a custom `QROAM` bloq in chemistry and we switched to using `SelectSwapQROM` after the unification of qrom interfaces; that may have triggered the change.

fdmalone:
Makes sense

PR #1212, opened 2024-07-30
Add a CI workflow to run pytest nightly

charlesyuan314:
This workflow runs `check/pytest` (all tests) at 4:00 AM PT (11:00 AM UTC) every Monday-Friday.

It can be a first step toward reducing the runtime of `pytest` on pull requests.


charlesyuan314:
Note from [GitHub](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows):

> When the last user to commit to the cron schedule of a workflow is removed from the organization, the scheduled workflow will be disabled. If a user with write permissions to the repository makes a commit that changes the cron schedule, the scheduled workflow will be reactivated. Note that, in this situation, the workflow is not reactivated by any change to the workflow file; you must alter the cron value and commit this change.

mpharrigan:
what counts as a user committing? if I press the merge button is it me?

mpharrigan (approved):

charlesyuan314:
Yes, I think you are a committer if you press merge.

charlesyuan314:
To be safe, I think you should also press merge on the PR so you show up as coauthor of the PR in the commit log.

mpharrigan:
> To be safe, I think you should also press squash-and-merge on the PR so you show up as coauthor of the PR in the commit log.

sorry, that's the button I meant. I think i'll be the "committer" but you will be the "author"

PR #1214, opened 2024-07-30
Remove redundant import_line from BloqDocSpec

charlesyuan314:
These `import_line`s are all equivalent to the default generated by `BloqDocSpec`.

There may be more that are technically redundant but are different from their current state in the notebook (i.e. import from a child rather than parent module). They can be addressed in a follow-up PR.

### DIFF
--- a/qualtran/bloqs/basic_gates/identity.py
+++ b/qualtran/bloqs/basic_gates/identity.py
@@ -100,8 +100,4 @@ def _identity() -> Identity:
     return identity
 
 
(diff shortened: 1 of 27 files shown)


mpharrigan:
> There may be more that are technically redundant but are different from their current state in the notebook (i.e. import from a child rather than parent module). They can be addressed in a follow-up PR.

sounds good

mpharrigan (approved):

PR #1215, opened 2024-07-30
Use QDType bit interfaces for classical simulation

### LINKED ISSUE #606: Refactor `classical_sim.py` bitwise functions to allow for signed values

### LINKED ISSUE #643: Support data types in classical simulation

### LINKED ISSUE #811: Consolidate bit iteration

anurudhp:
Final part of #1142!

- Remove classical sim `ints_to_bits` and `bits_to_ints`, and always use QDType bit interfaces (`from_bits, to_bits` etc.).
- Update all bookeeping bloqs to use the QDType bit features. 

Fixes #811 
Fixes #606 
Fixes #643

### DIFF
--- a/qualtran/_infra/data_types.py
+++ b/qualtran/_infra/data_types.py
@@ -547,6 +547,9 @@ class QFxp(QDType):
     int type is QUInt(6). So a true classical value of `10.0011` will have a raw
     integer representation of `100011`.
 
+    See https://github.com/quantumlib/Qualtran/issues/1219 for discussion on alternatives
+    and future upgrades.
+
 
     Attributes:
         bitsize: The total number of qubits used to represent the integer and
--- a/qualtran/bloqs/bookkeeping/cast.py
+++ b/qualtran/bloqs/bookkeeping/cast.py
@@ -26,7 +26,6 @@ from qualtran import (
     ConnectionT,
     DecomposeTypeError,
     QDType,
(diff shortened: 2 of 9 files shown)


mpharrigan:
Since this will close all those bugs, can you open a bug for classical simulation of fixed point. A brief overview of the current situation where you pass around ints because of deficiencies in fxpmath; and potential alternatives going forward

anurudhp:
> Since this will close all those bugs, can you open a bug for classical simulation of fixed point. A brief overview of the current situation where you pass around ints because of deficiencies in fxpmath; and potential alternatives going forward

Done!

mpharrigan (approved):

PR #1216, opened 2024-07-30
Remove more redundant import_line from BloqDocSpec

charlesyuan314:
Follow-up of #1214.

Removes `import_line`s that are essentially redundant with the default of `BloqDocSpec`, differing only by importing from a child rather than parent module. The notebooks change but still run.


anurudhp (approved):

mpharrigan:
thank you

PR #1223, opened 2024-07-30
Fix `alpha` in `LinearCombination`

charlesyuan314:
- Ensure that `alpha` in `LinearCombination` is correctly scaled by $||\lambda||$
- Correctly scale coefficient $\lambda_i$ by $\alpha_i$ when constituent block encoding has $\alpha_i \neq 1$
- Update tests to compare $A = (A/\alpha)\alpha$ with the top-left block of $\mathcal{B}[A]$


anurudhp (approved):
lgtm

PR #1227, opened 2024-07-31
Mark tests taking >10s with `pytest.mark.slow`

charlesyuan314:
These tests are not currently marked as slow but are in fact slower than many tests that are currently marked as slow.

### DIFF
--- a/qualtran/bloqs/arithmetic/bitwise.ipynb
+++ b/qualtran/bloqs/arithmetic/bitwise.ipynb
@@ -172,7 +172,7 @@
    },
    "outputs": [],
    "source": [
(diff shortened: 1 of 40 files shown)


charlesyuan314:
Please suggest a change if any test here is critical and should not be skipped on per-PR workflows if we were to skip slow tests.

mpharrigan (approved):

mpharrigan:
thank you for your vigilance 

now that we have the nightly tests -- if any of the nightly tests fail but the PR checks passed, I will mandate that the original author go back and write a "quick" test that would have caught the bug :) if possible

PR #1228, opened 2024-07-31
Mark tests taking >4s with `pytest.mark.slow`

charlesyuan314:
Continuation of #1227. There is a clear cluster of tests that take >4s whereas the vast majority of the remaining tests take <2s.


charlesyuan314:
Please suggest a change if any test here is critical and should not be skipped on per-PR workflows if we were to skip slow tests.

qualtran/bloqs/arithmetic/controlled_add_or_subtract_test.py
```diff
@@ -47,6 +47,7 @@ def test_examples(bloq_autotester):
     bloq_autotester(_ctrl_add_or_sub_signed_symb)
 
 
+@pytest.mark.slow
 @pytest.mark.parametrize("bitsize", [2, 4, pytest.param(5, marks=pytest.mark.slow)])
```
anurudhp, line 51: Here only `5` is slow right?

On my machine:
```
2: 77ms
4: 1.271s
5: 5.428s
```

charlesyuan314: You're right, I missed that `5` was already marked, so this one is good.

charlesyuan314: I'm looking through the rest as well.

qualtran/bloqs/basic_gates/power_test.py
```diff
@@ -87,5 +87,6 @@ def to_cirq_circuit(bloq: GateWithRegisters) -> cirq.Circuit:
     )
 
 
+@pytest.mark.slow
```
mpharrigan, line 90: these test_no_circular_import tests are all redundant with each other. I added it following a bug report where there was a tricky circular import that didn't usually get triggered; but I added one test because these were all in one module. During some reorganization work, the one module got split; and these tests got duplicated. Can you remove all but one of them (and mark as slow)

mpharrigan:
The cluster is probably because I go through every once in a while with a roughly 2s threshold but then more sneak in :)

qualtran/bloqs/basic_gates/power_test.py
charlesyuan314: I kept the ones in `qualtran.bloqs.basic_gates.power` and `qualtran.bloqs.bookkeeping.arbitrary_clifford`.

If those two are still redundant I can remove one of them.

mpharrigan: thank you

mpharrigan (approved):

PR #1233, opened 2024-07-31
Fix reflection in `ChebyshevPolynomial`

charlesyuan314:
- Move the position of the first reflection operator in the decomposition. This was from a different reference and appears to be wrong (generates a sign flip). The code now follows the stated reference and works correctly for order 1.
- Change the tests to use only Hermitian matrices $A$. The previous tests incorrectly implied that Chebyshev polynomials can be computed correctly for non-Hermitian matrices, which is not true in general.


qualtran/bloqs/block_encoding/chebyshev_polynomial.py
```diff
@@ -162,11 +160,10 @@ def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
 
 @bloq_example
 def _chebyshev_poly_even() -> ChebyshevPolynomial:
-    from qualtran.bloqs.basic_gates import Hadamard, XGate
-    from qualtran.bloqs.block_encoding import LinearCombination, Unitary
+    from qualtran.bloqs.basic_gates import XGate
+    from qualtran.bloqs.block_encoding import Unitary
 
-    bloq = LinearCombination((Unitary(XGate()), Unitary(Hadamard())), (0.5, 0.5), lambd_bits=1)
```
anurudhp, line 168: isn't $X + H$ is also hermitian?

qualtran/bloqs/block_encoding/chebyshev_polynomial_test.py
```diff
@@ -130,16 +125,15 @@ def test_scaled_chebyshev_poly_odd(bloq_autotester):
     bloq_autotester(_scaled_chebyshev_poly_odd)
 
 
-@pytest.mark.slow
-def test_scaled_chebyshev_even_tensors():
-    from_gate = t4((XGate().tensor_contract() + Hadamard().tensor_contract()))
+def test_scaled_chebyshev_poly_even_tensors():
+    from_gate = t2(XGate().tensor_contract() * 3.14)
```
anurudhp, line 129: Would be good to have a test where the input block-encoding doesn't satisfy $U^2 = I$. Perhaps using `MatrixGate.random`?

qualtran/bloqs/block_encoding/chebyshev_polynomial.py
charlesyuan314: Nice catch. It isn't, but $(X+H)/\sqrt{2+\sqrt{2}}$ is. That makes it suitable for a test for `ScaledChebyshevPolynomial` with $\alpha = \sqrt{2+\sqrt{2}}$, and I have made it one. It won't work for `ChebyshevPolynomial` though.

PR #1235, opened 2024-07-31
Add notebook tests for block encodings

charlesyuan314:


anurudhp:
Are these notebook tests needed anymore? The CI seems to just run `dev_tools/execute-notebooks.py` which collects and runs all notebooks.

https://github.com/quantumlib/Qualtran/blob/a49d0e03f931d50c337371d6f89d3c2fff666987/.github/workflows/ci.yaml#L53-L67

charlesyuan314:
This is what I got from @mpharrigan:

> there should be a pytest for each notebook marked with that mark so local users can easily check the notebooks alongside their unit tests and isolate failures

fdmalone:
The notebook tests are skipped for pytest during the CI (tested via execute_notebook) but will run locally (if you forget to -m "not notebook" or whatever

anurudhp (approved):

PR #1246, opened 2024-08-05
Remove legacy register methods from `BlockEncoding` interface

charlesyuan314:
These methods have been superceded by corresponding `bitsize` methods.


mpharrigan (approved):
woo!

Issue #1247, opened 2024-08-05
Implement `alpha` and `epsilon` for chemistry `BlockEncoding`s


charlesyuan314:
These methods from the `BlockEncoding` interface need to be implemented for `{Single,Double}Factorization{OneBody,BlockEncoding}` for these parameters to be automatically propagated to composite block encodings.

PR #1248, opened 2024-08-05
Make methods `abstractmethod` in `BlockEncoding`

charlesyuan314:
See #1247.

These methods were not previously marked as abstract due to not being implemented in chemistry block encodings.

### DIFF
--- a/qualtran/bloqs/chemistry/df/double_factorization.py
+++ b/qualtran/bloqs/chemistry/df/double_factorization.py
@@ -122,13 +122,13 @@ class DoubleFactorizationOneBody(BlockEncoding):
 
     @property
     def alpha(self) -> float:
-        # TODO: correct if necessary, see https://github.com/quantumlib/Qualtran/issues/1247
-        return 1.0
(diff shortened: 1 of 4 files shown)


qualtran/bloqs/chemistry/df/double_factorization.py
```diff
@@ -120,6 +120,28 @@ def control_registers(self) -> Iterable[Register]:
             Register("succ_p", QBit()),
         )
 
+    @property
+    def alpha(self) -> float:
+        # TODO: correct if necessary, see https://github.com/quantumlib/Qualtran/issues/1247
+        return 1.0
```
anurudhp, line 126: what happens if you raise NotImplemented here?

anurudhp (approved):

qualtran/bloqs/chemistry/df/double_factorization.py
charlesyuan314: We discussed; changing to raise exception here.

PR #1255, opened 2024-08-06
[prototype] Use rotation `eps` in `GateCounts`

anurudhp:


qualtran/surface_code/beverland_et_al_model.py
```diff
@@ -116,14 +116,12 @@ def n_discrete_logical_gates(
...
-    if n_rotations > 0:
-        ret = (
-            ret
-            + rotation_model.preparation_overhead(eps_syn)
-            + n_rotations * rotation_model.rotation_cost(eps_syn / n_rotations)
-        )
+    rotation_epsilons: dict[float, int] = alg.n_logical_gates.rotation_epsilons
+    ret = attrs.evolve(alg.n_logical_gates, rotation_epsilons={})
+    if rotation_epsilons:
+        rotation_model.preparation_overhead(min(eps for eps in rotation_epsilons.values()))
```
anurudhp, line 122: what's the correct way to pick the eps for this?

mpharrigan:
we had discussed binning by ceil(log(eps)) -- can you comment on that

anurudhp:
@tanujkhattar's comment:
> It's better to store frequency of `eps` by assuming we can represent float `eps` as a fraction `x / 2^{precision}` and store frequencies of `x`; with `precision` as a configurable parameter in the constructor of `QECGateCount` class. It's also fine to store frequency of `eps` directly for now and make this change as a future improvement. I'd advice against storing frequency of `ceil(log2(1/eps))` since it's a very aggressive approximation.

Binning by the above integer approximation can be implemented.

My bigger concern is the various epsilons floating around:
- `bloq.eps` for rotation bloqs
- `error_budget` in the surface code costing functions

Should we remove this `error_budget` and only use `bloq.eps`?

mpharrigan:
Certain physical cost functions make certain assumptions, and we've tried to implement them as described in the literature. The coupling between the logical costs and the physical cost models is done -- in this case -- via the methods on `GateCounts`. So, since the beverland physical cost model just counts "rotations" and gives one third of the total error budget to synthesizing them, the `get_beverland_counts` method should just return the number of rotations in the appropriate field. The gidney-fowler models consume `get_total_t_and_ccz_counts`, which can use the per-gate epsilon values to turn it into t and ccz counts

qualtran/surface_code/beverland_et_al_model.py
mpharrigan: `eps_syn` is passed in to this function

qualtran/surface_code/ui.py
```diff
@@ -522,24 +522,30 @@ def update(
...
+    rotation_model_name: str,
 ):
     """Updates the visualization."""
     if any(x is None for x in [physical_error_rate, error_budget, *algorithm_data, magic_count]):
         raise PreventUpdate
 
     # TODO: We implicitly rely on the order of the input components
     qubits, measurements, ts, toffolis, rotations, n_rotation_layers = algorithm_data
+
+    rotation_eps = 1e-3 / rotations  # TODO this should not be a magic number
```
mpharrigan, line 534: should we support "unknown precision" rotations. I guess in the rotation bloqs we have a default eps; but it's not hard to imagine a situation where a researcher has a bunch of rotations and they haven't computed an epsilon for each of them

anurudhp: I'm using `error_budget / rotations` here

qualtran/surface_code/beverland_et_al_model.py
anurudhp: fixed to ignore bloq eps, as discussed

anurudhp:
@mpharrigan ptal!

mpharrigan:
taking a look now. Can you update the original PR description with a summary of changes in this pr

qualtran/resource_counting/_bloq_counts.py
```diff
@@ -128,21 +135,69 @@ class GateCounts:
...
+    )
+    eps_bin_prec: int = 20
+
+    @classmethod
+    def from_rotation_with_eps(cls, eps: float, *, eps_bin_prec: int = 20, n_rotations: int = 1):
+        """Construct a GateCount with a rotation of precision `eps`.
+
+        Args:
+            eps: precision to synthesize the rotation(s).
+            eps_bin_prec: number of bits to approximate `eps` to, defaults to 10.
```
mpharrigan, line 150: doc mismatch re: default

qualtran/resource_counting/_bloq_counts.py
```diff
@@ -128,21 +135,69 @@ class GateCounts:
...
+        for eps_bin, n_rot in self.binned_rotation_epsilons.items():
+            yield eps_bin / 2**self.eps_bin_prec, n_rot
 
     def __add__(self, other):
         if not isinstance(other, GateCounts):
             raise TypeError(f"Can only add other `GateCounts` objects, not {self}")
 
+        eps_bin_prec = max(self.eps_bin_prec, other.eps_bin_prec)
+        this = self.with_rotation_eps_bin_prec(eps_bin_prec)
+        other = other.with_rotation_eps_bin_prec(other.eps_bin_prec)
```
mpharrigan, line 190: why do you do this for addition but not multiplication?

qualtran/resource_counting/_bloq_counts.py
```diff
@@ -174,36 +232,59 @@ def _is_nonzero(v):
...
+        """T-cost of a single Z rotation with precision `eps`.
+
+        References:
+            [Efficient synthesis of universal Repeat-Until-Success circuits](https://arxiv.org/abs/1404.5320)
+            Bocharov et. al. 2014. Page 4, Paragraph "Simulation Results."
+        """
+        return ceil(1.149 * log2(1.0 / eps) + 9.2)
+
+    @property
+    def rotation_to_t(self) -> SymbolicInt:
```
mpharrigan, line 256: I'm not sure I like this as a property. Even its name suggests an "action". It's name also suggests that it takes rotations and turns it to t. It sounds very similar to the static method name `rotation_t_cost` if I didn't see the code.

Suggest: `total_rotations_as_t()`

qualtran/resource_counting/_bloq_counts.py
```diff
@@ -174,36 +232,59 @@ def _is_nonzero(v):
...
-    def total_t_and_ccz_count(self, ts_per_rotation: int = 11) -> Dict[str, SymbolicInt]:
+    def total_t_and_ccz_count(self) -> Dict[str, SymbolicInt]:
         n_ccz = self.toffoli + self.cswap + self.and_bloq
-        n_t = self.t + ts_per_rotation * self.rotation
+        n_t = self.t + self.rotation_to_t
         return {'n_t': n_t, 'n_ccz': n_ccz}
 
+    def n_rotation_ignoring_eps(self) -> SymbolicInt:
+        """Total number of rotations, ignoring the individual precisions."""
+        return ssum(self.binned_rotation_epsilons.values())
```
mpharrigan, line 286: now _this_ could be a property. 

We don't use the `n_xx` prefix anywhere else in this class. Consider `rotation(s)_ignoring_eps`.

qualtran/resource_counting/_bloq_counts.py
```diff
@@ -128,21 +135,69 @@ class GateCounts:
     cswap: SymbolicInt = 0
     and_bloq: SymbolicInt = 0
     clifford: SymbolicInt = 0
-    rotation: SymbolicInt = 0
     measurement: SymbolicInt = 0
+    binned_rotation_epsilons: Counter[int] = field(
+        factory=Counter, converter=_mapping_to_counter, eq=lambda d: tuple(d.items())
+    )
+    eps_bin_prec: int = 20
```
mpharrigan, line 142: please update the docstring with a thorough description of the representation of rotations. 

Is it possible for someone to construct this class with inconsistent binned values vs the `eps_bin_prec`? What if someone puts in the epsilons as floats instead of 2**20*floats 

document that the `eps_bin_prec` behavior during addition. 

`eps_bin_prec` -> `epsilon_precision`? and then document that precision is measured by number of binary digits.

qualtran/resource_counting/_bloq_counts.py
```diff
@@ -128,21 +135,69 @@ class GateCounts:
...
+    def iter_rotations_with_epsilon(self) -> Iterator[tuple[float, SymbolicInt]]:
+        """Iterate through the rotation precisions (epsilon) and their frequency."""
+        for eps_bin, n_rot in self.binned_rotation_epsilons.items():
+            yield eps_bin / 2**self.eps_bin_prec, n_rot
 
     def __add__(self, other):
         if not isinstance(other, GateCounts):
             raise TypeError(f"Can only add other `GateCounts` objects, not {self}")
 
+        eps_bin_prec = max(self.eps_bin_prec, other.eps_bin_prec)
```
mpharrigan, line 188: What if you're adding one carefully-crafted rotation-containing GateCounts with a low eps_precision to a GateCounts that doesn't contain any rotations (it's just e.g. Ts and Toffolis) so it has the default eps_precision of 20

mpharrigan (commented):
some comments and questions, but I think this is generally the correct approach

qualtran/resource_counting/_bloq_counts.py
anurudhp: Multiplication only affects number of rotations, and not the value of epsilon

qualtran/resource_counting/_bloq_counts.py
anurudhp: used `rotations_ignoring_eps`, made it a property.

qualtran/resource_counting/_bloq_counts.py
anurudhp: Used `np.format_scientific_float` to avoid such issues.

qualtran/resource_counting/_bloq_counts.py
anurudhp: Used `np.format_scientific_float` to avoid such issues.

qualtran/resource_counting/_bloq_counts.py
anurudhp: done

qualtran/resource_counting/_bloq_counts.py
anurudhp: updated

anurudhp:
The code snippet I was trying to use (which works with this PR):
```py
import sympy
from qualtran.bloqs.basic_gates import ZPowGate
from qualtran.resource_counting import get_cost_value, QECGatesCost

t, eps = sympy.symbols(r"t \epsilon")
bloq = ZPowGate(exponent=t, eps=eps)
costs = get_cost_value(bloq, QECGatesCost())
costs.total_t_count()
```

qualtran/resource_counting/_bloq_counts.py
```diff
@@ -115,6 +116,16 @@ def __str__(self):
         return f'{self.gateset_name} counts'
 
 
+FloatRepr_T = Union[str, sympy.Expr]
```
mpharrigan, line 119: we usually don't put an underscore before the `T`

qualtran/resource_counting/_bloq_counts.py
```diff
@@ -128,8 +139,38 @@ class GateCounts:
...
+            eps_repr_prec: number of digits to approximate `eps` to. Uses 10 by default.
+                           See `np.format_float_scientific` for more details.
+                           If `eps` is symbolic, this parameter is ignored.
+            n_rotations: number of rotations, defaults to 1.
+        """
+        if is_symbolic(eps):
+            eps_bin: FloatRepr_T = eps
+        else:
+            eps_bin = np.format_float_scientific(eps, precision=eps_repr_prec, unique=False)
+        return cls(binned_rotation_epsilons=Counter({eps_bin: n_rotations}))
```
mpharrigan, line 167: this still desperately needs a description of the data contained within the `GateCounts` dataclass in the class docstring.

qualtran/resource_counting/_bloq_counts.py
```diff
@@ -152,8 +193,10 @@ def __mul__(self, other):
             cswap=other * self.cswap,
             and_bloq=other * self.and_bloq,
             clifford=other * self.clifford,
-            rotation=other * self.rotation,
             measurement=other * self.measurement,
+            binned_rotation_epsilons=Counter(
+                {eps_bin: other * n_rot for eps_bin, n_rot in self.binned_rotation_epsilons.items()}
+            ),
```
mpharrigan, line 199: so how does it work when adding or multiplying when there are values that differ only in precision

mpharrigan (commented):
I don't think this representation is particularly intuitive, so it needs solid documentation

qualtran/resource_counting/_bloq_counts.py
```diff
@@ -128,8 +139,38 @@ class GateCounts:
...
+            eps: precision to synthesize the rotation(s).
+            eps_repr_prec: number of digits to approximate `eps` to. Uses 10 by default.
+                           See `np.format_float_scientific` for more details.
+                           If `eps` is symbolic, this parameter is ignored.
+            n_rotations: number of rotations, defaults to 1.
+        """
+        if is_symbolic(eps):
+            eps_bin: FloatRepr_T = eps
+        else:
+            eps_bin = np.format_float_scientific(eps, precision=eps_repr_prec, unique=False)
```
mpharrigan, line 166: why `unique=False`?

qualtran/resource_counting/_bloq_counts.py
anurudhp: fixed

qualtran/resource_counting/_bloq_counts.py
anurudhp: the `__mul__` is to multiply with some int `other`, right? which just multilpies the frequencies, not affecting the `epsilons`.

For add, it just combines the two counters together, and equal keys will get merged. (handled by `Counter.__add__`). The precision of representing epsilon is only used when converting it to a string key in `from_rotation_with_eps`, but not stored in the GateCounts object itself.

qualtran/resource_counting/_bloq_counts.py
anurudhp: This was to ensure that the representation was more stable. from the docs:
> [... If precision is given fewer digits than necessary can be printed. ...](https://numpy.org/doc/stable/reference/generated/numpy.format_float_scientific.html#:~:text=If%20precision%20is%20given%20fewer%20digits%20than%20necessary%20can%20be%20printed.)

anurudhp: Actually, I'm not sure what the best way is to check if two rotation frequency tables are the same, as the epsilons could have been stored with different precisions, or be approximate.

PR #1256, opened 2024-08-07
Use `AutoPartition` in `BlackBox`

charlesyuan314:
Simplify `BlackBox{Select,Prepare}` decomposition using `AutoPartition`.


anurudhp (approved):
nice!

PR #1260, opened 2024-08-07
Implement `signal_state` for block encodings

charlesyuan314:
For `LinearCombination`, `Product`, and `ChebyshevPolynomial`, the signal state is `PrepareIdentity` and the bloq requires all constituent block encodings to also have `PrepareIdentity`. This is because the circuits implementing these block encodings either control or reflect on the ancilla register in a way that does not work if the signal state is not zero.

For `Unitary`, `LCU`, and `SparseMatrix`, the signal state is `PrepareIdentity` and there are no constituent block encodings.

For `Phase` and `TensorProduct`, the signal state is derived from the constituent block encoding. The case for `TensorProduct` where the constituent does not have `PrepareIdentity` will be implemented in a follow-up PR.

TODO:
- [x] add tests for signal state
- [x] add test for checks in `ChebyshevPolynomial`

### DIFF
--- a/qualtran/bloqs/block_encoding/chebyshev_polynomial.py
+++ b/qualtran/bloqs/block_encoding/chebyshev_polynomial.py
@@ -31,7 +31,7 @@ from qualtran import (
 from qualtran.bloqs.block_encoding import BlockEncoding
 from qualtran.bloqs.block_encoding.linear_combination import LinearCombination
 from qualtran.bloqs.reflections.reflection_using_prepare import ReflectionUsingPrepare
-from qualtran.bloqs.state_preparation.prepare_base import PrepareOracle
+from qualtran.bloqs.state_preparation.black_box_prepare import BlackBoxPrepare
 from qualtran.symbolics import is_symbolic, SymbolicFloat, SymbolicInt
 
 if TYPE_CHECKING:
@@ -108,7 +108,7 @@ class ChebyshevPolynomial(BlockEncoding):
         return self.block_encoding.epsilon * self.order
 
     @property
-    def signal_state(self) -> PrepareOracle:
+    def signal_state(self) -> BlackBoxPrepare:
         # This method will be implemented in the future after PrepareOracle
         # is updated for the BlockEncoding interface.
         # Github issue: https://github.com/quantumlib/Qualtran/issues/1104
@@ -242,7 +242,7 @@ class ScaledChebyshevPolynomial(BlockEncoding):
         return (self.signature.get_right("ancilla"),) if self.ancilla_bitsize > 0 else ()
 
     @property
-    def signal_state(self) -> PrepareOracle:
+    def signal_state(self) -> BlackBoxPrepare:
         # This method will be implemented in the future after PrepareOracle
         # is updated for the BlockEncoding interface.
         # Github issue: https://github.com/quantumlib/Qualtran/issues/1104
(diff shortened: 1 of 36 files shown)


qualtran/bloqs/block_encoding/chebyshev_polynomial.py
```diff
@@ -75,6 +76,10 @@ class ChebyshevPolynomial(BlockEncoding):
     def __attrs_post_init__(self):
         if self.order < 0:
             raise ValueError(f"order must be greater >= 0. Found {self.order}.")
+        if not isinstance(self.block_encoding.signal_state.prepare, PrepareIdentity):
+            raise ValueError(
+                "Cannot take Chebyshev polynomial of block encodings with non-zero signal state."
+            )
```
anurudhp, line 82: Is this also the case for `ScaledChebyshevPolynomial`?

qualtran/bloqs/block_encoding/product_test.py
```diff
@@ -63,11 +68,41 @@ def test_product_signature():
...
+
+    @property
+    def signature(self) -> Signature:
+        return Signature.build(system=1, ancilla=1, resource=1)
+
+    @property
+    def signal_state(self) -> BlackBoxPrepare:
+        return BlackBoxPrepare(TestPrepareOracle())
+
+
```
anurudhp, line 98: move them to `bloqs/for_testing/`?

qualtran/bloqs/block_encoding/unitary.py
```diff
@@ -74,10 +75,7 @@ def pretty_name(self) -> str:
 
     @property
     def signal_state(self) -> BlackBoxPrepare:
-        # This method will be implemented in the future after PrepareOracle
-        # is updated for the BlockEncoding interface.
-        # Github issue: https://github.com/quantumlib/Qualtran/issues/1104
-        raise NotImplementedError
+        return BlackBoxPrepare(PrepareIdentity((QAny(self.ancilla_bitsize),)))
```
anurudhp, line 78: Does this cause any issues when `ancilla_bitsize=0`?

anurudhp (approved):
lgtm! % nits

qualtran/bloqs/block_encoding/unitary.py
charlesyuan314: It would fail if we ever took that `signal_state` and did a decomposition. It's not something we do currently. 

I think a simple fix would be to add a check in `PrepareIdentity.build_composite_bloq` to not do anything if the bitsize is zero.

qualtran/bloqs/block_encoding/chebyshev_polynomial.py
charlesyuan314: It is captured when `ScaledChebyshevPolynomial` instantiates `ChebyshevPolynomial` with the same arguments.

I added the redundant check here though for clarity.

qualtran/bloqs/block_encoding/product_test.py
charlesyuan314: I could try in a follow-up PR. The thing is that these really are stub classes whose only purpose is to test edge cases of constructors. They're not really useful for independent testing of any other behavior.

qualtran/bloqs/block_encoding/unitary.py
charlesyuan314: If that (or some other change to `PrepareIdentity`) is desirable we can do in a follow-up PR.

Issue #914, opened 2024-05-01
MultiControlPauli improvements


mpharrigan:
- This works for any single-qubit operation. It does not require a pauli gate
 - `target_gate` is annotated as a `cirq.Pauli`. The interplay between cirq gates and bloqs is unnecessary. This should take any one-qubit bloq.
 - `on_classical_vals` should use the classical simulability of the target gate instead of special-casing certain values

PR #1263, opened 2024-08-07
Rename `multi_control_multi_target_pauli` to `multi_control_pauli`

anurudhp:
Use `MultiControlX` wherever possible instead of `MultiControlPauli(..., cirq.X)`.

Towards refactoring `MultiControlPauli` to accept any bloq (#914, #1131). Follow up will accept any bloq and remove dependency on cirq.


charlesyuan314 (approved):

charlesyuan314:
For my own knowledge: why was this previously called `multi_target`? Did it support multiple targets in the past?

mpharrigan:
the module was called multi_control_multi_target because it had bloqs for doing multi control xor multi target things. It lives on as `qualtran.mcmt` package

mpharrigan (approved):

Issue #1272, opened 2024-08-09
definition of `_add_ctrled` is too restrictive


NoureldinYosri:
The second return return value of `get_ctrl_system` is `_add_ctrled` which is defined as
  
https://github.com/quantumlib/Qualtran/blob/a4365b26a01956bb0341f12e673176e3a66ab33b/qualtran/_infra/bloq.py#L384C33-L389

This definition is too restrictive since it assumes that no new registers may be created which is not true. for example And().controlled() should be MultiAnd which creates an extra `junk` register.

mpharrigan:
I think this is out of scope for the notion of a "controlled" operation.

Controlled(A) = $\ket{0}\bra{0}\otimes I + \ket{1}\bra{1} \otimes A$

mpharrigan:
@NoureldinYosri can you provide more context on your use case

mpharrigan:
I wrote some more here: https://github.com/quantumlib/Qualtran/issues/1304#issuecomment-2307862322

The contract (irrespective of whether the registers are RIGHT or THRU) is that the controlled version can only add new registers that comport with `CtrlSpec.qdtypes`. The system has no idea what to do with the extra junk registers. If you wanted to encode the idea that "this junk needs to get cleaned up by And^dag" you'd use something like #1221

Issue #1271, opened 2024-08-09
resource_counting.query_costs counts controlled T and controlled And as rotations


NoureldinYosri:
```py
In [6]: b = Add(QUInt(4)).controlled()

In [7]: query_costs(b, [QECGatesCost()])
Out[7]: 
{Controlled(subbloq=ArbitraryClifford(n=2), ctrl_spec=CtrlSpec(qdtypes=(QBit(),), cvs=(array(1),))): {QECGatesCost(): GateCounts(t=0, toffoli=0, cswap=0, and_bloq=0, clifford=0, rotation=0, measurement=0)},
 Controlled(subbloq=TGate(is_adjoint=False), ctrl_spec=CtrlSpec(qdtypes=(QBit(),), cvs=(array(1),))): {QECGatesCost(): GateCounts(t=0, toffoli=0, cswap=0, and_bloq=0, clifford=0, rotation=1, measurement=0)},
 Controlled(subbloq=And(cv1=1, cv2=1, uncompute=False), ctrl_spec=CtrlSpec(qdtypes=(QBit(),), cvs=(array(1),))): {QECGatesCost(): GateCounts(t=0, toffoli=0, cswap=0, and_bloq=0, clifford=0, rotation=4, measurement=0)},
 Controlled(subbloq=And(cv1=1, cv2=1, uncompute=True), ctrl_spec=CtrlSpec(qdtypes=(QBit(),), cvs=(array(1),))): {QECGatesCost(): GateCounts(t=0, toffoli=0, cswap=0, and_bloq=0, clifford=0, rotation=0, measurement=0)},
 Toffoli(): {QECGatesCost(): GateCounts(t=0, toffoli=1, cswap=0, and_bloq=0, clifford=0, rotation=0, measurement=0)},
 Controlled(subbloq=Add(a_dtype=QUInt(bitsize=4), b_dtype=QUInt(bitsize=4)), ctrl_spec=CtrlSpec(qdtypes=(QBit(),), cvs=(array(1),))): {QECGatesCost(): GateCounts(t=0, toffoli=15, cswap=0, and_bloq=0, clifford=0, rotation=12, measurement=0)}}
```

---

update: desired output should be no rotations for this gate ... instead a complexity based on and_bloq and/or toffoli

mpharrigan:
can you describe the observed and desired behavior

NoureldinYosri:
updated

mpharrigan:
[...]

Issue #1221, opened 2024-07-30
Discussion: Add a `Conjugate` meta-bloq


anurudhp:
This captures the pattern `Conjugate(A, B)` $= A^\dagger \cdot B \cdot A$. This has an efficient controlled version: `C-Conjugate(A, B) = Conjugate(A, C-B)` $= A^\dagger \cdot C\text{-}B \cdot A$.

Some bloqs which have custom controlled version to support the above optimization:
- [`QubitizationWalkOperator`](https://github.com/quantumlib/Qualtran/blob/2a78ca1d3641dcfce4c4813456dfdee1c6c00749/qualtran/bloqs/qubitization/qubitization_walk_operator.py#L53)
- [`ReflectionUsingPrepare`](https://github.com/quantumlib/Qualtran/blob/2a78ca1d3641dcfce4c4813456dfdee1c6c00749/qualtran/bloqs/reflections/reflection_using_prepare.py#L38)
- [`MeanEstimationOperator`](https://github.com/quantumlib/Qualtran/blob/2a78ca1d3641dcfce4c4813456dfdee1c6c00749/qualtran/bloqs/mean_estimation/mean_estimation_operator.py#L68)
etc.

This would require carefully wiring up the registers, though. Especially when the registers of A and B don't match.
See [`LCUBlockEncodingZeroState`](https://github.com/quantumlib/Qualtran/blob/2a78ca1d3641dcfce4c4813456dfdee1c6c00749/qualtran/bloqs/block_encoding/lcu_block_encoding.py#L276) for a similar decomposition.

PR #1273, opened 2024-08-09
And.controlled and MultiAnd.controlled should return MultiAnd

NoureldinYosri:
related to https://github.com/quantumlib/Qualtran/issues/1271 and https://github.com/quantumlib/Qualtran/issues/1272

### DIFF
--- a/qualtran/bloqs/mcmt/and_bloq.py
+++ b/qualtran/bloqs/mcmt/and_bloq.py
@@ -23,7 +23,19 @@ to the and of its control registers. `And` will output the result into a fresh r
 """
 import itertools
 from functools import cached_property
-from typing import cast, Dict, Iterable, Iterator, List, Optional, Set, Tuple, TYPE_CHECKING, Union
+from typing import (
+    cast,
+    Dict,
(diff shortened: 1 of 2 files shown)


mpharrigan:
Why did you close this?

NoureldinYosri:
I thought this use case doesn't follow the controlled protocol ... the `_add_ctrled ` method for And and MultiAnd doesn't follow the expected signature from your comment https://github.com/quantumlib/Qualtran/issues/1272#issuecomment-2297466324

mpharrigan:
Can you offer your thoughts on #1304 or provide context for where you want to do And().controlled()

mpharrigan:
I wrote some more here: https://github.com/quantumlib/Qualtran/issues/1304#issuecomment-2307862322

The contract (irrespective of whether the registers are RIGHT or THRU) is that the controlled version can only add new registers that comport with `CtrlSpec.qdtypes`. The system has no idea what to do with the extra junk registers. If you wanted to encode the idea that "this junk needs to get cleaned up by And^dag" you'd use something like #1221

PR #1277, opened 2024-08-12
Improve support for CAdd

NoureldinYosri:
This is a followup PR to https://github.com/quantumlib/Qualtran/pull/864 that

- adds a notebook for CAdd
- fixes classical action and adds classical action tests
- replaces the decomposition. The old decomposition is buggy and failed classical tests. Instead I use a simple decomposition that uses $n$ ancillas.
- adds tests for decomposition and symbolic costs

PS: we should probably replace this decomposition with the old one when we fix the bug in the old one.

### DIFF
--- a/qualtran/bloqs/arithmetic/controlled_addition.py
+++ b/qualtran/bloqs/arithmetic/controlled_addition.py
@@ -108,9 +108,7 @@ class CAdd(Bloq):
             return {'ctrl': ctrl, 'a': a, 'b': b}
         else:
             if not isinstance(self.b_dtype.bitsize, int):
-                raise NotImplementedError(
-                    f'classical simulation is not supported for symbolic bloq {self}'
-                )
+                raise ValueError(f'classical simulation is not supported for symbolic bloq {self}')
             return {
                 'ctrl': ctrl,
                 'a': a,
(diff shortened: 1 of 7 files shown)


qualtran/bloqs/arithmetic/controlled_addition.py
```diff
@@ -91,50 +101,23 @@ def signature(self):
...
-        unsigned = isinstance(self.a_dtype, (QUInt, QMontgomeryUInt))
-        b_bitsize = self.b_dtype.bitsize
-        N = 2**b_bitsize if unsigned else 2 ** (b_bitsize - 1)
         ctrl = kwargs['ctrl']
         if ctrl != self.cv:
             return {'ctrl': ctrl, 'a': a, 'b': b}
         else:
-            return {'ctrl': ctrl, 'a': a, 'b': int(math.fmod(a + b, N))}
+            if not isinstance(self.b_dtype.bitsize, int):
+                raise NotImplementedError(
```
mpharrigan, line 111: we should only use `NotImplementedError` for cases that are theoretically possible but we just haven't done it _yet_. Is this the case here? If it's an error in any future, I'd suggest `ValueError`.

mpharrigan (approved):

mpharrigan:
Let's track somewhere the task that `Add().controlled()` returns this bloq

qualtran/bloqs/arithmetic/controlled_addition.py
NoureldinYosri: changed to ValueError ... from a code point of view a sympy symbol bitsize also works for this specific operation but it fails other classical operations ... this is why I wasn't sure

NoureldinYosri:
I created https://github.com/quantumlib/Qualtran/issues/1278 to track making Add.controlled() return CAdd

PR #1279, opened 2024-08-13
Update musical score diagram for adjoint and add more replacements for qpic

tanujkhattar:
The qpic drawing code uses musical score infrastructure to draw diagrams. Right now, the way to distinguish between a `bloq` and `Adjoint(bloq)` in a musical score diagram is to add a dagger in front of the title -- the wire symbols remain as is. The qpic infrastructure, however, does not make use of the title of a bloq and only uses wire symbols. As a result, the adjoint gets drawn as the same as the original bloq. This PR updates the musical score diagrams to add a dagger in front of every text symbol on every wire. 

I personally think this looks fine, but @mpharrigan let me know if you have concerns. The QPIC code gets updated nicely with this change.


mpharrigan (approved):
yes, I think this is fine/good.

PR #1267, opened 2024-08-08
Add qref and bartiq dependencies

mstechly:
As requested in #1194 by @mpharrigan , I added bartiq and qref dependency.
Let me know if I should add it anywhere else as well.

I used `==`, as given `qref` and `bartiq` are not stable, their update could potentially break some tests in Qualtran, which would be unnecessarily disrupting.

PR #1284, opened 2024-08-13
Update dependencies for newer qref & bartiq versions

mstechly:
Given compatibility issues that @mpharrigan mentioned [here](https://github.com/quantumlib/Qualtran/pull/1267#issuecomment-2278298079), I made a new Bartiq release that solves it and updated versions here.


mpharrigan (approved):
I re-ran the `pip-compile` routines in docker to avoid picking up the mac-only packages, but lgtm; thanks

mstechly:
@mpharrigan Good catch, thanks!

Issue #1135, opened 2024-07-12
Implement Programmable ancilla rotation


anurudhp:
"Simulating chemistry efficiently on fault-tolerant quantum computers" http://arxiv.org/abs/1204.0567, Fig 4.

![image](https://github.com/user-attachments/assets/5f32b874-ed7f-4e33-8ae5-1eecb4e135a5)

This would require a Measure bloq, classical registers and support for classically controlled bloq.

PR #1287, opened 2024-08-14
Add shim for measurement gate

anurudhp:
Introduces a simple shim bloq `Measure` for single qubit measurements. Useful for resource counting, e.g. #1135


mpharrigan (approved):
thanks, very useful

PR #1288, opened 2024-08-14
`SU2RotationGate`: add helper to construct from Euler ZXZ angles

anurudhp:
- add a helper `from_euler_zxz_angles`, which is another commonly used representation for SU2.
- simplify decomposition to use Rz and Rx gates.
- Use eps/3 per rotation, without any eps for global phase.

### DIFF
--- a/qualtran/bloqs/basic_gates/su2_rotation.py
+++ b/qualtran/bloqs/basic_gates/su2_rotation.py
@@ -20,7 +20,7 @@ from attrs import frozen
 from numpy.typing import NDArray
 
 from qualtran import bloq_example, BloqDocSpec, ConnectionT, GateWithRegisters, Register, Signature
-from qualtran.bloqs.basic_gates import GlobalPhase, Hadamard, Rz, ZPowGate
+from qualtran.bloqs.basic_gates import GlobalPhase, Hadamard, Rz
 from qualtran.cirq_interop.t_complexity_protocol import TComplexity
 from qualtran.drawing import Text, TextBox
 from qualtran.symbolics import is_symbolic, pi, SymbolicFloat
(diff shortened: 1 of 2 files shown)


qualtran/bloqs/basic_gates/su2_rotation.py
```diff
@@ -148,7 +193,7 @@ def adjoint(self) -> 'SU2RotationGate':
         )
 
     def _t_complexity_(self) -> TComplexity:
-        return TComplexity(rotations=3)
+        return TComplexity(rotations=3, clifford=2)
```
mpharrigan, line 196: xref https://github.com/quantumlib/Qualtran/issues/1309#issuecomment-2299825657

do you want this to always be 3 rotations or do you want the angles to affect the complexity? Please implement build_call_graph, remove _t_complexity_ override, and add a test with `qlt_testing.assert_equivalent_bloq_counts`. If you want it always to be three rotations, return that in the call graph and use `generalize_rotation_angle` generalizer

mpharrigan:
can you update the PR description to briefly motivate this change

qualtran/bloqs/basic_gates/su2_rotation.py
anurudhp: It was annotated with 3 rotations for simplicity, but I'll make it return the actual rotations so that classify_bloqs handles the special angles correctly.

anurudhp: I removed the t-complexity override and simplified the decomposition to be Rz Rx Rz.

Should I also implement a call graph? because the decomposition here is always possible and small - 1-qubit 4-gates.

mpharrigan: it should be fine since the decomposition is small and supports symbolics

qualtran/bloqs/basic_gates/su2_rotation.py
```diff
@@ -125,17 +165,20 @@ def _unitary_(self):
         return self.rotation_matrix
 
     def build_composite_bloq(self, bb: 'BloqBuilder', q: 'SoquetT') -> Dict[str, 'SoquetT']:
+        # TODO implement controlled version, and pass eps/4 to each rotation (incl. global phase)
```
mpharrigan, line 168: open and link a github issue

mpharrigan (approved):

PR #1291, opened 2024-08-15
Add `QROAMClean` and `QROAMCleanAdjoint` bloqs for data loading

tanujkhattar:
Addresses https://github.com/quantumlib/Qualtran/issues/368

The only variation left after this is a `QROAMDirty` bloq, which allocates `k-1` dirty registers and assumes target is in 0 state initially. Since the constant factors of this construction are _very_ similar to `SelectSwapQROM` (leading order terms are identical), it's not a priority to add this version right now.

More work needs to be done to add (a) Docstrings, (b) Replace uses of `QROAM` and `QROAMTwoRegs` in chemistry using the newly added Bloqs and eventually add a decomposition for `QROAMCleanAdjoint` that uses measurement based un-computation; but everything else is setup correctly. 

Here are some pictures:

image

image

### DIFF
--- a/qualtran/bloqs/data_loading/qroam_clean.py
+++ b/qualtran/bloqs/data_loading/qroam_clean.py
@@ -24,6 +24,7 @@ from numpy.typing import ArrayLike
 from qualtran import bloq_example, BloqDocSpec, GateWithRegisters, Register, Side, Signature
 from qualtran.bloqs.basic_gates import Toffoli
 from qualtran.bloqs.data_loading.qrom_base import QROMBase
+from qualtran.drawing import Circle, LarrowTextBox, RarrowTextBox, Text, TextBox, WireSymbol
 from qualtran.symbolics import ceil, is_symbolic, log2, prod, SymbolicFloat, SymbolicInt
 
 if TYPE_CHECKING:
@@ -76,6 +77,37 @@ def get_optimal_log_block_size_clean_ancilla(
 
 @attrs.frozen
 class QROAMCleanAdjoint(QROMBase, GateWithRegisters):  # type: ignore[misc]
+    r"""Measurement based uncomputation of a table lookup. Assumes target is left in |0> state.
+
+    Measurement based uncomputation of a table, as described in Ref[1]. Follows the usual API
+    of a QROM bloq. Assumes that the target register was initially in the $|0\rangle$ state before
+    the lookup was performed and must be deallocated and left in |0\rangle$ state after
+    uncomputation.
+
+    Reduces the problem of uncomputating a data lookup with $N$ elements and target bitsize $b$ to
+    doing another data lookup with $\frac{N}{2}$ elements, each of bitsize 2. The reduction is
+    explained in Ref[1] and uses X-basis measurements + classically controlled Cliffords. A variant
+    of `QROAM` bloq can be used to perform the resulting lookup and resulting junk registers can be
+    cleaned up with measurement based uncomputation of SwapWithZero.
+
+    Thus, cost of uncomputing a data lookup with $N$ elements and target bitsize $b$ is
+    $\frac{N}{K} + (K - 1)$ Toffoli gates (instead of $\frac{N}{K} + (K - 1) \times b$ used by the
+    original lookup).
+
+    Registers:
+        - control_registers: If control is specified, a THRU register to denote the control qubits.
+            Empty by default for uncontrolled version of the Bloq.
+        - selection_registers: $N$ THRU registers, each with shape (), to load $N$ dimensional
+            classical datasets.
+        - target_registers: $M$ LEFT registers to load $M$ different classical datasets. Each target
+            register is of bitsize $b$ and shape described by a tuple of length $N + S$. Here $S$ is
+            a parameter that describes the shape of the output to be loaded for each selection index.
+
+
+    References:
+        [Qubitization of Arbitrary Basis Quantum Chemistry Leveraging Sparsity and Low Rank Factorization](https://arxiv.org/abs/1902.02134).
+            Berry et. al. (2019). Appendix C.
+    """
     log_block_sizes: Tuple[SymbolicInt, ...] = attrs.field(
         converter=lambda x: tuple(x.tolist() if isinstance(x, np.ndarray) else x)
     )
@@ -143,10 +175,10 @@ class QROAMCleanAdjoint(QROMBase, GateWithRegisters):  # type: ignore[misc]
         return attrs.evolve(self, log_block_sizes=log_block_sizes)
 
     def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
-        X = prod([2**k for k in self.log_block_sizes])
-        N = prod(self.data_shape)
-        B = 2
-        cost = qroam_cost(X, ceil(N / 2), B, adjoint=True)
+        block_sizes = prod([2**k for k in self.log_block_sizes])
+        data_size = prod(self.data_shape)
+        target_bitsize = 2
+        cost = qroam_cost(block_sizes, ceil(data_size / 2), target_bitsize, adjoint=True)
         return {(Toffoli(), cost)}
 
     @cached_property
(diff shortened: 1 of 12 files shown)


qualtran/bloqs/data_loading/qroam_clean.py
```diff
@@ -0,0 +1,389 @@
...
+        return bb.allocate(dtype=dtype, dirty=dirty)
+    soqs = np.empty(shape, dtype=object)
+    ndindex_iter = np.ndindex(shape)
+    _ = next(ndindex_iter)  # Skip the first element.
+    for idx in ndindex_iter:
+        soqs[idx] = bb.allocate(dtype=dtype, dirty=dirty)
+    return soqs
+
+
+def qroam_cost(x, data_size: SymbolicInt, bitsize: SymbolicInt, adjoint: bool = False):
```
fdmalone, line 50: nit: docstring.

qualtran/bloqs/data_loading/qroam_clean.py
```diff
@@ -0,0 +1,389 @@
...
+    else:
+        return data_size / x + bitsize * (x - 1)
+
+
+def get_optimal_log_block_size_clean_ancilla(
+    data_size: SymbolicInt,
+    bitsize: SymbolicInt,
+    adjoint: bool = False,
+    qroam_block_size: Optional[SymbolicInt] = None,
+) -> SymbolicInt:
```
fdmalone, line 63: docstring.

qualtran/bloqs/data_loading/qroam_clean.py
tanujkhattar: this is mostly copied from the chemistry file.

qualtran/bloqs/data_loading/qroam_clean.py
```diff
@@ -0,0 +1,389 @@
...
+        if any(b > 1 for b in block_sizes):
+            ctrl, selection, qrom_targets = self._build_composite_bloq_with_swz_clean(
+                bb, ctrl, selection, qrom_targets
+            )
+        else:
+            ctrl, selection, qrom_targets = self._add_qrom_bloq(bb, ctrl, selection, qrom_targets)
+        # Construct and return dictionary of final soquets.
+        soqs |= {reg.name: soq for reg, soq in zip(self.control_registers, ctrl)}
+        soqs |= {reg.name: soq for reg, soq in zip(self.selection_registers, selection)}
+        soqs |= {reg.name: soq.flat[1:] for reg, soq in zip(self.junk_registers, qrom_targets)}  # type: ignore[union-attr]
```
fdmalone, line 318: Could you describe this convention somewhere? Maybe above where qrom_targets are built.

tanujkhattar: Yes, the idea is that all the `k-1` newly allocated registers are right registers and are kept around as junk. The adjoint bloq expects has the same API and will be responsible to deallocate the target + junk registers using measurement based uncomputation. I've also added some figures as part of the PR description to make it easier to see what's going on. 

The classical simulation is also implemented in a way that for each selection index `s`, the correct permutation of `target + junk` registers is computed and returned; and the adjoint bloq is instantiated so that uncomputing the original qroam read reduces to a  table lookup of size `N / 2` and target bitsize `b * k` -- the important thing to note is that cost of uncomputation does not dependent upon the target bitsize (`b * k`) and this is reflected in the costs; but the classical simulation tests assert that adjoint bloq has the correct table as input.

qualtran/bloqs/data_loading/qroam_clean_test.py
```diff
@@ -0,0 +1,141 @@
...
+    K = 2**k
+    expected_toffoli = ceil(N / K) + (K - 1) * b - 2
+    assert bloq.t_complexity().t == 4 * expected_toffoli
+    bloq_inv = bloq.adjoint()
+    assert isinstance(bloq_inv, QROAMCleanAdjoint)
+    inv_k = sympy.symbols('kinv')
+    inv_K = 2**inv_k
+    bloq_inv = bloq_inv.with_log_block_sizes(log_block_sizes=(inv_k,))
+    expected_toffoli_inv = ceil(N / 2) / inv_K + inv_K
+    assert bloq_inv.t_complexity().t == 4 * expected_toffoli_inv
```
fdmalone, line 45: Maybe out of scope but we should probably make a utility function which uses the new cost framework to get Toffoli counts directly but hides some of the boilerplate.

qualtran/bloqs/data_loading/qroam_clean_test.py
```diff
@@ -0,0 +1,141 @@
...
+)
+from qualtran.symbolics import ceil
+
+
+def test_bloq_examples(bloq_autotester):
+    bloq_autotester(_qroam_clean_multi_data)
+    bloq_autotester(_qroam_clean_multi_dim)
+
+
+def test_t_complexity_1d_data_symbolic():
```
fdmalone, line 32: Is it worth adding a test which uses QROAM then QROAM^ in its decomposition to make sure the allocated RIGHT registers match the LEFTs which will be deallocated in the inverse?

qualtran/bloqs/data_loading/select_swap_qrom.py
```diff
@@ -222,7 +225,7 @@ def batched_data(self) -> List[np.ndarray]:
         #
         # For data[N1][N2] with block sizes (k1, k2), you load batches of size `(k1, k2)` at once.
         # Thus, you load batch[N1/k1][N2/k2] where batch[i][j] = data[i*k1:(i + 1)*k1][j*k2:(j + 1)*k2]
-        batched_data = [np.empty(self.batched_data_shape) for _ in range(len(self.target_bitsizes))]
+        batched_data = [np.zeros(self.batched_data_shape, dtype=int) for _ in self.target_bitsizes]
```
fdmalone, line 228: Why this change?

qualtran/bloqs/data_loading/qroam_clean_test.py
tanujkhattar: The classical simulation tests assert this; and also the fact that adjoint has the right lookup table to load.

qualtran/bloqs/data_loading/select_swap_qrom.py
```diff
@@ -451,15 +454,15 @@ def wire_symbol(self, reg: Optional[Register], idx: Tuple[int, ...] = tuple()) -
...
+    qroam_multi_data = SelectSwapQROM.build_from_data(data1, data2)
     return qroam_multi_data
 
 
 @bloq_example
 def _qroam_multi_dim() -> SelectSwapQROM:
     data1 = np.arange(25).reshape((5, 5))
     data2 = (np.arange(25) + 1).reshape((5, 5))
-    qroam_multi_dim = SelectSwapQROM.build_from_data([data1, data2])
+    qroam_multi_dim = SelectSwapQROM.build_from_data(data1, data2)
```
fdmalone, line 465: Was this secretly failing?

qualtran/bloqs/data_loading/select_swap_qrom.py
tanujkhattar: `dtype=int` so we don't get floats by default. zeros instead of empty because the batched data size can be larger than actual data (to the closest power of 2)

qualtran/bloqs/data_loading/select_swap_qrom.py
tanujkhattar: This was secretly passing multi-dimensional data instead of multiple datasets. My intention at the time of writing the test was to load 2 classical datasets; each of 2 dimensions. This was loading 1 classical dataset, of shape `(2, 5, 5)`.

qualtran/bloqs/data_loading/qroam_clean.py
```diff
@@ -0,0 +1,389 @@
...
+        if log_block_sizes is None:
+            return self
+        if isinstance(log_block_sizes, (int, sympy.Basic, numbers.Number)):
+            log_block_sizes = (log_block_sizes,)
+        if not is_symbolic(*log_block_sizes):
+            assert all(1 <= 2**bs <= ilen for bs, ilen in zip(log_block_sizes, self.data_shape))
+        return attrs.evolve(self, log_block_sizes=log_block_sizes)
+
+    def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
+        X = prod([2**k for k in self.log_block_sizes])
```
fdmalone, line 146: not sure what our policy is on pep8 variable names I think `data_size` etc is a bit more readable though

tanujkhattar: will update.

qualtran/bloqs/data_loading/qroam_clean.py
```diff
@@ -0,0 +1,389 @@
...
+        return ret
+
+    @cached_property
+    def junk_registers(self) -> Tuple[Register, ...]:
+        # The newly allocated registers should be kept around for measurement based uncomputation.
+        junk_regs = []
+        block_size = prod(self.block_sizes)
+        for reg in self.target_registers:
+            assert reg.shape == ()
+            if is_symbolic(block_size) or block_size > 1:
```
fdmalone, line 262: Can you still get a saving for QROM inverse when block_size == 1 (regular old QROM)? I have a feeling you can but maybe that should be handled elsewhere?

qualtran/bloqs/data_loading/qroam_clean.py
```diff
@@ -0,0 +1,389 @@
...
+        selection = self._unpartition_sel_register(bb, sel_l, sel_k)
+        return ctrl, selection, qrom_targets
+
+    def build_composite_bloq(self, bb: 'BloqBuilder', **soqs: 'SoquetT') -> Dict[str, 'SoquetT']:
+        # Get the ctrl and target register for the SelectSwapQROM.
+        ctrl = [soqs.pop(reg.name) for reg in self.control_registers]
+        selection = [soqs.pop(reg.name) for reg in self.selection_registers]
+        if is_symbolic(*self.block_sizes):
+            raise ValueError(
+                f"Cannot decompose QROAM bloq with symbolic block sizes. Found {self.block_sizes=}"
```
fdmalone, line 293: Is that `=` meant to be there?

fdmalone (changes requested):
Awesome. Looking good just some questions / nits.

qualtran/bloqs/data_loading/qroam_clean.py
tanujkhattar: `f'{a=}'` in an f-string prints `'a=(value of a)'` whereas `f'{a}'` only prints `'(value of a)'`. The former is nicer IMO, so yes it's meant to be there.

qualtran/bloqs/data_loading/qroam_clean.py
fdmalone: I realized about half way in lol, sorry. You can assign it to me :(

qualtran/bloqs/data_loading/qroam_clean_test.py
fdmalone: Got it. Nice.

qualtran/bloqs/data_loading/select_swap_qrom.py
fdmalone: more on the empty -> zeros switch. Makes sense.

qualtran/bloqs/data_loading/qroam_clean.py
fdmalone: Neat. (I checked it locally but mistyped f'{a}=')

qualtran/bloqs/data_loading/qroam_clean.py
tanujkhattar: added detailed docstrings to both the classes.

qualtran/bloqs/data_loading/qroam_clean.py
tanujkhattar: done

qualtran/bloqs/data_loading/qroam_clean.py
tanujkhattar: Yes, it's already handled. You just pass log_block_size=0 to the `QROAMClean` bloq. 

image

fdmalone: Awesome.

fdmalone (approved):

PR #1302, opened 2024-08-19
Fix QROAMClean sum logic when determining total bitsize

fdmalone:


tanujkhattar (approved):

qualtran/bloqs/data_loading/qroam_clean.py
```diff
@@ -256,8 +256,8 @@ def _target_reg_side(self) -> Side:
 
     @log_block_sizes.default
     def _default_log_block_sizes(self) -> Tuple[SymbolicInt, ...]:
-        target_bitsize = sum(self.target_bitsizes) * sum(
-            prod(shape) for shape in self.target_shapes
+        target_bitsize = sum(
+            bs * prod(shape) for (bs, shape) in zip(self.target_bitsizes, self.target_shapes)
```
tanujkhattar, line 260: Can we add a test?

fdmalone: Added a test which would have failed previously

PR #1303, opened 2024-08-19
Profiling large composite bloq construction and qubit counting timings

mpharrigan:
This PR includes

 - A parametrically-large bloq for testing decomposition and qubit count _loosely_ modeled on a qrom-ish circuit
 - A series of potential optimizations, whose timing I've recorded below (check the second commit for the details)
 - Reversion of all of the optimizations _except_ those that are a meaningful win with limited code health burden

----

I timed qubit counting as well as just doing a decomposition. 

```python
len(LargeBloq(n_select=60, n_ops=120).decompose_bloq().bloq_instances)
14282
```

```python
%%timeit
val = get_cost_value(LargeBloq(n_select=60, n_ops=120), QubitCount())
```

```
baseline
1.91 s ± 8.29 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

qubit count operation re-ordering
    1.63 s ± 82.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

with cached shape property
    1.06 s ± 23.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

the first two optimizations, and changes to register sorting
    1.02 s ± 6.77 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

the first two optimizations, and removing some of the dispatch from register shape
    959 ms ± 12.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

the first two optimizations, and non-attrs classes for BloqInstance, Soquet, ... #130 
    733 ms ± 7.09 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

All optimizations
    609 ms ± 4.29 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```

```python
%%timeit
cb = LargeBloq(n_select=60, n_ops=120).decompose_bloq()
```

```
baseline
373 ms ± 10.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

qubit count changes, cached shape property: n/c

register sort
    346 ms ± 5.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

no fancy shape
    291 ms ± 16.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

non-attrs quantum graph
    276 ms ± 1.52 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

all optimizations
    171 ms ± 3.28 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

```

image

## Conclusions

 - marginal improvements for the things I tried when speeding up a decomposition, probably not worth it
 - 15k bloqs in ~1s is probably fine. Decompositions shouldn't have that many sub-operations
 - Qubit counting was doing a decomposition twice, which is not good for performance

### DIFF
--- a/qualtran/_infra/composite_bloq.py
+++ b/qualtran/_infra/composite_bloq.py
@@ -653,11 +653,7 @@ class _IgnoreAvailable:
         pass
 
 
-def _reg_to_soq(
-    binst: Union[BloqInstance, DanglingT],
-    reg: Register,
-    available: Union[Set[Soquet], _IgnoreAvailable] = _IgnoreAvailable(),
-) -> SoquetT:
+def _reg_to_soq(binst: Union[BloqInstance, DanglingT], reg: Register) -> SoquetT:
     """Create the soquet or array of soquets for a register.
 
     Args:
@@ -677,14 +673,12 @@ def _reg_to_soq(
         for ri in reg.all_idxs():
             soq = Soquet(binst, reg, idx=ri)
             soqs[ri] = soq
-            available.add(soq)
         return soqs
 
     # Annoyingly, this must be a special case.
     # Otherwise, x[i] = thing will nest *array* objects because our ndarray's type is
     # 'object'. This wouldn't happen--for example--with an integer array.
     soq = Soquet(binst, reg)
-    available.add(soq)
     return soq
 
 
@@ -692,7 +686,8 @@ def _process_soquets(
     registers: Iterable[Register],
     in_soqs: Mapping[str, SoquetInT],
     debug_str: str,
-    func: Callable[[Soquet, Register, Tuple[int, ...]], None],
+    cxn_list: List[Connection],
+    binst: BloqInstance,
 ) -> None:
     """Process and validate `in_soqs` in the context of `registers`.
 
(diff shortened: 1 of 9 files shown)


tanujkhattar:
> 15k bloqs in ~1s is probably fine. Decompositions shouldn't have that many sub-operations

From a theoretical perspective, the complexity for qubit counting should probably be linear; i.e. - `O(N + M)` where `N` is number of nodes in the composite graph and `M` is number of edges. A modern machine can execute ~1e8 operations per second. We should aim to support qubit counting for ~1e5-1e6 bloqs in the decomposition per ~1s; discounting for ~100x slowdown due to Python and constant factors wrt error checking / data structures etc. 

Right now this number is ~1.5e4 -- I think we should try to speed things up somewhere between ~10x - 100x as a reasonable goal. It'll be good to profile what is bringing the biggest slowdown in terms of constant factors.

tanujkhattar (approved):
LGTM. As noted in my previous comment, we should aim to construct composite bloqs and do qubit counting for composite bloqs with ~1e5-1e6 subbloqs; instead of 1.5e4; in ~1s.

Issue #1304, opened 2024-08-19
Controlled versions of non-THRU bloqs


mpharrigan:
With the discussion in #1272 #1273 , I have been reminded of how uncomfortable I feel about controlled versions of bloqs with RIGHT or LEFT (allocating / de-allocating) registers.

There are parts of the code that don't dis-allow this; Maybe we should dis-allow this. 

Controlling a bloq can be thought of a special case of the `Select` operation: our control bit toggles between two possible operations, the identity (do nothing) and the operation in question. The two operations **must** have the same tensor shape. 

From inspection of the tensor simulation code for `Controlled`, it 'supports' right- and left- registers. But it must make a choice for what the identity operation is. Assume the operation we're considering is a one-bit allocation. It says the "identity operation" is $\ket{0}$ -- which is a choice.

It's a strange choice though. Assume our control bit is in equal superposition. 

 - `Controlled(OneState)` -> $\ket{0} + \ket{1}$
 - `Controlled(ZeroState)` -> $\ket{0}$

This "preferred direction" seems odd

mpharrigan:
<img width="699" alt="image" src="https://github.com/user-attachments/assets/7c7eed9e-0055-40dc-89a5-4b7fefb5172f">

mpharrigan:
```python
Controlled(ZeroState(), CtrlSpec()).call_classically(ctrl=0)
```

```
KeyError                                  Traceback (most recent call last)
Cell In[45], line 1
----> 1 Controlled(ZeroState(), CtrlSpec()).call_classically(ctrl=0)

File ~/qutran/qutran/qualtran/_infra/bloq.py:233, in Bloq.call_classically(self, **vals)
    212 def call_classically(
    213     self, **vals: Union['sympy.Symbol', 'ClassicalValT']
    214 ) -> Tuple['ClassicalValT', ...]:
    215     """Call this bloq on classical data.
    216 
    217     Bloq users can call this function to apply bloqs to classical data. If you're
   (...)
    231         registers.
    232     """
--> 233     res = self.as_composite_bloq().on_classical_vals(**vals)
    234     return tuple(res[reg.name] for reg in self.signature.rights())

File ~/qutran/qutran/qualtran/_infra/composite_bloq.py:220, in CompositeBloq.on_classical_vals(self, **vals)
    217 """Support classical data by recursing into the composite bloq."""
    218 from qualtran.simulation.classical_sim import call_cbloq_classically
[...]

PR #1305, opened 2024-08-19
Restrict controlled from non-thru registers

mpharrigan:
For #1304 

 - Only thru registers allowed in `Controlled`
 - Prevents triggering the bug in classical simulation where controlled allocations are not supported

### DIFF
--- a/qualtran/_infra/controlled_test.py
+++ b/qualtran/_infra/controlled_test.py
@@ -385,62 +385,6 @@ def test_controlled_global_phase_tensor():
     np.testing.assert_allclose(bloq.tensor_contract(), should_be)
 
 
-@attrs.frozen
-class TestCtrlStatePrepAnd(Bloq):
-    """Decomposes into a Controlled-AND gate + int effects & targets where ctrl is active.
-
-    Tensor contraction should give the output state vector corresponding to applying an
-    `And(and_ctrl)`; assuming all the control bits are active.
-    """
-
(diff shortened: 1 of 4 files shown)


tanujkhattar:
Left a comment on the original issue - https://github.com/quantumlib/Qualtran/issues/1304#issuecomment-2297600946

mpharrigan:
Let's do this and relax later

qualtran/_infra/controlled_test.py
```diff
@@ -385,62 +366,6 @@ def test_controlled_global_phase_tensor():
...
-@attrs.frozen
-class TestCtrlStatePrepAnd(Bloq):
-    """Decomposes into a Controlled-AND gate + int effects & targets where ctrl is active.
-
-    Tensor contraction should give the output state vector corresponding to applying an
-    `And(and_ctrl)`; assuming all the control bits are active.
-    """
-
-    ctrl_spec: CtrlSpec
-    and_ctrl: Tuple[int, int]
```
tanujkhattar, line 397: Can we add a private flag to the `Controlled` class, something of the form `_THRU_REGISTERS_ONLY = True`, and keep this test around? Since we plan to relax this constraint later, it'll be nice to not delete the test.

tanujkhattar (approved):
LGTM % one request. Please add the restriction controlled by a flag set to TRUE by default.

PR #1205, opened 2024-07-28
Sparse state preparation via Dense + Permute

anurudhp:

PR #1307, opened 2024-08-19
Add more permutation tests

anurudhp:
Adding tests to catch the (fixed) infinite loop bug: https://github.com/quantumlib/Qualtran/pull/1205/files#r1722422347


mpharrigan (approved):
thank you!

PR #1308, opened 2024-08-20
SparseStatePrep: fix bug in using `scipy.sparse.dok_array`

anurudhp:
[`dok_array`](https://docs.scipy.org/doc/scipy-1.14.0/reference/generated/scipy.sparse.dok_array.html) should only accept only 2D arrays (based on docs), but seems to support 1D arrays in newer versions.

This bugfix assumes it only works with 2D, and invokes it correctly. Tested against scipy v1.10.0.

discussion: https://github.com/quantumlib/Qualtran/pull/1205#discussion_r1722411733


mpharrigan (approved):
thank you!

Issue #1290, opened 2024-08-15
Incorrect gate counts for controlled gates


anurudhp:
`Controlled` bloqs do not have correct costs when the subbloq provides specialized controlled versions for only a subset of CtrlSpec.

```py
from qualtran import CtrlSpec
from qualtran.bloqs.basic_gates import XGate
from qualtran.resource_counting import get_cost_value, QECGatesCost


def show_cost(bloq):
    cost = get_cost_value(bloq, QECGatesCost())
    print(f"{bloq} :: {cost}")


show_cost(XGate().controlled(CtrlSpec(cvs=(1, 1))))
show_cost(XGate().controlled(CtrlSpec(cvs=(1, 0))))
show_cost(XGate().controlled(CtrlSpec(cvs=(1, 1, 1))))
```
```
Toffoli :: toffoli: 1
C[2][XGate] :: -
C[3][XGate] :: -
```

mpharrigan:
It's sortof impossible for it to know the accurate counts for a `Controlled()` atomic bloq.

But there's a bug nevertheless -- it shouldn't report successfully zero counts; it should raise an error

mpharrigan:
How does the t_complexity protocol handle these?

anurudhp:
`.t_complexity()` raises a `TypeError` for the second and third examples.

mpharrigan:
fixed in #1313  -- this will cause an error to be raised. Open follow-up issues for missing costs for Controlled(Atomic)

PR #1311, opened 2024-08-20
Don't silently omit counts from missing callees

mpharrigan:
Addresses the buggy part of #1290 by re-raising the error. Better error messages from Controlled(Atomic).

### DIFF
--- a/qualtran/_infra/controlled.py
+++ b/qualtran/_infra/controlled.py
@@ -31,7 +31,7 @@ import cirq
 import numpy as np
 from numpy.typing import NDArray
 
-from .bloq import Bloq
+from .bloq import Bloq, DecomposeNotImplementedError, DecomposeTypeError
 from .data_types import QBit, QDType
 from .gate_with_registers import GateWithRegisters
 from .registers import Register, Side, Signature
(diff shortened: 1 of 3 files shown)


anurudhp (commented):
Do we distinguish between atomic bloqs and bloqs where the decomposition failed? Because in the former we treat it as a leaf bloq, and in the latter we cannot compute the correct cost. Is my understanding correct?

mpharrigan:
> Do we distinguish between atomic bloqs and bloqs where the decomposition failed? Because in the former we treat it as a leaf bloq, and in the latter we cannot compute the correct cost. Is my understanding correct?

If it's a leaf bloq, `QECGatesCost` should have already determined its cost before trying to peer into its decomposition. If you get to the point in the code where you're trying to look at callees to determine costs: if it's a `DecomposeTypeError` then it's an error since we've reached an atomic, leaf bloq that we don't know how to count; if it's a `DecomposeNotImplementedError` then there are costs that should be from the decomposition but there is no decomposition. So for resource counting, those are both an error condition where we can't return resource counts

anurudhp:
Ah, makes sense, thanks!

anurudhp (approved):

mpharrigan:
I was going to factor out this change from the fixes i've been making to QECGatesCost but they're somewhat intertwined since this change reveals the failures that need some of the other fixes. Moving to draft

mpharrigan:
subsumed by #1313

PR #1313, opened 2024-08-20
[QECGatesCost] strict & account for more leaf bloqs

mpharrigan:
The first commit is #1311 ; which can't be merged independently because it reveals errors that this PR fixes.

 - account for bookeeping, identity, global phase gates
 - partition rotations into clifford, t, or rotation based on their angle

### DIFF
--- a/qualtran/resource_counting/classify_bloqs.py
+++ b/qualtran/resource_counting/classify_bloqs.py
@@ -127,6 +127,7 @@ _CLIFFORD_ANGLES = np.array(
 _CLIFFORD_EXPONENTS = np.array([1.0, 0.5, 1.5, -1.0, -0.5, -1.5, 0.0, -2, 2])
 _T_ANGLES = np.array([np.pi / 4, -np.pi / 4])
 _T_EXPONENTS = np.array([0.25, -0.25])
+_ANGLE_ATOL = 1e-12
 
 
 def bloq_is_t_gate(b: Bloq) -> bool:
@@ -144,14 +145,14 @@ def bloq_is_t_gate(b: Bloq) -> bool:
     if isinstance(b, (Rz, Rx, Ry)):
         if is_symbolic(b.angle):
             return False  # Symbolic rotation
-        if np.any(np.abs(b.angle - _T_ANGLES) < b.eps):
+        if np.any(np.abs(b.angle - _T_ANGLES) < _ANGLE_ATOL):
             return True  # T hidden in a rotation bloq
         return False
 
     if isinstance(b, (ZPowGate, XPowGate, YPowGate)):
         if is_symbolic(b.exponent):
             return False  # Symbolic rotation
-        if np.any(np.abs(b.exponent - _T_EXPONENTS) < b.eps):
+        if np.any(np.abs(b.exponent - _T_EXPONENTS) < _ANGLE_ATOL):
             return True  # T hidden in a rotation bloq
         return False
 
@@ -206,14 +207,14 @@ def bloq_is_clifford(b: Bloq) -> bool:
     if isinstance(b, (Rz, Rx, Ry)):
         if is_symbolic(b.angle):
             return False  # Symbolic rotation
-        if np.any(np.abs(b.angle - _CLIFFORD_ANGLES) < b.eps):
+        if np.any(np.abs(b.angle - _CLIFFORD_ANGLES) < _ANGLE_ATOL):
             return True  # Clifford hidden in a rotation bloq
         return False
 
     if isinstance(b, (ZPowGate, XPowGate, YPowGate)):
         if is_symbolic(b.exponent):
             return False  # Symbolic rotation
-        if np.any(np.abs(b.exponent - _CLIFFORD_EXPONENTS) < b.eps):
+        if np.any(np.abs(b.exponent - _CLIFFORD_EXPONENTS) < _ANGLE_ATOL):
             return True  # Clifford hidden in a rotation bloq
         return False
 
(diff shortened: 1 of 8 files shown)


anurudhp (approved):
lgtm % concerns using `bloq.eps`

mpharrigan:
I don't see a comment, did you try to leave one? From offline: I'll use machine precision to compare the angles

qualtran/resource_counting/classify_bloqs.py
```diff
@@ -109,7 +111,62 @@ def classify_t_count_by_bloq_type(
...
+    """
+    from qualtran.bloqs.basic_gates import Rx, Ry, Rz, TGate, XPowGate, YPowGate, ZPowGate
+
+    if isinstance(b, TGate):
+        return True
+
+    if isinstance(b, (Rz, Rx, Ry)):
+        if is_symbolic(b.angle):
+            return False  # Symbolic rotation
+        if np.any(np.abs(b.angle - _T_ANGLES) < b.eps):
```
anurudhp, line 147: `bloq.eps` is currently not propagated properly, and in most places uses hardcoded values (usually 1e-11). Might be better to check against a configurable precision?

qualtran/resource_counting/classify_bloqs.py
```diff
@@ -109,7 +111,62 @@ def classify_t_count_by_bloq_type(
...
+            return True  # T hidden in a rotation bloq
+        return False
+
+    if isinstance(b, (ZPowGate, XPowGate, YPowGate)):
+        if is_symbolic(b.exponent):
+            return False  # Symbolic rotation
+        if np.any(np.abs(b.exponent - _T_EXPONENTS) < b.eps):
```
anurudhp, line 154: (eps issue above)

qualtran/resource_counting/classify_bloqs.py
```diff
@@ -145,12 +203,42 @@ def bloq_is_clifford(b: Bloq):
...
+        if is_symbolic(b.angle):
+            return False  # Symbolic rotation
+        if np.any(np.abs(b.angle - _CLIFFORD_ANGLES) < b.eps):
+            return True  # Clifford hidden in a rotation bloq
+        return False
+
+    if isinstance(b, (ZPowGate, XPowGate, YPowGate)):
+        if is_symbolic(b.exponent):
+            return False  # Symbolic rotation
+        if np.any(np.abs(b.exponent - _CLIFFORD_EXPONENTS) < b.eps):
```
anurudhp, line 216: (eps issue above)

qualtran/resource_counting/classify_bloqs.py
```diff
@@ -145,12 +203,42 @@ def bloq_is_clifford(b: Bloq):
     ):
         return True
 
+    if isinstance(b, (Rz, Rx, Ry)):
+        if is_symbolic(b.angle):
+            return False  # Symbolic rotation
+        if np.any(np.abs(b.angle - _CLIFFORD_ANGLES) < b.eps):
```
anurudhp, line 209: (eps issue above)

qualtran/resource_counting/classify_bloqs.py
```diff
@@ -160,4 +248,25 @@ def bloq_is_rotation(b: Bloq):
...
         return bloq_is_rotation(b.subbloq)
 
-    return isinstance(b, (Rz, Rx, Ry, ZPowGate, XPowGate, YPowGate))
+    if isinstance(b, CZPowGate):
+        return True
+
+    if isinstance(b, (Rz, Rx, Ry)):
+        if is_symbolic(b.angle):
+            return True
+        if np.any(np.abs(b.angle - _CLIFFORD_ANGLES) < b.eps):
```
anurudhp, line 257: (eps issue above)

qualtran/resource_counting/classify_bloqs.py
```diff
@@ -160,4 +248,25 @@ def bloq_is_rotation(b: Bloq):
...
+            return False  # Clifford
+        if np.any(np.abs(b.angle - _T_ANGLES) < b.eps):
```
anurudhp, line 259: (eps issue above)

qualtran/resource_counting/classify_bloqs.py
```diff
@@ -160,4 +248,25 @@ def bloq_is_rotation(b: Bloq):
...
+            return False  # T gate
+        return True
+
+    if isinstance(b, (ZPowGate, XPowGate, YPowGate)):
+        if is_symbolic(b.exponent):
+            return True
+        if np.any(np.abs(b.exponent - _CLIFFORD_EXPONENTS) < b.eps):
```
anurudhp, line 266: (eps issue above)

qualtran/resource_counting/classify_bloqs.py
```diff
@@ -160,4 +248,25 @@ def bloq_is_rotation(b: Bloq):
...
+            return False  # Clifford
+        if np.any(np.abs(b.exponent - _T_EXPONENTS) < b.eps):
```
anurudhp, line 268: (eps issue above)

anurudhp (commented):
oops, added the comment now.

qualtran/resource_counting/classify_bloqs.py
mpharrigan: comparing against a classical float-comparison value. If users want to "round" angles that should best be done by an external force. This tolerance is just for classical, python float comparison. Thanks!

qualtran/resource_counting/classify_bloqs.py
```diff
@@ -109,7 +111,63 @@ def classify_t_count_by_bloq_type(
...
+
+    if isinstance(b, TGate):
+        return True
+
+    if isinstance(b, (Rz, Rx, Ry)):
+        if is_symbolic(b.angle):
+            return False  # Symbolic rotation
+        if np.any(np.abs(b.angle - _T_ANGLES) < _ANGLE_ATOL):
+            return True  # T hidden in a rotation bloq
+        return False
```
tanujkhattar, line 150: I think `bloq_is_t_gate` should return True only if the unitary of the bloq is equal to that of a `T` gate. For other single qubit rotations, we should have default decompositions; which a user can choose to override if needed; to avoid undercounting cliffords.

mpharrigan: we generally don't have "default" decompositions that one can override. 

I can change this function name to `bloq_is_t_like`. Counting Rx(pi/4) as 1 T gate is much more accurate than 1 "rotation", so I'm going to proceed with the change. We can add a decomposition of Rx to Rz later _or_ change the workflow where a re-write rule is responsible for simplifying any Rx(theta) to X/4 or whatever we like (and remove this itchy logic from the gate counting code).

mpharrigan: it's also unclear (to me) how we'd want to count cliffords for a situation like Rx(pi/4). Using an auto-correcting circuit like Fig 17 of game-of-surface-codes implies you can do any pauli with the same number of gates

Issue #1315, opened 2024-08-20
QROAMCleanAdjoint signature 


fdmalone:
Would it make sense to have the signature of QROAMCleanAdjoint match that of QROAMClean so that for bloqs which use QROAMClean they don't have to implement a bespoke adjoint method and can just defer to the default Adjoint or is this even possible?

cc @tanujkhattar

Issue #1121, opened 2024-07-10
Make CSWAP a bloq of its own


ncrubin:
The cswap implementation

https://github.com/quantumlib/Qualtran/blob/3f325ce16ab44890a9e2150338e11cee0a4a8225/qualtran/bloqs/arithmetic/comparison.py#L258

should be a block of its own

mpharrigan:
we of course already have https://github.com/quantumlib/Qualtran/blob/95141395bdb1384a08b87874bba1977ddc28568a/qualtran/bloqs/basic_gates/swap.py#L107

PR #1316, opened 2024-08-20
Use QROAMClean in Chemistry Sparse Prepare bloq

fdmalone:
Uses the new QROAMClean bloq which allows the costs to match more closely. Some issues:

1. ~Delegating the sparse indices output from QROAM as junk registers causes an issue with building block encodings as they are no longer selection registers (xref: #549) so will not match SELECTs signature. This was a hack at the time but is a blocker now.~
2. I went back and forth on how to implement the adjoint with the current implementation functional but not pleasant and error prone (cc @mpharrigan for why I was asking those questions in the chat). I think #1315 **maybe** would allow use to use the Adjoint wrapper here by delegating the register wrangling logic to QROAM but not 100% sure.
3. There's an off by 2 Toffoli count difference between openfermion and qualtran's QROAM which I haven't tracked down yet.

flyby bug fix for QROAMClean wire symbols and includes #1302 which should be merged first.


fdmalone:
I guess on 2 I could call _build_composite_bloq, pick out the QROAM bloq and handle then handle the register mapping.

fdmalone:
I'm sort of troubled that all the tests passed given that there are sparse walk operator examples...

fdmalone:
> I'm sort of troubled that all the tests passed given that there are sparse walk operator examples...

After talking to @tanujkhattar  Point 1 is not valid, we're not feeding Prepares selection registers into SELECT, just grabbing soqs from **soqs. Point 2 probably needs a quick update to QROAMAdjoint

fdmalone:
@tanujkhattar the -2 difference stems from the -8 (/4 = -2) term in our QROAM which isn't in the openfermion calculation. My guess is ours is more correct? Do you remember where the -2 comes from?

Screenshot 2024-08-22 at 10 49 31 AM

Otherwise this is good to go!

tanujkhattar:
The `-2` is from the `QROM` which uses unary iteration and requires `N - 2` AND gates for doing (un-controlled) unary iteration on `N` qubits. It would be `N-1` for controlled-unary iteration.

So yes, we are more accurate.

qualtran/bloqs/chemistry/sparse/prepare.py
```diff
@@ -178,9 +182,8 @@ class PrepareSparse(PrepareOracle):
     one_body: Tuple[int, ...] = attrs.field(repr=False)
     keep: Tuple[int, ...] = attrs.field(repr=False)
     num_bits_rot_aa: int = 8
-    is_adjoint: bool = False
     sum_of_l1_coeffs: SymbolicFloat = 0.0
-    qroam_block_size: Optional[int] = None
+    qroam_block_size: int = 1
```
tanujkhattar, line 186: Should we just store the `log_block_size` and change `block_size` to be a computed property? It makes more sense to store the log_block_size to avoid potential bugs where `block_size` is not a power of 2.

qualtran/bloqs/chemistry/sparse/prepare.py
```diff
@@ -340,75 +345,48 @@ def build_qrom_bloq(self) -> 'Bloq':
...
-        alt_pqrs: NDArray[Soquet],  # type: ignore[type-var]
-        theta: NDArray[Soquet],  # type: ignore[type-var]
-        keep: 'SoquetT',
-        less_than: 'SoquetT',
-        alt_flag_1b: 'SoquetT',
-    ) -> Dict[str, 'SoquetT']:
+    def add_qrom(self, bb: 'BloqBuilder', **soqs: 'SoquetT') -> Dict[str, 'SoquetT']:
+        qrom = self.build_qrom_bloq()
+        # The qroam_junk_regs won't be present initially when building the
+        # composite bloq as they're RIGHT registers.
```
tanujkhattar, line 356: Ditto: Please update to `qroam_target` and `qroam_junk` correspondingly; instead of `qroam_junk` and `qroam_extra_junk`.

qualtran/bloqs/chemistry/sparse/prepare.py
```diff
@@ -340,75 +345,48 @@ def build_qrom_bloq(self) -> 'Bloq':
...
-            target12_=keep,
-        )
-        lte_bloq = LessThanEqual(self.num_bits_state_prep, self.num_bits_state_prep)
+        soqs |= self.add_qrom(bb, **soqs)
         # prepare uniform superposition over sigma
-        sigma = bb.add(OnEach(self.num_bits_state_prep, Hadamard()), q=sigma)
-        keep, sigma, less_than = bb.add(lte_bloq, x=keep, y=sigma, target=less_than)
-        less_than, theta[1] = bb.add(CZ(), q1=less_than, q2=theta[1])
+        soqs['sigma'] = bb.add(OnEach(self.num_bits_state_prep, Hadamard()), q=soqs['sigma'])
+        # Inequality test for alias sampling
```
tanujkhattar, line 378: Unrelated to the PR -- but do you know why are we implementing the alias sampling circuit explicitly here instead of directly relying on using `StatePreparationViaAliasSampling` bloq ? If it makes sense, Can you open an issue to document the requirements that would enable us to directly use the alias sampling bloq here?

qualtran/bloqs/chemistry/sparse/sparse_test.py
```diff
@@ -125,11 +77,8 @@ def test_sparse_costs_against_openfermion(num_spin_orb, num_bits_rot_aa):
     )
     # correct for SelectSwapQROM vs QROAM
     # https://github.com/quantumlib/Qualtran/issues/574
-    paper_qrom = qrom_cost(prep_sparse)
-    paper_qrom += qrom_cost(prep_sparse_adj)
-    qual_qrom_cost = get_sel_swap_qrom_toff_count(prep_sparse)
-    qual_qrom_cost += get_sel_swap_qrom_toff_count(prep_sparse_adj)
-    delta_qrom = qual_qrom_cost - paper_qrom
+    # The -2 comes from a more accurate calculation of the QROAM costs in Qualtran.
+    delta_qrom = -2
```
tanujkhattar, line 81: Feel free to elaborate -- "The -2 comes from a more accurate calculation of the QROAM costs in Qualtran. The difference arises from uncontrolled unary iteration used by QROM, which QROAMClean delegates to. "

tanujkhattar (approved):
LGTM % nits

qualtran/bloqs/chemistry/sparse/prepare.py
```diff
@@ -193,63 +196,61 @@ def selection_registers(self) -> Tuple[Register, ...]:
...
-            Register("flag_1b", BoundedQUInt(1)),
         )
 
     @cached_property
     def junk_registers(self) -> Tuple[Register, ...]:
+        extra_junk = (Register("less_than", QBit()),)
+        return extra_junk + self.qroam_junk_regs + self.qroam_extra_junk_regs
+
+    @cached_property
+    def qroam_junk_regs(self) -> Tuple[Register, ...]:
```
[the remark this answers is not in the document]
fdmalone: Eh, so the junk naming here is relative to Prepares junk registers, but maybe this makes more sense

qualtran/bloqs/chemistry/sparse/prepare.py
fdmalone: Done.

qualtran/bloqs/chemistry/sparse/prepare.py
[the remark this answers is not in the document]
fdmalone: Done.

qualtran/bloqs/chemistry/sparse/prepare.py
```diff
@@ -340,75 +345,48 @@ def build_qrom_bloq(self) -> 'Bloq':
             self.keep,
             target_bitsizes=target_bitsizes,
             log_block_sizes=log_block_sizes,
-            use_dirty_ancilla=False,
         )
+        # if self.is_adjoint:
+        #     return qrom.adjoint()
```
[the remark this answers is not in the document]
fdmalone: Done.

qualtran/bloqs/chemistry/sparse/prepare.py
fdmalone: Well in this case we're using QROAM not QROM and our QROAM is quite special. Not sure if there is a more general abstraction to be made for SparseStatePreparationAliasSampling, but maybe?

fdmalone: this has a single selection registers, multiple ind / alt registers, uses QROAM AND (not implemented here) uses swaps which can be inverted for free using MBUC (a la #1121 (I think))

fdmalone: Ideally interchanging the QROMs for all the state prep bloqs would be nice

fdmalone:
@tanujkhattar there was a hot fix in there related to point 1. which became an issue. My fix was to take the union of the selection registers for select and reflect (prepare) as the "selection" registers for select are "junk" registers from prepare in the sparse case.

fdmalone:
Whoops, auto-merge did its thing. I'll make the same change with THC shortly so can revert then if problematic.

Issue #1250, opened 2024-08-05
Propagate rotation epsilon in gate cost


mpharrigan:
The `t_complexity` protocol ignores the `eps` parameter when counting rotations.

The solution is to enhance the `QECGatesCost` cost and associated dataclass `GateCounts` to count rotation/eps pairs.

anurudhp:
Sample code snippet:
```py
import sympy
from qualtran.bloqs.basic_gates import ZPowGate

t, eps = sympy.symbols(r"t \epsilon")
bloq = ZPowGate(exponent=t, eps=eps)
print(bloq.t_complexity())
```
Prints `TComplexity(rotations=1)`.

One could do `bloq.t_complexity().t_incl_rotations(eps)`, but this `eps` is not the same as the bloq's `eps` (which in general could be different for each subbloq).

tanujkhattar:
Some history:

eps for rotations was introduced in the early days of Qualtran and I think is a useful parameter when constructing rotation bloqs. I agree it's probably not propagated correctly everywhere (though I remember implementing and correctly propagating `eps` in qvr, qft and hamming weight phasing bloq) -- we should file issues and try to get that fixed. 

Originally, the rotation bloqs had a `_t_complexity_` method implemented and would return the number of T gates scaling with `log(1/eps)`; so the T-counts for call graphs with rotation bloqs would take the eps in account. (eg: https://github.com/quantumlib/Qualtran/pull/469)

As part of unifying / deprecating the T-complexity protocol; a temporary fix was to return rotations=1 for all rotation bloqs and add a method on the TComplexity class to convert rotations into T-gates. Eventually, the idea was to use the new costing framework to keep track of bloq counts and then use the eps to convert rotation bloqs to t-gates like we used to do originally. Here are some relevant discussions - 
https://github.com/quantumlib/Qualtran/issues/662#issuecomment-1949215534
https://github.com/quantumlib/Qualtran/pull/678#discussion_r1492944645

Using a single / hardcoded eps to convert all rotation bloqs to t gates was never meant to be a permanent solution. We've just not gotten around completing the migration yet I guess.

tanujkhattar:
@mpharrigan As per our discussion offline, it'll be nice to have this as part of the v1.0 milestone.

Issue #878, opened 2024-04-16
Add explicit `controlled-U` bloqs for single qubit unitaries `U` and annotate them with costs / decompositions


tanujkhattar:
Some bloqs which need this are
- Rotations - Rx, Ry, Rz
- Hadamard - CH

etc.

mpharrigan:
CHadamard is implemented. I've annotated it with 2 rotations and 4 cliffords with the following description:

>This is based on the decomposition provided by `cirq.decompose_multi_controlled_rotation`
> which uses three cirq.MatrixGate's to do a controlled version of any single-qubit gate.
> The first MatrixGate happens to be a clifford, Hadamard operation in this case.
> The other two are considered 'rotations'.

tanujkhattar:
We should update the annotation with 2 T gates + cliffords instead of 2 rotations as per my comment in https://github.com/quantumlib/Qualtran/issues/233

[...]

PR #1317, opened 2024-08-20
Cleanup redundant code for computing t complexity

anurudhp:
Cleaning up some redundant code to compute t-complexity.

- remove `_get_all_rotation_types` (in favor of `bloq_is_rotation`).
- reduce uses of `t_counts_from_sigma` (in favor of `QECGatesCost`).

TODOs in follow up:
- Currently, only `t_counts_from_sigma` uses eps correctly, but it's only used in a handful of places. This should be replaced with `QECGatesCost` after adding support for eps (#1250).
- We could add a simple `GateCounts.total_t_and_rotation_count` helper to support existing use cases which only take number of rotations (e.g. Beverland-et-al model).

Relevant discussions: 
- https://github.com/quantumlib/Qualtran/issues/662#issuecomment-1949215534
- https://github.com/quantumlib/Qualtran/pull/678#discussion_r1492944645


qualtran/resource_counting/t_counts_from_sigma_test.py
```diff
@@ -11,66 +11,54 @@
...
-    }
-    expected_t_count = (
-        +100
-        + 1 * TComplexity.rotation_cost(z_eps1)
-        + 5 * TComplexity.rotation_cost(z_eps2)
-        + 9 * TComplexity.rotation_cost(x_eps)
-        + 13 * TComplexity.rotation_cost(y_eps)
-        + 20 * TComplexity.rotation_cost(cz_eps)
-    )
-    assert t_counts_from_sigma(sigma) == expected_t_count
```
mpharrigan, line 76: can you revert this test. This actually tests the expected behavior of `t_counts_from_sigma`, i.e. it has a sigma with more than one entry

mpharrigan (commented):
Generally very reasonable changes; unsure about that one test deletion

qualtran/resource_counting/t_counts_from_sigma_test.py
anurudhp: ah makes sense, reverted.

qualtran/resource_counting/t_counts_from_sigma.py
```diff
@@ -11,45 +11,26 @@
...
+    from qualtran.resource_counting.classify_bloqs import bloq_is_rotation
 
-    if rotation_types is None:
-        rotation_types = _get_all_rotation_types()
     ret = sigma.get(TGate(), 0) + sigma.get(TGate().adjoint(), 0)
     for bloq, counts in sigma.items():
-        if isinstance(bloq, rotation_types) and not cirq.has_stabilizer_effect(bloq):
+        if bloq_is_rotation(bloq) and not cirq.has_stabilizer_effect(bloq):
+            if isinstance(bloq, Controlled):
+                # TODO native controlled rotation bloqs missing (CRz, CRy etc.)
```
mpharrigan, line 32: can you link that issue #878

mpharrigan (approved):

PR #1322, opened 2024-08-20
Make `Identity` a bookkeeping bloq

### LINKED ISSUE #1207: Add n-bit Identity, implement C[I] = I
We have a one qubit `Identity` bloq. This should probably let you do an n-bit identity operation and the controlled version of it should be an `n+m` bit identity for m control bits.

anurudhp:
Fixes #1207


mpharrigan:
Thanks. This does indeed solve the issue as posed, but I'm not sure it's exactly what we need. The C[I] is indeed equivalent to `I` on the second qubit and no operation on the control qubit-- but then again, `I` is equivalent to no operation on the qubit. The purpose (I'm making an educated guess) of an `Identity` bloq is to have control over the adjacency matrix of the compute graph without _doing_ anything to the data. Currently, the `I` bloq is used in a list of bloqs to select from. You can imagine wanting to select from a list of $n$-qubit operations, and put the identity at some of the indices. You can imagine using a multi-qubit identity bloq to form a "barrier" in the circuit to time-align operations. If the answer to "why do we have an identity bloq" is indeed "to control the adjacency matrix of the compute graph without doing anything" then I think turning a two bit Identity operation into a one bit identity operation as done in this PR is not the exact right move.

Organizationally, I also think it belongs back where it was; alongside its pauli friends.

anurudhp:
closed in favor of #1327

Issue #1292, opened 2024-08-16
Constructing QROAMCleanAdjoint is slow


fdmalone:
Just reporting for the moment couldn't investigate much but it's thousands of times slower:

``` python
import time

import numpy as np

from qualtran.bloqs.data_loading.qroam_clean import QROAMClean
from qualtran.bloqs.data_loading.select_swap_qrom import SelectSwapQROM

for i in range(100, 1100, 100):
    x = np.random.randint(0, 100, i)
    start = time.time()
    bloq = SelectSwapQROM.build_from_data(x, target_bitsizes=(30,)).adjoint()
    sel_swp = time.time() - start
    start = time.time()
    bloq = QROAMClean.build_from_data(x, target_bitsizes=(30,)).adjoint()
    qroam_swp = time.time() - start
    print(i, sel_swp, qroam_swp, qroam_swp / sel_swp)
```

It's almost certainly the triple loop in batched_data_permuted. This is problematic for the chemistry bloqs which often contain tensors with 10^5-10^6 elements. Granted we should probably be using symbolics for resource estimates, but not quite yet.

PR #1324, opened 2024-08-21
Add `QROAMCleanAdjointWrapper` to match signature of `QROAMClean` and delegate implementation to `QROAMCleanAdjoint`

tanujkhattar:
Fixes https://github.com/quantumlib/Qualtran/issues/1315

This also fixes https://github.com/quantumlib/Qualtran/issues/1292 since construction of `batched_data_permuted` is now done lazily; only when the decomposition is required. Duplication of call graph mitigates the slow computation for the moment. 

cc @fdmalone

### DIFF
--- a/qualtran/bloqs/data_loading/qroam_clean.py
+++ b/qualtran/bloqs/data_loading/qroam_clean.py
@@ -21,7 +21,7 @@ import numpy as np
 import sympy
 from numpy.typing import ArrayLike
 
-from qualtran import bloq_example, BloqDocSpec, GateWithRegisters, Register, Side, Signature
+from qualtran import Bloq, bloq_example, BloqDocSpec, GateWithRegisters, Register, Side, Signature
 from qualtran.bloqs.basic_gates import Toffoli
 from qualtran.bloqs.data_loading.qrom_base import QROMBase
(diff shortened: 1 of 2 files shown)


qualtran/bloqs/data_loading/qroam_clean.py
```diff
@@ -205,6 +205,94 @@ def wire_symbol(self, reg: Optional[Register], idx: Tuple[int, ...] = tuple()) -
         raise ValueError(f'Unknown register name {name}')
 
 
+@attrs.frozen
+class QROAMCleanAdjointWrapper(Bloq):
```
fdmalone, line 209: Should we keep this as QROAMCleanAdjoint and rename QROAMCleanAdjoint TableLookupAdjoint (or something appropriate)

fdmalone (approved):

qualtran/bloqs/data_loading/qroam_clean.py
tanujkhattar: I thought about it and I think `TableLookupAdjoint` would be too general, especially as we add more adjoint bloqs that have different tradeoffs for using dirty vs clean ancillas; along with using measurement based uncomputation. 

the current name is closely tied to its implementation; so I think it's fine to keep it as is.

PR #1327, opened 2024-08-21
Make `Identity` an `n` qubit bloq

### LINKED ISSUE #1207: Add n-bit Identity, implement C[I] = I

anurudhp:
fixes #1207 

- `Identity` now accepts a `bitsize`, and has a single register `q` of `QAny(bitsize)`.
- `Identity(n).controlled(m)` is `Identity(n + m)`.

more discussion: https://github.com/quantumlib/Qualtran/pull/1322#issuecomment-2299989824

### DIFF
--- a/qualtran/bloqs/basic_gates/identity.ipynb
+++ b/qualtran/bloqs/basic_gates/identity.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "92a37b86",
+   "id": "6cc35310",
    "metadata": {
     "cq.autogen": "title_cell"
    },
@@ -13,7 +13,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "c6021f9f",
+   "id": "0097397d",
    "metadata": {
     "cq.autogen": "top_imports"
    },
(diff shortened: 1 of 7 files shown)


fdmalone:
I'm still a little confused about why the Identity being a single qubit bloq is problematic, does OnEach(Identity, n).controlled(m) not do the same thing?

tanujkhattar:
I think `OnEach(Identity(), n).controlled(m)` would probably do the same thing; but I think it's nicer and cleaner to have an n-bit Identity operation that returns itself when controlled.

fdmalone:
Ok, should this still be a basic gate? I think the justification for moving it back to basic gates from #1322 was that it was more logically consistent to be among $I$, $X$, $Y$ and $Z$ and basic gates should probably be restricted to one and two qubit gates maybe? Please ignore me if it makes things easier / better / cleaner.

anurudhp:
Another subtle issue with `OnEach(Identity(), n).controlled(m)` is that it would produce `m` bloqs of `OnEach(Identity(), n + 1)` (because the control qubits are common), and therefore `m * (n + 1)` `Identity()` bloqs. Having a single `Identity(n)` makes it cleaner imo.

fdmalone:
SGTM. Do we need to have both? Probably not right?

anurudhp:
both as in?

mpharrigan:
> I'm still a little confused about why the Identity being a single qubit bloq is problematic, doesOnEach(Identity(), n).controlled(m) not do the same thing?

The motivation is that we _have_ an identity bloq (that is somewhat out of our control; e.g. passed as a class attribute; e.g. in applyLthgate) that we want to control. We could return `OnEach`; but the spirit of qualtran is to give names to things

mpharrigan (approved):
very nice, thanks so much for this. Lovely use of `add_and_partition`

mpharrigan:
>  SGTM. Do we need to have both? Probably not right?

Not 100% sure what's being asked here, but this PR keeps one `Identity` bloq that has a bitsize that defaults to 1, so there's only one thing to contend with.

Issue #1309, opened 2024-08-20
Counting rotations notes


mpharrigan:

mpharrigan:
We treat Rz(pi/4) as a T gate. I'm not sure how I feel about that; but for expediency I will reproduce this behavior in the QECGatesCost cost.

However, the t_complexity protocol treats Rx(pi/4) as a rotation instead of a T and a clifford. I think we should be consistent and count these as a T gate. I'm not going to count them as a T+clifford because that complicates the code.

The Rx and Ry bloq examples use T angles. We should change them to be arbitrary angle

mpharrigan:
LPRSInterimPrep -- its t_complexity method does not match its build_call_graph, or at least not how we usually treat it. Elsewhere, we treat a CRz as one "rotation"; here' each is treated as 2 rotations and 3 cliffords

mpharrigan:
QvrZPow.qvr_zpow uses six ZPowGates. But the exponent is less than epsilon and my wip code treats it as six cliffords

mpharrigan:
I think the t complexity of `QuantumVariableRotation` is very incorrect

https://github.com/quantumlib/Qualtran/blob/03d09a01f0e246bbd30573c817c60becd49f0621/qualtran/bloqs/chemistry/trotter/grid_ham/qvr.py#L61

since the 0 angle gets turned into a clifford

mpharrigan:
LPResourceState also relies on simplifying CZ**1.0 to CZ

mpharrigan:
https://github.com/quantumlib/Qualtran/blob/7823bdbf51e596babfd242726b5aba04d8abc93f/qualtran/bloqs/basic_gates/su2_rotation.py#L151

SU2RotationGate always returns 3 rotations, no matter the angles; even if they turn into cliffords or Ts. Indeed the bloq_examples all use angles that can be specialized.

mpharrigan:
https://github.com/quantumlib/Qualtran/blob/ec85731ec6e0e498797b8412b39f7bdd75bc20f0/qualtran/bloqs/arithmetic/controlled_addition.py#L163

bloq_example "cmodadd_symbolic" has CAdd as one of its callees. The t complexity protocol reports the number of toffolis as $8n+4$, which would be $2n + 1$ AND's. The call graph reports n AND + 1 `Add` with $n$ ands; giving a total of $2n$ ANDs

@NoureldinYosri can you investigate

mpharrigan:
[...]

PR #1329, opened 2024-08-21
Fix  and improve CAdd tests

NoureldinYosri:
context: https://github.com/quantumlib/Qualtran/issues/1309#issuecomment-2299843007

The old tcomplexity annotation represented the buggy original decomposition from https://github.com/quantumlib/Qualtran/pull/864


mpharrigan:
thanks!

PR #1333, opened 2024-08-22
[QECGatesCost] Adjustments part 3

mpharrigan:
This logic is slightly inconsistent within itself, but it matches the previous behavior from the `t_complexity` protocol. We count Controlled(Rotation) as a rotation no matter what the rotation angle is. The ultimate fix is blocked by having explicit representations for all leaf controlled locations (https://github.com/quantumlib/Qualtran/issues/878).

 - Also make debugging unexpected qubit count failures easier with a helpful wrapping error message that includes the problematic bloq
 - Don't need a hack for Controlled(GlobalPhase) following #1113 
 - Improve hacky logic: multi-bit controls aren't counted as rotations


qualtran/resource_counting/classify_bloqs.py
```diff
@@ -244,10 +244,12 @@ def bloq_is_rotation(b: Bloq) -> bool:
     if isinstance(b, Controlled):
```
anurudhp, line 244: This branch captures any `CtrlSpec` right? For non-trivial ctrl-spec, shouldn't we error out? (or use the and-ladder, but perhaps for later) #1290

anurudhp (approved):
lgtm % doubt regarding Controlled override

qualtran/resource_counting/classify_bloqs.py
mpharrigan: yes, that is an issue; but not a _new_ issue introduced by this PR. Let me see if I can sneak a quick fix in or if it'll be a whole thing; in which case, I will defer to another PR

mpharrigan: I could -- there aren't any places in the unit tests or bloq examples that relies on `Controlled(rot, #cvs>1)`

PR #1340, opened 2024-08-23
Remove pretty_name part 1

mpharrigan:
Change the default implementation of `Bloq.pretty_name` to go via `__str__`. Start removing `pretty_name` implementations and moving them to `__str__` or `wire_symbols` as necessary.

towards #791


qualtran/bloqs/factoring/ecc/ec_phase_estimate_r.py
```diff
@@ -70,11 +70,8 @@ def build_composite_bloq(self, bb: 'BloqBuilder', x: Soquet, y: Soquet) -> Dict[
     def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
         return {(ECAddR(n=self.n, R=self.point), self.n), (MeasureQFT(n=self.n), 1)}
 
-    def pretty_name(self) -> str:
-        point_str = str(self.point)
-        if len(point_str) < 5 or True:
-            return f'PE${point_str}$'
-        return 'PE(R)'
+    def __str__(self) -> str:
```
dstrain115, line 73: Any concerns about dropping the long string condition?

dstrain115 (approved):

qualtran/bloqs/factoring/ecc/ec_phase_estimate_r.py
mpharrigan: if you look closely, this seems to actually have been some debug code that snuck through (mea culpa) because of the `or True` condition.

mpharrigan:
Test failure is due to different precision of string representations of floating point values locally vs ci. I can choose an arbitrary precision, but then the format string fails for sympy objects

PR #1345, opened 2024-08-26
Fix invocation of `PrepareIdentity` in block encodings

charlesyuan314:
Use `PrepareIdentity.from_bitsizes` to ensure that the signature contains `Register`s rather than `QAny`s.


fdmalone:
.... How was this not causing test failures / mypy to complain lol

qualtran/bloqs/block_encoding/chebyshev_polynomial_test.py
```diff
@@ -221,6 +221,7 @@ def test_chebyshev_matrix():
```
fdmalone, line 221: TestBlockEncoding in this file is still passing QAny rather than a tuple of registers

fdmalone (approved):
Double check you got them all. Would a walk operator test for one of these block encodings have caught this bug too (as it would use the signal state explicitly). It's a bit unfortunate the the auto tester isn't catching these during the decompose test.

fdmalone:
> .... How was this not causing test failures / mypy to complain lol

no test failures makes sense, the lack of mypy errors is surprising.

charlesyuan314:
> Would a walk operator test for one of these block encodings have caught this bug too (as it would use the signal state explicitly).

Yes, what is necessary is for the test to explicitly invoke the decomposition of the signal state.

fdmalone:
Do you know why mypy wasn't catching this?

charlesyuan314:
No, I don't. I suspect that mypy is insufficiently strict with attributes of `attr.frozen`.

fdmalone:
Very troubling.

Issue #1346, opened 2024-08-26
Make `And` a leaf bloq


mpharrigan:
Specific changes that are needed to do this part of #873 :

 - Any unit test that relies on the "sigma" return value of `call_graph` will change. This is brittle to begin with. Need to port to `QECGatesCost` 
 - Cirq simulation needs the cirq-style decomposition unless we override `_kraus_` with custom kraus operators. 
 - Shim over the `_t_complexity_` values for the time being.

mpharrigan:
note: the `keep` argument can be removed from the unit tests in `permutation_test.py` after this is done

mpharrigan:
There is now a flag called `FLAG_AND_AS_LEAF` which -- when set to True -- will turn `And` into a leaf bloq. There is a method called `to_clifford_t_circuit()` which returns a cirq circuit. 

I've kept the `decompose_from_registers` method to support cirq simulation, because it's not straightforward to override the `_kraus_` method

fdmalone:
@mpharrigan can you remind me what the annoying chemistry things were?

mpharrigan:
I don't think "annoying" is the right word; but many of the chemistry tests define a reference toffoli count, multiply it by 4, and assert that it matches deduced T counts. This is usually done via the `sigma` return value of `bloq.call_graph()`. Changing the status of `And` changes the contents of `sigma`. The tests need to be re-worked to compare the toffoli counts directly via new sigma values or porting to `QECGatesCost`.

fdmalone:
@mpharrigan is there something else I can do here?

anurudhp:
@fdmalone are you working on this? If not I have some wip code that i'd written for my other PR. I'm happy to polish that and make a PR

fdmalone:
Go for it, I was mainly focused on cleaning up the tech debt of the
chemistry bloqs, let me know if you hit anything.

mpharrigan:
I'm not sure what else needs doing. If tests pass after setting / removing the `FLAG_AND_AS_LEAF` then it's probably pretty close to being done-done

PR #1347, opened 2024-08-26
First preparations for making And an atomic/leaf bloq

mpharrigan:
Initial steps towards #1346 

I can only do so many of these changes in an afternoon, so all the changes in this PR make tests work for either `And` as leaf or non-leaf; and I put the leafiness behind a flag.


fdmalone (approved):

fdmalone:
That test failure is a bit confusing... is it secretly a float or something, the amount printed is 276

mpharrigan:
The proximate cause was changing the return value of classify bloqs from defaultdict to a dictionary. The "other" key didn't actually exist -- everything is categorized now. Idk if we should customize the categorization scheme to combine the toffolis and multi_control whatever into an "other" category so they match. I took the liberty of changing the notebook code to print the union of keys in the paper costs, binned costs (and default to zero)

PR #1350, opened 2024-08-27
Fix `selection_registers` in `qubitization_walk_test`

charlesyuan314:
Fix a typing problem that is causing the [nightly tests to fail](https://github.com/quantumlib/Qualtran/actions/runs/10576856467/job/29303562603).


charlesyuan314:
@fdmalone Would it make sense to add a non-slow test that catches the same behavior as the failing slow test?

fdmalone (approved):

fdmalone:
hmmm in this case I think this is a vestige of an older qualtran (returning a Signature and not a tuple of Registers). I'm not sure what we would be testing...

PR #1355, opened 2024-08-28
HammingWeightCompute: replace ArbitraryClifford with CNOT, support symbolic bitsizes

anurudhp:


qualtran/bloqs/arithmetic/hamming_weight.py
```diff
@@ -102,13 +118,6 @@ def decompose_from_registers(
...
-        return {(And(), num_and), (ArbitraryClifford(n=2), num_clifford)}
-
-    def __pow__(self, power: int):
-        if power == 1:
-            return self
-        if power == -1:
-            return self.adjoint()
-        raise NotImplementedError("HammingWeightCompute.__pow__ defined only for +1/-1.")
+        num_and = self.junk_bitsize
+        num_clifford = num_and * 5 + self.bit_count_of_bitsize
```
tanujkhattar, line 122: nit: 
```suggestion
        num_cnot = num_and * 5 + self.bit_count_of_bitsize
```

qualtran/bloqs/arithmetic/hamming_weight.py
```diff
@@ -63,6 +64,21 @@ def signature(self):
...
+
+    @cached_property
+    def out_bitsize(self) -> SymbolicInt:
+        return bit_length(self.bitsize)
+
+    @cached_property
+    def bit_count_of_bitsize(self) -> SymbolicInt:
+        """lower bound on number of 1s in bitsize"""
+        if is_symbolic(self.bitsize):
+            return 1  # worst case
```
tanujkhattar, line 79: Can you open an issue, and link it here, with a request that we should add support for custom sympy functions so we don't need to assume a worst case bound here? Ideally, we would define a `bit_count(x)` function in `symbolics/` and report the costs in terms of `bit_count(N)`. When users have values of `N`, one could substitute it in the symbolic cost expression and `bit_count(N)` would evaluate to a number as expected.

tanujkhattar (approved):
LGTM % nits

PR #1356, opened 2024-08-28
Change return type of build_call_graph to set | dict

dstrain115:
- Changes build_call_graph to return a Union of Dict[Bloq, Union[int, Expr]] and Set[BloqCountT]
- Adds checks in main call sites to check both cases.


qualtran/_infra/controlled.py
```diff
@@ -396,7 +395,10 @@ def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
                 f"Could not build call graph for {self}: {e2}"
             ) from e2
 
-        return {(bloq.controlled(self.ctrl_spec), n) for bloq, n in sub_cg}
+        if isinstance(sub_cg, dict):
+            return {(bloq.controlled(self.ctrl_spec), n) for bloq, n in sub_cg.items()}
+        else:
+            return {(bloq.controlled(self.ctrl_spec), n) for bloq, n in sub_cg}
```
mpharrigan, line 402: maybe err on the side of returning the new-style dict rather than casting everything to old-style set

qualtran/resource_counting/_call_graph.py
```diff
@@ -69,15 +70,19 @@ def build_cbloq_call_graph(cbloq: CompositeBloq) -> Set[BloqCountT]:
...
 
     This calls `generalizer` on each of the callees returned from that function,
     and filters out cases where `generalizer` returns `None`.
     """
     callee_counts: Dict[Bloq, Union[int, sympy.Expr]] = defaultdict(lambda: 0)
-    for callee, n in raw_callee_counts:
+    if isinstance(raw_callee_counts, dict):
+        raw_callee_iterator: Iterable[BloqCountT] = raw_callee_counts.items()
+    else:
+        raw_callee_iterator = raw_callee_counts
```
mpharrigan, line 84: this would be a good place to emit a deprecation warning if a set is encountered

mpharrigan (commented):
this is generally what I'm looking for 👍

qualtran/_infra/controlled.py
dstrain115: Done.

qualtran/resource_counting/_call_graph.py
dstrain115: Do you think we should wait on this until migration of most of the bloqs is complete?
I can add this, but it might be a bit noisy for a while.

qualtran/_infra/adjoint.py
```diff
@@ -158,9 +158,17 @@ def adjoint(self) -> 'Bloq':
...
         """The call graph takes the adjoint of each of the bloqs in `subbloq`'s call graph."""
-        return {(bloq.adjoint(), n) for bloq, n in self.subbloq.build_call_graph(ssa=ssa)}
+        sub_cg = self.subbloq.build_call_graph(ssa=ssa)
+        if isinstance(sub_cg, dict):
+            return {bloq.adjoint(): n for bloq, n in sub_cg.items()}
+        else:
+            rtn: 'BloqCountDictT' = {}
+            for bloq, n in sub_cg:
+                adj = bloq.adjoint()
+                rtn[adj] = n + rtn.get(adj, 0)
```
anurudhp, line 170: You can use `Counter`/`defaultdict(lambda: 0)` here instead of `dict`, makes the code much cleaner.

```py
counts = Counter[Bloq]()
for bloq, n in sub_cg:
    counts[bloq.adjoint()] += n
```

qualtran/resource_counting/_call_graph.py
mpharrigan: we can wait

qualtran/bloqs/chemistry/df/prepare_test.py
```diff
@@ -50,10 +50,12 @@ def test_outerprep_t_counts():
         num_aux, num_bits_state_prep=num_bits_state_prep, num_bits_rot_aa=num_bits_rot_aa
     )
     toff = get_cost_value(outer_prep, QECGatesCost()).total_t_and_ccz_count()['n_ccz']
+    print(toff)
```
mpharrigan, line 53: debug prints

qualtran/_infra/controlled.py
```diff
@@ -396,7 +395,14 @@ def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
...
             ) from e2
 
-        return {(bloq.controlled(self.ctrl_spec), n) for bloq, n in sub_cg}
+        if isinstance(sub_cg, dict):
+            return {bloq.controlled(self.ctrl_spec): n for bloq, n in sub_cg.items()}
+        else:
+            rtn: 'BloqCountDictT' = {}
+            for bloq, n in sub_cg:
+                controlled = bloq.controlled(self.ctrl_spec)
+                rtn[controlled] = n + rtn.get(controlled, 0)
```
mpharrigan, line 404: consider using Counter here too

mpharrigan (approved):
consider using Counter, and remove the print statements; but after that lgtm

qualtran/_infra/adjoint.py
dstrain115: Good idea, done.

qualtran/bloqs/chemistry/df/prepare_test.py
dstrain115: Removed

qualtran/_infra/controlled.py
dstrain115: Done.

qualtran/resource_counting/_call_graph.py
```diff
@@ -16,14 +16,15 @@
...
-from typing import Callable, cast, Dict, List, Optional, Sequence, Set, Tuple, Union
+from typing import Callable, cast, Dict, Iterable, List, Optional, Sequence, Set, Tuple, Union
 
 import networkx as nx
 import sympy
 
 from qualtran import Bloq, CompositeBloq, DecomposeNotImplementedError, DecomposeTypeError
 
 BloqCountT = Tuple[Bloq, Union[int, sympy.Expr]]
+BloqCountDictT = Dict[Bloq, Union[int, sympy.Expr]]
```
anurudhp, line 27: We could perhaps use `Mapping` here instead of `dict`? As `build_call_graph` is an internal function, would make it easier for bloq authors to return any mapping (eg. `Counter` instead of casting to `dict` for mypy).

dstrain115: Good idea, done.

qualtran/_infra/adjoint.py
```diff
@@ -158,9 +159,16 @@ def adjoint(self) -> 'Bloq':
...
         """The call graph takes the adjoint of each of the bloqs in `subbloq`'s call graph."""
-        return {(bloq.adjoint(), n) for bloq, n in self.subbloq.build_call_graph(ssa=ssa)}
+        sub_cg = self.subbloq.build_call_graph(ssa=ssa)
+        if isinstance(sub_cg, set):
+            counts = Counter['Bloq']()
+            for bloq, n in sub_cg:
+                counts[bloq.adjoint()] += n
+            return counts
+        else:
+            return {bloq.adjoint(): n for bloq, n in sub_cg.items()}
```
anurudhp, line 171: Should we be worried about a situation where different bloqs return the same `adjoint()`?  I think it should never happen, but am not so sure. (same concern for `bloq.controlled()`).

@mpharrigan @tanujkhattar?

anurudhp (approved):
one minor concern about uniqueness of adjoints/controlleds, otherwise lgtm!

qualtran/_infra/adjoint.py
mpharrigan: that's an interesting corner case to consider; and the type of thing we need to be concerned about now that we are being strict about having only unique bloqs in the callees. I don't _think_ it should happen. I defer to @dstrain115 on whether he thinks it's worth defending against that vs blocking progress on the overall migration

dstrain115: I've added the Counter for both sets and dicts, which should alleviate the concern.  Will push tomorrow unless there are any further issues.

Issue #1318, opened 2024-08-20
`bloq_is_clifford` treats `MultiTargetCNOT` as a single clifford gate


anurudhp:
https://github.com/quantumlib/Qualtran/blob/683472a25a1bafb64fd5e5c58f79405737a4afd4/qualtran/resource_counting/classify_bloqs.py#L140

Is this the expected behaviour due to some possibly optimized implementation?  

@tanujkhattar @mpharrigan

mpharrigan:
instead of $n$ cliffords? 

The notion of clifford counting is pretty poorly defined. In the surface code, you can do a multi-target cnot in 1 step (d cycles) -- the same speed as any other single clifford gate

anurudhp:
Ah! I think we should document this in the code clearly.

mpharrigan:
The old t complexity code counts all the constituent CNOTs. Since it _has_ a decomposition (isn't an atomic/leaf bloq) maybe we should keep counting them. However: there's no support for symbolic bitsize, which causes tests to fail. This bloq could probably do with a re-write to make it more readable and including a `build_call_graph`

wjhuggins:
This is an example where I think we would benefit from counting things in terms of qubits * surface code cycles, as advocated in #1454.

PR #1359, opened 2024-08-28
[QECGatesCost] Port everything to QECGatesCost

mpharrigan:
This changes the default implementation of `t_complexity` to go via `QECGatesCost`. This builds on the prior changes #1313 #1323 #1333 and the even longer history of bloq-ifying the t_complexity protocol

Specifically in this PR there are

 - some changes to get matching clifford counts (t & rotation counts were already settled)
 - A flag to `QECGatesCost` that can toggle "legacy" behavior for exact matching of counts. You can read the code to see exactly what this is
 - Converters to the "legacy" `TComplexity` object.

With these changes: all bloq examples match their T complexities. All tests pass. 

Next steps will be to remove all `_t_complexity_` overrides. This should be seamless, as I did most of the testing without the legacy shim to priorities `_t_complexity_` override and fixed most of the issues (usually by moving the costs to `my_static_costs`). 

Pending resolution of #1318, we can change the way MultiTargetCNOT is handled. In this PR: the legacy flag makes it count as multiple cliffords; otherwise the "new" behavior of counting it as one object is maintained. If we want it to always count as multiple, then a symbolic call graph is needed.


qualtran/cirq_interop/t_complexity_protocol.py
```diff
@@ -236,6 +239,9 @@ def _t_complexity_for_bloq(bloq: Bloq) -> Optional[TComplexity]:
     return _t_complexity_from_strategies(bloq, strategies)
 
 
+USE_NEW_GATE_COUNTING_FLAG = True
```
fdmalone, line 242: Is this a temporary thing?

qualtran/drawing/bloq_counts_graph.py
```diff
@@ -241,6 +241,7 @@ def format_qec_gates_cost(cls, val: 'GateCounts', agg: Optional[str] = None) ->
             'and_bloq': 'Ands',
             'clifford': 'Cliffords',
             'rotation': 'Rotations',
+            'rotations': 'Rotations',
```
fdmalone, line 244: rotation and rotations map to Rotations correct? There's no extra s on the previous Rotation

qualtran/resource_counting/_bloq_counts.py
```diff
@@ -209,6 +211,38 @@ def total_t_and_ccz_count(self, ts_per_rotation: int = 11) -> Dict[str, Symbolic
...
 
+    def to_legacy_t_complexity(
+        self,
+        ts_per_toffoli: int = 4,
+        ts_per_cswap: int = 7,
+        ts_per_and_bloq: int = 4,
+        cliffords_per_and_bloq: int = 9,
+        cliffords_per_cswap: int = 10,
+    ) -> 'TComplexity':
+        """Return a legacy `TComplexity` object.
```
fdmalone, line 222: nit: should have Args in docstring no?

qualtran/resource_counting/_bloq_counts.py
```diff
@@ -262,14 +314,35 @@ def compute(self, bloq: 'Bloq', get_callee_cost: Callable[['Bloq'], GateCounts])
...
+            if self.legacy_shims:
+                return GateCounts(and_bloq=1, clifford=2 * n_inverted_controls)
+            else:
+                return GateCounts(and_bloq=1)
 
         # CSwaps aka Fredkin
         if isinstance(bloq, TwoBitCSwap):
             return GateCounts(cswap=1)
 
+        if isinstance(bloq, MultiTargetCNOT):
```
fdmalone, line 339: Link to issue?

fdmalone (approved):

qualtran/cirq_interop/t_complexity_protocol.py
mpharrigan: yes

qualtran/drawing/bloq_counts_graph.py
mpharrigan: this matches all the field names in the various "aggregated" formats. Since I added support for `TComplexity`, we need the with-an-s version as well since its `TComplexity.rotations` (although the other fields don't have the s 🤷‍♀️ )

qualtran/cirq_interop/t_complexity_protocol.py
mpharrigan: follow-up PR will remove the vestigial code in this file, this flag, and `_t_comlpexity_` overrides

qualtran/bloqs/basic_gates/rotation_test.py
```diff
@@ -20,13 +20,28 @@
 from qualtran._infra.gate_with_registers import get_named_qubits
 from qualtran.bloqs.basic_gates import CZPowGate, Rx, Ry, Rz, XPowGate, YPowGate, ZPowGate
 from qualtran.bloqs.basic_gates.rotation import _rx, _ry, _rz
+from qualtran.resource_counting import GateCounts, get_cost_value, QECGatesCost
+from qualtran.resource_counting.classify_bloqs import bloq_is_rotation, bloq_is_t_like
 
 
 def test_rotation_gates():
```
anurudhp, line 27: rename this to `test_rotation_with_t_angle` or sth?

anurudhp (approved):

qualtran/bloqs/basic_gates/rotation_test.py
mpharrigan: done

qualtran/resource_counting/_bloq_counts.py
mpharrigan: done

PR #1369, opened 2024-08-30
Support more symbolic decompositions

anurudhp:
- permutation bloqs: decompose for symbolic `N`
- stateprep rotation - check phase bitsize, raise `DecomposeTypeError` when not possible.

### DIFF
--- a/qualtran/bloqs/arithmetic/permutation_test.py
+++ b/qualtran/bloqs/arithmetic/permutation_test.py
@@ -27,6 +27,7 @@ import numpy as np
 import pytest
 import sympy
 
+import qualtran.testing as qlt_testing
 from qualtran import QBit
 from qualtran.bloqs.arithmetic.permutation import (
     _permutation,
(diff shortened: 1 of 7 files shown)


tanujkhattar (approved):

tanujkhattar:
I think it'll be nice to add some tests, but LGTM overall.

mpharrigan:
yeah, this could use some symbolic `bloq_example`s

anurudhp:
Added some partially symbolic bloq examples along with a `assert_valid_decomposition` test for them

qualtran/bloqs/arithmetic/permutation_test.py
```diff
@@ -44,16 +48,38 @@
...
+        _permutation,
+        _sparse_permutation,
+        _permutation_cycle_symb_N,
+        _permutation_symb_with_cycles,
+        _sparse_permutation_with_symbolic_N,
+    ],
+    ids=lambda bloq_ex: bloq_ex.name,
+)
+def test_decomposition(bloq_ex):
+    qlt_testing.assert_valid_bloq_decomposition(bloq_ex.make())
```
mpharrigan, line 82: this is redundant with the `bloq_autotester`

anurudhp: I want to verify that the bloq actually decomposes (i.e. without raising a `DecomposeTypeError`). `bloq_autotester` treats `DecomposeTypeError` as a success right?

mpharrigan: oh good point. (technically it would report it as a skip)

PR #1373, opened 2024-09-01
Flatten nested controls in `ControlledViaAnd`

anurudhp:
Override `ControlledViaAnd.get_ctrl_system` so that `ControlledViaAnd(bloq, spec1).controlled(spec2)` now returns `ControlledViaAnd(bloq, spec)` where `spec` is the sequence of `spec1, spec2`.


qualtran/_infra/gate_with_registers.py
```diff
@@ -371,8 +371,9 @@ def __pow__(self, power: int) -> 'GateWithRegisters':
             return Power(bloq, abs(power))
         raise NotImplementedError(f"{self} does not implemented __pow__ for {power=}.")
 
+    @classmethod
     def _get_ctrl_spec(
-        self,
+        cls,
```
mpharrigan, line 376: where is this called? doesn't the call site need to be updated?

qualtran/bloqs/mcmt/controlled_via_and.py
```diff
@@ -126,6 +126,29 @@ def build_call_graph(self, ssa: 'SympySymbolAllocator') -> set['BloqCountT']:
 
         return set(counts.items())
 
+    def get_ctrl_system(self, ctrl_spec: 'CtrlSpec') -> tuple['Bloq', 'AddControlledT']:
+        """flatten nested controlled bloqs"""
```
mpharrigan, line 130: Either write a complete docstring or remove or relegate to a line comment

mpharrigan (commented):
add a test and ideally a little demo to the doc notebook.

also can you add a description to the PR

qualtran/_infra/gate_with_registers.py
anurudhp: It is only used in one location, to convert cirq style controls to `CtrlSpec`:
https://github.com/quantumlib/Qualtran/blob/db8b9938d8210024bb6b92942376363586165efc/qualtran/_infra/gate_with_registers.py#L499-L501

I made this a class method so that I can call it in #1348 (as the code doesn't depend on the instance).

qualtran/bloqs/mcmt/controlled_via_and.py
anurudhp: removed

qualtran/_infra/gate_with_registers.py
mpharrigan: can you please update that existing call in `gate_with_registers.py` to call it as a class method, i.e. `GateWithRegisters._get_ctrl_spec` instead of `self.`

mpharrigan (approved):

qualtran/_infra/gate_with_registers.py
anurudhp: done

PR #1375, opened 2024-09-02
Raise `DecomposeTypeError` for symbolic register shapes

### LINKED ISSUE #1374: Attempting to decompose bloq with symbolic shaped registers raises an incorrect `ValueError`
Cause: `decompose_bloq` calls [`_decompose_from_build_composite_bloq`](https://github.com/quantumlib/Qualtran/blob/64acb5280f43071afe7b00769aa30126bc45ba87/qualtran/_infra/bloq.py#L46-L51), which first initializes the registers before calling `build_composite_bloq`.
This raises a `ValueError` when a symbolic register shape is encountered

Causes most of the "Cannot get real-valued shape." issues in #1367 

```py
from attrs import frozen
import sympy
from qualtran import Bloq, Signature, Register, QBit, DecomposeTypeError
from qualtran.bloqs.basic_gates import XGate
from qualtran.symbolics import SymbolicInt, is_symbolic


@frozen
class XGateOnEach(Bloq):
    n: SymbolicInt

    @property
    def signature(self):
        return Signature([Register('q', QBit(), shape=(self.n,))])

    def build_composite_bloq(self, bb, q):
        if is_symbolic(self.n):
            raise DecomposeTypeError(f"cannot decompose symbolic {self}")

        for i in range(self.n):
            q[i] = bb.add(XGate(), q=q[i])

        return {'q': q}


def test_decomposition():
    n = sympy.Symbol("n")
    _ = XGateOnEach(n).decompose_bloq()


test_decomposition()
```

```
  File "qualtran/_infra/registers.py", line 85, in shape
    raise ValueError(f"{self} is symbolic. Cannot get real-valued shape.")
ValueError: Register(name='q', dtype=QBit(), _shape=(n,), side=<Side.THRU: 3>) is symbolic. Cannot get real-valued shape.
```

anurudhp:
fixes #1374

### DIFF
--- a/qualtran/_infra/composite_bloq.py
+++ b/qualtran/_infra/composite_bloq.py
@@ -846,6 +846,8 @@ class BloqBuilder:
             initial, left-dangling soquets for the register. Otherwise, this is a RIGHT register
             and will be used for error checking in `finalize()` and nothing is returned.
         """
+        from qualtran.symbolics import is_symbolic
+
         if not self.add_register_allowed:
             raise ValueError(
                 "This BloqBuilder was constructed from pre-specified registers. "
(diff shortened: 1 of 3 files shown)


mpharrigan (approved):
Thanks

Issue #1251, opened 2024-08-05
Migrate t_complexity to `QECGatesCost` cost key


mpharrigan:
With the goal of #539 there are now two built-in ways of counting gates: one where you have to provide which gates to count and another, more-opinionated one that makes some assumptions about the relevant counts for a surface code / lattice surgery / qec execution model. This latter one is `QECGatesCost` and is similar to the `t_complexity` protocol. These two protocols should be unified.

 - Reduce the amount of dynamacism and magic in existing t_complexity while verifying no regression in t count. This is mostly done, see https://github.com/quantumlib/Qualtran/pull/678 https://github.com/quantumlib/Qualtran/pull/740 https://github.com/quantumlib/Qualtran/pull/803 https://github.com/quantumlib/Qualtran/pull/1015 https://github.com/quantumlib/Qualtran/pull/1012 https://github.com/quantumlib/Qualtran/pull/996 https://github.com/quantumlib/Qualtran/pull/994
 - Split the cirq t-complexity from pure-bloqs t-complexity https://github.com/quantumlib/Qualtran/pull/835
 - Remove remaining `_t_complexity_` overrides and move logic either to `from_directly_countable_bloqs` (probably most of them) or duplicate to the `my_static_costs` generic method otherwise
 - make the `t_complexity` function and method be a wrapper around `QECGateCost` computation; verify identical resource counts

mpharrigan:
https://github.com/quantumlib/Qualtran/pull/1359 makes the default go via QECGateCost

PR #1377, opened 2024-09-03
Remove _t_complexity_

mpharrigan:
Final part of #1251 

This PR removes the legacy `_t_complexity_` override from all bloqs. These were already redundant: during development of previous PRs I had removed use of these anyways. Those PRs did careful accounting of all `bloq_example` and unit test t complexity values and made sure they matched the `QECGatesCost` values (with some small, contained legacy shims). This PR shows that off!

See my inline comments, but I left in a `getattrs`-based shim so that the `_t_complexity_` override will still be used if `QECGatesCost()` is in legacy mode. The only thing left in qualtran that uses it is a specific unit test I added to test it.

### DIFF
--- a/qualtran/resource_counting/_bloq_counts.py
+++ b/qualtran/resource_counting/_bloq_counts.py
@@ -12,7 +12,6 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 import logging
-import warnings
 from collections import defaultdict
 from typing import Callable, cast, Dict, Sequence, Tuple, TYPE_CHECKING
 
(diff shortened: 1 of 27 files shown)


qualtran/resource_counting/_bloq_counts.py
```diff
@@ -288,18 +287,6 @@ def compute(self, bloq: 'Bloq', get_callee_cost: Callable[['Bloq'], GateCounts])
...
-            if legacy_val is not NotImplemented:
-                warnings.warn(
-                    "Please migrate explicit cost annotations to the general "
-                    "`Bloq.my_static_costs` method override.",
-                    DeprecationWarning,
-                )
-                return GateCounts(
-                    t=legacy_val.t, clifford=legacy_val.clifford, rotation=legacy_val.rotations
-                )
-
```
mpharrigan, line 302: Please see that all tests pass without this shim; but I think I'll put this back in (using `getattr(bloq, '_t_complexity_', None)` for a bit of backwards compatibility support

mpharrigan: Latest commit puts this back in

mpharrigan:
End of an era! ptal @tanujkhattar

tanujkhattar (approved):

PR #1383, opened 2024-09-03
Use QECGatesCost for phase_estimation_of_quantum_walk

mpharrigan:
Remove dependence on the `t_complexity_compat` from a (large) example so we make sure the gate costing is always cached. This is a shim. A full solution would be to replace the phase estimation circuit with a phase estimation bloq and remove the helper code from the notebook.


qualtran/bloqs/phase_estimation/phase_estimation_of_quantum_walk.ipynb
```diff
@@ -129,6 +117,44 @@
...
+    "    # Usually, you'd combine this into a bloq of its own, but we\n",
+    "    # use this helper function to add up the costs of the bloqs\n",
+    "    # found within the circuit.\n",
+    "    cost_key = QECGatesCost()\n",
+    "    costs_cache = {}\n",
+    "    total_cost = cost_key.zero()\n",
+    "    for op in circuit.all_operations():\n",
+    "        if not isinstance(op.gate, Bloq):\n",
+    "            # Skip state prep and QFT for now\n",
+    "            continue\n",
```
tanujkhattar, line 148: Maybe add a `print("Skipping {op.gate=}")` to make sure only state prep and QFT are being skipped.

qualtran/bloqs/phase_estimation/phase_estimation_of_quantum_walk.ipynb
```diff
@@ -178,19 +204,7 @@
...
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "from qualtran.bloqs.phase_estimation.qubitization_qpe import _qubitization_qpe_hubbard_model_large\n",
-    "qpe = _qubitization_qpe_hubbard_model_large.make()\n",
-    "%time result = qpe.t_complexity()\n",
```
tanujkhattar, line 193: Why are we removing this? AFAIR, `_qubitization_qpe_hubbard_model_large` actually returns a bloq which is equivalent to the cirq circuit being constructed above and the two t_complexity's should match.

tanujkhattar (approved):
LGTM % nits

qualtran/bloqs/phase_estimation/phase_estimation_of_quantum_walk.ipynb
mpharrigan: yeah I had this when I wrote it but the stateprepchannel repr is very long and this would get printed every time

qualtran/bloqs/phase_estimation/phase_estimation_of_quantum_walk.ipynb
mpharrigan: oh, I didn't realize that. I thought this was just the walk operator or sth. I think this was originally included to make sure the two disparate t_complexity code paths were executed; but now there's only one. We should remove the cirq stuff if we already have this as bloqs and bloq examples. I'll revert and open an issue

PR #1392, opened 2024-09-06
Change build_call_graph in bloqs to return dict

dstrain115:
- This changes the build_call_graph function within bloqs in qualtran to return a dictionary of cost counts rather than a set.
- This will allow the ordering of cost counts to be deterministic

Note that this requires some slight code changes for bloqs that have multiple set items since (Toffoli(), 1) and (Toffoli(), 2) would have two different items in a set, but share an index in the dictionary.

This also may alter counts (i.e. fix a bug) where set items clobber each other.  For instance, adding (Toffoli(), self.bits_a) and (Toffoli(), self.bits_b) will previously give the wrong count if bits_a == bits_b since the two items would be the same in the set.

### DIFF
--- a/qualtran/bloqs/chemistry/pbc/first_quantization/projectile/select_uv.py
+++ b/qualtran/bloqs/chemistry/pbc/first_quantization/projectile/select_uv.py
@@ -80,17 +80,21 @@ class SelectUVFirstQuantizationWithProj(Bloq):
 
     def build_call_graph(self, ssa: 'SympySymbolAllocator') -> 'BloqCountDictT':
         cost = Counter['Bloq']()
+        # tc_p and tc_n
         # C8 and C9, one of the registers of size num_bits_n so need to account for this.
         cost[SignedIntegerToTwosComplement(self.num_bits_p)] += 3
         cost[SignedIntegerToTwosComplement(self.num_bits_n)] += 3
         # Adding nu into p / q. Nu is one bit larger than p.
         cost[Add(QInt(self.num_bits_p + 1))] += 3
         cost[Add(QInt(self.num_bits_n + 1))] += 3
+        # ctrl_add_p and ctrl_add_n
         cost[Toffoli()] += 3 * (self.num_bits_p + 1)
         cost[Toffoli()] += 3 * (self.num_bits_n + 1)
+        # inv_tc_p and inv_tc_n
         # + 2 as these numbers are larger from addition of $\nu$
         cost[SignedIntegerToTwosComplement(self.num_bits_p + 2)] += 3
         cost[SignedIntegerToTwosComplement(self.num_bits_n + 2)] += 3
+        # cost for phase:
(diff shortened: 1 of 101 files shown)


qualtran/bloqs/chemistry/pbc/first_quantization/projectile/select_uv.py
```diff
@@ -77,32 +78,22 @@ def signature(self) -> Signature:
...
-        cost_ctrl_add_n = (Toffoli(), 3 * (self.num_bits_n + 1))
+        cost[Add(QInt(self.num_bits_p + 1))] += 3
+        cost[Add(QInt(self.num_bits_n + 1))] += 3
+        cost[Toffoli()] += 3 * (self.num_bits_p + 1)
+        cost[Toffoli()] += 3 * (self.num_bits_n + 1)
         # + 2 as these numbers are larger from addition of $\nu$
-        cost_inv_tc_p = (SignedIntegerToTwosComplement(self.num_bits_p + 2), 3)
-        cost_inv_tc_n = (SignedIntegerToTwosComplement(self.num_bits_n + 2), 3)
+        cost[SignedIntegerToTwosComplement(self.num_bits_p + 2)] += 3
+        cost[SignedIntegerToTwosComplement(self.num_bits_n + 2)] += 3
```
mpharrigan, line 93: i feel like there may be some information lost by removing those variable names? Not blocking, but maybe include in comments, eg the cost[Toffoli()] += 3 * self.num_bits_p is for `ctrl_add_p` which isn't otherwise obvious

qualtran/bloqs/chemistry/pbc/first_quantization/select_t.py
```diff
@@ -60,15 +60,15 @@ def signature(self) -> Signature:
...
+    def build_call_graph(self, ssa: 'SympySymbolAllocator') -> 'BloqCountDictT':
         # Cost is $5(n_{p} - 1) + 2$ which comes from copying each $w$ component of $p$
         # into an ancilla register ($3(n_{p}-1)$), copying the $r$ and $s$ bit of into an
         # ancilla ($2(n_{p}-1)$), controlling on both those bit perform phase flip on an
         # ancilla $|+\rangle$ state. This requires $1$ Toffoli, Then erase which costs
         # only Cliffords. There is an additional control bit controlling the application
         # of $T$ thus we come to our total.
-        # Eq 73. page
-        return {(Toffoli(), (5 * (self.num_bits_p - 1) + 2))}
+        # Eq 73, page 20.
```
mpharrigan, line 70: nice

mpharrigan (approved):
wow! great

qualtran/bloqs/chemistry/pbc/first_quantization/projectile/select_uv.py
dstrain115: Done.

mpharrigan:
didn't mean to conflict you. I'll try to fix them up quick

PR #1394, opened 2024-09-09
Bugfix in `QROAMCleanAdjointWrapper` to correctly pass `num_controls` to `QROAMCleanAdjoint`

tanujkhattar:
See updated test, which would fail earlier but passes now. 

Also made other drive by improvements like:

1. Fix the logic of `get_optimal_log_block_size_clean_ancilla` so it returns an integer >=0.
2. Pass ` target_bitsizes` to `QROAMCleanAdjoint` from` QROAMCleanAdjointWrapper` so that target bitsize remains the same post delegation when constructing the bloq for data. 
3. Update `find_optimal_log_block_size` implementation for `SelectSwapQROM` to find the optimal bitsize taking into account a factor of 2 for dirty ancillas. Added a test to verify the new block size leads to smaller counts. 
4. Raise a `DecomposeTypeError` for `QROM` decomposition when `self.has_data()` is false. This makes qubit counting protocols deal with symbolic qroms better.

### DIFF
--- a/qualtran/bloqs/data_loading/qroam_clean.py
+++ b/qualtran/bloqs/data_loading/qroam_clean.py
@@ -71,6 +71,8 @@ def get_optimal_log_block_size_clean_ancilla(
         k = log2(qroam_block_size)
     if is_symbolic(k):
         return k
+    if k < 0:
+        return 0
     k_int = np.array([np.floor(k), np.ceil(k)])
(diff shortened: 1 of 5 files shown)


fdmalone (approved):

tanujkhattar:
@fdmalone I have some follow up updates coming to this PR, so don't merge just yet.

mpharrigan (approved):

fdmalone (approved):

PR #1400, opened 2024-09-09
Remove pretty_name part 2

dstrain115:
Change the default implementation of Bloq.pretty_name to go via __str__. Removing pretty_name implementations and moving them to __str__ or wire_symbols as necessary.


mpharrigan (approved):
thank you!

PR #1402, opened 2024-09-10
Remove pretty_name part 3

dstrain115:
Change the default implementation of Bloq.pretty_name to go via str. Removing pretty_name implementations and moving them to str or wire_symbols as necessary.

This change removes all references to pretty_name


fdmalone (approved):

mpharrigan:
nice. can you fix the two notebooks as well?

mpharrigan (changes requested):
Hold off on merging until I cut v0.5

dstrain115:
Ok, notebooks fixed.  Feel free to merge when ready

qualtran/_infra/adjoint.py
```diff
@@ -167,10 +167,6 @@ def build_call_graph(self, ssa: 'SympySymbolAllocator') -> 'BloqCountDictT':
                 counts[bloq.adjoint()] += n
         return counts
 
-    def pretty_name(self) -> str:
-        """The subbloq's pretty_name with a dagger."""
-        return self.subbloq.pretty_name() + '†'
-
     def __str__(self) -> str:
         """Delegate to subbloq's `__str__` method."""
         return f'Adjoint(subbloq={str(self.subbloq)})'
```
anurudhp, line 172: could we use `str(subbloq) + dagger` in `__str__` instead of `Adjoint(subloq=...)`? It would make the diagrams easier to read.

dstrain115: Done.

anurudhp (approved):

mpharrigan (approved):

qualtran/_infra/adjoint.py
mpharrigan: +1, I wanted to do this for a while; but feared the test changes that would need to accompany it

PR #1406, opened 2024-09-12
Implement decomposition of ECAddR, ECWindowAddR, and ECAdd bloqs

fpapa250:
Implements decomposition of ECAddR and ECPhaseEstimateR to match that of https://arxiv.org/abs/2306.08585.

Removes MeasureQFT() bloq in favor of an actual QFT() bloq and a shimmed in Measure() bloq.


qualtran/bloqs/factoring/ecc/ec_add_r.py
```diff
@@ -67,6 +84,49 @@ def signature(self) -> 'Signature':
...
+
+        bb.add(Free(QUInt(self.n)), reg=a)
+        bb.add(Free(QUInt(self.n)), reg=b)
+        bb.add(Free(QUInt(self.n)), reg=lam_r)
+
+        return {'ctrl': ctrl, 'x': x, 'y': y}
+
+    def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
+        ctrl_spec = CtrlSpec(qdtypes=QBit(), cvs=(1))
+        return {(ECAdd(self.n, self.R.mod).controlled(ctrl_spec=ctrl_spec), 1)}
```
mpharrigan, line 128: I think you can just do `.controlled()` .. these are all default args

qualtran/bloqs/factoring/ecc/ec_add_r.py
```diff
@@ -67,6 +84,49 @@ def signature(self) -> 'Signature':
...
+            XorK(QUInt(bitsize=self.n), self.R.y).controlled(ctrl_spec=ctrl_spec), ctrl=ctrl, x=b
+        )
+
+        bb.add(Free(QUInt(self.n)), reg=a)
+        bb.add(Free(QUInt(self.n)), reg=b)
+        bb.add(Free(QUInt(self.n)), reg=lam_r)
+
+        return {'ctrl': ctrl, 'x': x, 'y': y}
+
+    def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountT']:
```
mpharrigan, line 126: although this is still supported for backwards compatibility, we've transitioned to returning a dictionary for this method. can you please update

mpharrigan:
tests?

fpapa250:
> tests?

I added a classical simulation test for ECAddR(). Any recommendations on what tests I should add for ECPhaseEstimateR()?

Also, for the call graph should I include the XorK() bloqs in the counts? I notice you don't include the IntState counts and this is kind of a roundabout way of initializing an intstate in a controlled manor. I can add call graph tests once that is sorted (and once I fix the .controlled() bug on the call graph).

qualtran/bloqs/factoring/ecc/ec_add.py
```diff
@@ -33,6 +37,8 @@ class ECAdd(Bloq):
     Args:
         n: The bitsize of the two registers storing the elliptic curve point
         mod: The modulus of the field in which we do the addition.
+        curve_a: The $a$ coefficient of an elliptic curve given in the standard form of
```
mpharrigan, line 40: does this actually affect the decomposition?

This is a little confusing. In litinski they use (x, y) + (a, b) for the curve points and c1, c2 for the curve coefficients. Other references use (x1, y1) + (x2, y2) and a, b for the curve points. Maybe we should switch to using this second scheme.

qualtran/bloqs/factoring/ecc/ec_add_r.py
```diff
@@ -67,6 +85,41 @@ def signature(self) -> 'Signature':
...
+        ctrl, b = bb.add(XorK(QUInt(bitsize=self.n), self.R.y).controlled(), ctrl=ctrl, x=b)
+
+        bb.add(Free(QUInt(self.n)), reg=a)
+        bb.add(Free(QUInt(self.n)), reg=b)
+        bb.add(IntEffect(bitsize=self.n, val=lam), val=lam_r)
+
+        return {'ctrl': ctrl, 'x': x, 'y': y}
+
+    def build_call_graph(self, ssa: 'SympySymbolAllocator') -> Set['BloqCountDictT']:
+        return {(ECAdd(self.n, self.R.mod, self.R.curve_a).controlled(), 1)}
```
mpharrigan, line 121: please update to return a dictionary instead of a set

qualtran/bloqs/factoring/ecc/ec_add_r.py
```diff
@@ -67,6 +85,41 @@ def signature(self) -> 'Signature':
...
+        lam_denom = (2 * self.R.y) % self.R.mod
+
+        lam = (lam_num * pow(lam_denom, -1, mod=self.R.mod)) % self.R.mod
+        lam_r = bb.add(IntState(bitsize=self.n, val=lam))
+
+        a, b, x, y, lam_r = bb.add(
+            ECAdd(self.n, self.R.mod, self.R.curve_a), a=a, b=b, x=x, y=y, lam=lam_r
+        )
+
+        # Call graph doesn't like the controlled() operator on arbitrary clifford.
```
mpharrigan, line 110: what does this mean? does this decomposition not work?

qualtran/bloqs/factoring/ecc/ec_add_r.py
```diff
@@ -67,6 +85,41 @@ def signature(self) -> 'Signature':
             [Register('ctrl', QBit()), Register('x', QUInt(self.n)), Register('y', QUInt(self.n))]
         )
 
+    def build_composite_bloq(
```
mpharrigan, line 88: where does this decomposition come from? it does not match the references.

qualtran/bloqs/factoring/ecc/ec_phase_estimate_r.py
```diff
@@ -64,11 +66,21 @@ def build_composite_bloq(self, bb: 'BloqBuilder', x: Soquet, y: Soquet) -> Dict[
         for i in range(self.n):
             ctrl[i], x, y = bb.add(ECAddR(n=self.n, R=2**i * self.point), ctrl=ctrl[i], x=x, y=y)
 
-        bb.add(MeasureQFT(n=self.n), x=ctrl)
+        ctrls = bb.join(np.array(ctrl), dtype=QUInt(bitsize=self.n))
+        ctrls = bb.add(QFTTextBook(bitsize=self.n), q=ctrls)
+        ctrl = bb.split(ctrls)
```
mpharrigan, line 71: I think we should keep `MeasureQFT` factored out. 1) makes the circuit cleaner at high level of abstraction 2) we'd want to actually do the semi-classical qft which sortof requires that you're measuring it.

You can just implement the decomposition for MeasureQFT as the code you've inlined here

mpharrigan (commented):
Let's also add a concrete bloq example for ECAdd, which I think will trigger the decomposition bug (?) You can also use this bloq_example to write a minimal unit test that you can indeed `bloq.call_classically(...)` although you won't be able to assert anything particularly insightful about the results

qualtran/bloqs/factoring/ecc/ec_add.py
fpapa250: I no longer need curve_a for the classical simulation, I can get it from a, b, and lambda_r. I think a and b for the curve coefficients is much more clear than c1, c2. I can swap from (x,y) + (a,b) to (x1,y1) + (x2,y2), but I think (x, y) + (a, b) is easier to understand. I think since we use curve_a instead of just a it is clear that it is different. I'll defer judgement to you on the naming convention and I can change it if you think that's better.

qualtran/bloqs/factoring/ecc/ec_add_r.py
fpapa250: Fixed bug in decomposition that was triggering this I think it's okay now.

qualtran/bloqs/factoring/ecc/ec_add_r.py
fpapa250: I'm a little confused myself on whether ECAddR and ECWindowAddR are necessary distinctions. Is ECAddR not just supposed to be a single window ECWindowAddR? In that case we can make it one bloq with a var window_size that changes decomposition. Otherwise I couldn't find a special decomposition for the ECAddR circuit so I put what I imagined it would be (initialize a,b from R.x,R.y; ECPointAdd; uninitialize a,b;).

fpapa250:
@mpharrigan I'll spend some time today finishing the PR (more tests and fix the call graph for ECAdd). I was wondering if you think ECAdd should be split into sub-bloqs since the decomposition is long and messy?

NoureldinYosri:
@fpapa250 can you break down this PR into smaller pieces? e.g. "Add Equals", "support serialization of ECPoint", ...etc

qualtran/bloqs/factoring/ecc/ec_add.py
mpharrigan: point taken on `curve_a` but we'll probably want to write latex expressions in some places and then $a$ will be ambiguous.

qualtran/bloqs/factoring/ecc/ec_add_r.py
mpharrigan: I think we chatted in person, but the Roetteler / q# reference has a different circuit for this. What do you want to do here? I think we discussed having the "main call graph" go via ECWindowAddR. We'll keep the reference to the special circuit for adding a constant (without additional register allocation) but not use it

mpharrigan:
Can you please link all the sub-PRs that you broke this into for the historical record

fpapa250:
Broke into the following PRs:
https://github.com/quantumlib/Qualtran/pull/1411
https://github.com/quantumlib/Qualtran/pull/1412
https://github.com/quantumlib/Qualtran/pull/1424
https://github.com/quantumlib/Qualtran/pull/1425

Will make separate new PR for ECWindowAddR

fpapa250:
One more PR:

https://github.com/quantumlib/Qualtran/pull/1477

Now that they are all merged this PR can be closed as all of it's pieces are merged.

mpharrigan:
wooooo

PR #1408, opened 2024-09-17
Create linear half comparison bloqs

NoureldinYosri:
This PR creates all versions of half a comparison operation in $n$ toffoli complexity. The uncomputation of these bloqs has zero cost.

The half comparason operation is needed in the modular inversion bloq.

I also modify the OutOfPlaceAdder to allow it to not compute the extra And when it's not needed

---

This PR also shows how we can implement all versions of comparison by implementing just one. for example creating a half greater than bloq that use logarithmic depth is enough to implement all the others.


qualtran/bloqs/arithmetic/addition.py
```diff
@@ -272,6 +272,7 @@ class OutOfPlaceAdder(GateWithRegisters, cirq.ArithmeticGate):  # type: ignore[m
 
     bitsize: 'SymbolicInt'
     is_adjoint: bool = False
+    include_most_significant_bit: bool = True
```
mpharrigan, line 275: document

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -280,7 +281,9 @@ def signature(self):
             [
                 Register('a', QUInt(self.bitsize)),
                 Register('b', QUInt(self.bitsize)),
-                Register('c', QUInt(self.bitsize + 1), side=side),
+                Register(
+                    'c', QUInt(self.bitsize + int(self.include_most_significant_bit)), side=side
```
mpharrigan, line 285: update Registers: doc as well

mpharrigan: consider doing ` + (1 if self.include_most_significant_bit else 0)` for explicitness

qualtran/bloqs/arithmetic/addition.py
```diff
@@ -328,12 +336,22 @@ def decompose_from_registers(
                 cirq.CX(a[i], c[i + 1]),
                 cirq.CX(b[i], c[i]),
             ]
-            for i in range(self.bitsize)
+            for i in range(self.bitsize - 1 + int(self.include_most_significant_bit))
         ]
+        if not self.include_most_significant_bit:
```
mpharrigan, line 341: a comment would be useful here

qualtran/bloqs/arithmetic/comparison_test.py
```diff
@@ -378,3 +387,99 @@ def test_clineardepthgreaterthan_tcomplexity(ctrl, dtype):
...
+        LinearDepthHalfGreaterThanEqual,
+        LinearDepthHalfLessThan,
+        LinearDepthHalfLessThanEqual,
+    ],
+)
+@pytest.mark.parametrize('dtype', [QInt, QUInt, QMontgomeryUInt])
+@pytest.mark.parametrize('bitsize', range(2, 5))
+def test_linear_half_comparison_classical_action(comp_cls, dtype, bitsize):
+    b = comp_cls(dtype(bitsize))
+    if dtype is QInt:
```
mpharrigan, line 439: dtype.get_classical_domain() ?

qualtran/bloqs/arithmetic/comparison.py
```diff
@@ -1154,3 +1156,423 @@ def _clineardepthgreaterthan_example() -> CLinearDepthGreaterThan:
 _CLinearDepthGreaterThan_DOC = BloqDocSpec(
     bloq_cls=CLinearDepthGreaterThan, examples=[_clineardepthgreaterthan_example]
 )
+
+_SUPPORTED_DTYPES = Union[QInt, QUInt, QMontgomeryUInt]
```
mpharrigan, line 1160: this is only used in one place (?) consider un-factoring out. Just put it inline

qualtran/bloqs/arithmetic/comparison.py
```diff
@@ -1154,3 +1156,423 @@ def _clineardepthgreaterthan_example() -> CLinearDepthGreaterThan:
...
+                SignExtend(self.dtype, QInt(self.dtype.bitsize + 1)).adjoint(): 2,
+            }
+        if self.comparison_dtype in ('>=', '<='):
+            extra_ops[XGate()] = 1
+        adder = self._half_greater_than_bloq
+        if self.uncompute:
+            adder = adder.adjoint()
+        adder_call_graph = adder.build_call_graph(ssa)
+        assert isinstance(adder_call_graph, dict)
+        counts = {**adder_call_graph}
```
mpharrigan, line 1418: defaultdict with default of 0 so you can just add the extra counts?

qualtran/bloqs/arithmetic/comparison.py
```diff
@@ -1154,3 +1156,423 @@ def _clineardepthgreaterthan_example() -> CLinearDepthGreaterThan:
...
+                Register('target', QBit(), side=side),
+            ]
+        )
+
+    def adjoint(self) -> '_HalfComparisonBase':
+        return attrs.evolve(self, uncompute=self.uncompute ^ True)
+
+    @cached_property
+    @abc.abstractmethod
+    def _half_greater_than_bloq(self) -> Bloq:
```
mpharrigan, line 1326: why can't this just be an attribute of `_HalfComparison[Base]`

qualtran/bloqs/arithmetic/comparison.py
```diff
@@ -1154,3 +1156,423 @@ def _clineardepthgreaterthan_example() -> CLinearDepthGreaterThan:
...
+            adder = adder.adjoint()
+        counts[adder] = 1
+
+        return counts
+
+
+@frozen
+class _HalfComparisonBase(Bloq):
+    dtype: _SUPPORTED_DTYPES
+    comparison_dtype: str = attrs.field(
```
mpharrigan, line 1304: comparison data type? This isn't a data type, right? It's the comparison operation. Consider renaming to `comparison_op`

qualtran/bloqs/arithmetic/comparison.py
```diff
@@ -1154,3 +1156,423 @@ def _clineardepthgreaterthan_example() -> CLinearDepthGreaterThan:
...
+    Registers:
+        a: first input register.
+        b: second input register.
+        c: ancilla register that will contain $b-a$ and will be used for uncomputation.
+        target: A single bit output register to store the result of a > b.
+
+    References:
+        [Halving the cost of quantum addition](https://arxiv.org/abs/1709.06648).
+    """
+    comparison_dtype: str = attrs.field(default='>', init=False)
```
mpharrigan, line 1456: I'm not sure this is how we want to do this sort of inheritance. It's a little confusing. And presumably this attribute will show up in the repr and when you tab complete on available attributes.

qualtran/bloqs/arithmetic/comparison.py
```diff
@@ -1154,3 +1156,423 @@ def _clineardepthgreaterthan_example() -> CLinearDepthGreaterThan:
...
+                counts[k] = 0
+            counts[k] += v
+        return counts
+
+
+@frozen
+class _LinearDepthHalfComparisonBase(_HalfComparisonBase):
+    @cached_property
+    def _half_greater_than_bloq(self) -> Bloq:
+        return _HalfLinearDepthGreaterThan(self.dtype, uncompute=False)
```
mpharrigan, line 1430: is this always going to be the value used here? why even factor it out at all?

mpharrigan (commented):
some things are missing docs. Can you simplify some of the complicated inheritance by introducing one or more "has a" relationships instead of relying on "is a"

mpharrigan:
See `ZeroState(_ZVector)` for an example of how to specialize a private base implementation

qualtran/bloqs/arithmetic/comparison.py
NoureldinYosri: I was thinking `comparison_type` must have written `dtype` out of habbit :'D

mpharrigan:
@NoureldinYosri any updates here?

qualtran/bloqs/arithmetic/addition.py
NoureldinYosri: done

qualtran/bloqs/arithmetic/addition.py
NoureldinYosri: done

qualtran/bloqs/arithmetic/addition.py
NoureldinYosri: done

qualtran/bloqs/arithmetic/comparison.py
NoureldinYosri: it will differ between different falvours ... the ones I'm implementing here have linear depth and linear complexity, this is why their names start with LinearDepth.... There is another variant with log depth but nlog complexity which we might want to add later (we already have one instance of it in `LessThanEqual`)

qualtran/bloqs/arithmetic/comparison.py
NoureldinYosri: changed to `_op_symbol`

qualtran/bloqs/arithmetic/comparison.py
NoureldinYosri: how is this different from the ZeroState bloq? both set the value of a property of a base class and hide it

qualtran/bloqs/arithmetic/comparison.py
NoureldinYosri: same as above ... it's a property of the LinearDepth variant but we might want to add the _LogDepth variant

NoureldinYosri (commented):
@mpharrigan ptal.

as for the ZeroState example. I don't see how it and this implementation differ?

mpharrigan (commented):
Can you provide a short description of the interplay between the private classes in docstrings (see my comments); and a couple small nits. Thanks!!

qualtran/bloqs/arithmetic/comparison.py
mpharrigan: but looks good now

mpharrigan (approved):

PR #1414, opened 2024-09-19
Add `my_static_costs` override for accurate `QubitCount`s for qrom bloqs

tanujkhattar:
Adds support for correctly counting number of qubits for cases when the exact data is not available, and hence it's not possible to create the circuit decomposition.


mpharrigan (approved):

mpharrigan:
for testing sympy equality, they recommend subtracting the two expressions and asserting that it simplifies to zero. using `==` tests for strict equality of the expression itself (i.e. its format or in-memory rep)

qualtran/bloqs/mod_arithmetic/mod_multiplication_test.py
```diff
@@ -183,11 +183,11 @@ def test_dirtyoutofplacemontgomerymodmul_symbolic_cost(uncompute):
 
     # Litinski 2023 https://arxiv.org/abs/2306.08585
     # Figure/Table 8. Lists modular multiplication as 2.25n^2+9n toffoli.
-    # The following formula is 2.25n^2+8.25n-1 written with rationals because sympy comparison fails with floats.
+    # The following formula is 2.25n^2+7.25n-1 written with rationals because sympy comparison fails with floats.
```
tanujkhattar, line 186: @NoureldinYosri Can you please verify that this change is expected? I've simply updated the test assuming the constant factor reduction in qroam uncomputation is translating into reduced constant factors here.

NoureldinYosri: yes, the constant factor from qroam gets multiplied by $n/4$ here so it reduces the coefficient of $n$

PR #1409, opened 2024-09-19
Update Bartiq demo to work with version 0.6.0

mstechly:
This PR contains changes necessary for Baritq demo to work with Bartiq 0.6.0, which we plan to release soon.

PR #1415, opened 2024-09-20
2024-09 Bump dependencies, particularly qref

mpharrigan:
- new qref version pin, cc #1409
 - general dependency bump for the ci


mpharrigan:
I merged this PR into #1409 which contains the fixes to the failing notebook due to the version bump

PR #1424, opened 2024-09-27
Maked intermediate additions unsigned in ModAdd

fpapa250:
Replicate bug with ModAdd(bitsize=8, mod=17).adjoint().on_classical_vals(x=0,y=0). Basically, we do signed integer addition with registers that have unsigned dtypes. Change the intermediate register dtypes to QInt to perform modular reductions then restore to unsigned int dtypes.

PR #1425, opened 2024-09-27
Add ECAdd() Bloq

fpapa250:
Adds ECAdd bloq with classical simulation. I broke the bloq into the 6 steps shown in https://arxiv.org/abs/2306.08585. I classically simulated all steps. I put in a shimmed ModInv classical implementation (it's not montgomery mod inversion, but it still works on ECPoints not in montgomery form; I'm a little stuck on understanding what to do for R in montgomery mod inversion).


fpapa250:
Dependent on https://github.com/quantumlib/Qualtran/pull/1424. For testing purposes I put the bug fix in this branch too, but it should be merged first.

mpharrigan (commented):
A momentus piece of work! I did not check the exact circuits but the classical testing should give us a great deal of confidence. It would give more confidence if we had matching toffoli costs (and/or defensible reasons for any differences). Some small-ish things for improvement.

qualtran/bloqs/factoring/ecc/ec_add.py
```diff
@@ -30,6 +50,939 @@
...
+                Register('x', QMontgomeryUInt(self.n)),
+                Register('y', QMontgomeryUInt(self.n)),
+            ]
+        )
+
+    def adjoint(self) -> '_ECAddStepOne':
+        return attrs.evolve(self, uncompute=self.uncompute ^ True)
+
+    def on_classical_vals(
+        self, a: 'ClassicalValT', b: 'ClassicalValT', x: 'ClassicalValT', y: 'ClassicalValT'
```
NoureldinYosri, line 111: the classical action of the adjoint will fail since the adjoint has those sided register which the on_classica_vals doesn't expect

NoureldinYosri (dismissed):

qualtran/bloqs/mod_arithmetic/_shims.py
```diff
@@ -103,6 +117,37 @@ def build_call_graph(self, ssa: 'SympySymbolAllocator') -> 'BloqCountDictT':
...
+        garbage1 = x
+        garbage2 = x
+
+        # When P = Q the circuit computes lambda with a mod inversion, but doesn't use the output
+        # value. Here we will just do nothing to x in that case because it won't matter in the
+        # circuit anyway.
+        try:
+            x = pow(int(x) * pow(2, self.n, self.mod), -1, mod=self.mod) % self.mod
+        except ValueError:
+            pass
```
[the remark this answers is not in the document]
fpapa250: In the circuit we may have x=0; the inverse is going to cause a value error in classical simulation. If this is happening in the ECAdd circuit then we are going to ignore the value and use the precomputed lambda_r as the lambda/slope value. It doesn't matter what we do here it won't affect the circuit; I can change it to just do nothing if x=0 if that is preferred.

qualtran/bloqs/factoring/ecc/ec_add.py
fpapa250: I removed adjoint stuff - it's not required at all for this circuit and I think it'd be better to focus my efforts on other things for now.

qualtran/bloqs/factoring/ecc/ec_add_test.py
```diff
@@ -12,13 +12,155 @@
...
+        (32, 17, 4, 0, 0, 15, 13, 0, 0, 7, (0, 0, 15, 13, 2, 0, 0, 7)),
+    ],
+)
+def test_ec_add_step_two_classical(n, p, window_size, f1, ctrl, a, b, x, y, lam_r, result):
+    bloq = _ECAddStepTwo(n=n, mod=p, window_size=window_size)
+    ret1 = bloq.call_classically(f1=f1, ctrl=ctrl, a=a, b=b, x=x, y=y, lam_r=lam_r)
+    ret2 = bloq.decompose_bloq().call_classically(f1=f1, ctrl=ctrl, a=a, b=b, x=x, y=y, lam_r=lam_r)
+    assert ret1 == ret2
+    assert ret1 == result
+
```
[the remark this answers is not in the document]
NoureldinYosri: > Is the bloq_autotester sufficient for bloq decomp/equivalent bloq counts? 

no bloq_autotester tests only the given bloq and doesn't run all of the tests above. We need parameterized tests for a varity of instances of the bloqs.

> Only certain points on the curve are valid points

testing for prime modulus should be good enough.

qualtran/_infra/data_types.py
```diff
@@ -809,6 +809,18 @@ def assert_valid_classical_val_array(
         if np.any(val_array >= 2**self.bitsize):
             raise ValueError(f"Too-large classical values encountered in {debug_str}")
 
+    def montgomery_inverse(self, xm, n, p):
+        return xm ** (p - 2) * 2 ** (2 * n) % p
```
NoureldinYosri, line 814: These values will be huge and O(n) to compute .. prefer the following to keep all intermediate values in $[0, p)$ and the complexity to $\mathcal{O}(\log{p})$ here and below
```suggestion
        return (pow(xm, p - 2, p)) * pow(2, 2 * n, p)) % p
```

qualtran/_infra/data_types.py
```diff
@@ -810,6 +810,18 @@ def assert_valid_classical_val_array(
         if np.any(val_array >= 2**self.bitsize):
             raise ValueError(f"Too-large classical values encountered in {debug_str}")
 
+    def montgomery_inverse(self, xm, n, p):
+        return ((pow(xm, p - 2, p)) * pow(2, 2 * n, p)) % p
+
+    def montgomery_product(self, xm, ym, n, p):
+        return (xm * ym * pow(2, n * (p - 2), p)) % p
```
mpharrigan, line 817: can you write some unit tests for these operations? maybe check that prod(x, inv(x)) = 1

qualtran/_infra/data_types.py
```diff
@@ -810,6 +810,18 @@ def assert_valid_classical_val_array(
...
+
+    def montgomery_to_uint(self, xm, n, p):
+        return (xm * pow(2, n * (p - 2), p)) % p
+
+    def uint_to_montgomery(self, x, n, p):
+        return (x * pow(2, n, p)) % p
```
mpharrigan, line 823: sorry, you've opened a can-of-worms with these methods :)

Right now, these can be turned into `@staticmethod`s since they don't use `self`. Why don't they use `self`? Shouldn't `n` be `self.bitsize`? Why is the prime `p` not an attribute on the data type class? 

 - please open and link-with-a-todo and issue that says the datatype should learn these parameters
 - in the interim, change these to static methods and document(!) the input parameters

qualtran/bloqs/factoring/ecc/ec_add_test.py
```diff
@@ -12,13 +12,456 @@
...
+from qualtran.resource_counting.generalizers import ignore_alloc_free, ignore_split_join
+
+
+@pytest.mark.slow
+@pytest.mark.parametrize(
+    ['n', 'm'], [(n, m) for n in range(7, 10) for m in range(1, n + 1) if n % m == 0]
+)
+@pytest.mark.parametrize('a,b', [(15, 13), (2, 10), (8, 3), (0, 0)])
+@pytest.mark.parametrize('x,y', [(15, 13), (2, 10), (8, 3), (0, 0)])
+def test_ec_add_steps_classical_fast(n, m, a, b, x, y):
```
mpharrigan, line 42: this says fast but is slow and is marked as slow

mpharrigan: please consider taking a sampling of configurations rather than taking the "outer product" of point1 x point2 x n x m

qualtran/bloqs/factoring/ecc/ec_add_test.py
```diff
@@ -12,13 +12,456 @@
...
+        (5, 8),
+        (10, 15),
+        (1, 12),
+        (3, 0),
+        (1, 5),
+        (10, 2),
+        (0, 0),
+    ],
+)
+def test_ec_add_steps_classical(n, m, a, b, x, y):
```
mpharrigan, line 198: the parameterization here seems a little gratuitous. do we need the outer product of all possible n x m? What are we probing with different n values? it either fits or it doesn't, no? how about picking some extremal pairs? Do we need the complete outer product of point1 x point2? Maybe cross all possible point1 with some select extremal point2's

mpharrigan (commented):
it's tough to judge the right trade-offs for testing these. Left my comments. Thanks for opening and linking the other issues!

qualtran/_infra/data_types.py
fpapa250: Opened https://github.com/quantumlib/Qualtran/issues/1471 for adding modulus p to QMontgomeryUInt(). I think this would require too many changes for the scope of this PR. Happy to do the follow up PR after my project is completed.

Documented + tested - didn't make static cause I opted to use self.bitsize.

qualtran/bloqs/factoring/ecc/ec_add_test.py
fpapa250: Marked correctly and reduced points to a few interesting cross products.

qualtran/bloqs/factoring/ecc/ec_add_test.py
fpapa250: I pretty much halved the testing by removing a big chunk of the pairs. I did it in a way such that all 4 bugs from https://github.com/quantumlib/Qualtran/issues/1461 would still trigger if I made the frees non-dirty. Let me know if you want me to further reduce the tests. All-in-all it runs in 1 min 30 s on my pc with the slow tests running.

qualtran/bloqs/mod_arithmetic/_shims.py
```diff
@@ -103,6 +117,29 @@ def build_call_graph(self, ssa: 'SympySymbolAllocator') -> 'BloqCountDictT':
             Swap(self.n): 1,
         }
 
+    def on_classical_vals(
```
NoureldinYosri, line 120: why use the shim instead of the actual implementation?

qualtran/bloqs/arithmetic/_shims.py
```diff
@@ -39,6 +42,20 @@ def build_call_graph(self, ssa: SympySymbolAllocator) -> BloqCountDictT:
         return {Toffoli(): self.n - 2}
 
 
+@frozen
```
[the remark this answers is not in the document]
fpapa250: I agree it's easy to implement, but I think it is out of scope for this PR - it is a component of ModInv not ECAdd. I just changed it in this PR to make the Toffoli count test more accurate.

qualtran/bloqs/mod_arithmetic/_shims.py
fpapa250: I only implemented enough to test that ECAdd works - when ModInverse bloq is merged I'm happy to replace the shim with that in the construction/tests.

mpharrigan (approved):

PR #1428, opened 2024-09-29
Add RSA Phase Estimate Bloq and Move ModExp to rsa/ subdirectory

fpapa250:
I made a higher level top bloq with the phase estimation for shors for RSA. I also moved the bloqs to an rsa/ directory and moved the MeasureQFT shim to the factoring/ directory. Last, I made the ModExp bloq take an (n_exponent,) shaped QBit() register because the controls are used in the plus state in the phase estimation circuit so it can't be of type QUint (I think).


mpharrigan:
I think we discussed offline, but +1 to the top-level bloq. This should be set up as "phase estimation of the modular multiply operation". -1 to the signature change of ModExp, but maybe we don't include modexp as part of the decomposition if we frame the overall algorithm as phase estimation of modmul

qualtran/bloqs/factoring/rsa/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+#
+# isort:skip_file
+
+r"""Bloqs for breaking RSA cryptography systems via integer factorization.
+
+RSA cryptography is a form of public key cryptography based on the difficulty of
+factoring the product of two large prime numbers.
+
+Using RSA, the cryptographic scheme chooses two large prime numbers p, q, the product
+N, λ(n) = lcm(p - 1, q - 1), an integer e such that 1 < e < λ(n),
```
mpharrigan, line 23: why is lambda written as a function of n when the dependency doesn't show up in its definition? You don't actually define little n anywhere

qualtran/bloqs/factoring/rsa/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+and finally d as d ≡ e^-1 (mod λ(n)). The public key consists of the modulus N and the
```
mpharrigan, line 24: why does d get the tripple-equals and the others don't?

qualtran/bloqs/factoring/rsa/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+public (or encryption) exponent e. The private key consists of the private (or decryption)
+exponent d, which must be kept secret. p, q, and λ(n) must also be kept secret because
+they can be used to calculate d.
```
mpharrigan, line 27: is there a reference you can include where you got this from?

qualtran/bloqs/factoring/rsa/rsa_mod_exp.py
```diff
@@ -59,17 +61,17 @@ class ModExp(Bloq):
 
     Registers:
         exponent: The exponent
-        x [right]: The output register containing the result of the exponentiation
+        x: The output register containing the result of the exponentiation
```
mpharrigan, line 64: you didn't like my "[right]"? I think we should document sided registers somehow

qualtran/bloqs/factoring/rsa/rsa_mod_exp.py
```diff
@@ -85,33 +87,39 @@ def signature(self) -> 'Signature':
...
             little_n = sympy.ceiling(sympy.log(big_n, 2))
         else:
             little_n = int(math.ceil(math.log2(big_n)))
         if g is None:
-            g = random.randint(2, big_n)
+            if is_symbolic(big_n):
+                g = sympy.symbols('g')
+            else:
+                while True:
+                    g = random.randint(2, int(big_n))
```
mpharrigan, line 107: thanks for fixing this, but I'm going to be very picky:

please add an optional `rs: np.RandomState = None` argument to this method that can make this deterministic. If it's not provided, make a new `np.random.RandomState()`. Use a seeded, deterministic random state in all tests.

qualtran/bloqs/factoring/_factoring_shims.py
```diff
@@ -23,7 +24,7 @@
 
 @frozen
 class MeasureQFT(Bloq):
-    n: int
+    n: Union[int, sympy.Expr]
```
mpharrigan, line 27: SymbolicInt?

qualtran/serialization/resolver_dict.py
```diff
@@ -97,7 +97,9 @@
 import qualtran.bloqs.data_loading.qroam_clean
 import qualtran.bloqs.data_loading.qrom
 import qualtran.bloqs.data_loading.select_swap_qrom
-import qualtran.bloqs.factoring.mod_exp
+
+qualtran.bloqs.factoring._factoring_shims
```
mpharrigan, line 101: what happened here?

qualtran/bloqs/factoring/rsa/rsa_phase_estimate.py
```diff
@@ -0,0 +1,141 @@
...
+    QUInt,
+    Signature,
+    SoquetT,
+)
+from qualtran.bloqs.basic_gates import IntState, PlusState
+from qualtran.bloqs.bookkeeping import Free
+from qualtran.bloqs.factoring._factoring_shims import MeasureQFT
+from qualtran.bloqs.mod_arithmetic.mod_multiplication import CModMulK
+from qualtran.resource_counting import BloqCountDictT, SympySymbolAllocator
+from qualtran.symbolics.types import is_symbolic, SymbolicInt
```
mpharrigan, line 39: just import from qualtran.symbolics

qualtran/bloqs/factoring/rsa/rsa_phase_estimate.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+@frozen
+class RSAPhaseEstimate(Bloq):
+    """Perform a single phase estimation of ModExp for the given base.
```
mpharrigan, line 44: aren't we phase-estimating ModMul?

qualtran/bloqs/factoring/rsa/rsa_phase_estimate.py
```diff
@@ -0,0 +1,141 @@
...
+        if is_symbolic(big_n):
+            little_n = sympy.ceiling(sympy.log(big_n, 2))
+        else:
+            little_n = int(math.ceil(math.log2(big_n)))
+        if g is None:
+            if is_symbolic(big_n):
+                g = sympy.symbols('g')
+            else:
+                while True:
+                    g = random.randint(2, int(big_n))
```
mpharrigan, line 85: see other comment on random

qualtran/bloqs/factoring/rsa/rsa_phase_estimate.py
```diff
@@ -0,0 +1,141 @@
...
+from qualtran.resource_counting import BloqCountDictT, SympySymbolAllocator
+from qualtran.symbolics.types import is_symbolic, SymbolicInt
+
+
+@frozen
+class RSAPhaseEstimate(Bloq):
+    """Perform a single phase estimation of ModExp for the given base.
+
+    Computes the phase estimation of a single run of Modular Exponentiation with
+    an optional, pre-set base or a random, valid base.
```
mpharrigan, line 47: be more precise here. "The constructor requires a pre-set base, see the `make_for_shor` factory method for picking a random, valid base"

qualtran/bloqs/factoring/rsa/rsa_phase_estimate.py
```diff
@@ -0,0 +1,141 @@
...
+
+    Args:
+        n: The bitsize of the modulus N.
+        mod: The modulus N; a part of the public key for RSA.
+        base: A base for modular exponentiation.
+
+    References:
+        [Circuit for Shor's algorithm using 2n+3 qubits](https://arxiv.org/abs/quant-ph/0205095).
+        Stephane Beauregard. 2003.
```
mpharrigan, line 56: Just put their last name. This is Figure 1, ya? please put

mpharrigan (commented):
some nits, but looking good!

qualtran/bloqs/factoring/rsa/__init__.py
fpapa250: I just copied from wikipedia, not gonna lie (although I'm familiar I've taken multiple classes which discussed RSA). d gets triple equals because it is congruent mod lam(n) not equal. The others are actually equal not just congruent mod lam(n).

qualtran/bloqs/factoring/rsa/rsa_mod_exp.py
fpapa250: Oops, my bad. I forgot to add this back when I reconfigured mod exp to not be used in the phase estimation.

qualtran/bloqs/factoring/rsa/rsa_mod_exp.py
```diff
@@ -104,7 +110,9 @@ def make_for_shor(cls, big_n: 'SymbolicInt', g: Optional['SymbolicInt'] = None):
                 g = sympy.symbols('g')
             else:
                 while True:
-                    g = random.randint(2, int(big_n))
+                    if rs is None:
+                        rs = np.random.RandomState()
```
mpharrigan, line 114: sorry again, but these sorts of things can be a _huge_ headache if they slip through:

the random state must be seeded/initialized outside the loop or untold bad things will happen; including repeatedly trying the same number millions of times depending on how the seeding works

mpharrigan (commented):
random issues

mpharrigan (approved):

PR #1411, opened 2024-09-19
Add Equals() bloq

fpapa250:
Adds bloq to check if two quantum registers are equal.

PR #1429, opened 2024-09-30
fix call graph for `Equals`

anurudhp:
bloq introduced in #1411 had some extra bloqs in the call graph


anurudhp:
@NoureldinYosri @mpharrigan ptal

PR #1431, opened 2024-10-02
et. al. -> et al.

mpharrigan:
This one is mea culpa

https://www.merriam-webster.com/dictionary/et%20al.

"et" is a full word; "al" is an abbreviation.


fdmalone:
Your latin teacher's et alia must be so disappointed. Pro dolor!

fdmalone (approved):

Issue #1419, opened 2024-09-26
Add a new QDType for Galois Fields


tanujkhattar:
Galois Fields are useful in the context of optimization algorithms like DQI - https://arxiv.org/abs/2408.08292

The Galois library is a useful numpy extension that supports Galois fields - https://galois.readthedocs.io/en/v0.4.0/

This issue is a feature request to add a new QDType for Galois Fields and add integration with the galois library so that classical simulation protocols can pass around numpy arrays of galois type.

PR #1433, opened 2024-10-02
Add QGF data type for Galois Fields

tanujkhattar:
Part of https://github.com/quantumlib/Qualtran/issues/1419

Serialization support for new QDType will be added in a follow up PR

### DIFF
--- a/qualtran/__init__.py
+++ b/qualtran/__init__.py
@@ -53,6 +53,7 @@ from ._infra.data_types import (
     QUInt,
     BQUInt,
     QMontgomeryUInt,
+    QGF,
 )
 
 # Internal imports: none
--- a/qualtran/_infra/data_types.py
+++ b/qualtran/_infra/data_types.py
@@ -50,14 +50,16 @@ respectively.
 
 import abc
 from enum import Enum
+from functools import cached_property
 from typing import Any, Iterable, List, Sequence, Union
 
 import attrs
 import numpy as np
 from fxpmath import Fxp
+from galois import GF
 from numpy.typing import NDArray
 
-from qualtran.symbolics import is_symbolic, SymbolicInt
+from qualtran.symbolics import bit_length, is_symbolic, SymbolicInt
 
 
 class QDType(metaclass=abc.ABCMeta):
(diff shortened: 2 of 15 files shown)


tanujkhattar:
Ugh, looks like both galois and quimb use numba and that's causing issues with notebook tests that I'm unable to reproduce locally. @mpharrigan Do you have any idea how to fix this?

mpharrigan:
usually depending on the same package would be a boon for compatibility! I'll investigate

tanujkhattar:
From a brief look, I think quimb setting the number of numba threads explicitly may be causing the issue:

https://github.com/jcmgray/quimb/blob/8ecf9d6ef7d2055ebbae5f7dbb5aeb0e65c0d422/quimb/core.py#L37-L51

Numba also has a default but one uses `os.cpu_count()` and other uses `psutil.cpu_count()`. Both of these should be the same but looks like there have been [discrepancies in the past](https://bugs.python.org/issue33166).

https://github.com/numba/numba/blob/717308e5cb6f7ddcdbab9fadd9972b1bf1b83240/numba/core/config.py#L524C28-L524C37

I couldn't find anything in galois that explicitly sets number of threads for numba, but maybe there's some sneaky logic living somewhere. 

Thanks for looking into this and let me know what you find!

mpharrigan:
can't reproduce locally, but seems to be https://github.com/nasa/hybridq/blob/main/README.md#runtimeerror-cannot-set-numba_num_threads-to-a-different-value-once-the-threads-have-been-launched

tanujkhattar:
I guess the question is who is changing the NUMBA_NUM_THREADS after its precached, and galois is probably doing that given the failures are seen in this PR.

mpharrigan:
I tried just importing the packages, which didn't trigger the bug. I tried disabling the parallel execution of notebooks which didn't fix the bug. 

From the error message and synthesizing some of the info from @tanujkhattar 's investigation above: it looks like the original value of 2 is incorrect if we're trying to set it to the number of cores, which github documents as 4 https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories which is the second value

mpharrigan:
How about we patch the CI failure by explicitly setting the environment variable (see current status) and keep an eye out for this showing up on an actual computer in a reproducible way.

tanujkhattar:
Thanks, that sounds like a good plan to me

qualtran/_infra/data_types.py
```diff
@@ -810,8 +812,130 @@ def assert_valid_classical_val_array(
...
+    perform arithmetic over Galois Fields. By default, the Conway polynomial $C_{p, m}$ is used
+    as the irreducible polynomial.
+
+    Attributes:
+        characteristic: The characteristic $p$ of the field $GF(p^{m})$.
+            The characteristic must be prime.
+        degree: The degree $m$ of the field $GF(p^{m})$. The degree must be a positive integer.
+
+    References
+        [Finite Field](https://en.wikipedia.org/wiki/Finite_field)
```
mpharrigan, line 842: how does this get rendered in the api docs? Usually we separate individual references with a blank newline; but I admit that that is primarily for the notebook-based doc system

qualtran/_infra/data_types.py
```diff
@@ -810,8 +812,130 @@ def assert_valid_classical_val_array(
...
 
 
+@attrs.frozen
+class QGF(QDType):
+    r"""Galois Field type to represent elements of a finite field.
+
+    A Finite Field or Galois Field is a field that contains finite number of elements. The order
+    of a finite field is the number of elements in the field, which is either a prime number or
+    a prime power. For every prime number $p$ and every positive integer $m$ there are fields of
+    order $p^{m}0$, all of which are isomorphic. When m=1, the finite field of order p can be
```
mpharrigan, line 822: is this 0 supposed to be here? what does it mean?

qualtran/_infra/data_types.py
```diff
@@ -810,8 +812,130 @@ def assert_valid_classical_val_array(
...
+    multiplication is defined as polynomial multiplication modulo an irreducible polynomial of
+    degree $m$. The selection of the specific irreducible polynomial affects the representation
+    of the given field, but all fields of a fixed size are isomorphic.
+
+    The data type uses the [Galois library](https://mhostetter.github.io/galois/latest/) to
+    perform arithmetic over Galois Fields. By default, the Conway polynomial $C_{p, m}$ is used
+    as the irreducible polynomial.
+
+    Attributes:
+        characteristic: The characteristic $p$ of the field $GF(p^{m})$.
```
mpharrigan, line 837: personal preference, but I generally remove rendundant latex `{ }`. You can just write `$p^m$`.

qualtran/simulation/classical_sim.py
```diff
@@ -81,16 +81,23 @@ def _numpy_dtype_from_qdtype(dtype: 'QDType') -> Type:
     return object
 
 
+def _numpy_empty_reg_from_reg(reg: Register) -> np.ndarray:
```
mpharrigan, line 84: "nump" and "from reg" makes sense; but we're not returning an empty register per se. `_numpy_empty_array_from_reg`? `_np_empty_from_reg`? `_empty_ndarray_from_reg`?

mpharrigan: Should this function be an overridable method on QDType? Not blocking this PR but if you think it is a good idea we can open an issue

qualtran/__init__.py
```diff
@@ -53,6 +53,7 @@
     QUInt,
     BQUInt,
     QMontgomeryUInt,
+    QGF,
```
mpharrigan, line 56: any concerns about effects on import time? I know nothing will be slower than quimb but you can theoretically avoid importing quimb sometimes

mpharrigan (approved):
we love types!

qualtran/__init__.py
tanujkhattar: Moved `from galois import GF` to a local import inside a method of `QGF`. Verified that `time python -c 'import qualtran'` has no degradation from this PR now.

qualtran/simulation/classical_sim.py
tanujkhattar: Yes, I think it's a good idea to have this function on then QDType. It would also make `Fxp` integration better I think; since `Fxp` also derives from numpy array and we don't make use of that. This is also why `Fxp` is slow for us I think. 

I'll open an issue.

qualtran/_infra/data_types.py
tanujkhattar: done

qualtran/_infra/data_types.py
tanujkhattar: added blank newline between references.

tanujkhattar:
Thanks for the review, I've addressed the nits. Merging now.

qualtran/simulation/classical_sim.py
mpharrigan: Did you end up opening one? There's something during review of #1436 that I'd like to note on the issue as well

tanujkhattar: Opened https://github.com/quantumlib/Qualtran/issues/1437

PR #1439, opened 2024-10-05
Update copyright checker and switch to pylint-3

pavoljuhas:
* Bump up to the latest pylint-3.3.1

* Adjust `pylint_copyright_checker` for the pylint-3 API
  Follow the `examples/custom_raw.py` in the pylint repository.

* Refresh development dependencies

* Temporarily disable rules that started failing in pylint-3

Cherry-picked from https://github.com/quantumlib/Cirq/pull/6751
Related to #1440


mpharrigan:
what's the deal with the `abstract-class-instantiated`. Those classes aren't abstract. We're instantiating them

mpharrigan (approved):

pavoljuhas:
> what's the deal with the `abstract-class-instantiated`. Those classes aren't abstract. We're instantiating them

I checked one of such instances - [Bloq.signature](https://github.com/quantumlib/Qualtran/blob/9ee553b9deaff980013936aee921ef6c907bf4d4/qualtran/_infra/bloq.py#L99-L101) is an abstract property which is replaced by attrs field, but not redefined in its [CompositeBloq](https://github.com/quantumlib/Qualtran/blob/9ee553b9deaff980013936aee921ef6c907bf4d4/qualtran/_infra/composite_bloq.py#L109) subclass.  I suspect pylint checks for a property named "signature", but `type(CompositeBloq.signature)` is `member_descriptor`, not `property`.

micpap25:
> I checked one of such instances - [Bloq.signature](https://github.com/quantumlib/Qualtran/blob/9ee553b9deaff980013936aee921ef6c907bf4d4/qualtran/_infra/bloq.py#L99-L101) is an abstract property which is replaced by attrs field, but not redefined in its [CompositeBloq](https://github.com/quantumlib/Qualtran/blob/9ee553b9deaff980013936aee921ef6c907bf4d4/qualtran/_infra/composite_bloq.py#L109) subclass. I suspect pylint checks for a property named "signature", but `type(CompositeBloq.signature)` is `member_descriptor`, not `property`.

This observation is corroborated by the similar errors on `_PyCell` and `_MarkdownCell` in regards to `_Cell.cell_id`.
However, `Unitary` also gives the `abstract-class-instantiated` issue and it does attempt to instantiate `signature` as a property, so I'm not confident...

Any has anything been attempted since this issue was last commented on?

mpharrigan:
I don't think so. Let's continue discussion on the issue page, #1440

PR #1441, opened 2024-10-06
Add `GF2Square` bloq for squaring over GF($2^m$)

tanujkhattar:
Part of a larger effort to support galois field arithmetic in qualtran. xref https://github.com/quantumlib/Qualtran/issues/1419

Currently on top of https://github.com/quantumlib/Qualtran/pull/1438

PR #1442, opened 2024-10-06
Add `GF2Inverse` bloq for computing inverse over GF($2^m$)

tanujkhattar:
Part of a larger effort to support galois field arithmetic in qualtran. xref https://github.com/quantumlib/Qualtran/issues/1419

Currently on top of https://github.com/quantumlib/Qualtran/pull/1441


qualtran/bloqs/gf_arithmetic/gf2_inverse.py
```diff
@@ -0,0 +1,167 @@
...
+        a^{-1} = a^{2^m - 2}
+    $$
+
+    Thus, the inverse can be obtained via $m - 1$ squaring and multiplication operations.
+
+    Args:
+        bitsize: The degree $m$ of the galois field $GF(2^m)$. Also corresponds to the number of
+            qubits in the input register whose inverse should be calculated.
+
+    Registers:
```
mpharrigan, line 71: is there a reference for this construction?

qualtran/bloqs/gf_arithmetic/gf2_inverse_test.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def test_gf2_multiplication_symbolic(bloq_autotester):
+    bloq_autotester(_gf2_inverse_symbolic)
+
+
+def test_gf2_multiplication_symbolic_toffoli_complexity():
+    bloq = _gf2_inverse_symbolic.make()
+    m = bloq.bitsize
+    assert get_cost_value(bloq, QECGatesCost()).toffoli - m**2 * (m - 2) == 0
```
mpharrigan, line 39: nice. is there a reference for this expression? consider including the equation number as line comment and/or the reference in the references section

mpharrigan (approved):
lgtm

qualtran/bloqs/gf_arithmetic/gf2_inverse.py
```diff
@@ -0,0 +1,167 @@
...
+        result = bb.allocate(dtype=self.qgf)
+        if self.bitsize == 1:
+            x, result = bb.add(GF2Addition(self.bitsize), x=x, y=result)
+            return {'x': x, 'result': result}
+
+        x = bb.add(GF2Square(self.bitsize), x=x)
+        x, result = bb.add(GF2Addition(self.bitsize), x=x, y=result)
+
+        junk = []
+        for i in range(2, self.bitsize):
```
NoureldinYosri, line 119: nit: I would argue that `range(self.bitsize - 2)` is clearer but it's up to you

qualtran/bloqs/gf_arithmetic/gf2_inverse_test.py
NoureldinYosri, line 39: is the cost just toffolis? if so,  then in other tests we have been calling `total_t_and_ccz_count` and assertign that `n_t` == 0 and then checking `n_ccz` == the symbolic cost

qualtran/bloqs/gf_arithmetic/gf2_inverse.py
tanujkhattar: This one follows directly from Fermat's little theorem and the standard repeated squaring for exponentiation, that I've explained in the docstring. Dmitri told me there's a more optimized version explain in https://www.sciencedirect.com/science/article/pii/0890540188900247, which reduces the cost from O(m^3) to O(m^2log(m)) via a smarter way to implement repeated squaring for multiplication. 

I'll implement this in a follow-up and add a reference.

qualtran/bloqs/gf_arithmetic/gf2_inverse_test.py
tanujkhattar: I don't have a reference, this is an unoptimized implementation that follows from the docstring.

qualtran/bloqs/gf_arithmetic/gf2_inverse.py
tanujkhattar: I prefer this because $i$ has a significance (we are computing $x^{2^{i}}$ in the $i$'th iteration), so I'll keep it as is.

qualtran/bloqs/gf_arithmetic/gf2_inverse_test.py
tanujkhattar: Used `total_toffoli_only` which implicitly asserts that `n_t == 0`

Issue #645, opened 2024-02-08
Implement Oracles from "Compilation of Fault-Tolerant Quantum Heuristics for Combinatorial Optimization"


tanujkhattar:
[Compilation of Fault-Tolerant Quantum Heuristics for Combinatorial Optimization](https://arxiv.org/abs/2007.07391) gives detailed costings and constructions of oracles for useful optimization problems. This issue is to track the addition of these oracles to Qualtran. The following is a running list of useful oracles that we should add

* [ ] Hamming Weight Phasing with a configurable number of ancilla qubits: We already have an implementation for hamming weight phasing that uses `n - n.bit_count()` additional ancilla qubits described in https://arxiv.org/abs/1709.06648. This task is to expand this implementation s.t. the the primitive expects number of ancilla qubits as a configurable parameter and follows the strategy described in Appendix A.2 of https://arxiv.org/pdf/1902.10673.pdf
* [ ]  Direct energy oracle for SK model: Using the learnings from the above task, implement the direct energy oracle for computing the cost function for SK model described in Section II.A.2 of https://arxiv.org/pdf/2007.07391.pdf
* [ ] Arithmetic primitives for multiplication via repeated addition described in Appendix D of https://arxiv.org/pdf/2007.07391.pdf
  * [ ] Scale integer by real number (example demonstrating addition into phase gradient register)
  * [ ] Multiply two real numbers
  * [ ] Square a real number

senecameeks:
I can take this one on

dobbse42:
I figure this is a good issue to implement in order to familiarize myself with qualtran and the development workflow in general. I've written a bloq which I believe implements the simplest method for Hamming weight phasing with a configurable number of ancilla, but I am unsure what the best-practice method is for sending a subset of a register to a component bloq. At the moment I split the entire register, take a slice of that list corresponding to my subset, then join the slice and send that to the component bloq. I then split the output of that component bloq and add them to an initially empty list. Following this, the elements of the list look like:
Split<0>.reg\[0\], Split<0>.reg\[1\], ...
Now when I run bb.join on this list, I get ValueError("'join' expects a 1-d array of input soquets to join."). My understanding was that bb.split yielded exactly a 1-d array of soquets, so intuitively sticking the results of a bunch of splits together and running bb.join would work. I think I am misunderstanding how join and split work/are meant to be used, and I am sure there is a better way of doing what I am trying to do. Any suggestions would be greatly appreciated.

[...]

PR #1450, opened 2024-10-08
Hamming weight phasing with configurable number of ancilla

dobbse42:
#645: Added HammingWeightPhasingWithConfigurableAncilla and a temp test file, still debugging. TODO: add unit tests to hamming_weight_phasing_test.py

Currently having issues joining the slices of the phased register back together with
`x = bb.join(x_parts, dtype=QUInt(self.bitsize.bit_length()))`
at the end of build_composite_bloq() (line 272).


mpharrigan:
Thanks for opening a PR! Please ping me or @tanujkhattar when this is ready for review or if you have any specific questions

qualtran/bloqs/rotations/hamming_weight_phasing.py
```diff
@@ -209,3 +209,83 @@ def _hamming_weight_phasing_via_phase_gradient() -> HammingWeightPhasingViaPhase
...
+        '''        x_parts = [
+            a
+            for x_part in x_parts
+            for a in x_part
+        ]
+        '''
+        #print("shape after flatten: ", np.shape(x_parts))
+        for part in x:
+            print("next elem: ", part)
+        x = bb.join(x_parts, dtype=QUInt(self.bitsize.bit_length()))
```
mpharrigan, line 272: try `bb.join(np.array(x_parts), ...)`

Elsewhere, we tried to be careful about accepting either lists or numpy arrays but seem to have missed this one. 
https://github.com/quantumlib/Qualtran/issues/1470

dobbse42:
@mpharrigan This should be in a functional state at least. There are a few TODO comments which indicate things I couldn't immediately tell how to do but aren't essential, so suggestions for those would be appreciated. I also have a couple of questions:
- Do we want to perform input validation (e.g. ancillasize < bitsize-1)? If so, what is the preferred method?
- Should I write gatecount assertions to get gatecounts of subcomponents, or just hardcode them? e.g. should I just write "4\*num_toffolis" or is there some way to get the T-count of a specified toffoli decomposition in tests?
- Is there any particular logic to the choice of parameter values to test, or just common sense?
- Are we testing against what we logically expect or what the paper says should happen? i.e. we can often get more precise gatecounts by just thinking about the situation (in which case we might assert that counts are exactly equal to what we expect), whereas papers are often just concerned with asymptotics or upper bounds (in which case we would assert that gatecounts are <= what the paper reports). Further, there are situations where computing the precise gatecount for every input is tedious and not very helpful for testing compared to just saying "less than" or "approx equal to".
- When is a build_call_graph override necessary? For example, HWphasegradient doesn't have one but I got errors when I tried to test HWConfigurableancilla without one.

All feedback is welcome, especially on the tests as I do not have much experience with pytest. I expect to have to make quite a few modifications to this.

anurudhp:
Hi @dobbse42! Here is some feedback on the PR, hope it is helpful!

> Do we want to perform input validation (e.g. ancillasize < bitsize-1)? If so, what is the preferred method?

You can use the `__attrs_post_init__` method - raise a `ValueError` on invalid inputs (https://www.attrs.org/en/stable/init.html#post-init)

> Should I write gatecount assertions to get gatecounts of subcomponents, or just hardcode them? e.g. should I just write "4*num_toffolis" or is there some way to get the T-count of a specified toffoli decomposition in tests?

The current standard way is to use `QECGatesCost`, e.g. see https://github.com/quantumlib/Qualtran/blob/7bbec3b8e41bbefcfd5820830ca5bc582ff75a29/qualtran/resource_counting/_bloq_counts_test.py#L82-L85
If you only care about the T-cost, you can use the [`.total_t_count`](https://github.com/quantumlib/Qualtran/blob/7bbec3b8e41bbefcfd5820830ca5bc582ff75a29/qualtran/resource_counting/_bloq_counts.py#L186-L192) method to get it (see class [`GateCounts`](https://github.com/quantumlib/Qualtran/blob/7bbec3b8e41bbefcfd5820830ca5bc582ff75a29/qualtran/resource_counting/_bloq_counts.py#L126) for more such helper functions)

> Is there any particular logic to the choice of parameter values to test, or just common sense?

We usually pick a few random small ones (especially for simulation tests), and a few slightly larger sizes (e.g. for gate cost tests). If you feel there are some edge cases that require careful testing, it's good to add more tests in!

> Are we testing against what we logically expect or what the paper says should happen? i.e. we can often get more precise gatecounts by just thinking about the situation (in which case we might assert that counts are exactly equal to what we expect), whereas papers are often just concerned with asymptotics or upper bounds (in which case we would assert that gatecounts are <= what the paper reports). Further, there are situations where computing the precise gatecount for every input is tedious and not very helpful for testing compared to just saying "less than" or "approx equal to".

If the paper only has upper bounds, you can add "less than equal" tests against the reference formulas. Sometimes the formulas may only be asymptotically valid, or be missing constant factors, in which case you can leave out precise gate count tests. Another useful contribution is to plot the T-costs vs input size in the bloq notebook, just to visually verify. 

> When is a build_call_graph override necessary? For example, HWphasegradient doesn't have one but I got errors when I tried to test HWConfigurableancilla without one.

The override is needed when the bloq cannot be decomposed (e.g. because of symbolic bitsizes). It is also good to add the override if the call graph is much simpler (and efficient to compute) compared to the bloq decomposition. E.g. in this particular case, the decomposition requires a lot of splits and joins, but the call graph is much simpler. So I'd say keep the override for convenience. 

In case the bloq has a valid decomposition but the call graph still gives an error, that is a bug, please report it!

NoureldinYosri:
@dobbse42 thank you for contributing to Qualtran ... I left a couple of code style comments

mpharrigan:
hi @dobbse42 , I see you've pushed new commits. What's the status of this? Have you addressed the open review comments or are these still some outstanding?

NoureldinYosri:
@dobbse42 to fix the formating CI you can run `./check/format-incremental --apply`, for the others you can check the log to see why they complain, for example the pylint is complaining because in `qualtran/bloqs/rotations/hamming_weight_phasing_test.py` the class `GateCounts` is imported but never used

mhucka:
@dobbse42 @NoureldinYosri Just circling around to different PRs and trying to do a bit of housekeeping. For planning purposes, what is the status of this PR?

PR #1451, opened 2024-10-08
Improve interface for bloqs with specialized single-qubit-controlled versions

anurudhp:
- Adds a method `get_ctrl_system_1bit_cv` that bloqs can use in their `get_ctrl_system`
- Reduces multiple controls to single qubit using `ControlledViaAnd`
- Handles nested controls (e.g. `bloq.controlled().controlled()`) by using a singly-controlled bloq and merging the controls as above

I have reduced the uses of the old interface (`SpecializedSingleQubitControlledExtension`), but some of the uses are not easy to update (as there is no easy way to get an uncontrolled bloq from the controlled bloq in these cases). These should eventually be updated carefully and the old interface removed.


anurudhp:
@mpharrigan thoughts on the design?

mpharrigan:
I took a very brief look and this looks really nice! I'll likely have more specific questions when I take a closer look. How does single-bit zero-control work? it looks like the protocol interface lets the bloq specify its control value but the logic seemingly only works for cv=1 (at least in contrast to SpecializedSingleQubitControl)

anurudhp:
> How does single-bit zero-control work? it looks like the protocol interface lets the bloq specify its control value but the logic seemingly only works for cv=1 (at least in contrast to SpecializedSingleQubitControl)

Yes, this is true for now. I was considering updating `ControlledViaAnd`'s decomp in the single control case to pass the `cv` through (instead of applying `X` gates for 0-ctrl as it does now), and in the above method also checking for `CtrlSpec(cvs=0)`.

There is also this nuance that there's a lot of ways to specify a single 0 or 1 (e.g. `QUInt(1)` etc. instead of `QBit()`; and shapes like `(1, 1, 1)` instead of `()`). Maybe I can check if `ctrl_spec.num_qubits` is 1 and extract the correct bit (like in `ControlledViaAnd`)

> Is ControlledViaAnd a complete drop-in replacement for qualtran.Controlled? Is there any reason why we wouldn't want to replace Controlled with ControlledViaAnd?

This was the idea, but `ControlledViaAnd` uses the `And` bloq, so I was worried putting it in `_infra` could create cyclic dependencies. (I think everything else in `_infra/` is agnostic of `bloqs/`, which keeps the design clean imo)

anurudhp:
> Is ControlledViaAnd a complete drop-in replacement for qualtran.Controlled? Is there any reason why we wouldn't want to replace Controlled with ControlledViaAnd?

One solution with the least changes would be to update the default `Bloq.get_ctrl_system` to:

```py
    from qualtran import Controlled
    from qualtran.bloqs.mcmt import ControlledViaAnd

    if ctrl_spec.num_qubits != 1:
        return ControlledViaAnd.make_ctrl_system(self, ctrl_spec=ctrl_spec)
    return Controlled.make_ctrl_system(self, ctrl_spec=ctrl_spec)
```

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
```diff
@@ -0,0 +1,102 @@
...
+
+    if ctrl_spec.num_qubits != 1:
+        return _get_default_fallback()
+
+    assert isinstance(
+        bloq, BloqWithSpecializedControl
+    ), f"{bloq} must implement protocol {BloqWithSpecializedControl}"
+
+    # extract the single bit control value
+    (cv,) = ctrl_spec._cvs_tuple
```
mpharrigan, line 53: I'm always confused by getting a particular control value from a CtrlSpec. The type annotation is really confusing because of the variety of valid `cv` values. Is it worth adding a property to CtrlSpec that returns a single bit control value (or errors if there are more than one)

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def _get_default_fallback():
+        assert isinstance(bloq, Bloq)
+        return ControlledViaAnd.make_ctrl_system(bloq=bloq, ctrl_spec=ctrl_spec)
+
+    if ctrl_spec.num_qubits != 1:
+        return _get_default_fallback()
+
+    assert isinstance(
+        bloq, BloqWithSpecializedControl
```
mpharrigan, line 49: so this function needs to get a bloq's current `cv`, it's control register name, and a callable that returns a controlled version.

Did you consider just taking these as arguments to the function? i.e. at the call site you could just pass a cv and control register name.

Maybe this is awkward for the `with_cv` callable. But do you expect this to basically always be `lambda cv: attrs.evolve(bloq, **{ctrl_attr_name: cv})` for a given `ctrl_attr_name`? if so: you can specialize to this case and take everything as arguments instead of a duck-typing `Protocol`.

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
```diff
@@ -0,0 +1,102 @@
...
+    ), f"{bloq} must implement protocol {BloqWithSpecializedControl}"
+
+    # extract the single bit control value
+    (cv,) = ctrl_spec._cvs_tuple
+
+    if bloq.cv is None:
+        # the easy case: use the controlled bloq
+        ctrl_bloq = bloq.with_cv(cv=cv)
+        if ctrl_bloq is None:
```
mpharrigan, line 58: this behavior is not obvious. Is it needed? please document in the protocol method

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
```diff
@@ -0,0 +1,102 @@
...
+            return _get_default_fallback()
+
+        assert isinstance(ctrl_bloq, BloqWithSpecializedControl)
+        ctrl_reg_name = ctrl_bloq.ctrl_reg_name
```
mpharrigan, line 62: do we expect this to always be the same as `bloq.ctrl_reg_name`?

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
```diff
@@ -0,0 +1,102 @@
...
+
+        def _adder(
+            bb: 'BloqBuilder', ctrl_soqs: Sequence['SoquetT'], in_soqs: dict[str, 'SoquetT']
+        ) -> tuple[Iterable['SoquetT'], Iterable['SoquetT']]:
+            assert ctrl_bloq is not None
```
mpharrigan, line 67: redundant?

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
```diff
@@ -0,0 +1,102 @@
...
+            in_soqs |= {ctrl_reg_name: ctrl}
+
+            out_soqs = bb.add_d(ctrl_bloq, **in_soqs)
+
+            ctrl = out_soqs.pop(ctrl_reg_name)
+            return [ctrl], out_soqs.values()
+
+    else:
+        # the difficult case: must combine the two controls into one
+        un_ctrl_bloq = bloq.with_cv(cv=None)
```
mpharrigan, line 79: explicitly document this behavior in the protocol

mpharrigan (commented):
sum initial detailed thoughts. Curious to hear your responses

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
anurudhp: Yes this makes sense. Perhaps `CtrlSpec.get_single_control_bit`?

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
anurudhp: Yeah I think this is always the pattern. I realized it doesn't even need to be a callable, I just need 3 bloqs: `ctrl_bloq`, `unctrl_bloq`, `ctrl_0_bloq`. I can just pass those three bloqs here, along with the control register name and the current `cv`.

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
anurudhp: The older `SelectOracle` interfaces have a `.control_registers` property which could store arbitrary registers, so we can't figure out the control register name from `bloq` itself.

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
anurudhp: this was for `mypy`, not sure why it didn't pick up the previous `if` condition.

anurudhp: removed now

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
anurudhp: added a method to extract the bit correctly

anurudhp:
@mpharrigan I simplifed the interface using your suggestions: I now simply pass the controlled bloq, the ctrl reg name, and the un-controlled bloq. (removed the protocol). I updated the two unittest examples as well.

Let me know if it looks good, and I'll update the other examples.

qualtran/_infra/controlled.py
```diff
@@ -250,6 +250,18 @@ def from_cirq_cv(
             bloq_cvs.append(curr_cvs)
         return CtrlSpec(tuple(qdtypes), tuple(bloq_cvs))
 
+    def get_single_control_bit(self) -> Union[0, 1]:
```
mpharrigan, line 253: cool! Add a little unit test?

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
```diff
@@ -11,61 +11,58 @@
...
+    """Build the control system for a bloq with a specialized single-qubit controlled variant.
+
+    Args:
+        bloq: The current bloq of which a controlled version is requested.
+        ctrl_spec: The control specification
+        current_control_bit: The control bit of the current bloq, one of `0, 1, None`.
+        bloq_with_ctrl: The variant of this bloq with control bit `1`.
+        ctrl_reg_name: The name of the control qubit register.
+        bloq_without_ctrl: The variant of this bloq without a control.
+        bloq_with_ctrl_0: (optional) the variant of this bloq controlled by a qubit in the 0 state.
```
mpharrigan, line 44: what happens if you don't provide this (please document)

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
```diff
@@ -11,61 +11,58 @@
...
 ) -> tuple['Bloq', 'AddControlledT']:
+    """Build the control system for a bloq with a specialized single-qubit controlled variant.
+
+    Args:
+        bloq: The current bloq of which a controlled version is requested.
+        ctrl_spec: The control specification
+        current_control_bit: The control bit of the current bloq, one of `0, 1, None`.
+        bloq_with_ctrl: The variant of this bloq with control bit `1`.
+        ctrl_reg_name: The name of the control qubit register.
+        bloq_without_ctrl: The variant of this bloq without a control.
```
mpharrigan, line 43: is this necessary? or more specifically: do we need all three of: `bloq`, `current_control_bit`, `bloq_without_ctrl`? maybe we can get rid of `bloq`? I just started looking at this so I haven't read the code yet; just my instinct.

edit: I think we can get rid of `bloq`

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
```diff
@@ -11,61 +11,58 @@
...
+    ctrl_reg_name: str,
+    bloq_without_ctrl: 'Bloq',
+    bloq_with_ctrl_0: Optional['Bloq'] = None,
 ) -> tuple['Bloq', 'AddControlledT']:
+    """Build the control system for a bloq with a specialized single-qubit controlled variant.
+
+    Args:
+        bloq: The current bloq of which a controlled version is requested.
+        ctrl_spec: The control specification
+        current_control_bit: The control bit of the current bloq, one of `0, 1, None`.
```
mpharrigan, line 40: I'd recommend keeping the abbreviation everywhere `current_ctrl_bit`

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
mpharrigan: ah now my comment shall be updated to "please document in the docstring"

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
```diff
@@ -0,0 +1,97 @@
...
+        bloq_with_ctrl: The variant of this bloq with control bit `1`.
+        ctrl_reg_name: The name of the control qubit register.
+        bloq_without_ctrl: The variant of this bloq without a control.
+        bloq_with_ctrl_0: (optional) the variant of this bloq controlled by a qubit in the 0 state.
+    """
+    from qualtran import Bloq, CtrlSpec, Soquet
```
mpharrigan, line 46: my IDE says the Bloq import is not used

mpharrigan (commented):
cool! What do you think? I think I prefer this style. Would be nice to see one of the examples upgraded to use the just-a-function version

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control_test.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def get_ctrl_system(self, ctrl_spec: 'CtrlSpec') -> Tuple['Bloq', 'AddControlledT']:
+        return get_ctrl_system_for_bloq_with_specialized_single_qubit_control(
+            self,
+            ctrl_spec,
+            current_control_bit=self.cv,
+            bloq_with_ctrl=attrs.evolve(self, cv=1),
+            ctrl_reg_name='ctrl',
+            bloq_without_ctrl=attrs.evolve(self, cv=None),
+            bloq_with_ctrl_0=attrs.evolve(self, cv=0),
```
mpharrigan, line 58: I guess one complaint would be that this makes a couple copies of the bloq, when we really only need one(-ish?) of them. The attrs.evolve calls could be seen as redundant as well. What if we have a base function that accepts a callable and then two variants of the interface that provide a friendlier(?) interface to the callable version

```python
def get_special_1bit_ctrl(unctrl_bloq, get_ctrl_bloq: Callable[]):
  ctrl_bloq = get_ctrl_bloq(ctrl_spec.get_single_ctrl_bit())

def get_attrs_special_1bit_ctrl(unctrl_bloq, cv_attr_name: str = 'cv')
  return get_special_1bit_ctrl(unctrl_bloq, lambda x: attrs.evolve(unctrl_bloq, **{cv_attr_name: x}))

def get_special_1bit_ctrl_from_list(unctrl_bloq, bloq_with_1_ctrl, bloq_with_0_ctrl):
  return get_special_1bit_ctrl(unctrl_bloq, lambda x: {0: bloq_with_1_ctrl, 1: bloq_with_0_ctrl, None: unctrl_bloq}[x])
```

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
anurudhp: it falls back to the default controlled by 1. added to docstring

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
anurudhp: removed `bloq`

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
anurudhp: done

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
anurudhp: fixed

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control_test.py
anurudhp: This makes sense. The main method now takes a callable, and I added a helper (third case) to accept the bloq variants.

The second case seems like more work than it is worth to support, as the control register name may depend on the bloq etc. I've updated the examples to use the new helper (third case) instead.

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
```diff
@@ -0,0 +1,149 @@
...
+) -> tuple['Bloq', 'AddControlledT']:
+    """Build the control system for a bloq with a specialized single-qubit controlled variant.
+
+    Uses the provided specialized implementation when a singly-controlled variant of the bloq is
+    requested. When controlled by multiple qubits, the controls are reduced to a single qubit
+    and the singly-controlled bloq is used.
+
+    The user can provide specializations for the bloq controlled by `1` and (optionally) by `0`.
+    The specialization for control bit `1` must be provided.
+    In case a specialization for a control bit `0` is not provided, the default fallback instead,
```
mpharrigan, line 38: instead -> is used instead"""

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def _get_default_fallback():
+        current_bloq: 'Bloq'
+        if current_ctrl_bit is None:
+            current_bloq = bloq_without_ctrl
+        else:
+            ctrl_bloq_and_reg = get_ctrl_bloq_and_ctrl_reg_name(current_ctrl_bit)
+            if ctrl_bloq_and_reg is None:
+                raise ValueError(
+                    f"Expected a controlled bloq (self) matching the current control bit {current_ctrl_bit}, got None"
```
mpharrigan, line 60: what does (self) mean here?

mpharrigan: I'm actually generally confused about this error message. From the docstring it sounds like `get_ctrl_bloq_and_ctrl_reg_name` is allowed to return None. But not if current_ctrl_bit is set to something?

mpharrigan (commented):
This is now really elegant, imo. You can see the flow of information through the function calls clear as day. 

A couple small nits, then LGTM _except_

We've got to pare down the name somehow :p The module name is very long. 
```python
        from qualtran.bloqs.mcmt.bloq_with_specialized_single_qubit_control import (
            get_ctrl_system_for_bloq_with_specialized_single_qubit_control,
        )
```

this is an absolute unit of an import line haha. Let me brainstorm some ideas. This function is a helper function for `get_ctrl_system` for the case where the user has a way of constructing a single-bit controlled version of the bloq. The module lives in `qualtran.mcmt` because it handles multi-controlled stuff and uses and ladders. What parts of the name can be removed without causing ambiguity? We can drop "bloq". I don't think "specialized" adds much. Single qubit seems important, but maybe we can abbreviate to `1bit`. `get_ctrl_system` should probably be kept verbatim since it is meant to be used as a one-liner within a method of that name. "control" can be "ctrl". 

`qualtran.mcmt.specialized_ctrl.get_ctrl_system_1bit` would be my overall suggestion. Or maybe `get_ctrl_system_1cv` or `one_cv` or `1bit_cv`

qualtran/bloqs/mcmt/bloq_with_specialized_single_qubit_control.py
anurudhp: it can only return `None` for `cv=0`. For `cv=1` a specialization must be provided. The type is a bit confusing now after merging the two cases into a single callable, but I don't think there's a cleaner way of doing it (using a single callable)

anurudhp: One option could be to make the main function private, and only expose the helpers:
```py
def _get_ctrl_sys(unctrl_bloq, ctrl_getter: Callable[[ControlBit], Optional[Bloq]]): ...

def get(unctrl_bloq, ctrl_getter: Callable[[ControlBit], Bloq]): ...
def get_from_list(unctrl_bloq, ctrl_bloq: Bloq, ctrl_0_bloq: Optional[Bloq]=None): ...
```

anurudhp: Ah sorry looks like I misunderstood your question.

`get_ctrl_bloq_and_ctrl_reg_name(current_ctrl_bit)` should return the bloq from which this is called (i.e. `self` of the caller), so it doesn't make sense if it returns `None`.

anurudhp:
I simplified it further - avoided passing `bloq_without_ctrl` by instead using a intermediate bloq to construct CCU from CU. (previously I was using `ControlledViaAnd(U, [cv1, cv2])` which inturn used `U.get_ctrl_system()`).

Now one needs to only pass the current ctrl bit, and a getter which takes `0` or `1` and returns the ctrl bloq (and ctrl reg name).

qualtran/bloqs/mcmt/specialized_ctrl.py
```diff
@@ -0,0 +1,255 @@
...
+        current_ctrl_bit: The control bit of the current bloq, one of `0, 1, None`.
+        get_ctrl_bloq_and_ctrl_reg_name: A callable that accepts a control bit (`0` or `1`),
+            and returns the controlled variant of this bloq and the name of the control register.
+    """
+    return _get_ctrl_system_1bit_cv(
+        bloq,
+        ctrl_spec,
+        current_ctrl_bit=current_ctrl_bit,
+        get_ctrl_bloq_and_ctrl_reg_name=get_ctrl_bloq_and_ctrl_reg_name,
+    )
```
mpharrigan, line 216: this is literally just the same method? why not just make the private one public? is it because you wanted a different, private docstring?

qualtran/bloqs/mcmt/specialized_ctrl.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+def get_ctrl_system_1bit_cv_from_bloqs(
+    bloq: 'Bloq',
+    ctrl_spec: 'CtrlSpec',
+    *,
+    current_ctrl_bit: Optional['ControlBit'],
+    bloq_with_ctrl: 'Bloq',
```
mpharrigan, line 224: is this now just `ctrl_bloq`? I think the "with" phrasing was to support passing in both positive and negative control

qualtran/bloqs/mcmt/specialized_ctrl.py
anurudhp: The public method does not have `Optional` in the return type of `get_ctrl_bloq`.

qualtran/bloqs/mcmt/specialized_ctrl.py
anurudhp: yes, this is the controlled variant of the bloq. I wanted to not confuse it with `bloq.controlled()` (i.e. when `current_ctrl_bit` is `1`, `bloq_with_ctrl` is `bloq`), so was hesitant to name it `ctrl_bloq`. But I think `ctrl_bloq` is also okay.

mpharrigan: yeah that might be worth keeping distinct

qualtran/bloqs/mcmt/specialized_ctrl.py
mpharrigan: I see

mpharrigan (approved):
let's merge this puppy!

PR #1455, opened 2024-10-08
Single-qubit and Controlled Z rotations

### LINKED ISSUE #1446: BloqAsCirqGate.__pow__ probably broken
It will return `BloqAsCirqGate(Power(...))` but `BloqAsCirqGate` asserts that the wrappee is not a `cirq.Gate` but `Power` is a `GateWithRegisters`.

mpharrigan:
Update single-qubit Z rotations `ZPowGate` and `Rz`. These are atomic operations.

Introduce actual constructions for `CZPowGate` and `CRz` for better resource counting. The previous hack would just count them as "one rotation". Now the former is a rotation and two Toffolis (TODO: can these be `And`?); the latter is two rotations and two cliffords. 

Connect the single-qubit gates to their controlled versions via the controlled protocol. 

There's still hacks for Controlled-S and Controlled-T, which I will port in a follow-up. 

How should we handle X and Y rotations? One possibility is to make them compile to ZPow/Rz. Otherwise: duplicate the controlled construction for each flavor.

@tanujkhattar there are two tests failing that I'd like your help with

 - the TextbookQFT uses cirq `controlled_by` which is causing breaking simulation tests in a weird way. This triggers #1446 but even after hack-fixing that it runs into a different issue. Switching to qualtran tensor contraction fixes the test but there may be outstanding issues
 - I don't understand the flamegraph format. `test_get_flame_graph_data_qft_textbook` has different numbers now but I don't know how to verify that these are correct. 
 - fixes #1446


anurudhp:
> TODO: can these be And?

Yep you can use CZPow = And^dagger . ZPow . And

qualtran/bloqs/qft/qft_text_book_test.py
```diff
@@ -42,23 +44,28 @@ def test_qft_text_book(n: int, without_reverse: bool):
...
 @pytest.mark.parametrize('n', [10, 123])
 def test_qft_text_book_t_complexity(n: int):
     qft_bloq = QFTTextBook(n)
-    qft_t_complexity = qft_bloq.t_complexity()
-    assert qft_t_complexity.rotations == (n * (n - 1)) // 2
-    assert qft_t_complexity.t == 0
+    qlt_testing.assert_equivalent_bloq_counts(qft_bloq, generalizer=[ignore_split_join])
+    gate_counts = get_cost_value(qft_bloq, QECGatesCost())
+    # special angle ZPow gets turned into clifford or T
+    rots = ((n - 3) * (n - 2)) // 2
```
mpharrigan, line 56: I don't know why this is failing specifically for `n=123` (and by a lot). There's (n-1) phase gradients, each of size `i` and we want to skip the first two (S, T). cc @tanujkhattar

qualtran/bloqs/basic_gates/rotation.py
```diff
@@ -325,6 +461,83 @@ def __str__(self):
...
+
+    def build_composite_bloq(
+        self, bb: 'BloqBuilder', ctrl: 'Soquet', q: 'Soquet'
+    ) -> Dict[str, 'SoquetT']:
+        from qualtran.bloqs.basic_gates import CNOT
+
+        t = self.angle / np.pi
+        q = bb.add(ZPowGate(t / 2), q=q)
+        ctrl, q = bb.add(CNOT(), ctrl=ctrl, target=q)
+        q = bb.add(ZPowGate(-t / 2), q=q)
```
tanujkhattar, line 522: Please propagate `eps` to the `ZPowGate`s as `self.eps / 2`

qualtran/bloqs/qft/qft_text_book_test.py
mpharrigan: I didn't look at the individual ones. Do you mind taking a look?

mpharrigan:
@tanujkhattar do you have any thoughts on the cirq simulation failure

qualtran/bloqs/qft/qft_text_book_test.py
mpharrigan: oooh interesting

qualtran/_infra/controlled.py
```diff
@@ -516,3 +516,34 @@ def _circuit_diagram_info_(
...
+        def adder(
+            bb: 'BloqBuilder', ctrl_soqs: Sequence['SoquetT'], in_soqs: dict[str, 'SoquetT']
+        ) -> tuple[Iterable['SoquetT'], Iterable['SoquetT']]:
+            (ctrl_soq,) = ctrl_soqs
+            soqs = {ctrl_reg_name: ctrl_soq} | in_soqs
+            soqs = bb.add_d(ctrl_bloq, **soqs)
+            ctrl_soqs = [soqs.pop(ctrl_reg_name)]
+            return ctrl_soqs, soqs.values()
+
+        return ctrl_bloq, adder
```
anurudhp, line 549: I think this can be replaced with the new `get_ctrl_system_1bit_cv`

qualtran/bloqs/basic_gates/rotation.py
mpharrigan: done

qualtran/bloqs/basic_gates/rotation.py
```diff
@@ -319,6 +466,83 @@ def __str__(self):
...
+    a = sympy.Symbol('a')
+    rz = Rz(a)
+    return rz
+
+
+_RZ_DOC = BloqDocSpec(bloq_cls=Rz, examples=[_rz], call_graph_example=None)
+
+
+@frozen
+class CRz(Bloq):
```
anurudhp, line 480: please override the ctrl system for this (and other controlled bloqs), and add a test to verify gate counts for `CRz().controlled()` (should be `CRz() + And() + And.adjoint()`)

mpharrigan: why would I need to override the ctrl system? there isn't a native CCRz bloq

anurudhp: Because right now, `CRz.controlled` would return `Controlled(CRz)`, i.e. does not merge the two controls using `And`. 

The second case of `get_ctrl_system_1bit_cv` essentially handles this

anurudhp: (sync discussion) I did not notice the decomposition, in which case the override is not needed.

But I still think it's good to override to get a much more cost effective circuit. The current `CRz(2*pi*t).controlled()` gives
download
`toffoli: 2, and_bloq: 2, clifford: 2, rotation: 2, measurement: 2`

But with the override it becomes
download
`and_bloq: 1, clifford: 3, rotation: 2, measurement: 1`

---
full code

```py
import attrs
import numpy as np
import sympy

from qualtran import Bloq, DidNotFlattenAnythingError
from qualtran.bloqs.basic_gates import CRz
from qualtran.bloqs.mcmt import And
from qualtran.resource_counting import get_cost_value, QECGatesCost

def display_bloq_and_cost(bloq, *, flat=True):
    def _pred(binst):
        return not binst.bloq_is(And)
        
    cbloq = bloq.as_composite_bloq()
    if flat: cbloq = cbloq.flatten(pred=_pred)

    show_bloq(cbloq, 'latex')
    cost = get_cost_value(bloq, QECGatesCost())
    print(cost)

theta = 2 * np.pi * sympy.Symbol("t")
crz = CRz(theta)
display_bloq_and_cost(crz.controlled())

@attrs.frozen
class CRzWithOverride(CRz):
    def get_ctrl_system(self, ctrl_spec):
        from qualtran.bloqs.mcmt.specialized_ctrl import get_ctrl_system_1bit_cv_from_bloqs

        return get_ctrl_system_1bit_cv_from_bloqs(
            self,
            ctrl_spec,
            current_ctrl_bit=1,
            bloq_with_ctrl=self,
            ctrl_reg_name='ctrl'
        )

display_bloq_and_cost(CRzWithOverride(theta).controlled())
print(cost.total_t_count())
```

mpharrigan: that can be done in a follow-up. I don't think it's worth blocking this PR any more than it already has been

qualtran/bloqs/reflections/reflection_using_prepare.py
```diff
@@ -187,7 +187,9 @@ def build_call_graph(self, ssa: 'SympySymbolAllocator') -> 'BloqCountDictT':
         if self.global_phase != 1:
             phase_op: Bloq = GlobalPhase.from_coefficient(self.global_phase, eps=self.eps)
             if self.control_val is not None:
-                phase_op = phase_op.controlled(ctrl_spec=CtrlSpec(cvs=self.control_val))
+                phase_op = phase_op.controlled()
+                if self.control_val == 0:
+                    costs[XGate()] = 2
```
anurudhp, line 192: should this be `+= 2`? (as there's already an `XGate()` in L186)

anurudhp (approved):

qualtran/bloqs/reflections/reflection_using_prepare.py
mpharrigan: that's guarded by `if control_val is None` and this line test `if control_val == 0`

NoureldinYosri (approved):

PR #1460, opened 2024-10-09
Loosen coupling to Cirq 2024-10

mpharrigan:
Qualtran generally depends on `cirq`, but we usually take a little bit of care to move `cirq` to local imports where it's only needed if you're doing really cirq-y things (like converting to a cirq op). This PR does general maintenance to move cirq imports to local imports in `_infra/` submodules and a couple other easy places. 

This PR also patches in changes to some of the `_infra/` test modules (and a couple other easy places) to use `pytest.importorskip('cirq')` inside tests that require cirq. In general, we'll always have tests that depend on cirq because it's a battle-tested "reference implementation" but with these changes you can run a couple sanity checks on modules without hard dependencies on cirq. Otherwise pytest just fails to collect the tests and gives up immediately.


qualtran/symbolics/types.py
```diff
@@ -15,17 +15,16 @@
...
-from cirq._doc import document
 from typing_extensions import TypeIs
 
 SymbolicFloat = Union[float, sympy.Expr]
-document(SymbolicFloat, """A floating point value or a sympy expression.""")
+"""A floating point value or a sympy expression."""
 
 SymbolicInt = Union[int, sympy.Expr]
-document(SymbolicFloat, """A floating point value or a sympy expression.""")
+"""A floating point value or a sympy expression."""
```
anurudhp, line 24:
```suggestion
"""An integer value or a sympy expression."""
```

tanujkhattar (approved):

PR #1464, opened 2024-10-11
Create KaliskiModInverse

NoureldinYosri:
Implemented Kaliski modular inversion from appendix C5 https://arxiv.org/abs/2302.06639. The Toffoli complexity is $26n^2 + 7n - 1$ is higher than the formula in Litenski of $26n^2 + 2n$ because litenski drops constant factors in his estimates ... but these constant factors get multiplied by 2n (the number of iterations).

That's the cost of a single iteration is $13n + 3$ but litenski calculates it as $13n$


qualtran/bloqs/mod_arithmetic/mod_division_test.py
```diff
@@ -0,0 +1,86 @@
...
+def test_kaliski_mod_inverse_classical_action(bitsize, mod):
+    blq = KaliskiModInverse(bitsize, mod)
+    cblq = blq.decompose_bloq()
+    p2 = pow(2, bitsize, mod)
+    for x in range(1, mod):
+        if math.gcd(x, mod) != 1:
+            continue
+        x_montgomery = (x * p2) % mod
+        inv_x = pow(x, -1, mod)
+        inv_x_montgomery = (inv_x * p2) % mod
```
mpharrigan, line 37: can you use @fpapa250 's new utilities `QMontgomery.to_montgomery()` or whatever it's called to make this more clear/explicit?

qualtran/bloqs/mod_arithmetic/mod_division.py
```diff
@@ -0,0 +1,660 @@
...
+        b ^= a
+        if f:
+            assert v % 2 == 0, f'{u=} {v=} {r=} {s=} {a=} {b=} {m=} {f=}'
+            v >>= 1
+        r = (r << 1) % self.mod
+        if a:
+            u, v = v, u
+            s, r = r, s
+        a ^= s == 0
+        return {'u': u, 'v': v, 'r': r, 's': s, 'a': a, 'b': b, 'm': m, 'f': f}
```
mpharrigan, line 484: This is a broader comment, but attaching it to these lines since this is the place where it's most acute.

Is this just a transcription of the individual bloq operations? ideally the classical implementation would be inspectably-correct so that we can write effective unit tests against it. If possible, we would like to avoid a unit test where one inscrutable routine produces the same output as another inscrutable routine (although this is better than nothing!). Some concrete suggestions and requests:

 - Can we have a unit test that proves that the bloqs in this PR are computing a modular inverse? You can take a bunch of numbers, run them through the bloq simulator, and assert that `x * (x^-1) = 1` or use `pow(x, -1, p)` to validate the output.
 - Maybe some inline comments in this `on_classical_vals`. Presumably there's some method to this madness :)

mpharrigan (commented):
generally lgtm, but curious to hear your response to my comment

qualtran/bloqs/mod_arithmetic/mod_division.py
NoureldinYosri: The reason  I wrote it like this is that I didn't want to write individual tests for the `_KaliskiIterationStepX` bloqs. Instead I prefered a blackbox style test where I test the thing that we will actually use `KaliskiModInverse` which produces two register, the first has the modular inverse and the second has the ancillas. The problem here is that the value in the ancilla register is not trivial.

Reimplementing this by calling the individual _KaliskiIterationStepX.on_each is not a good idea since that will make the test useless. The test now compares the output of `_KaliskiIteration.on_classical_data` with the output of the individual steps. but if `_KaliskiIteration.on_classical_data` calls those steps then the test becomes useless.

----

> Can we have a unit test that proves that the bloqs in this PR are computing a modular inverse? 

That's what `test_kaliski_mod_inverse_classical_action` does. The modular inverse in motegomry form is such that $x_m x_m^{-1} = (x R) (x^{-1} R) R^{-1} = R = 2^{n} \mod p$ rather than $1$; it also comes from the identity $a_m \cdot b_m = (a \cdot b)_m$. I made the test clearer. 

Screenshot from 2024-10-29 10-42-51

> Maybe some inline comments in this on_classical_vals. Presumably there's some method to this madness

I added a docstring explaining what happens here.

qualtran/bloqs/mod_arithmetic/mod_division_test.py
NoureldinYosri: done

qualtran/bloqs/mod_arithmetic/mod_division.py
NoureldinYosri: > Maybe some inline comments in this on_classical_vals. Presumably there's some method to this madness

nvm I updated the method to follow the pseudocode of the original algorithm from Fig7b in https://arxiv.org/abs/2001.09580 which is alot more readable

mpharrigan (approved):
Thanks for the changes! I agree with all the motivation and blackbox style testing. Good to know you already had a "systems test" and even better to see it written out in a readable way now.

PR #1472, opened 2024-10-16
Hone Bloq<->Cirq adapters

mpharrigan:
CirqGateAsBloq only has bloq API. BloqAsCirqGate only has cirq API.

I also took the liberty of removing a level of indirection that had BloqAsCirqGate bopping over to composite_bloq.py only to immediately come back to the interop module. 

Changed CirqGateAsBloqBase to CirqGateAsBloqMixin and wrote about its intended use case(s).


mpharrigan:
subsumed by #1603

PR #1473, opened 2024-10-17
[qwa] Export docs for bloqs without examples

shab5:
Previously, the JSON export for the web app did not export documentation for Bloqs with no examples. This fixes that, which is a prerequisite for b/372965588.

### DIFF
--- a/dev_tools/ui-export.ipynb
+++ b/dev_tools/ui-export.ipynb
@@ -17,7 +17,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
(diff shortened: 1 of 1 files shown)


dev_tools/ui-export.ipynb
```diff
@@ -71,7 +72,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 3,
```
mpharrigan, line 75: can you "clear outputs" from jupyter

mpharrigan (approved):

Issue #1476, opened 2024-10-24
Change all integer arguments in FindECCPrivateKey bloq (and downstream bloqs) to SymbolicInt


fpapa250:
Higher level bloqs (from ECAdd up to FindECCPrivateKey) have int arguments instead of SymbolicInt arguments. I would like to merge the decomposition for ECWindowAddR before fixing this for all the bloqs.

PR #1477, opened 2024-10-24
Add Decomposition of ECWindowAddR

fpapa250:
- Adds working decomposition of ECWindowAddR (with classical simulation tests compatible with existing APIs)
- Updates higher level bloqs to have window size args for modular multiplication and windowed EC Add
- Temporary int() casting all over because upper level bloqs and APIs (QMontgomeryUInt functions) only accept ints and not SymbolicInts (tagged in issue: https://github.com/quantumlib/Qualtran/issues/1476, will fix in future PR)


qualtran/bloqs/factoring/ecc/ecc.ipynb
```diff
@@ -109,7 +111,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 3,
```
mpharrigan, line 114: please clear outputs

qualtran/bloqs/factoring/ecc/ec_point.py
```diff
@@ -49,6 +49,8 @@ def __add__(self, other):
         if (other.mod != self.mod) or (other.curve_a != self.curve_a):
             raise ValueError('Use consistent mod and curve')
 
+        if is_symbolic(self.x, self.y, other.x, other.y, self.mod, self.curve_a):
+            return self
```
mpharrigan, line 53: this is misleading. This should be an error probably

qualtran/bloqs/factoring/ecc/ec_phase_estimate_r.py
```diff
@@ -48,27 +50,54 @@ class ECPhaseEstimateR(Bloq):
...
         ctrl = [bb.add(PlusState()) for _ in range(self.n)]
-        for i in range(self.n):
-            ctrl[i], x, y = bb.add(ECAddR(n=self.n, R=2**i * self.point), ctrl=ctrl[i], x=x, y=y)
+
+        if self.ec_window_size == 1:
+            for i in range(self.n):
+                ctrl[i], x, y = bb.add(self.ec_add(R=2**i * self.point), ctrl=ctrl[i], x=x, y=y)
+        else:
+            ctrls = np.split(np.array(ctrl), self.n // self.ec_window_size)
+            for i in range(self.n // self.ec_window_size):
```
mpharrigan, line 87: optional: you can probably factor out `self.n // self.ec_window_size` as a quick lil property on this bloq

qualtran/bloqs/factoring/ecc/ec_phase_estimate_r.py
```diff
@@ -48,27 +50,54 @@ class ECPhaseEstimateR(Bloq):
     Args:
         n: The bitsize of the elliptic curve points' x and y registers.
         point: The elliptic curve point to phase estimate against.
+        ec_window_size: The number of bits in the ECAdd window.
```
mpharrigan, line 53: to my ear, these two parameters are named inconsistently. ECAdd -> ec; ModMul -> Mul.

mpharrigan: document what happens if window size is 1

qualtran/bloqs/factoring/ecc/ec_add_r_test.py
```diff
@@ -12,16 +12,59 @@
...
+    qlt_testing.assert_equivalent_bloq_counts(bloq, [ignore_alloc_free, ignore_split_join])
+
+
+@pytest.mark.parametrize(
+    ['n', 'm'], [(n, m) for n in range(7, 9) for m in range(1, n + 1) if n % m == 0]
+)
+@pytest.mark.parametrize('a,b', [(15, 13), (0, 0)])
+@pytest.mark.parametrize('x,y', [(15, 13), (5, 8)])
+@pytest.mark.parametrize('ctrl', [0, 1, 5, 8])
+def test_ec_window_add_r_classical(n, m, ctrl, x, y, a, b):
```
mpharrigan, line 60: can you mark this as slow (even though its on the cusp) and add a "quick" check

mpharrigan (commented):
cool! some comments

qualtran/bloqs/factoring/ecc/ec_point.py
fpapa250: The reason I did this was to keep tests for the ec_phase_estimate example bloqs. If we change it to raise an error the bloq_autotester fails. I figured if it was symbolic it doesn't really matter what the output is. For now I made it less misleading - if you want me to change to raise an error, I can do that, just have to forego the example bloq tests.

mpharrigan: where are we adding symbolic EC points in the bloq examples?

mpharrigan: I see: we manipulate the ECPoint in the decomposition. So is the regression due to the insertion of `int()` wrappers?

mpharrigan: PTAL at the commit I pushed

fpapa250: Yes exactly, we have to use the int wrappers to avoid a mypy typing error and then when there is an int() wrapper on a symbolic value we now also need to have a special case to deal with symbolic values before the wrapper. To remedy that I just added a symbolics check and return a new symbolic value as the result of the addition. Solves all the above problems.

fpapa250: LGTM

qualtran/bloqs/factoring/ecc/ec_add_r.py
```diff
@@ -113,33 +130,133 @@ class ECWindowAddR(Bloq):
...
+    def on_classical_vals(self, ctrl, x, y) -> Dict[str, Union['ClassicalValT', sympy.Expr]]:
+        # TODO(https://github.com/quantumlib/Qualtran/issues/1476): make ECAdd accept SymbolicInt.
+        A = ECPoint(
+            QMontgomeryUInt(self.n).montgomery_to_uint(int(x), int(self.R.mod)),
+            QMontgomeryUInt(self.n).montgomery_to_uint(int(y), int(self.R.mod)),
+            mod=self.R.mod,
+            curve_a=self.R.curve_a,
+        )
+        ctrls = QUInt(self.n).from_bits(ctrl)
+        result: ECPoint = A + (ctrls * self.R)
```
NoureldinYosri, line 253: asking for a clarification, can it happen that ctrls be 5 and we would compute `A + 5 * self.R` ?

fpapa250: Yes, the circuit uses QROM to load in what 5*self.R is and then adds that to x,y and we window it into groups of window_size ctrl bits

qualtran/bloqs/factoring/ecc/ec_add_r.py
```diff
@@ -113,33 +130,133 @@ class ECWindowAddR(Bloq):
...
     References:
         [How to compute a 256-bit elliptic curve private key with only 50 million Toffoli gates](https://arxiv.org/abs/2306.08585).
         Litinski. 2013. Section 1, eq. (3) and (4).
     """
 
     n: int
-    window_size: int
     R: ECPoint
+    add_window_size: int
+    mul_window_size: int = 1
```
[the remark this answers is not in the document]
fpapa250: Correct

mpharrigan (approved):

PR #1479, opened 2024-10-29
Add `SparseMatrixHermitian` block-encoding

anurudhp:
Implements a Hermitian block-encoding $U$ of a sparse matrix $A$. i.e. the unitary $U$ is Hermitian, or equivalently, a reflection. This is useful to directly build an efficient walk operator, for example in #1348 

Ref: https://arxiv.org/abs/2201.08309, Ch. 6.5. Proposition 6.8, Fig 6.7.


anurudhp:
@tanujkhattar @fdmalone ptal!

mpharrigan (approved):
Thanks! @tanujkhattar may have some comments later, which I endeavor you to address if and when they occur; but to keep things moving, I'm going to LGTM and merge this

Issue #1266, opened 2024-08-08
QubitizationWalkOperator should be a block encoding?


fdmalone:
Need to consider control registers for BlockEncoding base class potentially.

PR #1481, opened 2024-10-29
Simplify `QubitizationWalkOperator`: remove specializations for control/adjoint and use defaults instead 

### LINKED ISSUE #1480: `Adjoint` prevents specialized control bloqs from being constructed
The default `Adjoint` wrapper does not propagate controls to its subbloq. And rightly so in general, as `Controlled.adjoint` pushes the adjoint to the subbloq, so enabling both would lead to an infinite recursion.

But in cases where we provide a specialized control (see e.g. #1451), it would be desirable if `bloq.adjoint().controlled()` simplified to `Adjoint(bloq.controlled())`.

To illustrate, `.controlled().adjoint()` works as expected:
```py
adj_ctrl_bloq = BloqWithSpecializedCtrl().controlled().adjoint()
g, _ = adj_ctrl_bloq.call_graph()
show_call_graph(g)
```
download

But `.adjoint().controlled()` fails to use the specialization:
```py
ctrl_adj_bloq = BloqWithSpecializedCtrl().adjoint().controlled()
g, _ = ctrl_adj_bloq.call_graph()
show_call_graph(g)
```
download

In cases where a simple adjoint is known (e.g. self-adjoint), we can override it. But otherwise, the default solution has been to hack in a `is_adjoint: bool` parameter, but this doubles the amount of code, especially in `build_composite_bloq`. For example, https://github.com/quantumlib/Qualtran/blob/7c6715b7e64de12257b3959faa3caff5c3cdfd8b/qualtran/bloqs/qubitization/qubitization_walk_operator.py#L150-L155

**my suggestion** We can provide a new `AdjointSpecialized` metabloq that pushes the controls into the subbloq, but this must be used with care - only for bloqs with provide a specialized single-bit-ctrl implementation.

(ran into this in #1481)

---
full code to reproduce the issue:
```py
from typing import Optional
import attrs
from qualtran import Bloq, Signature, QAny
from qualtran.bloqs.basic_gates import TGate, XGate, CNOT
from qualtran.drawing import show_call_graph
from qualtran.resource_counting import QECGatesCost

@attrs.frozen
class BloqWithSpecializedCtrl(Bloq):
    """Implements the unitary `T^\dagger X T`"""

    cv: Optional[int] = None

    @property
        n_ctrl = 1 if self.cv is not None else 0
        return Signature.build_from_dtypes(
            ctrl=QAny(n_ctrl),
            target=QAny(1),
        )

    def build_composite_bloq(self, bb, target, **soqs):
        ctrl = soqs.pop("ctrl", None)

        target = bb.add(TGate(), q=target)

        if self.cv is None:
            target = bb.add(XGate(), q=target)
        else:
            if self.cv == 0:
                ctrl = bb.add(XGate(), q=ctrl)
            ctrl, target = bb.add(CNOT(), ctrl=ctrl, target=target)
            if self.cv == 0:
                ctrl = bb.add(XGate(), q=ctrl)

        target = bb.add(TGate().adjoint(), q=target)

        out_soqs = {"target": target}
        if self.cv is not None:
            out_soqs |= {"ctrl": ctrl}
        return out_soqs

    def get_ctrl_system(self, ctrl_spec: "CtrlSpec") -> tuple["Bloq", "AddControlledT"]:
        from qualtran.bloqs.mcmt.specialized_ctrl import get_ctrl_system_1bit_cv

        return get_ctrl_system_1bit_cv(
            self,
            ctrl_spec,
            current_ctrl_bit=self.cv,
            get_ctrl_bloq_and_ctrl_reg_name=lambda cv: (
                attrs.evolve(self, cv=cv),
                "ctrl",
            ),
        )

    def __str__(self):
        return "U" if self.cv is None else f"C_{self.cv}[U]"

adj_ctrl_bloq = BloqWithSpecializedCtrl().controlled().adjoint()
g, _ = adj_ctrl_bloq.call_graph()
show_call_graph(g)

ctrl_adj_bloq = BloqWithSpecializedCtrl().adjoint().controlled()
g, _ = ctrl_adj_bloq.call_graph()
show_call_graph(g)
```

anurudhp:
towards #1266 

- `QubitizationWalkOperator` does not need a specialized control, it is handled automatically by the select and reflect ops it decomposes to. The bloq/cirq diagrams also match the default in `Controlled`.
- Implement ctrl systems for basic bloqs `TwoBitSwap`, `ZGate`, `YGate`.
-  [new] Add a `AdjointWithSpecializedCtrl` metabloq, which pushes single bit controls to the subbloq. This fixes #1480 
- Above is used in `SelectHubbard`. Also fix a bug in it: the decomp does not account for `control_val=0`, so use default for that case.


anurudhp:
@mpharrigan @tanujkhattar @fdmalone

qualtran/bloqs/basic_gates/z_basis.py
```diff
@@ -254,20 +254,17 @@ def my_tensors(
...
     def get_ctrl_system(self, ctrl_spec: 'CtrlSpec') -> Tuple['Bloq', 'AddControlledT']:
-        if ctrl_spec != CtrlSpec():
-            # Delegate to the general superclass behavior
-            return super().get_ctrl_system(ctrl_spec=ctrl_spec)
+        from qualtran.bloqs.bookkeeping import AutoPartition
+        from qualtran.bloqs.mcmt.specialized_ctrl import get_ctrl_system_1bit_cv_from_bloqs
 
-        bloq = CZ()
+        cz = CZ()
+        cz_wrapped = AutoPartition(
```
mpharrigan, line 261: this is kindof a regression, no? Before; you'd get a nice idiomatic CZ and now there's another level of indirection (?)

qualtran/bloqs/meta/adjoint_with_ctrl.py
mpharrigan, line 1: is this a new top-level module? if yes: it needs an init.py. I also don't really want a new top-level module for this. This seems "controlly". Can we put it in `mcmt` near the other controlly stuff?

mpharrigan (commented):
initial review, thanks for the ping

qualtran/bloqs/meta/adjoint_with_ctrl.py
anurudhp: Yes, I think we can put it in the `mcmt.specialized_ctrl` file, as this is only required in that case.

qualtran/bloqs/basic_gates/z_basis.py
mpharrigan: > Btw, in the get_ctrl_sys_1bit, I return the unwrapped ctrl_bloq, so you'd only get CZ in the call graph.

can you explain this more? where does it get unwrapped? in the first return value but not when used as part of the `adder` method?

> it's kinda annoying that the target registers of Z and CZ do not match

For generality, I think it's important to keep this case supported. 

My question is: is it worth porting these gates to `get_ctrl_system_1bit_cv_from_bloqs`? Can you explicitly list what we gain? Could we keep the old code and only delegate to the new function for previously-unhandled situations? 

The `specilized_ctrl` simplifications are great for bloq authors (our "users") to make it simple to quickly support a broad set of use cases. In core parts of the library (like a Z gate) the onus is instead on the developers of the library (i.e. us) to absorb a lot of the complexity to make the users experience as straightforward as possible. To me, that means not introducing any unnecessary partitioning into the circuit

qualtran/bloqs/mcmt/specialized_ctrl.py
```diff
@@ -253,3 +260,80 @@ def get_ctrl_bloq_and_ctrl_reg_name(cv: 'ControlBit') -> Optional[tuple['Bloq',
...
+    """
+
+    NONE = enum.auto()
+    ZERO = enum.auto()
+    ONE = enum.auto()
+    BOTH = ZERO | ONE
+
+
+@attrs.frozen()
+class AdjointWithSpecializedCtrl(Adjoint):
```
mpharrigan, line 278: this looks reasonable to me. Will need a helpful class docstring to motivate its use and document how to use it. Consider adding a small section to either/both the Controlled.ipynb or Adjoint.ipynb documentation notebooks

qualtran/bloqs/qubitization/qubitization_walk_operator.py
```diff
@@ -119,58 +116,30 @@ def junk_registers(self) -> Tuple[Register, ...]:
...
-            yield self.reflect.on_registers(**reflect_reg)
-
-    def get_single_qubit_controlled_bloq(self, control_val: int) -> 'QubitizationWalkOperator':
-        assert self.control_val is None
-        c_block = self.block_encoding.controlled(ctrl_spec=CtrlSpec(cvs=control_val))
-        if not isinstance(c_block, (SelectBlockEncoding, LCUBlockEncoding)):
-            raise TypeError(
-                f"controlled version of {self.block_encoding} = {c_block} must also be a SelectOracle"
-            )
-        return attrs.evolve(self, block_encoding=c_block, control_val=control_val)
```
mpharrigan, line 164: nice 😎

qualtran/bloqs/mcmt/specialized_ctrl.py
anurudhp: added docstring and a notebook `specialized_ctrl.ipynb` explaining the usage

mpharrigan (approved):

qualtran/bloqs/chemistry/hubbard_model/qubitization/select_hubbard.py
```diff
@@ -90,6 +90,8 @@ class SelectHubbard(SelectOracle):
     def __attrs_post_init__(self):
         if self.x_dim != self.y_dim:
             raise NotImplementedError("Currently only supports the case where x_dim=y_dim.")
+        if self.control_val == 0:
+            raise ValueError("control_val=0 not supported")
```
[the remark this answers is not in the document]
anurudhp: makes sense, I updated the error

PR #1486, opened 2024-11-04
Modify KaliskiModInverse to support zero

NoureldinYosri:
By construction KaliskiModInverse assumes its inputs are valid (i.e. gcd(x, mod) = 1). The case of zero is interesting however since for the bloq to be unitary it has to send zero to itself. the default circuit either does that or sends it to $p$ depending on whether the opreation $p - r$ is implemented as `ModNeg` $\mathcal{O}(2n)$ 
toffoli or as ~r -> AddK($p + 1$)  $\mathcal{O}(n)$ toffoli (which is what we do).

Changing the final step to use ModNeg solves the problem (but increases the toffoli costs by $n$). it also prevents us from freeing the $u$ and $s$  registers so we will have to keep them around for uncomputation. This means we keep $6n$ junk qubits.

Alternatively we can do this at no additional cost and by keeping only $4n$ junk qubits. The condition $x=0$ or more precisely $v=0$ is checked at the beginning of each iteration. so I just keep the result of that check for uncomputation. The case where $x=0$ corresponds to having the first check yield 1. Also the first check is 1 if and ony if $x=0$.

### DIFF
--- a/qualtran/bloqs/mod_arithmetic/mod_division.py
+++ b/qualtran/bloqs/mod_arithmetic/mod_division.py
@@ -17,7 +17,7 @@ from typing import cast, Dict, List, Optional, Tuple, TYPE_CHECKING, Union
 
 import numpy as np
 import sympy
-from attrs import evolve, frozen
+from attrs import evolve, field, frozen
 
 from qualtran import (
     Bloq,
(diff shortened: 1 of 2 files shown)


mpharrigan:
What does it mean to send it to mod p?

Can you clarify how this relates to the paper references: did they assume a `ModNeg`-based implementation? or is this overlooked in the paper implementation?

NoureldinYosri:
> What does it mean to send it to mod p?
bad formatting ... I meant to say that the result is either 0 or $p$ depending how the last step is done.

> did they assume a ModNeg-based implementation? 
yes they did. they either overlooked the that $u$ and $s$ can't be freed or they kept it regardless.  on the other hand the source of the circuit https://arxiv.org/abs/2302.06639, assumes valid inputs and so it frees those registers

mpharrigan (approved):

NoureldinYosri:
@mpharrigan the litenski paper overlooks that they can't free the $u$ and $s$ registers. the figure they have (fig 7) shows the terminal value that only happens when the input is valid 

image

Issue #1336, opened 2024-08-23
Tensor contraction 4-5x slower than `cirq.unitary` for hamiltonian simulation bloq


anurudhp:
`cirq.unitary(BloqAsCirqGate(W_e_iHt))` is 4-5x faster than `W_e_iHt.tensor_contract()`. Code used in: https://github.com/quantumlib/Qualtran/blob/0d9d97413e0d7bf49a3a70b553179af39a91b185/qualtran/bloqs/hamiltonian_simulation/hamiltonian_simulation_by_gqsp_test.py#L72-L82

Discussion: https://github.com/quantumlib/Qualtran/pull/1328#discussion_r1728052654

One possible optimization is to add 0 state/effect for ancilla and then tensor contract to directly get the block encoded matrix

mpharrigan:
I could roughly reproduce the original numbers and with some preliminary optimizations I can get

<img width="441" alt="Image" src="https://github.com/user-attachments/assets/41c2d204-4466-4d6b-8cc7-e72c28e11e42" />

**quimb thing.** I included a heuristic in the tensor contraction code to try to optimize the tensor network if the contraction width is sufficiently large. This is important for some large networks to make them tractable at all, but in these test cases the "optimization" was taking several seconds but naively contracting takes like 0.08s . Removing the tensor network simplifier shaves off a couple of seconds


**qualtran thing.** Most of the time is spent flattening the circuit. This involves `map_soqs` to re-wire-up the flattened version. This involves putting a lot of soquets in dictionaries. I sped up flattening by 1) cacheing an intermediate mapping dictionary rather than reconstructing it each time 2) changing the hash function of `Soquet`, which was itself using a lot of time... but I did this out of order so maybe (2) isn't necessary given (1). 

In the end, most of the time is still spent flattening the circuit and there's no part that's taking an obviously large amount of time, i.e. we're at the whims of Python.

These examples require flattening 9 times to get to the fully-flat form. We might be able to go faster by depth-first-searching the compute graph *as* we build the tensor network rather than factoring it into two steps: flatten; convert to quimb; and factoring that first step into 9 steps: `flatten_once` nine times. This would require more work

PR #1487, opened 2024-11-06
fix QPE test failure, add non-slow bloq example autotest

anurudhp:


anurudhp:
@mpharrigan @fdmalone ptal!

qualtran/_infra/controlled.py
```diff
@@ -458,9 +458,12 @@ def _unitary_(self):
             # subbloq is a cirq gate, use the cirq-style API to derive a unitary.
             import cirq
 
-            return cirq.unitary(
-                cirq.ControlledGate(self.subbloq, control_values=self.ctrl_spec.to_cirq_cv())
-            )
+            try:
```
fdmalone, line 461: Why are we just passing? This is because the subbloq might not be unitary right? Maybe we should at least print a warning to indicate the user is doing something unsupported etc.

fdmalone (approved):
LGTM % nit about this silent exception

qualtran/_infra/controlled.py
fdmalone: Re-reading, if a warning is too annoying (although I guess we could leverage logging here) at least add a comment to the source code

anurudhp: The ideal would be to always use `tensor_contract`, but at the moment, it's about 5-10x slower than `cirq.unitary` (e.g. #1336). So my interim fix was to try cirq, and use tensor if it failed.

I'll add this as a comment in the source code

fdmalone:
LGTM

anurudhp:
could you merge it? (I don't have the permissions)
thanks!

Issue #1461, opened 2024-10-10
Fix bugs in ECAdd Circuit


fpapa250:
There are a few bugs with the ECAdd circuit from https://arxiv.org/abs/2306.08585. The inputs and outputs are fine, but in steps 2, 5, and 6 some of the flag bits are freed before they are set correctly back to zero. The circuit shows that none of the qubits are freed when dirty, however this is not the case. The following bugs exist:

1. Step 2: lam = lam_r undoes f1 when it should only happen if lam is set to lam_r
2. Step 5: free dirty reg lam when a,b=x,y (not sure why)
3. Step 6: free dirty reg when x,y = 0,0
    1. when b = y = 0, but x != 0: f2 never gets flipped back
    2. when p1 = p2 and f4 not set f4 gets flipped on

PR #1489, opened 2024-11-07
Fix bugs in ECAdd bloq

fpapa250:
The circuit described in the paper had 4 bugs (https://github.com/quantumlib/Qualtran/issues/1461) which caused (with edge case inputs) flags f1, f2, f4 and lambda to be freed (considered = 0 in the paper) while having values > 0.

Bugs:
1 - Step 2: lam = lam_r undoes f1 when it should only happen if lam is set to lam_r (very rare case)
BUGFIX:
1. Set a new flag bit when computed lam = lam_r (before writing lam from computed z register to lam register)
2. Control existing equals bloq to clear f1 only when lam = lam_r AND the ancilla flag is set
3. Unset ancilla and free when z4 (containing computed lam) = lam_r

2 - Step 5: free dirty reg lam when a,b=x,y
BUGFIX:
1. set a new flag if x (denominator of lambda computation) = 0
2. Clear lam using lam_r controlled on ctrl and new flag
3. unset new flag when x = 0

3 - Step 6: free dirty reg when x,y = 0,0 and b = y = 0, but x != 0: f2 never gets flipped back
BUGFIX:
1. unset f2 when x,y = 0 and b = 0 OR a,b = 0 and y = 0

4 - Step 6: when p1 = p2 and f4 not set f4 gets flipped on
BUGFIX:
1. Move CModSub and CModAdd to before Equals(xy, ab) bloq


qualtran/bloqs/factoring/ecc/ec_add.py
```diff
@@ -32,6 +32,7 @@
     Soquet,
     SoquetT,
 )
+from qualtran._infra.controlled import CtrlSpec
```
anurudhp, line 35:
```suggestion
from qualtran import CtrlSpec
```

fpapa250:
@mpharrigan Here are the bug fixes to the ECAdd circuit that the paper didn't consider. They are thoroughly tested with the existing unit tests.

mpharrigan:
Nice!

The issue and PR description describes the modifications and why they were needed -- but I suspect only the most dedicated users will find them! Do you want to put a description of the modifications into the library somewhere? Ideally this would be in the docstrings of the public bloq classes so they'd get rendered into the docs

mpharrigan:
an alternative would be to have a dedicated jupyter notebook written as a mini-paper that goes through all the crypto bloqs and includes a section on the modifications. This would be more work but this is the sort of thing we can slap a byline on and get a zenodo doi to make it citable

qualtran/bloqs/factoring/ecc/ec_add.py
```diff
@@ -865,6 +913,21 @@ def build_composite_bloq(
...
+        a = bb.join(aby_arr[0], dtype=QMontgomeryUInt(self.n))
+        b = bb.join(aby_arr[1], dtype=QMontgomeryUInt(self.n))
+        y = bb.join(aby_arr[2], dtype=QMontgomeryUInt(self.n))
+
+        xyb_arr = np.concatenate([bb.split(x), bb.split(y), bb.split(b)])
+        xyb_arr, f2 = bb.add(MultiControlX(cvs=[0] * 3 * self.n), controls=xyb_arr, target=f2)
+        xyb_arr = np.split(xyb_arr, 3)
+        x = bb.join(xyb_arr[0], dtype=QMontgomeryUInt(self.n))
+        y = bb.join(xyb_arr[1], dtype=QMontgomeryUInt(self.n))
+        b = bb.join(xyb_arr[2], dtype=QMontgomeryUInt(self.n))
```
anurudhp, line 929: I think you can replace this with the default controlled to avoid manual splits and joins:
```py
mcx = XGate().controlled(CtrlSpec(qdtypes=QMontgomeryUInt(self.n), cvs=[0, 0, 0]))
[a, b, y], f2 = bb.add(mcx, ctrl=[a, b, y], q=f2)
[x, y, b], f2 = bb.add(mcx, ctrl=[x, y, b], q=f2)
```

Though I suspect the types _may_ not be propagated correctly yet. In case you try the above suggestion and it fails, could you please open an issue?

p.s. this would also enable decomposing for symbolic `self.n` which would be an added benefit.

fpapa250:
> an alternative would be to have a dedicated jupyter notebook written as a mini-paper that goes through all the crypto bloqs and includes a section on the modifications. This would be more work but this is the sort of thing we can slap a byline on and get a zenodo doi to make it citable

Sorry for the radio silence, took a nice break for the holidays. I'll find some time in the next two weeks to make a dedicated jupyter notebook. Sounds like a good idea!

qualtran/bloqs/cryptography/ecc/ec_add.py
```diff
@@ -729,9 +748,31 @@ def build_composite_bloq(
...
+            ctrl = ctrls[0]
+            ancilla = ctrls[1]
+            lam_r_split[i] = ctrls[2]
+        lam_r = bb.join(lam_r_split, dtype=QMontgomeryUInt(self.n))
+        lam = bb.join(lam_split, dtype=QMontgomeryUInt(self.n))
+        x_split, ancilla = bb.add(
+            MultiControlX(cvs=[0] * int(self.n)), controls=x_split, target=ancilla
+        )
+        x = bb.join(x_split, dtype=QMontgomeryUInt(self.n))
+        bb.free(ancilla)
```
anurudhp, line 774:
```suggestion
        clear_lam = (
            Xor(QMontgomeryUInt(self.n))
            .controlled(CtrlSpec(qdtypes=QMontgomeryUInt(self.n), cvs=0))
            .controlled()
        )
        ctrl, x, lam_r, lam = bb.add(clear_lam, ctrl1=ctrl, ctrl2=x, x=lam_r, y=lam)
```

You can also factor the `clear_lam` bloq into a class property, so it can also be used in the call graph.

qualtran/bloqs/cryptography/ecc/ec_add.py
```diff
@@ -771,6 +817,8 @@ def build_call_graph(self, ssa: SympySymbolAllocator) -> BloqCountDictT:
             KaliskiModInverse(bitsize=self.n, mod=self.mod).adjoint(): 1,
             ModAdd(self.n, mod=self.mod): 1,
             MultiControlX(cvs=[1, 1]): self.n,
+            MultiControlX(cvs=cvs): 2,
+            MultiControlX(cvs=[1, 1, 1]): self.n,
```
anurudhp, line 821: can be replaced by the above `clear_lam` bloq

qualtran/bloqs/cryptography/ecc/ec_add.py
```diff
@@ -939,33 +992,35 @@ def build_composite_bloq(
...
+            cvs2 = [0] * 2 * self.n
+            cvs3 = [0] * 3 * self.n
         else:
-            cvs = HasLength(2 * self.n)
+            cvs2 = HasLength(2 * self.n)
+            cvs3 = HasLength(3 * self.n)
         return {
-            MultiControlX(cvs=cvs): 1,
+            MultiControlX(cvs=cvs2): 1,
+            MultiControlX(cvs=cvs3): 2,
```
anurudhp, line 1015: is this the newly refactored `mcx` gate? if yes, replacing will also simplify the symbolic case

anurudhp (commented):
sorry for all the nitpicking, it can be done in a follow-up as well!

qualtran/bloqs/cryptography/ecc/ec_add.py
fpapa250: I tried replacing the code as suggested with the clear_lam bloq, but it gives me an error: qualtran._infra.composite_bloq.BloqError: During finalization, {Soquet(binst=BloqInstance(bloq=ControlledViaAnd(subbloq=Xor(dtype=QMontgomeryUInt(bitsize=8, modulus=None)), ctrl_spec=CtrlSpec(qdtypes=(QBit(), QMontgomeryUInt(bitsize=8, modulus=None)), cvs=(array(1), array(0)))), i=14), reg=Register(name='y', dtype=QMontgomeryUInt(bitsize=8, modulus=None), _shape=(), side=<Side.THRU: 3>), idx=())} Soquets were not used.

Do you know what's going wrong here? I'm not really understanding why it's complaining

anurudhp: Hey, could you tell me which test raises this error? I replaced it locally and the tests seem to all pass.

anurudhp: Did you accidentally remove the next line that frees `lam`? That may have caused the above issue

qualtran/bloqs/cryptography/ecc/ec_add.py
```diff
@@ -939,33 +998,32 @@ def build_composite_bloq(
         y = bb.join(xy_arr[1], dtype=QMontgomeryUInt(self.n))
 
         # Free all ancilla qubits in the zero state.
-        # TODO(https://github.com/quantumlib/Qualtran/issues/1461): Fix bugs in circuit where f1,
```
mpharrigan, line 942: make sure we remember to close this bug 🎉

qualtran/bloqs/cryptography/ecc/ec_add_test.py
```diff
@@ -417,12 +423,12 @@ def test_ec_add_symbolic_cost():
...
     # Litinski 2023 https://arxiv.org/abs/2306.08585
     # Based on the counts from Figures 3, 5, and 8 the toffoli count for ECAdd is 126.5n^2 + 189n.
-    # The following formula is 126.5n^2 + 195.5n - 31. We account for the discrepancy in the
+    # The following formula is 126.5n^2 + 215.5n - 34. We account for the discrepancy in the
     # coefficient of n by a reduction in the toffoli cost of Montgomery ModMult, an increase in the
     # toffoli cost for Kaliski Mod Inverse, n extra toffolis in ModNeg, 2n extra toffolis to do n
-    # 3-controlled toffolis in step 2. The expression is written with rationals because sympy
-    # comparison fails with floats.
-    assert total_toff == sympy.Rational(253, 2) * n**2 + sympy.Rational(407, 2) * n - 31
+    # 3-controlled toffolis in step 2, and a few extra gates added to fix bugs found in the
```
mpharrigan, line 429: "...to fix bugs found in the circuit (see class docstrings)"

mpharrigan (approved):
very nice 👌

mpharrigan:
Fixes #1461

qualtran/bloqs/cryptography/ecc/ec_add.py
fpapa250: Will do!

Issue #1488, opened 2024-11-06
`cirq.unitary` fails for bloq with `And` followed by `And^dagger`


anurudhp:
```py
from qualtran import Bloq, Signature
from qualtran.bloqs.mcmt import And
from qualtran.cirq_interop import BloqAsCirqGate

@attrs.frozen
class MyBloq(Bloq):
    @property
    def signature(self):
        return Signature.build(a=1, b=1)

    def build_composite_bloq(self, bb, a, b):
        [a, b], c = bb.add(And(), ctrl=[a, b])
        [a, b] = bb.add(And().adjoint(), ctrl=[a, b], target=c)
        return dict(a=a, b=b)

MyBloq().tensor_contract() # works
cirq.unitary(BloqAsCirqGate(MyBloq())) # fails
```

```
TypeError: cirq.unitary failed. Value doesn't have a (non-parameterized) unitary effect.

type: <class 'qualtran.cirq_interop._bloq_to_cirq.BloqAsCirqGate'>
value: BloqAsCirqGate(MyBloq)
```

mpharrigan:
What's the expected behavior? And^dag doesn't have a unitary effect. The tensor contraction projects the target into the zero state, which isn't physical -- but keeps everything in state-vector-land

anurudhp:
There are bloqs which use `And` in compute-uncompute pairs, which then fails with the cirq simulator (even state vector). I think some old `GateWithRegisters` code return a `cirq.ControlledGate` in some cases, but newer bloq code using `BloqAsCirqGate` or `ControlledViaAnd` does not always do this.

Ideally it would be good if such cases (compute-uncompute pairs) can work with the cirq simulators as expected.

mpharrigan:
It's tricky because And^dag is **not** a unitary operation. If we want to treat pairs of compute-uncompute, we'd need to have a mechanism for grouping such pairs of bloqs

mpharrigan:
The issue as written is fixed, but see #1763 for follow-on issues

PR #1490, opened 2024-11-07
cleanup remaining uses of old control specialization interface

anurudhp:
Removes the finally uses of the old `SpecializedSingleQubitControlledExtension` interface in favour of the new interface from #1451.

For controlled versions of `LCUBlockEncoding`, `SelectBlockEncoding` etc, do not update the subbloqs to controlled. Instead, use the appropriate subbloq/controlled-subbloq when decomposing/building call graph.


anurudhp:
@fdmalone

qualtran/bloqs/multiplexers/select_pauli_lcu.py
```diff
@@ -142,6 +142,9 @@ def get_ctrl_system(self, ctrl_spec: 'CtrlSpec') -> Tuple['Bloq', 'AddControlled
     def adjoint(self) -> 'Bloq':
         return self
 
+    def _has_unitary_(self):
```
fdmalone, line 145: Why is this necessary in this PR?

qualtran/bloqs/mean_estimation/mean_estimation_operator_test.py
```diff
@@ -63,7 +62,7 @@ class BernoulliEncoder(SelectOracle, SpecializedSingleQubitControlledExtension):
 
     @cached_property
     def control_registers(self) -> Tuple[Register, ...]:
-        return () if self.control_val is None else (Register('control', QBit()),)
+        return ()
```
fdmalone, line 65: Do we need to keep around this property if it's just empty? (I missed the older PRs so probably need to read those)

fdmalone (approved):
LGTM

qualtran/bloqs/mean_estimation/mean_estimation_operator_test.py
anurudhp: `SelectOracle` has this as an abstract property, but as there's no special controlled version I just return `()`.

Ideally the select/prepare interfaces should be updated to the newer `BlockEncoding` style interfaces (would be a major refactor), but as of now as callees use `.control_registers`, I'm keeping it as-is.

qualtran/bloqs/multiplexers/select_pauli_lcu.py
anurudhp: `cirq.has_unitary` fails to realize that this is unitary because when it decomposes, it gets `And^dagger` gates which aren't unitary (#1488)

mpharrigan (approved):
nice

PR #1492, opened 2024-11-10
Deprecate `MultiControlPauli`

### LINKED ISSUE #914: MultiControlPauli improvements

anurudhp:
All users now use the derived `MultiControlX` or `MultiControlZ` instead. I've kept the class around as it has a simpler interface (list of control bits instead of a CtrlSpec), and also supports zero controls (#679).

Fixes #914


anurudhp:
@mpharrigan ptal

mpharrigan:
usually, we'd keep the name as-is and use the `warnings.warn(..., DeprecationWarning)` thing to show that it's deprecated

mpharrigan (approved):

PR #1494, opened 2024-11-10
Fix QPE cirq simulation tests 

anurudhp:
Cirq simulator was unable to simulate `Controlled(QubitizationWalkOperator)`, so decompose one level further to get simulable gates.

Also made a subset fast tests to catch issues early on. Fixes the current Nightly test failure.

### DIFF
--- a/qualtran/bloqs/phase_estimation/qubitization_qpe_test.py
+++ b/qualtran/bloqs/phase_estimation/qubitization_qpe_test.py
@@ -74,6 +74,9 @@ def test_qubitization_phase_estimation_of_walk(num_terms: int, use_resource_stat
     )
     qpe_bloq = QubitizationQPE(walk, state_prep)
 
+    # TODO cirq simulation seems to fail for controlled `QubitizationWalkOperator`.
+    #      the following code decomposes a few levels till it gets only simulable bloqs.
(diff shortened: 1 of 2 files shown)


anurudhp:
@fdmalone ptal

mpharrigan:
is there a more fundamental issue at play here? this seems like a hacky fix on top of the existing tests

qualtran/bloqs/phase_estimation/qubitization_qpe_test.py
```diff
@@ -70,21 +72,42 @@ def test_qubitization_phase_estimation_of_walk(num_terms: int, use_resource_stat
     state_prep = (
         LPResourceState(precision) if use_resource_state else RectangularWindowState(precision)
     )
-    gh = GateHelper(QubitizationQPE(walk, ctrl_state_prep=state_prep))
-    qpe_reg, selection, target = (gh.quregs['qpe_reg'], gh.quregs['selection'], gh.quregs['target'])
+    qpe_bloq = QubitizationQPE(walk, state_prep)
+
+    def should_decompose(binst):
```
mpharrigan, line 77: can you open an issue and link it here. I assume this isn't ideally how things would work?

mpharrigan (approved):

PR #1496, opened 2024-11-13
Fix bug in KaliskiStep3 and add tests for all steps

NoureldinYosri:
There was bug in the decomposition of KaliskiStep3 where a variable `junk` was assigned twice and freed twice.

there were also a few minor things in classical action of other bloqs. the reason these were not caught before is that the tests are written for the top level KaliskiModInverse which decomposes into _KaliskiIteration ... so the classically action of the _KaliskiIteration was tested, but the decomposition of _KaliskiIteration into steps wasn't. so I added tests to check all decomposition down the decomposition of each step


mpharrigan:
can you add a PR description

NoureldinYosri:
> can you add a PR description

added

mpharrigan (approved):
lgtm

PR #1497, opened 2024-11-13
Fix symbolic call graphs for factoring phase estimates

fpapa250:
- Adds symbolic type checking for remaining ECC bloqs (https://github.com/quantumlib/Qualtran/issues/1476)
- Adds DecomposeTypeErrors for bloqs which can't be symbolically decomposed
- Both call graphs for factoring phase estimates work for symbolic arguments
- Adds decomposition for MeasureQFT bloq to allow for call graphs of the phase estimation bloqs

### DIFF
--- a/qualtran/bloqs/factoring/_factoring_shims.py
+++ b/qualtran/bloqs/factoring/_factoring_shims.py
@@ -13,13 +13,29 @@
 #  limitations under the License.
 
 from functools import cached_property
-from typing import Optional, Tuple
+from typing import Dict, Optional, Tuple
 
+import numpy as np
+import sympy
 from attrs import frozen
 
-from qualtran import Bloq, CompositeBloq, DecomposeTypeError, QBit, Register, Side, Signature
+from qualtran import (
+    Bloq,
+    BloqBuilder,
+    DecomposeTypeError,
+    QBit,
+    QUInt,
+    Register,
+    Side,
+    Signature,
+    Soquet,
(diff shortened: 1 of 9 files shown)


qualtran/bloqs/factoring/_factoring_shims.py
```diff
@@ -30,8 +46,21 @@ class MeasureQFT(Bloq):
     def signature(self) -> 'Signature':
         return Signature([Register('x', QBit(), shape=(self.n,), side=Side.LEFT)])
 
-    def decompose_bloq(self) -> 'CompositeBloq':
-        raise DecomposeTypeError('MeasureQFT is a placeholder, atomic bloq.')
+    def build_composite_bloq(self, bb: 'BloqBuilder', x: Soquet) -> Dict[str, 'SoquetT']:
```
NoureldinYosri, line 49: since this shim has a decomposition now should it become a bloq ?

fpapa250: I'll leave that up to @mpharrigan. We added this decomposition a few PRs back (or meant to), but it got clobbered when merging two separate PRs touching this file (my mistake). I think the goal eventually is to use the phase_estimation bloqs, but for now I just added this so it could decompose entirely into leaf bloqs.

NoureldinYosri (approved):
apart from the `MeasureQFT` which I think should become a bloq, LGTM

PR #1500, opened 2024-11-21
More explicit documentation of endianness

mpharrigan:
This needs to be _so_ explicit in class docstrings, as it is a common source of confusion and error. 

@ncrubin


ncrubin (approved):
LGTM

mhucka (approved):
These are so straightforward that I can probably offer an approval even though I'm not knowledgeable about Qualtran.

LGTM.

tanujkhattar (approved):

PR #1502, opened 2024-11-30
Self adjoint for `MCX` and `MCZ`

anurudhp:
- Self adjoint for `MCX` and `MCZ`
- fix deprecated call graph in `SparseMatrixHermitian`


anurudhp:
@mpharrigan @tanujkhattar @NoureldinYosri ptal

NoureldinYosri (approved):

PR #1503, opened 2024-11-30
Add list primitives from Quartic speedups paper

anurudhp:
implement the following primitives from https://arxiv.org/abs/2406.19378
- `SortInPlace`
- `SymmetricDifference`
- `HasDuplicates`

part of #1348


anurudhp:
@tanujkhattar ptal

tanujkhattar (approved):

PR #1509, opened 2024-12-18
Make `BloqBuilder.join` accept array-like of soquets

### LINKED ISSUE #1470: BloqBuilder.join chokes on lists, requires specifically ndarrays
Elsewhere, we try to code defensively so you can theoretically pass in array-like objects instead of requiring specifically numpy arrays. This got missed in `BloqBuilder.join`

anurudhp:
fixes #1470


anurudhp:
@mpharrigan ptal

mpharrigan (approved):
glad it was this easy!

PR #1510, opened 2024-12-18
Allow computing cost from decomposition of cirq gates

### LINKED ISSUE #1426: get_cost_value fails when applied to a CirqGateAsBloq with a known decomposition into gates that should be countable.
When I execute the code:
```
import sympy
import cirq

from qualtran.cirq_interop import CirqGateAsBloq
from qualtran.resource_counting import QECGatesCost, get_cost_value

theta = sympy.Symbol("theta", real=True)
cirq_swappow = cirq.SwapPowGate(exponent=theta)

# Would be cool if it worked:
swappow_bloq = CirqGateAsBloq(cirq_swappow)

get_cost_value(swappow_bloq, QECGatesCost())
```

I would expect that get_cost_value would automatically use cirq.decompose to break the SwapPowGate into gates that can be ingested by Qualtran and costed out, but instead I get an error message:

`ValueError: Cirq gate must be directly countable, not SWAP**theta`.

When I run (with the appropriate imports of course)
```
g, sigma = swappow_bloq.call_graph()

show_call_graph(g)

show_counts_sigma(sigma)
```

I get the expected behaviour:
image

anurudhp:
When computing the t-complexity, if `_from_directly_countable_cirq` returns `None`, it now defaults to using the decomposition to compute the gate counts.

fixes #1426 - added the test described here. (the exact cost values would need to updated once #1455 is merged)

### DIFF
--- a/qualtran/bloqs/basic_gates/rotation.py
+++ b/qualtran/bloqs/basic_gates/rotation.py
@@ -108,7 +108,7 @@ _Z_POW_DOC = BloqDocSpec(bloq_cls=ZPowGate, examples=[_z_pow])
 
 @frozen
 class CZPowGate(CirqGateAsBloqBase):
-    exponent: float = 1.0
+    exponent: SymbolicFloat = 1.0
     global_shift: float = 0.0
     eps: SymbolicFloat = 1e-11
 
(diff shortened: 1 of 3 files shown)


anurudhp:
@mpharrigan ptal

mpharrigan (approved):
Thanks!

PR #1512, opened 2024-12-18
Be a bit careful about complexity of the sparse Hamiltonian

### LINKED ISSUE #1511: Resource estimation notebook for chemistry has (I think) a typo
In /bloqs/chemistry/resource_estimation.ipynb, the size of the data that needs to be loaded for the sparse simulation approach is given as O(N^3) but it should be O(N^4).

fdmalone:
Fixes #1511


mpharrigan (approved):
Thanks @wjhuggins does this satisfy your concerns

wjhuggins:
Yeah, although the sparsity can be even lower in some situations, but I think this is good.

Issue #1368, opened 2024-08-29
big_O is buggy


anurudhp:
big_O only takes into account all variables in the input expression (setting their limit to inf), but not once multiplied later.

```py
import sympy
from qualtran.resource_counting import big_O

n = sympy.Symbol("n", positive=True)

print(big_O(1) * n) # O(n)
print(big_O(1) * (1 / n)) # O(1/n)
print(big_O(1) * (n + 1)) # O(1)
print(big_O(1) * (1 / n + 1)) # O(1/n)
```

Also, sympy does not support multivariate orders. And as we usually have variables tending to both limits, e.g. `n, m` etc to inf, and `eps, delta` etc to 0, using big_O might lead to incorrect expressions.

---

One option would be to instead return `C * E` instead of `big_O(E)` in call graphs etc. And while viewing expressions, the user can take `big_O` of the whole thing. Could possibly use `SympySymbolAllocator` to generate these constants.

mpharrigan:
You mean just include an arbitrary constant $C$ in the returned expression for a bloq?

PR #1513, opened 2024-12-20
Make `And` a leaf bloq

### LINKED ISSUE #1346: Make `And` a leaf bloq

anurudhp:
fixes #1346 

also removes the only use of big O in call graph #1368

- first commit sets `FLAG_AND_IS_LEAF=True` and fixes all tests
- second one removes it altogether and cleans up the `And` bloq code
- fix `t_counts_from_sigma` to explicitly count `And`s

### DIFF
--- a/qualtran/bloqs/chemistry/df/double_factorization.py
+++ b/qualtran/bloqs/chemistry/df/double_factorization.py
@@ -48,9 +48,8 @@ from qualtran import (
     Soquet,
     SoquetT,
 )
-from qualtran.bloqs.basic_gates import CSwap, Hadamard, Toffoli
+from qualtran.bloqs.basic_gates import CSwap, Hadamard
 from qualtran.bloqs.block_encoding import BlockEncoding
-from qualtran.bloqs.bookkeeping import ArbitraryClifford
 from qualtran.bloqs.chemistry.black_boxes import ApplyControlledZs
 from qualtran.bloqs.chemistry.df.prepare import (
     InnerPrepareDoubleFactorization,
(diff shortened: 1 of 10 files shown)


anurudhp:
@mpharrigan @fdmalone ptal!

qualtran/bloqs/chemistry/df/double_factorization.py
```diff
@@ -280,10 +279,10 @@ def build_call_graph(self, ssa: 'SympySymbolAllocator') -> 'BloqCountDictT':
             in_prep_dag: 1,  # in_prep_l^dag
             rot: 1,  # rotate into system basis  listing 4 pg 54
             # apply CCZ first then CCCZ, the cost is 1 + 2 Toffolis (step 4e, and 7)
-            Toffoli(): 1,
+            ApplyControlledZs(cvs=(1, 1), bitsize=5): 1,
```
mpharrigan, line 282: was this incorrect before? I see the change makes it match the comment

mpharrigan (approved):
excellent. very nice

qualtran/bloqs/chemistry/df/double_factorization.py
anurudhp: The previous test was only checking the `sigma` for the call graph. I updated it to check for equivalent bloq counts, which flagged this inconsistency.

Issue #1515, opened 2024-12-31
Incorrect `on_classical_vals` overrides for non classical bloqs


anurudhp:
I ran into this while implementing #1514. The tensors of these bloq examples seem to have non-trivial phases, which makes them non-classical. The following bloq examples fail in #1514 (I've restricted to bloqs with up to 9 qubits):
```
sub_from_small

E Mismatched elements: 144 / 65536 (0.22%)
E Max absolute difference: 1.
E Max relative difference: 1.
E x: array([[1., 0., 0., ..., 0., 0., 0.],
E [0., 1., 0., ..., 0., 0., 0.],
E [0., 0., 1., ..., 0., 0., 0.],...
E y: array([[1.+0.j, 0.+0.j, 0.+0.j, ..., 0.+0.j, 0.+0.j, 0.+0.j],
E [0.+0.j, 1.+0.j, 0.+0.j, ..., 0.+0.j, 0.+0.j, 0.+0.j],
E [0.+0.j, 0.+0.j, 1.+0.j, ..., 0.+0.j, 0.+0.j, 0.+0.j],...
```
```
cmod_add_k_small

E Mismatched elements: 18 / 1024 (1.76%)
E Max absolute difference: 1.
E Max relative difference: 0.
E x: array([[1., 0., 0., ..., 0., 0., 0.],
E [0., 1., 0., ..., 0., 0., 0.],
E [0., 0., 1., ..., 0., 0., 0.],...
E y: array([[1.+0.j, 0.+0.j, 0.+0.j, ..., 0.+0.j, 0.+0.j, 0.+0.j],
E [0.+0.j, 1.+0.j, 0.+0.j, ..., 0.+0.j, 0.+0.j, 0.+0.j],
E [0.+0.j, 0.+0.j, 1.+0.j, ..., 0.+0.j, 0.+0.j, 0.+0.j],...
```
```
moddbl_small

E Mismatched elements: 3 / 256 (1.17%)
E Max absolute difference: 1.
E Max relative difference: 1.
E x: array([[1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
E [0., 0., 0., 0., 0., 0., 0., 1., 0., 0., 0., 0., 0., 0., 0., 0.],
E [0., 1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],...
[...]

PR #1518, opened 2025-01-08
restrict classical action of certain arithmetic bloqs

NoureldinYosri:
- SubtractFrom.on_classical_values was using `math.fmod` which fails in the case of overflow ... using `simulation.classical_sim.add_ints` instead
- bloqs for modular arithmetic assume that their inputs are in the correct range $[0, \mod)$ ... fixing the classical action is easy for them ... but the decompositions/constructions will fail on those inputs since they were not designed to handle them.

fixes https://github.com/quantumlib/Qualtran/issues/1515


qualtran/bloqs/mod_arithmetic/mod_addition.py
```diff
@@ -87,7 +87,9 @@ def signature(self) -> 'Signature':
     def on_classical_vals(
         self, x: 'ClassicalValT', y: 'ClassicalValT'
     ) -> Dict[str, 'ClassicalValT']:
-        return {'x': x, 'y': (x + y) % self.mod}
+        if x < self.mod and y < self.mod:
+            y = (x + y) % self.mod
+        return {'x': x, 'y': y}
```
anurudhp, line 92: should we instead raise a `ValueError` for out-of-bounds inputs? (similar to e.g. `BQUInt.assert_valid_classical_val`).

NoureldinYosri: depends on how you define the bloq ... if we want the bloq to be unitary then the classical action is correct and we don't need to raise an error 

however constructions assume valid range ... I don't feel strongly towards either ... although I'm a bit leaning towards raising an error to match the construction

mpharrigan: I'd prefer that the classical action raises an error. The tensors-from-classical-action function would need to handle this, of course

NoureldinYosri: done

anurudhp (approved):

mpharrigan (approved):

PR #1521, opened 2025-01-10
Setting min version of attrs

### LINKED ISSUE #1520: Set minimum attrs version
I found that upgrading attrs to 23.2.0 or higher would resolve errors such as
```
  File "/Users/maxradin/anaconda3/lib/python3.11/functools.py", line 994, in __get__
    raise TypeError(msg) from None
TypeError: No '__dict__' attribute on 'CtrlSpec' instance to cache 'shapes' property.
```
when using qualtran. (See https://github.com/isi-usc-edu/pyLIQTR/issues/35 for example code and full traceback.)

My guess is that this is because earlier versions of attrs did not support cached properties on slotted classes, as described in https://github.com/python-attrs/attrs/pull/1200. I'd suggest updating qualtran's requirements to specify a minimum version of 23.2.0 for attrs.

max-radin:
Fixes #1520.


mpharrigan:
Thanks for investigating this. I think we've run into this exact issue before, so this definitely should be included in our runtime requirements.

mpharrigan (approved):

PR #1522, opened 2025-01-13
add option to generate QREF from callgraphs

mstechly:
This PR introduces an option to create QREF routines from callgraphs.

I have not updated the dependencies, as I decided it will be better to ask @mpharrigan for assistance :)
Since we're using strict bounds, `qref==0.9.0` and `bartiq==0.9.0` should be fine. If you want me to provide some looser bounds, I can find out what these should be.


mpharrigan (approved):
Thanks! I updated the dependencies

mstechly:
Thank you @mpharrigan  <3

PR #1529, opened 2025-01-25
Fix memory overflow in sparse matrix stress test

anurudhp:
Reduces the input size for `test_matrix_stress` so that the tensors can be computed.

For the old parameters, the nightly fails with a memory error:
```
FAILED qualtran/bloqs/block_encoding/sparse_matrix_test.py::test_matrix_stress - numpy.core._exceptions._ArrayMemoryError: Unable to allocate 64.0 GiB for an array with shape (16, 131072, 2048) and data type complex128
```

### DIFF
--- a/qualtran/bloqs/block_encoding/sparse_matrix_test.py
+++ b/qualtran/bloqs/block_encoding/sparse_matrix_test.py
@@ -244,23 +244,25 @@ def test_counts():
     )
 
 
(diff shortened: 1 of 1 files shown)


anurudhp:
@mpharrigan ptal

mpharrigan (approved):
thanks!

PR #1530, opened 2025-01-27
Some documentation for proto serialization

mpharrigan:
- Consistent naming: switched to `bloq_id` from a mixture of `bloq_id` and `bloq_idx`. `bloq_id` is used in the proto; and makes more sense since it's an identifier and not an index
- Add a couple details to the docstrings and comments


mpharrigan:
@tanujkhattar can you ptal

tanujkhattar (approved):

PR #1531, opened 2025-01-28
Create a notebook for quantum modular arithmetic

NoureldinYosri:
- **Fix bug in KaliskiStep3 and add tests for all steps**
- **cost**
- **Create a notebook for quantum modular arithmetics**

### DIFF
--- a/qualtran/bloqs/arithmetic/comparison.py
+++ b/qualtran/bloqs/arithmetic/comparison.py
@@ -1462,20 +1462,17 @@ class _HalfComparisonBase(Bloq):
         c: Optional['ClassicalValT'] = None,
         target: Optional['ClassicalValT'] = None,
     ) -> Dict[str, 'ClassicalValT']:
+        if self._op_symbol in ('>', '<='):
+            c_val = add_ints(-int(a), int(b), num_bits=self.dtype.bitsize + 1, is_signed=False)
+        else:
+            c_val = add_ints(int(a), -int(b), num_bits=self.dtype.bitsize + 1, is_signed=False)
         if self.uncompute:
-            assert c == add_ints(
-                int(a),
-                int(b),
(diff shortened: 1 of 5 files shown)


mpharrigan:
neat! isn't there a helper function you introduced that will automatically exhaustively test classical gates? 

Would probably be neat to have a small, exhaustive and large, fuzzed example for each 

maybe the "leading order" symbolics helper can be factored out into the sympy support utils

mpharrigan:
What about clearly segregating the "reference data" from the checks. 

```python
add1 = {
  bloq: ModAdd(bitsize=8, mod=3),
  domain: {'x': range(3), 'y': range(3)},
  tof: sympify('4*n'),  # Ref 1, Fig 8, "mod add"
  exhaustive: True
}

add2 = {
  bloq: ModAdd(bitsize=2048, mod=13*15),
  ...
  exhaustive: False
}

...

for data in [add1, add2, ...]:
  check_classical(add1) 
      # >> checking ModAdd(bitsize=8, mod=3)....
      # >> exhaustive classical action ✓
  check_cost(add1)
      # >> reference cost: 4n
      # >> computed toffli cost: 4n + 32 + lg(mod)
      # >> leading order cost: 4n ✓

```

mpharrigan:
Is there a straightforward way to get the classical domain from the bloq? Ideally you could just query the classical domain from the data types but e.g. for mod add there are additional restrictions. Should we be using a bounded quint here? 

It would also be nice if the constructors always took consistent-ish attributes so we could always have one example where we pass in a the same, small qdtype with exhaustive checking and one example where we pass in the same, large qdtype with fuzz testing

tanujkhattar:
A passing comment -- We should also consider using the QGF(q) type where q is a prime for mod add. QGF is the correct general type for arbitrary prime and extension fields. We should prefer that over bounded QUINT.

PR #1532, opened 2025-01-29
Rename factoring -> cryptograph and create a Root Bloqs docs section

NoureldinYosri:
- `factoring/` contains code for both `ECC` and `RSA` so a better name for it is `cryptography`
- created a section in the docs for `Root Bloqs` with the first one being `FindECCPrivateKey` which already exists


dev_tools/autogenerate-bloqs-notebooks-v2.py
```diff
@@ -63,11 +63,17 @@
...
-    'factoring/factoring-via-modexp.ipynb',
+    'cryptography/rsa/factoring-via-modexp.ipynb',
     'state_preparation/state_preparation_via_rotation_tutorial.ipynb',
 ]
 
 
+# --------------------------------------------------------------------------
+# -----   Root Bloqs   -----------------------------------------------------
+# --------------------------------------------------------------------------
+ROOT_BLOQS = ['cryptography/ecc/ecc.ipynb']
```
mpharrigan, line 74: why can't these be bloq notebooks like the rest of the bloq notebooks

mpharrigan (approved):

dev_tools/autogenerate-bloqs-notebooks-v2.py
NoureldinYosri: I don't understand the comment. this is an existing bloq notebook and I'm just putting a second link to it in the new section

mpharrigan: The idiomatic way to include bloq documentation in the TOC is by adding a `NotebookSpecV2` to the correct section `List[NotebookSpecV2]`. Each section list is added to `NB_BY_SECTION: Dict[str, List[...]]`

NoureldinYosri: yes I get that and that notebook is already there https://github.com/quantumlib/Qualtran/blob/89e916e27a9f60b59339e36bb3fa26ea2685da3f/dev_tools/qualtran_dev_tools/notebook_specs.py#L557-L565
image

---

what I'm doing here is creating a new section 
image

PR #1533, opened 2025-01-29
Update gf2.MultiplyPolyByConstantMod to support QGF dtype

NoureldinYosri:
- rename `MultiplyPolyByConstantMod` to `GF2MultiplyByConstantMod`
- change signature to use `QGF` instead of an array of qubits
- update `QGF` to include an optional field for the galois field


qualtran/bloqs/gf_arithmetic/gf2_multiplication.ipynb
```diff
@@ -144,7 +144,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "b26814e0",
+   "id": "94ab395f",
```
mpharrigan, line 147: why do these always get overwritten? they should generally be maintained. The autogenerate-bloqs-notebooks script should maintain these

mpharrigan (approved):

qualtran/_infra/data_types.py
```diff
@@ -899,6 +904,7 @@ class QGF(QDType):
 
     characteristic: SymbolicInt
     degree: SymbolicInt
+    galois_field: Optional['galois.FieldArrayMeta'] = None
```
tanujkhattar, line 907: This type is not hashable, as discussed offline please consider updating the class attributes so you only store hashable types and construct the corresponding galois field object by passing in the values to the constructor.

tanujkhattar: @NoureldinYosri Can you please confirm if this was addressed ?

NoureldinYosri: QGF was made hashable ... MultiplyPolyByConstantMod is hashable because I overload its __hash__ method ... or do you want to do the same for MultiplyPolyByConstantMod?

Issue #1525, opened 2025-01-24
Nightly tests are failing


anurudhp:
https://github.com/quantumlib/Qualtran/actions/runs/12927816318 (jan 23)

cc @NoureldinYosri looks like the mod-sub classical action test is failing after #1518

NoureldinYosri:
the tests pass locally 

```sh
$ pytest qualtran/bloqs/mod_arithmetic/mod_subtraction_test.py -k test_cmodsub_classical_action -v
qualtran/bloqs/mod_arithmetic/mod_subtraction_test.py::test_cmodsub_classical_action[13-4-QUInt-0] PASSED                                                                                                                                                                                                  [  5%]
qualtran/bloqs/mod_arithmetic/mod_subtraction_test.py::test_cmodsub_classical_action[13-4-QUInt-1] PASSED                                                                                                                                                                                                  [ 11%]
qualtran/bloqs/mod_arithmetic/mod_subtraction_test.py::test_cmodsub_classical_action[13-4-QMontgomeryUInt-0] PASSED                                                                                                                                                                                        [ 17%]
qualtran/bloqs/mod_arithmetic/mod_subtraction_test.py::test_cmodsub_classical_action[13-4-QMontgomeryUInt-1] PASSED                                                                                                                                                                                        [ 23%]
qualtran/bloqs/mod_arithmetic/mod_subtraction_test.py::test_cmodsub_classical_action[13-5-QUInt-0] PASSED                                                                                                                                                                                                  [ 29%]
qualtran/bloqs/mod_arithmetic/mod_subtraction_test.py::test_cmodsub_classical_action[13-5-QUInt-1] PASSED                                                                                                                                                                                                  [ 35%]
qualtran/bloqs/mod_arithmetic/mod_subtraction_test.py::test_cmodsub_classical_action[13-5-QMontgomeryUInt-0] PASSED                                                                                                                                                                                        [ 41%]
qualtran/bloqs/mod_arithmetic/mod_subtraction_test.py::test_cmodsub_classical_action[13-5-QMontgomeryUInt-1] PASSED                                                                                                                                                                                        [ 47%]
qualtran/bloqs/mod_arithmetic/mod_subtraction_test.py::test_cmodsub_classical_action[17-5-QUInt-0] PASSED                                                                                                                                                                                                  [ 52%]
[...]

PR #1537, opened 2025-02-03
Relax restriction on classical inputs for modular addition to fix nightly CI

NoureldinYosri:
The constructions still work when at most one of their inputs `==mod`. While this looks like a hack, this is necessarly since many higher level construction assume this since  `x == mod` can appear in lower level bloqs when the input to higher level bloq is `x=0`

fixes the nightly CI & fixes https://github.com/quantumlib/Qualtran/issues/1525


qualtran/bloqs/mod_arithmetic/mod_addition.py
```diff
@@ -87,13 +87,15 @@ def signature(self) -> 'Signature':
     def on_classical_vals(
         self, x: 'ClassicalValT', y: 'ClassicalValT'
     ) -> Dict[str, 'ClassicalValT']:
-        if not (0 <= x < self.mod):
+        # The construction still works when at most one of inputs equals `mod`.
+        special_case = (x == self.mod) ^ (y == self.mod)
```
mpharrigan, line 91: can we use `or` instead of `^`

mpharrigan (approved):
Thanks for investigating and fixing! One nit, then lgtm

mpharrigan:
but please wait until tomorrow to merge to see if I can receive notifications of failed nightly workflows.

qualtran/bloqs/mod_arithmetic/mod_addition.py
NoureldinYosri: this is `at most` not `at least` ... although I can simply say `not (x == y == self.mod)` and update the inequalities below from `x < mod` to `x <= mod`, what do you prefer?

PR #1542, opened 2025-02-04
Implement PRODUCT decomposition

christopherkang:
I have performed a quick implementation of the `PRODUCT` function using the implementation suggested in https://arxiv.org/pdf/2007.07391 (page 69).

This addresses issue #485. 

Some notes:
- This achieves the desired complexity and relies upon the `AND` bloq. It is important to use the `AND` bloq because the `AND.adjoint` natively allows us to uncompute without additional Toffoli complexity (by using the strategies from [Gidney](https://arxiv.org/abs/1709.06648))
- It also includes a brute force test checking all implementations with input sizes ranging from 1-3 bits. 
- My code as of now is not fully commented + notated --- I would appreciate feedback on making it better!

### DIFF
--- a/qualtran/bloqs/arithmetic/multiplication.py
+++ b/qualtran/bloqs/arithmetic/multiplication.py
@@ -23,6 +23,7 @@ from qualtran import (
     Bloq,
     bloq_example,
     BloqDocSpec,
+    CompositeBloq,
     ConnectionT,
     DecomposeTypeError,
     GateWithRegisters,
--- a/qualtran/bloqs/arithmetic/multiplication.py
+++ b/qualtran/bloqs/arithmetic/multiplication.py
@@ -12,10 +12,11 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from typing import Dict, Iterable, List, Optional, Sequence, Tuple, TYPE_CHECKING, Union
+from typing import Dict, Iterable, Iterator, List, Optional, Sequence, Tuple, TYPE_CHECKING, Union
 
 import cirq
 import numpy as np
+from numpy.typing import NDArray
 from attrs import evolve, frozen
 
 from qualtran import (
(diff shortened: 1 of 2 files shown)


mpharrigan:
@NoureldinYosri can you take a look

Ideally, we'd be adding new decompositions using the idiomatic qualtran style (`build_composite_bloq` with `BloqBuilder`). This is particularly useful when calling bloqs that have non-thru registers like how `And` will allocate its own target

qualtran/bloqs/arithmetic/multiplication.py
```diff
@@ -343,6 +351,51 @@ def build_call_graph(self, ssa: 'SympySymbolAllocator') -> 'BloqCountDictT':
         # See: https://github.com/quantumlib/Qualtran/issues/217
         num_toff = 2 * self.a_bitsize * self.b_bitsize - max(self.a_bitsize, self.b_bitsize)
         return {Toffoli(): num_toff}
+    
+    def decompose_from_registers(
```
NoureldinYosri, line 355: please use the idiomatic way of defining the construction instead of the legacy way... the idiomatic way makes the construction more readable. you can use this as an example

https://github.com/quantumlib/Qualtran/blob/60037bcb9e088d0e22ef436360e6ec262f4cfe88/qualtran/bloqs/arithmetic/comparison.py#L988-L1006

qualtran/bloqs/arithmetic/multiplication_test.py
```diff
@@ -119,6 +120,94 @@ def test_product():
     num_toff = 2 * bitsize * mbits - max(bitsize, mbits)
     assert t_complexity(cbloq) == TComplexity(t=4 * num_toff)
 
+def test_product_gates():
```
NoureldinYosri, line 123: we have methods that do most of the testing, you can use these as example

https://github.com/quantumlib/Qualtran/blob/60037bcb9e088d0e22ef436360e6ec262f4cfe88/qualtran/bloqs/mod_arithmetic/mod_subtraction_test.py#L209-L248

NoureldinYosri (changes requested):
@christopherkang thank you for your contribution. I see that you styled your code in the same way as the other bloqs in the file which is great ... however this file was written in the early days of Qualtran and now we have new ways for doing things. please use the examples I left to remodel your code

mhucka:
@christopherkang @NoureldinYosri  Just circling around to different PRs and trying to do a bit of housekeeping. For planning purposes, what is the status of this PR?

Issue #1471, opened 2024-10-16
Add modulus p as a class member of QMontgomeryUInt


fpapa250:
As a part of https://github.com/quantumlib/Qualtran/pull/1425, we are adding class functions which can be used to convert in and out of montgomery form and also to calculate the montgomery product or inverse. These functions all require bitsize and modulus p. For now it is easy to use the bitsize in the calculations, but adding the modulus as a class member would be require all instances of QMontgomeryUInt to instantiate with p. We leave that as a TODO action item for now.

PR #1543, opened 2025-02-05
Add a modulus parameter to QMontgomeryUInt

NoureldinYosri:
This allows us to use `.get_classical_domain()` in classical tests

fixes https://github.com/quantumlib/Qualtran/issues/1471


mpharrigan (approved):
nice!

PR #1552, opened 2025-02-13
Override ctrl system for Toffoli

### LINKED ISSUE #1544: Override `.controlled()` for common atomic gates like `Toffoli` gate so that we can construct controlled versions of higher level bloqs
The following code snippet currently throws an error and it should not:

```python
> Toffoli().controlled().t_complexity()
```

As a result, any bloq the decomposes into a `Toffoli` gate does not support resource estimation for `bloq.controlled()`

anurudhp:
fixes #1544

### DIFF
--- a/qualtran/bloqs/basic_gates/toffoli.py
+++ b/qualtran/bloqs/basic_gates/toffoli.py
@@ -13,7 +13,7 @@
 #  limitations under the License.
 import itertools
 from functools import cached_property
-from typing import Dict, List, Optional, Tuple, TYPE_CHECKING, Union
+from typing import cast, Dict, Iterable, List, Optional, Sequence, Tuple, TYPE_CHECKING, Union
 
 import numpy as np
 from attrs import frozen
@@ -25,6 +25,7 @@ from qualtran import (
     BloqDocSpec,
     CompositeBloq,
     Connection,
+    CtrlSpec,
     DecomposeTypeError,
     QBit,
     Register,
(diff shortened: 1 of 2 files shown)


anurudhp:
@tanujkhattar @mpharrigan ptal

mpharrigan (approved):
nice!

qualtran/bloqs/basic_gates/toffoli.py
```diff
@@ -127,6 +129,29 @@ def wire_symbol(self, reg: Optional[Register], idx: Tuple[int, ...] = tuple()) -
...
+            ctrl0, ctrl1 = cast(NDArray, in_soqs.pop('ctrl'))
+
+            (new_ctrl, ctrl0), ctrl1, target = bb.add(
+                cc_cnot, ctrl2=np.array([new_ctrl, ctrl0]), ctrl=ctrl1, target=in_soqs.pop('target')
+            )
+
+            return [new_ctrl], [ctrl0, ctrl1, target]
+
+        return cc_cnot, add_controlled
+
```
tanujkhattar, line 154: I was thinking the natural thing to do here should be to return an instance of the `MultiControlX` gate if you do `Toffoli().controlled()`,  but then I checked that we don't return `MultiControX` from `XGate.controlled()` or `CNOT.controlled()` either. 

This works for now. Eventually, it'll be nice to have the same bloq constructed upon calling 
```python
>>> XGate().controlled(CtrlSpec(cvs=[1, 1, 1, 1, 1]))
>>> CNOT().controlled(CtrlSpec(cvs=[1, 1, 1, 1]))
>>> Toffoli().controlled(CtrlSpec(cvs=[1, 1, 1]))
```

and the natural bloq to return for all three would be `MultiControlX(cvs=(1, 1, 1, 1, 1))`.

mpharrigan: fair enough -- I opened an issue to track

PR #1559, opened 2025-02-19
Pennylane-Qualtran interoperability

austingmhuang:
In this PR, we set out to define a new method `as_pl_op()` for the base `Bloq` class as well as several "basic gates" that have direct 1:1 translations with PennyLane operators. This is a work in progress and relies on [another PR inside of PennyLane](https://github.com/PennyLaneAI/pennylane/pull/6921) that contains the `qml.FromBloq` adapter class. We open this now for visibility and discussion purposes, but the work is still on-going.


qualtran/_infra/bloq.py
```diff
@@ -466,6 +467,25 @@ def as_cirq_op(
             bloq=self, cirq_quregs=cirq_quregs, qubit_manager=qubit_manager
         )
 
+    def as_pl_op(self, wires: 'pennylane.Wires') -> 'pennylane.Operation':
+        """Override this method to support conversion to a Cirq operation.
+
+        If this method is not overriden, the default implementation will wrap this bloq
+        in a `BloqAsCirqGate` shim.
```
mpharrigan, line 474: vestigial copy-paste docstring?

qualtran/_infra/bloq.py
```diff
@@ -466,6 +467,25 @@ def as_cirq_op(
...
+
+        Args:
+            wires: the wires that the op acts on
+
+        Returns:
+            ~pennylane.Operation: A PennyLane operation corresponding to this bloq acting on the 
+                provided wires or None. This method should return None if and only if the bloq 
+                instance truly should not be included in the Cirq circuit (e.g. for reshaping 
+                bloqs). A bloq with no cirq equivalent should raise an exception instead.
```
mpharrigan, line 483: "cirq"

mpharrigan (commented):
nice! will probably have to add some tests. We'll also need to add pennylane to the requirements, which I can help with (since we have a convoluted system for pinning the CI dependencies)

qualtran/_infra/bloq.py
austingmhuang: 😅 Yes, will fix.

austingmhuang:
> nice! will probably have to add some tests. We'll also need to add pennylane to the requirements, which I can help with (since we have a convoluted system for pinning the CI dependencies)

For the tests, I'm not sure where I'm supposed to add them. I took a look `qualtran/_infra/bloq_test.py` but I get the feeling that that's probably not where to go... do you mind pointing me in the right direction as to where to put these tests 😄 ?

mpharrigan:
Yeah, that's a good question. Usually we'd put them alongside the code for the adapter classes; but that will reside in the Pennylane repository. 

I think in this case it will suffice to have some basic unit tests for the particular examples for which you implemented the method (which can live in the test files associated with those bloqs)

mpharrigan:
You can also add a test to `_infra/bloq_test.py` with a very simple test of the fallback, adapter behavior. I know it's scary that there's nothing in there, but in this case it should be fine

austingmhuang:
Hi @mpharrigan, I added some very basic unit tests for each of the atomic bloqs I wrote an `as_pl_op` function for. If you think there should be additional tests e.g. tensor_contract()/matrix(), feel free to let me know and I can add them as well (Comparing tensor_contract/matrix is also tested for in PL itself, but not quite for every single bloq). I think the PR is ready for review, but I'll need the CI running to be sure :)

mpharrigan:
Sorry, since you're a first-time contributor I have to press the "approve running the CI" button every time you push a commit, sadly

mpharrigan:
oh, I have to update our dependencies. I will do that tomorrow

mpharrigan:
Can the [WIP] tag be removed from the title of the PR?

qualtran/_infra/bloq.py
```diff
@@ -466,6 +468,25 @@ def as_cirq_op(
...
+        Args:
+            wires: the wires that the op acts on
+
+        Returns:
+            ~.Operation: A PennyLane operation corresponding to this bloq acting on the 
+                provided wires or None. This method should return None if and only if the bloq 
+                instance truly should not be included in the PennyLane circuit (e.g. for reshaping 
+                bloqs). A bloq with no PennyLane equivalent should raise an exception instead.
+        """
+        from pennylane.io import FromBloq
```
mpharrigan, line 486: Actually, is this circular dependancy going to be a problem while trying to merge these two PRs across two packages? To merge this current, Qualtran PR; we need to depend on a release of Pennylane that includes this adapter. But that adapter needs a release of qualtran with `Bloq.as_pl_op`. Am I capturing this correctly?

austingmhuang:
> oh, I have to update our dependencies. I will do that tomorrow

Perhaps the workflow file should be edited instead for this PR specifically to get the CI passing. We can then remove it when qml.FromBloq is merged into pennylane master.

mpharrigan:
Yes, I can update our dependencies to that specific branch for the time being. I'll open an issue to switch it back to a proper release after the qualtran features are picked up in a release. Can you ensure that that branch will survive until the next pennylane release?

mpharrigan:
alright, if you merge in `main` it should pick up the branch

austingmhuang:
> Yes, I can update our dependencies to that specific branch for the time being. I'll open an issue to switch it back to a proper release after the qualtran features are picked up in a release. Can you ensure that that branch will survive until the next pennylane release?

Awesome thanks. Yes the branch will survive until the next pennylane release. I'll let you know when it's ready to merge and we can coordinate the process.

Jaybsoni (commented):
Looks good to me for the most part. I left a few comments to be addressed. Otherwise its good from my side 👍🏼 

One question for @mpharrigan, I am not sure how paranoid we should be about conventions? For most of the tests all we are checking is that if we map a Bloq using FromBloq, we get the expected PL operator. We DONT check if the Qualtran definition of an operation is the same as the PL one. For example, I noticed in the global_phase the extra factor of pi that is used in Qualtran. Should we compare the matrix produced via tensor_contract against the qml.matrix function?

qualtran/bloqs/basic_gates/cnot.py
```diff
@@ -38,6 +38,8 @@
 if TYPE_CHECKING:
     import cirq
     import quimb.tensor as qtn
+    from pennylane.operation import Operation
+    from pennylane.wires import Wires
```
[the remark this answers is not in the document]
austingmhuang: To my knowledge yes.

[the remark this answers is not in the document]
mpharrigan: these are under the `if TYPE_CHECKING` guard, so they won't actually be imported at runtime. In fact, if you use non-string type annotations you will get a runtime error

mpharrigan:
> We DONT check if the Qualtran definition of an operation is the same as the PL one. For example, I noticed in the global_phase the extra factor of pi that is used in Qualtran. Should we compare the matrix produced via tensor_contract against the qml.matrix function?

My opinion is that we should test that they're equivalent (where feasible). In the global phase op, the `as_pl_op` method should take care of any pi differences, yes?

austingmhuang:
> My opinion is that we should test that they're equivalent (where feasible). In the global phase op, the `as_pl_op` method should take care of any pi differences, yes?

Yes, the `as_pl_op` method takes care of the pi difference.

qualtran/bloqs/basic_gates/cnot.py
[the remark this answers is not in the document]
Jaybsoni: Ahh got it 👍🏼

Jaybsoni:
> > We DONT check if the Qualtran definition of an operation is the same as the PL one. For example, I noticed in the global_phase the extra factor of pi that is used in Qualtran. Should we compare the matrix produced via tensor_contract against the qml.matrix function?
> 
> My opinion is that we should test that they're equivalent (where feasible). In the global phase op, the `as_pl_op` method should take care of any pi differences, yes?

Yes the `as_pl_op` method handles this, I guess my point was in general do we need to check that there aren't other sneaky factors of pi or (1/2) that crop up due to different conventions? Although now that I've looked at it some more, I am convinced we've covered our bases.

austingmhuang:
> > We DONT check if the Qualtran definition of an operation is the same as the PL one. For example, I noticed in the global_phase the extra factor of pi that is used in Qualtran. Should we compare the matrix produced via tensor_contract against the qml.matrix function?
> 
> My opinion is that we should test that they're equivalent (where feasible).

For the tests, I purposefully avoided comparing tensor_contract with matrix because the behaviour of the two functions are not 100% equivalent. The shape of PL's matrix always has the shape of (2^len(wires), 2^len(wires)) whereas that's not true for tensor_contract. E.g. for GlobalPhase, the matrix() will be
```
array([[6.123234e-17-1.j, 0.000000e+00+0.j],
       [0.000000e+00+0.j, 6.123234e-17-1.j]])
```

whereas for qualtran's GlobalPhase().tensor_contract(), we can expect:
```
(6.123233995736766e-17+1j)
```

austingmhuang:
I think this PR looks pretty much ready. If you are okay with the PR @mpharrigan please merge 😺

Jaybsoni (approved):
Looks goof from me 👍🏼

mpharrigan (approved):

Issue #1558, opened 2025-02-17
`QGF.to_bits` fails for large inputs because it falls back to `QUInt.to_bits`


NoureldinYosri:
Doing operations in GF($2^n$) is valid and should be supported as long as there is enough memory since that's the bottleneck. however simple operations break due to the default fall back to $QUInt$ which can't/needn't handle  numbers with 1000s of bits

```py3
>>> import galois
>>> from qualtran import QGF
>>> dtype = QGF(2, 1014, galois.Poly.Degrees([1014,385,0]), 'poly')
>>> dtype.to_bits(dtype.gf_type(0))  # should be [0]*1014
AttributeError: 'int' object has no attribute 'log2'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/google/home/noureldinyosri/Desktop/quantumlib/Qualtran/qualtran/_infra/data_types.py", line 967, in to_bits
    return self._quint_equivalent.to_bits(int(x))
           ^^^^^^^^^^^^^^^^^^^^^^
  File "<attrs generated getattr qualtran._infra.data_types.QGF>", line 6, in __getattr__
  File "/usr/local/google/home/noureldinyosri/Desktop/quantumlib/Qualtran/qualtran/_infra/data_types.py", line 948, in _quint_equivalent
    return QUInt(self.num_qubits)
                 ^^^^^^^^^^^^^^^
  File "<attrs generated getattr qualtran._infra.data_types.QGF>", line 6, in __getattr__
  File "/usr/local/google/home/noureldinyosri/Desktop/quantumlib/Qualtran/qualtran/_infra/data_types.py", line 939, in num_qubits
    return self.bitsize
           ^^^^^^^^^^^^
  File "<attrs generated getattr qualtran._infra.data_types.QGF>", line 6, in __getattr__
  File "/usr/local/google/home/noureldinyosri/Desktop/quantumlib/Qualtran/qualtran/_infra/data_types.py", line 934, in bitsize
    return bit_length(self.order - 1)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/google/home/noureldinyosri/Desktop/quantumlib/Qualtran/qualtran/symbolics/math_funcs.py", line 153, in bit_length
    return ceil(log2(floor(x) + 1))
                ^^^^^^^^^^^^^^^^^^
[...]

PR #1560, opened 2025-02-19
Update math funcs to use functions from math instead of numpy to support arbitrary length integers

tanujkhattar:
Fixes https://github.com/quantumlib/Qualtran/issues/1558

### DIFF
--- a/qualtran/symbolics/math_funcs_test.py
+++ b/qualtran/symbolics/math_funcs_test.py
@@ -16,15 +16,31 @@ from typing import cast
 import numpy as np
 import pytest
 import sympy
+from sympy.codegen.cfunctions import log
 from sympy.codegen.cfunctions import log2 as sympy_log2
 
-from qualtran.symbolics import bit_length, ceil, is_zero, log2, sarg, sexp, smax, smin
+from qualtran.symbolics import bit_length, ceil, is_zero, ln, log2, sarg, sexp, smax, smin, ssqrt
 
 
 def test_log2():
     assert log2(sympy.Symbol('x')) == sympy_log2(sympy.Symbol('x'))
     assert log2(sympy.Number(10)) == sympy_log2(sympy.Number(10))
     assert log2(10) == np.log2(10)
+    assert log2(2**10000) == 10000
+
+
+def test_ln():
+    assert ln(sympy.Symbol('x')) == log(sympy.Symbol('x'))
+    assert ln(sympy.Number(10)) == log(sympy.Number(10))
+    assert ln(10) == np.log(10)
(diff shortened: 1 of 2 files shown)


qualtran/symbolics/math_funcs_test.py
```diff
@@ -16,15 +16,31 @@
...
+    assert ln(sympy.Number(10)) == log(sympy.Number(10))
+    assert ln(10) == np.log(10)
+    assert np.isclose(ln(2**1000) / ln(2), 1000)
+
+
+def test_ssqrt():
+    assert ssqrt(sympy.Symbol('x')) == sympy.sqrt(sympy.Symbol('x'))
+    assert ssqrt(sympy.Number(10)) == sympy.sqrt(sympy.Number(10))
+    assert ssqrt(10) == np.sqrt(10)
+    assert ssqrt(2**1000) == 2**500
```
NoureldinYosri, line 43: `math.sqrt` fails for larger inputs e.g. 
```py3
>>> math.sqrt(2**10000)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
OverflowError: int too large to convert to float
```

NoureldinYosri (changes requested):
I don't think bit conversion should depend on `math.log`

qualtran/symbolics/math_funcs_test.py
tanujkhattar: This is because return type of `math.sqrt()` is a float, but its input types can be large integers. `np.sqrt()` fails even on 2 ** 100 because it expects the input type to be 64 bit integers at-most.

tanujkhattar:
The thing that depends upon math.log right now is the `QGF.num_qubits` function, which determines how many qubits are required to represent an instance of the given `QGF` type. For binary fields, it would be equal to the `self.degree` but for arbitrary fields, it's currently calculated using the log2 of `self.characterisitic ** self.degree`. We could decide to change the encoding so we calculate it as `bit_length(self.characterisitc) * self.degree` instead, but we'll have to use `math.log` for the calculation anyway. 

Rest of the logic to convert to bits has no dependence on `math.log`.

NoureldinYosri:
> calculate it as bit_length(self.characterisitc) * self.degree
shouldn't it be `math.ceil(self.degree * math.log2(self.characterisitc))` which will be $\leq$ the above expression

would you like to add it in the is PR or in a follow up PR?

tanujkhattar:
This PR is independent and unlocks the linked issue so let's merge this. The remaining optimization can be done later

NoureldinYosri (approved):

PR #1571, opened 2025-02-27
Add .markdownlintrc config file for markdownlint

mhucka:
This configures `markdownlint`, a Markdown linter/formatter, to follow the Google style guide for Markdown as closely as possible. This allows users to use `markdownlint` when writing and editing Markdown files, and gets us ready to add a Markdown linter check in the CI workflow.

### DIFF
--- a/.markdownlintrc
+++ b/.markdownlintrc
@@ -1,0 +1,152 @@
+{ // Summary: markdownlint config file for Cirq                    -*- jsonc -*-
+  //
+  // Note: there are multiple programs programs named "markdownlint". For
+  // Cirq, we use https://github.com/igorshubovych/markdownlint-cli/, which is
+  // the one you get with "brew install markdownlint" on MacOS.
+  //
+  // These settings try to stay close to the Google Markdown Style as
+  // described at https://google.github.io/styleguide/docguide/style.html
+  //
+  // For a list of configuration options, see the following page:
+  // https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md
+  // (Beware that the above looks similar but is NOT the same as the page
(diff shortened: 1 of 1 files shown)


.markdownlintrc
```diff
@@ -0,0 +1,152 @@
+{ // Summary: markdownlint config file for Cirq                    -*- jsonc -*-
```
mpharrigan, line 1: "cirq"

mpharrigan:
what's the intended use? we have very few true markdown files in the repository. (a lot more if you count docstrings, jupyter notebooks, and autogenerated api docs)

.markdownlintrc
mhucka: Face palm. Thanks for catching that.

mhucka:
> what's the intended use? we have very few true markdown files in the repository. (a lot more if you count docstrings, jupyter notebooks, and autogenerated api docs)

That's a fair question. I'm about to add several more md files; in addition, with the other Quantumlib projects using Markdown more heavily, I _hope_ to see the Qualtran project from rST to md for their docs/ too … (But even if that doesn't happen, there will always be several Markdown files, and IMHO it's worth automating style checks if there's more than one.)

mpharrigan:
We don't actually use rst either. We use the myst-nb sphinx plugin to take jupyter notebooks (which indeed use markdown cells) and turn them into docs pages via the sphinx doc builder

mhucka:
> We don't actually use rst either. We use the myst-nb sphinx plugin to take jupyter notebooks (which indeed use markdown cells) and turn them into docs pages via the sphinx doc builder

Oh, I see. Good to know! 

I'm examining ipynb linters too, but don't know yet if markdownlint can be made part of that. (Hopefully it can, so that the same config will work everywhere.)

PR #1572, opened 2025-02-27
Add yamllint config file

mhucka:
This configures `yamllint`, a YAML file linter/formatter, to follow a common style for Quantumlib repos. This allows users to use `yamllint` when writing and editing YAML files, and gets us ready to add a YAML file linter check in the CI workflow.


.yamllint.yaml
```diff
@@ -0,0 +1,14 @@
+# Summary: yamllint configuration for Cirq.
```
mpharrigan, line 1: cirq!

mpharrigan (approved):

.yamllint.yaml
mhucka: How many times will I do that today.

PR #1575, opened 2025-03-03
Add explicit versions and timeouts to GHA workflows

mhucka:
This updates the GitHub Actions workflows to conform to some Google requirements and best practices:

- pin versions of GitHub Actions to commit SHAs
- pin versions of runners instead of using *-latest
- set default permissions for workflow
- set timeouts on jobs, to prevent running for 6 hrs (the default) in case of hangs or loops

This also fixes the following Scorecard security warnings:

- https://github.com/quantumlib/Qualtran/security/code-scanning/17
- https://github.com/quantumlib/Qualtran/security/code-scanning/18
- https://github.com/quantumlib/Qualtran/security/code-scanning/20
- https://github.com/quantumlib/Qualtran/security/code-scanning/21
- https://github.com/quantumlib/Qualtran/security/code-scanning/22
- https://github.com/quantumlib/Qualtran/security/code-scanning/23
- https://github.com/quantumlib/Qualtran/security/code-scanning/24
- https://github.com/quantumlib/Qualtran/security/code-scanning/25
- https://github.com/quantumlib/Qualtran/security/code-scanning/26
- https://github.com/quantumlib/Qualtran/security/code-scanning/27
- https://github.com/quantumlib/Qualtran/security/code-scanning/28
- https://github.com/quantumlib/Qualtran/security/code-scanning/29
- https://github.com/quantumlib/Qualtran/security/code-scanning/30
- https://github.com/quantumlib/Qualtran/security/code-scanning/31
- https://github.com/quantumlib/Qualtran/security/code-scanning/32
- https://github.com/quantumlib/Qualtran/security/code-scanning/33
- https://github.com/quantumlib/Qualtran/security/code-scanning/34
- https://github.com/quantumlib/Qualtran/security/code-scanning/35
- https://github.com/quantumlib/Qualtran/security/code-scanning/36
- https://github.com/quantumlib/Qualtran/security/code-scanning/37
- https://github.com/quantumlib/Qualtran/security/code-scanning/38
- https://github.com/quantumlib/Qualtran/security/code-scanning/39

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -16,13 +16,25 @@ on:
   schedule:
     - cron: "0 11 * * 1-5"  # Mon-Fri at 11:00 AM UTC
 
+  # Allow manual invocation – useful for debugging.
+  workflow_dispatch:
+
+concurrency:
+  # Cancel any previously-started but still active runs on the same branch.
+  cancel-in-progress: true
+  group: ${{github.workflow}}-${{github.event.pull_request.number||github.ref}}
+
+# Declare default permissions as read only.
+permissions: read-all
+
 jobs:
   pytest:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
+    timeout-minutes: 60
     steps:
-      - uses: actions/checkout@v3
-      - uses: ts-graphviz/setup-graphviz@v1
-      - uses: actions/setup-python@v4
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      - uses: ts-graphviz/setup-graphviz@c001ccfb5aff62e28bda6a6c39b59a7e061be5b9 # v1
+      - uses: actions/setup-python@42375524e23c412d93fb67b49958b491fce71c38 # v5
         with:
           python-version: "3.10"
       - name: Install dependencies
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -17,13 +17,22 @@ on:
     branches:
       - main
 
+concurrency:
+  # Cancel any previously-started but still active runs on the same branch.
+  cancel-in-progress: true
+  group: ${{github.workflow}}-${{github.event.pull_request.number||github.ref}}
+
+# Declare default permissions as read only.
+permissions: read-all
+
 jobs:
   pytest:
-    runs-on: ubuntu-latest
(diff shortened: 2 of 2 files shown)


.github/workflows/nightly.yaml
```diff
@@ -16,13 +16,25 @@ on:
   schedule:
     - cron: "0 11 * * 1-5"  # Mon-Fri at 11:00 AM UTC
 
+  # Allow manual invocation – useful for debugging.
+  workflow_dispatch:
```
mpharrigan, line 20: there's no value for this key?

.github/workflows/pr.yaml
```diff
@@ -17,13 +17,22 @@ on:
...
+  group: ${{github.workflow}}-${{github.event.pull_request.number||github.ref}}
+
+# Declare default permissions as read only.
+permissions: read-all
+
 jobs:
   pytest:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
+    timeout-minutes: 20
```
mpharrigan, line 31: how did you decide on the timeouts?

mpharrigan (approved):

.github/workflows/nightly.yaml
mhucka: That's correct. It doesn't need one unless one wants to pass arguments to the workflow. (In this case, the workflow doesn't take arguments, so there's nothing to add.)

I agree it looks weird :-). But it does work, since it's used in other repos.

.github/workflows/pr.yaml
mhucka: An excellent question. I looked at maybe 6-10 of the existing runs of these workflows, and looked at the times the steps took (GitHub lists the times in the GUI), then came up with numbers by making the timeouts very generous. E.g., for short steps that took < 1m, I used a timeout of 5-10 minutes, while for longer (I think the longest that any step took was < 9m), I set them to 20 minutes.

What makes this a little easier is that the goal of this is only to stop processes in the unlikely case that something makes them not stop naturally. So, we don't have to guess accurately, just guess a reasonable outer limit.

If you're at all worried, we could go higher, like 30 minutes or even higher. Right now, looking at past run histories, it seemed that if a job went longer than 20 minutes, it would probably be an indication of something going wrong. Getting an error then might be useful to finding out something was even happening. But it's your project, and you would know better! So please let me know.

mpharrigan: sounds good, just curious

PR #1580, opened 2025-03-04
Add problem-matchers for Pylint, MyPy, and Black

mhucka:
GitHub [Problem Matchers](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md) are a mechanism that enable workflow steps to scan the outputs of GitHub Actions for regex patterns and automatically write annotations in the workflow summary page. Using Problem Matchers allows information to be displayed more prominently in the GitHub user interface.

This add problem matchers for Pylint, Mypy and Black output.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -94,7 +94,7 @@ jobs:
           python -m pip install --upgrade pip
           pip install -r dev_tools/requirements/envs/format.env.txt
           pip install --no-deps -e .
-      - name: Set up Black output problem matcher
+      - name: Set up problem matcher for Black output
         run: echo '::add-matcher::.github/problem-matchers/black.json'
       - run: |
           check/format-incremental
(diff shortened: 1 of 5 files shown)


.github/workflows/pr.yaml
```diff
@@ -130,6 +134,8 @@ jobs:
           python -m pip install --upgrade pip
           pip install -r dev_tools/requirements/envs/dev.env.txt
           pip install --no-deps -e .
+      - name: Set up problem matcher for Mypy output
+        run: echo '::add-matcher::.github/problem-matchers/mypy.json'
```
mpharrigan, line 138: just echoing things triggers this?

mpharrigan (approved):

.github/workflows/pr.yaml
mhucka: I know it looks bonkers, but believe it or not, that bit with the `::add-matcher::` is the special sauce that GitHub uses to detect what's happening. Why? ¯\_(ツ)_/¯

PR #1581, opened 2025-03-05
Mark notebook unit tests

mpharrigan:
So they don't get executed as part of the quick suite. They get tested by the dedicated notebook CI job


mhucka (approved):
LGTM

Issue #445, opened 2023-10-25
Measurement and Classical Data


mpharrigan:
We need a measurement bloq. Measure-and-throw-out-the-result can be simulated by CNOTing to a measure qubit and then tracing it out (xref quantum channels roadmap item).
 
When we measure and keep the result, we want the result to be available in the compute graph for further processing or conditioning other gates on it. Measuring and keeping the result can be codified by CNOTing to a fresh register. This is a new type of register: a classical data register. From a simulation and tensor contraction standpoint it behaves identically to a new qubit. The restriction comes from the fact that we can only apply classical-logic bloqs to it. I.e.: all gates must be hadamard-free. Classical data can be copied and deleted but that’s just CNOT and tracing out, respectively. 

The way measurements are used in Cirq is with a stochastic simulator and statistics are gathered by running the simulation many times. This must correspond to cutting or projecting your tensor network stochastically. Before we try to support this NISQy view of measurement, I would like to have a very clear idea of why we’re using it. Presumably we use these simulation techniques to write unit tests. The full everettian simulation of classical feedback could get too expensive and projecting the classical wires to specific values could let us “fuzz test” non-exhaustively. But if we’re projecting those things out for performance reasons; why wouldn’t we also project out qubits as well?

mpharrigan:
I'm working on this

PR #1584, opened 2025-03-05
[open-systems] QCDType, CDType, QDType

mpharrigan:
This pull request is the first in a series for supporting open system (classical data & measurement), roadmap item #445.

## `QCDType`

 - The existing `QDType` base class for quantum data types now descends from the more general `QCDType`. `CDType` is introduced to represent classical data types.
 - `CBit` is added, which is a single classical bit (analogous to `QBit`). 
 - Python type annotations are updated as needed. In particular, `Register`'s `dtype` argument can be either quantum or classical. 
 - This existing, runtime, Qualtran-based type checking basically works without modification. It won't let you plug a quantum wire into a classical bloq (or vice versa)

## `_ControlledBase`

 - I've refactored the class hierarchy of `Controlled`. The circuit-implementation-agnostic bit is now in `_ControlledBase` from which `Controlled` (which uses a "total control" circuit) and `ControlledViaAnd` descend.
 - `CtrlSpec` supports classical data types (`CDType`)
 - The logic for choosing the correct default controlled meta-bloq based on control spec is factored out into a function
 - This is to support plugging in a new, forthcoming `ClassicallyControlled` metabloq

## Drawing

 - The musical score diagrams draw classical wires in blue.

## Documentation

 - The existing `QDType` system was not really documented (at least not with its own notebook). I went ahead and wrote docs for the data type system.
 - There is a section at the end about the `QCDType` class hierarchy
 
## Naming

 - Before this PR and throughout the framework, `qdtype: QDType` and `dtype: QDType` have been used interchangeably. After this PR, idiomatically: `qdtype` should be the variable name used for a `QDType` and `dtype` should be reserved for variables that can be either quantum or classical (i.e. `QCDType`). Where the variable name isn't backwards-compatibility breaking, I've made those changes in this PR; but there are many places where `dtype` is used where `qdtype` would be more appropriate
 - Before this PR and throughout the framework, `bitsize` and `num_qubits` (as variable/method names) have been used interchangeably. This PR adds methods/properties that aim to disambiguate these quantities. `bitsize` or `num_bits` is the total number of quantum and/or classical bits. `num_qubits` is strictly qubits (resource counting still counts strictly qubits). `num_cbits` is added; which is the number of strictly classical qubits. 

## Non-changes (yet)

 - While `CDType`, `CBit`, and `CtrlSpec(CBit())` are introduced, there's no real way to use them (with the exception of `Cast`, which is likely not what you really want).


qualtran/_infra/registers_test.py
```diff
@@ -135,6 +143,13 @@ def test_signature_symbolic():
     assert str(signature.n_qubits()) == 'n_x + n_y'
 
 
+def test_partial_classical_signature_n_qubits():
+    sig = Signature(
+        [Register('x', QUInt(5)), Register('y', QUInt(5), side=Side.RIGHT), Register('c', CBit())]
+    )
+    assert sig.n_qubits() == 10
```
tanujkhattar, line 150: Also assert for `sig.n_cbits()` (and if doesn't exist?)

qualtran/bloqs/bookkeeping/cast.py
```diff
@@ -43,30 +46,51 @@
...
-        shape: shape of the register to cast.
+        inp_dtype: Input QCDType to cast from.
+        out_dtype: Output QCDType to cast to.
+        shape: Optional multidimensional shape of the register to cast.
+        allow_quantum_to_classical: Whether to allow (potentially dangerous) casting a quantum
+            value to a classical value and vice versa. If you cast a classical bit to a qubit
+            that was originally obtained by casting a qubit to a classical bit, the program
+            will model unphysical quantum coherences that can give you fundamentally incorrect
+            resource estimates. Use a measurement operation to convert a qubit to a classical
+            bit (and correctly model the decoherence that results).
```
tanujkhattar, line 62: Why do we need this? What's the motivation to allow casting a classical wire as a quantum wire?

tanujkhattar (approved):
LGTM % one main question about casting classical wires to quantum wires. Off the top of my head, I think it shouldn't be allowed but curious to hear why you think otherwise.

qualtran/bloqs/bookkeeping/cast.py
mpharrigan: - You can imagine a circuit where you do a non-destructive measurement of a qubit followed by subsequent quantum operations. The measurement destroys coherences but it doesn't truly stop you from doing subsequent unitaries. https://algassert.com/quirk#circuit={%22cols%22:[[%22H%22],[%22Measure%22],[%22H%22]]} Quirk doesn't let you do this, but it's apologetic about it. This is logically equivalent to allocating a new qvar, doing a classically-controlled-X into it, and discarding the measurement bit; so it's a valid action.  
 - I explicitly use it in the documentation for the open-system simulation code https://github.com/quantumlib/Qualtran/pull/1596/files#diff-42bc0fab1608b2d3159b19895323bc971b200c018018857cf4dcba6dee062320R223 -- you can write a measurement operation with only unitary bloqs and `Discard` (but if we do it this way; we need to explicitly tell the type system that's what we're doing)
 - Philosophically, `Cast` is always a bit of a hack; and its usage should always be minimized. But it exists (much like Python `typing.cast` and cpp `reinterpret_cast`) for niche situations if you know what you're doing. Since this type of casting might give you incorrect resource estimates, I made the judgement call to guard it behind this verbose kwarg 
 - On the pure-hack side, I also am using it temporarily to prototype things where the "classical version" of all the operations don't exist yet.

qualtran/_infra/registers_test.py
mpharrigan: added `sig.n_cbits()` and `sig.n_bits()` for consistency with `Register` and `QCDType` and asserted them here; thanks. 

I kept the spellings consistent _within_ the class, so it's  `QCDType.num_xbits`, `Register.total_xbits`, `Signature.n_xbits`. Maybe we can provide a consistent spelling (in addition to the historical one(s)) in a follow-on PR.

qualtran/bloqs/bookkeeping/cast.py
tanujkhattar: > This is logically equivalent to allocating a new qvar, doing a classically-controlled-X into it, and discarding the measurement bit; so it's a valid action.

As a mental model, this looks cleaner to me as well. I think this would also play nicely when converting to lower level tools like Cirq, where the idea of applying quantum operations on classical bits does not exist. I don't see the benefits of this approach and I haven't seen any papers where people describe circuits using such operations. 

This might cause more confusion than benefit, so if this is not standard notation we should try to avoid it and stick to describing operations explicitly using qubit allocations + classically controlled operations.

mpharrigan: Well, one could always write it that way. For points 2-4, I'm going to keep it in for now

PR #1586, opened 2025-03-05
Add Pennylane to dependencies for interop

mpharrigan:
See #1559


mpharrigan:
We'll need to update this upon the next release of pennylane, opened #1589 to track

NoureldinYosri (approved):

PR #1595, opened 2025-03-13
Add ctrl system for `LinearCombination`

anurudhp:
Fixes the nightly test fail (`qualtran/bloqs/block_encoding/chebyshev_polynomial_test.py::test_scaled_chebyshev_even_tensors`)

Added two cost tests to ensure that no `C[And]` etc are in the call graph.


mpharrigan:
Thanks for doing this. When I investigated, I traced the ultimate problem to trying to do `Controlled(LessThanEqual)` as the proximate operation that tries to do `C[And]`. Would implementing `LessThanEqual.controlled()` fix the bug or is there something else wrong that requires the fix to be at the level of `LinearCombination` ? 

(note that we do have `CLinearDepthGreaterThan`)

anurudhp:
I think so, but `LinearCombination` uses `StatePreparationAliasSampling` (which uses `LessThanEqual`) in a compute-uncompute pair, so this override (i.e. only control the `select`) results in a lower cost.

mpharrigan:
Very well, thank you!

mpharrigan (approved):

PR #1596, opened 2025-03-14
[open-systems] Superoperator tensor contraction

mpharrigan:
This pull request is the third in a series for supporting open system (classical data & measurement), roadmap item https://github.com/quantumlib/Qualtran/issues/445.

This branch is based on #1590 (for my convenience; it doesn't use any of that functionality)

## `Bloq.tensor_contract(superoperator=True)`

This is the new interface. It will return a 0-, 2-, or 4-dimensional tensor for constants, density matrices, and superoperators (resp) to support simulating programs with measurements and non-unitary maps. Read the docs for the indexing convention for the superoperator; but if you're really doing something with the superoperators you may want to operate on the quimb tensor network with friendly indices itself.

## `cbloq_to_superquimb`

The workhorse of the new functionality is this function that converts a composite bloq to a tensor network that represents the density matrix and/or superoperator of the program. Some operations like measurement or discarding qubits cannot be represented in a pure-state statevector / unitary picture.

`cbloq_to_superquimb` still uses `Bloq.my_tensors` to construct the network. It adds each tensor twice: once in the "forward" direction and again in the "backward" (adjoint) direction. 

## `DiscardInd`

The `Bloq.my_tensors` method can now return a `DiscardInd` object, which signifies that that tensor index should be discarded by tracing it out. 

It turns out that this simple extension lets you represent any CPTP map using the "system+environment" modeling approach.

## `MeasZ`, `Discard`

This PR includes two non-unitary maps: measuring in the Z basis and discarding a qubit or classical bit. 

## Manipulating quimb tensor networks

This PR makes some changes to the way we handle outer indices in the quimb tensor network. This should only be of interest to folks who use the "hands on" approach of directly manipulating the tensor networks. The changes make sense and clean up some of the docs.

The outer indices are now re-mapped to tuples of qualtran objects that actually make sense rather than keeping them as `Connection`s to `DanglingT` objects. There is a new flag `friendly_indices=True` which will make the outer indices friendly strings.

**breaking change** I renamed and made-private the `get_left_and_right_inds` function. The original idea was to use this as a public way of getting a handle on the outer indices, but it was always tricky to use; and needed to be changed to support the new variety of outer indices present in the superoperator networks. No one should need a replacement for this function since the `qtn.TensorNetwork` object outer indices now just intrinsically make sense. 

## Non-changes

This includes the minimal set of non-unitary CPTP bloqs to support the docs notebooks.

### DIFF
--- a/qualtran/bloqs/basic_gates/discard.py
+++ b/qualtran/bloqs/basic_gates/discard.py
@@ -1,0 +1,68 @@
+#  Copyright 2024 Google LLC
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+from functools import cached_property
+from typing import Dict, List, TYPE_CHECKING
+
+from attrs import frozen
+
+from qualtran import Bloq, CBit, ConnectionT, QBit, Register, Side, Signature
+from qualtran.simulation.classical_sim import ClassicalValT
+
+if TYPE_CHECKING:
+    from qualtran.simulation.tensor import DiscardInd
+
+
+@frozen
+class Discard(Bloq):
+    """Discard a classical bit.
+
+    This is an allowed operation.
+    """
+
+    @cached_property
+    def signature(self) -> 'Signature':
+        return Signature([Register('c', CBit(), side=Side.LEFT)])
+
+    def on_classical_vals(self, c: int) -> Dict[str, 'ClassicalValT']:
+        return {}
+
+    def my_tensors(
+        self, incoming: Dict[str, 'ConnectionT'], outgoing: Dict[str, 'ConnectionT']
+    ) -> List['DiscardInd']:
+
+        from qualtran.simulation.tensor import DiscardInd
+
+        return [DiscardInd((incoming['c'], 0))]
+
+
+@frozen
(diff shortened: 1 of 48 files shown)


tanujkhattar (commented):
@mpharrigan Is this blocked on something?

@NoureldinYosri Can you please take a look? 

I'll be happy to LGTM and merge it since I want to use the functionality in this PR.

qualtran/bloqs/basic_gates/discard.py
```diff
@@ -0,0 +1,68 @@
+#  Copyright 2024 Google LLC
```
NoureldinYosri:
```suggestion
#  Copyright 2025 Google LLC
```

qualtran/bloqs/basic_gates/discard_test.py
```diff
@@ -0,0 +1,82 @@
+#  Copyright 2024 Google LLC
```
NoureldinYosri:
```suggestion
#  Copyright 2025 Google LLC
```

qualtran/bloqs/basic_gates/z_basis.py
```diff
@@ -379,6 +391,44 @@ def _cz() -> CZ:
...
+
+    Registers:
+        q [LEFT]: The qubit to measure.
+        c [RIGHT]: The classical measurement result.
+    """
+
+    @cached_property
+    def signature(self) -> 'Signature':
+        return Signature(
+            [Register('q', QBit(), side=Side.LEFT), Register('c', CBit(), side=Side.RIGHT)]
```
NoureldinYosri, line 406: nit: if the measurement is nondemolition measurement then the qubit can be reused. which would be simpler than allocating a new qubit in the state indicated by the classical bit. 

perhabs a boolean that controls whether the qubit is LEFT or TRHOUGH can be used here?

mpharrigan: That is certainly a valid operation, although I'd advocate for that being its own bloq.

The current construct is simpler if you're measuring a qubit to get its bit value

PR #1603, opened 2025-03-26
Strict `BloqAsCirqGate` and `CirqGateAsBloq`

### LINKED ISSUE #1570: [BUG] TextbookQPE does not decompose successfully
**Expected Behavior:**
We expect `TextbookQPE().decompose_bloq()` to execute without error

**Actual Behavior:**
When we combine GQSP with QPE, we get a `ValueError` when we try `decompose_bloq()`
```
from qualtran.bloqs.phase_estimation import RectangularWindowState, TextbookQPE
from qualtran.bloqs.hamiltonian_simulation.hamiltonian_simulation_by_gqsp import HamiltonianSimulationByGQSP
from qualtran.bloqs.chemistry.hubbard_model.qubitization import (
    get_walk_operator_for_hubbard_model,
)

walk_op = get_walk_operator_for_hubbard_model(2, 2, 1, 1)
hubbard_time_evolution_by_gqsp = HamiltonianSimulationByGQSP(walk_op, t=5, precision=1e-7)

textbook_qpe_w_gqsp = TextbookQPE(hubbard_time_evolution_by_gqsp, RectangularWindowState(3))

textbook_qpe_w_gqsp.decompose_bloq()
```

Raises:
`ValueError: Cannot control channel with non-unitary operators: bloq.HamiltonianSimulationByGQSP`

mpharrigan:
We use the adapter pattern for adapting cirq and qualtran (bloq) obejcts, namely `BloqAsCirqGate` and `CirqGateAsBloq`. This PR does not modify `qualtran.GateWithRegisters`, which acts simultaneously as a cirq gate and qualtran bloq.

## `BloqAsCirqGate` is now _only_ a `cirq.Gate`.

This PR removes the `signature` and `on_registers` methods from `BloqAsCirqGate(cirq.Gate)`. These methods are bloq-y and one shouldn't rely on them being available on something that purports to be a cirq gate.

`BloqAsCirqGate` also learns a new `_has_unitary_` implementation, which fixes #1570. This fix was revealed when I encountered a related bug while writing this PR: one of the unit tests started taking a long (indefinite) time; which was again traced to `cirq.Gate.controlled_by` trying to validate that the gate was unitary. 

## `CirqGateAsBloq` is now  _only_ a `Bloq`.

This PR removes the `GateWithRegisters(Bloq, cirq.Gate)` functionality from the adapter. Only the bloq API is available on `CirqGateAsBloq` (so things like `_unitary_`, diagram info, ... are removed). 

The functionality for `CirqGateAsBloq` is (still) contained in `CirqGateAsBloqBase` which can be used to bootstrap bloq definitions.


qualtran/cirq_interop/_cirq_to_bloq_test.py
```diff
@@ -77,16 +78,18 @@ def test_cirq_gate_as_bloq_for_trivial_gates():
...
 
 def test_cirq_gate_as_bloq_tensor_contract_for_and_gate():
     and_gate = And()
     bb = BloqBuilder()
     ctrl = [bb.add(OneState()) for _ in range(2)]
-    ctrl, target = bb.add(CirqGateAsBloq(and_gate), ctrl=ctrl)
-    cbloq = bb.finalize(ctrl=ctrl, target=target)
+    target = bb.add(ZeroState())
+    q = [*ctrl, target]
+    c0, c1, target = bb.add(CirqGateAsBloq(and_gate), q=q)
```
tanujkhattar, line 91: This seems like a backwards incompatible change? The CirqGateAsBloq doesn't preserve the signature anymore if the underlying gate is a `GateWithRegisters` ?

In theory, this should be fine since `GateWithRegisters` are bloqs so we should use them without wrapping in `CirqGateAsBloq`, but I'm nervous that this might break stuff in the interop. The tests are passing so that's a good indication, but have you looked into any potential consequence of this?

tanujkhattar: One way to be full proof could be to raise an error if a user tries to wrap a `GateWithRegisters` into a `CirqGateAsBloq` (or any other object that satisfies `isinstance(gate, Bloq)`). 

I don't like the current behavior where it silently works, but does the wrong thing. It's better to explicitly raise an error if we don't want / expect users to wrap bloq objects in this wrapper

tanujkhattar (changes requested):
Main feedback is to explicitly raise an error if `CirqGateAsBloq` is passed a `Bloq` object, instead of silently ignoring the LEFT / RIGHT registers and making everything a THRU register.

qualtran/cirq_interop/_cirq_to_bloq_test.py
mpharrigan: Yes, I can add some error checking since this changes the behavior of using `CirqGateAsBloq` with a `GateWithRegisters`. 

It was on purpose to make `CirqGateAsBloq` only use the `cirq.Gate` API, which doesn't have a notion of signatures. If we wanted to preserve the "special casing", I would (theoretically) argue for a different adapter: `GateWithRegistersAsBloq`... but this is redundant :)

mpharrigan: actually, I don't know if an error is appropriate: since `GateWithRegisters` is a `cirq.Gate`, you should be able to wrap it -- but it will treat it as a cirq gate. Maybe I'll emit a warning?

mpharrigan: It won't silently fail anyways: if you were wrapping a sided GateWithRegisters with CirqGateAsBloq, the change will change the signature; which will raise an error when you're trying to wire it up. 

And of course the fix is simple: remove the adapter! I've added a warning

mpharrigan:
This PR triggers #1336

tanujkhattar (approved):

mpharrigan:
I'm going to revert the part of `BloqAsCirqGate._unitary_` changes that forces the `tensor_contract` path

mpharrigan:
pytest time back to normal (lower than normal?)

PR #1614, opened 2025-04-02
[v0.6 branch] Optional new pennylane functionality

mpharrigan:
Since we can't push a PyPI package that depends on a url-specified dev branch of pennylane, make a patch against the v0.6 branch to fail gracefully if you only have the existing pennylane release.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -16,6 +16,7 @@ on:
   pull_request:
     branches:
       - main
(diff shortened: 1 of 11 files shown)


mpharrigan:
This isn't ready yet; I need to update the dependency file

mpharrigan:
this is ready

mhucka (approved):
LGTM

PR #1620, opened 2025-04-07
[QGFPoly] Add `QGFPoly`, a new data type to represent polynomials over galois fields

tanujkhattar:
Part of a larger effort to add bloqs that perform arithmetic over polynomials with coefficients in a galois field.

### DIFF
--- a/qualtran/_infra/data_types.py
+++ b/qualtran/_infra/data_types.py
@@ -1078,7 +1078,7 @@ class QGFPoly(QDType):
     def to_gf_coefficients(self, f_x: galois.Poly) -> galois.Array:
         """Returns a big-endian array of coefficients of the polynomial f(x)."""
         f_x_coeffs = self.qgf.gf_type.Zeros(self.degree + 1)
-        f_x_coeffs[self.degree - f_x.degree:] = f_x.coeffs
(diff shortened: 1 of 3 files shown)


qualtran/_infra/data_types.py
```diff
@@ -1025,6 +1025,100 @@ def __str__(self):
...
+    def assert_valid_classical_val(self, val: Any, debug_str: str = 'val'):
+        """Raises an exception if `val` is not a valid classical value for this type.
+
+        Args:
+            val: A classical value that should be in the domain of this QDType.
+            debug_str: Optional debugging information to use in exception messages.
+        """
+        if not isinstance(val, galois.Poly):
+            raise ValueError(f"{debug_str} should be a {galois.Poly}, not {val!r}")
+        if not val.field == self.qgf.gf_type:
```
NoureldinYosri, line 1100: nit: galois library uses singleton fields

```suggestion
        if val.field is not self.qgf.gf_type:
```

NoureldinYosri (approved):

qualtran/_infra/data_types.py
tanujkhattar: done.

PR #1622, opened 2025-04-08
[QGFPoly] Add `GF2PolyAddK`, `GFPolySplit` and `GFPolyJoin` bloqs

tanujkhattar:
Part of a larger effort to add bloqs that perform arithmetic over polynomials with coefficients in a galois field.

Adds bloqs for 
- `GFPolySplit`: Split a register of type `QGFPoly` into `n + 1` registers of type `QGF` where `n` is the degree of the polynomial.
- `GFPolyJoin`: Join `n + 1` registers of type `QGF`, where `n` is the degree of the polynomial, into a register of type `QGFPoly`.
- `GF2PolyAddK`: Bloq to add a constant polynomial into a quantum register of type `QGFPoly`.


tanujkhattar:
@NoureldinYosri I've added tests for gate counts, as you had pointed out in the review for the bigger PR. Can you LGTM this first so I can merge this first, rebase the second PR, and then merge that as well ?

NoureldinYosri (approved):

PR #1625, opened 2025-04-08
Add affiliations & ORCIDs to authors in CITATION.cff file

mhucka:
This adds affiliations and ORCID numbers to all the authors. It also makes use of YAML anchors and aliases (a feature I didn't know existed until recently) to avoid having to duplicate the author list in 2 places.


mpharrigan:
> YAML anchors and aliases to avoid having to duplicate the author list in 2 places.

is that change included in this PR?

mhucka:
> > YAML anchors and aliases to avoid having to duplicate the author list in 2 places.
> 
> is that change included in this PR?

Ahem. It was _supposed_ to be, but apparently I didn't do a git push after that last change :-(.  Thanks for catching that.

PR #1638, opened 2025-04-29
Modify Sphinx theme to use Qualtran theme colors and icons

mhucka:
This changes the colors to be the ones used by QAI and the Qualtran logo. It also adds the Qualtran logo in the upper left and the QAI logo on the bottom of the page. Finally, it changes the section headings in the sidebar to be the name of the current page instead of always being "Section navigation".

### DIFF
--- a/dev_tools/requirements/deps/docs.txt
+++ b/dev_tools/requirements/deps/docs.txt
@@ -2,6 +2,7 @@ ipywidgets
 nbconvert
 nbformat
 sphinx
+sphinx-autobuild
 pydata-sphinx-theme
 myst-nb
 tensorflow-docs~=2023.5.24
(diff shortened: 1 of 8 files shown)


mpharrigan (approved):
nice!

mpharrigan:
how did you re-generate the dependencies? any change should also update `dev.env.txt` which includes all the possible dependencies and serves as a "constraint" for the other subsets

mhucka:
> how did you re-generate the dependencies? any change should also update `dev.env.txt` which includes all the possible dependencies and serves as a "constraint" for the other subsets

I used the pip-compile script, but only checked in the docs env file, under the (apparently mistaken) assumption that the changes were limited to that. Guess that was wrong …

PR #1642, opened 2025-05-05
4 Ts per CSwap

### LINKED ISSUE #1463: Incorrect (or at least inconsistent) T cost for a TwoBitCSwap gate.
To summarize a discussion I had with @tanujkhattar, @mpharrigan, and @NoureldinYosri:

We currently count the T and Clifford costs for a TwoBitCSwap (Fredkin) gate using the construction from https://arxiv.org/pdf/1308.4134. However, we can implement this gate using two CNOT gates and a Toffoli gate, which has a four T-gate construction using ancilla qubits (https://arxiv.org/pdf/1212.5069).

I think we should default to the ancilla-assisted construction for the Fredkin gate. 

The simplest solution might be to no longer treat a Fredkin gate as a leaf node at all and instead decompose it into a Toffoli gate and two CNOT gates. If not, then we should at least be consistent in how we treat the resource counts for Fredkin and Toffoli gates.

mpharrigan:
Fixes #1463 

 - change default ts per cswap to 4
 - leaves it as inconsistently 7 in `to_legacy_t_complexity` with a note


qualtran/bloqs/chemistry/df/double_factorization_test.py
```diff
@@ -80,8 +80,7 @@ def test_compare_cost_to_openfermion():
     inner_prep_qrom_diff = 8
     prog_rot_qrom_diff = 60
     missing_toffoli = 4  # need one more toffoli for second application of CZ
-    swap_cost = 4 * (7 - 4) * num_spin_orb // 2
```
fdmalone, line 83: nice!

qualtran/bloqs/chemistry/hubbard_model/qubitization/select_hubbard_test.py
```diff
@@ -35,7 +35,7 @@ def test_select_t_complexity(dim):
     assert cost == GateCounts(
         cswap=2 * logN, and_bloq=5 * (N // 2) - 2, measurement=5 * (N // 2) - 2, clifford=ANY
     )
-    assert cost.total_t_count() == 10 * N + 14 * logN - 8
+    assert cost.total_t_count() == 10 * N + 8 * logN - 8
```
fdmalone, line 38: What does the linear T paper quote again / why does this change?

fdmalone (approved):

qualtran/bloqs/chemistry/hubbard_model/qubitization/select_hubbard_test.py
mpharrigan: per above, we assert cswap=2*logN. This changes the T count from 7*2*logN to 4*2*logN, so I think it's legit

PR #1643, opened 2025-05-07
feat: multi-level decomposition for (non-symbolic) bloqs

pqvr:
## Description

Each `Bloq` object has a `decompose_bloq` method. Consider the following snippet

```py
from qualtran.bloqs.qsp.generalized_qsp import GeneralizedQSP
from qualtran.bloqs.basic_gates import XPowGate

bloq = GeneralizedQSP.from_qsp_polynomial(XPowGate(), (0.5, 0.5))
decomp = bloq.decompose_bloq()
```

What I want to do is to continually decompose all of the subsequent bloqs. However, the type of the decomposed bloq is

```
CompositeBloq([3 subbloqs...])
```

and this does *not* have a `decompose_bloq` method. This feature allows one to completely decompose it into its constituent parts.

## Example

Consider the following example of defining a `Bloq` object and decomposing it.

```py
from qualtran.bloqs.basic_gates import XPowGate
from qualtran.bloqs.qsp.generalized_qsp import GeneralizedQSP
from qualtran.qref_interop._bloq_to_qref import bloq_to_qref
from qref.experimental.rendering import to_graphviz


bloq_qsp = GeneralizedQSP.from_qsp_polynomial(XPowGate(), (0.5, 0.5))
```

The following generates the decomposition (as was done before) without decomposing recursively:

```py
# Non-preserve bloqs on decomposition (default).
schema_qsp = bloq_to_qref(bloq_qsp)
plot_qsp = to_graphviz(schema_qsp)
plot_qsp.render(directory="doctest-output/qsp", format="png", cleanup=True)
```

Digraph gv

Using the new feature, we can select to "blow out" the subsequent `Bloq` objects for subsequent viewing:

```py
# Preserve bloqs on decomposition.
schema_qsp_preserve = bloq_to_qref(bloq_qsp, decomposition_rules=True)
plot_qsp_preserve = to_graphviz(schema_qsp_preserve)
plot_qsp_preserve.render(directory="doctest-output/qsp_preserve", format="png", cleanup=True)
```

Digraph gv

## Limitations

This works well when the inputs are numeric, but it does not work well for symbolic inputs. As this is outside of scope of this issue, the description of that will be described in a subsequent issue. 

Tagging @mstechly


qualtran/qref_interop/_bloq_to_qref.py
```diff
@@ -129,22 +131,104 @@ def _bloq_instance_name(instance: BloqInstance) -> str:
 
 
 def bloq_to_qref(
-    obj: Union[Bloq, CompositeBloq, BloqInstance], from_callgraph: bool = False
+    obj: Union[Bloq, CompositeBloq, BloqInstance],
+    *,
+    from_callgraph: bool = False,
+    preserve: Union[bool, Iterable[type[Bloq]]] = False,
```
mstechly, line 137: **suggestion**: I'd rename it to "decomposition_rules" or sth like that? "preserve" seems a bit too vague.

mstechly:
@mpharrigan Could I also ask you to update versions of `bartiq` & `qref` to `0.12.1` and `0.11.0` respectively?
Or remind us which magic scripts need to be used to do that 😅

qualtran/qref_interop/_bloq_to_qref.py
pqvr: Yes, I agree that this is a much better and more specific name. Updated in 353a73c0b7c84b81087a993f00dc79e5e135a6e4

mpharrigan:
Nice! Did you see that `CompositeBloq` has a `flatten` (and `flatten_once`) method that does something similar?

mpharrigan:
@pqvr can you sign the CLA

pqvr:
> @pqvr can you sign the CLA

Signed!

mpharrigan:
Thanks for your patience while I push the new dependencies!

 - `flatten` will indeed call decompose on subbloqs. See, e.g. https://github.com/quantumlib/Qualtran/blob/511b77645e4ebf97973397227e3dfefcde302444/qualtran/_infra/composite_bloq.py#L374
 - Yes, there is indeed a missing method that will flatten a `Bloq`. The trick is to "wrap" your bloq into a size-1 composite bloq; that is: `bloq.as_composite_bloq().flatten()` 

just fyi!

mpharrigan:
you can run the formatter with `check/format-incremental --apply`

pqvr:
> Thanks for your patience while I push the new dependencies!
> 
> * `flatten` will indeed call decompose on subbloqs. See, e.g. https://github.com/quantumlib/Qualtran/blob/511b77645e4ebf97973397227e3dfefcde302444/qualtran/_infra/composite_bloq.py#L374
> * Yes, there is indeed a missing method that will flatten a `Bloq`. The trick is to "wrap" your bloq into a size-1 composite bloq; that is: `bloq.as_composite_bloq().flatten()`
> 
> just fyi!

Thanks for the tip! You’re absolutely right that

```py
cb = my_bloq.as_composite_bloq().flatten()
```

will take any existing `CompositeBloq` and collapse all its nested subbloqs into one flat structure—because `flatten()` under the hood does call each child’s `.decompose_bloq()`.

However, there are two key reasons that doesn’t solve the symbolic‐width case:



> Thanks for your patience while I push the new dependencies!
> 
> * `flatten` will indeed call decompose on subbloqs. See, e.g. https://github.com/quantumlib/Qualtran/blob/511b77645e4ebf97973397227e3dfefcde302444/qualtran/_infra/composite_bloq.py#L374
> * Yes, there is indeed a missing method that will flatten a `Bloq`. The trick is to "wrap" your bloq into a size-1 composite bloq; that is: `bloq.as_composite_bloq().flatten()`
> 
> just fyi!

Thanks again! Yes, I did try using `as_composite_bloq().flatten()` early on, but it unfortunately leads to infinite recursion in some cases — especially when the `Bloq` being flattened internally uses `as_composite_bloq()` as part of its own decomposition (e.g. `GeneralizedQSP`).

Also, this patch addresses a slightly different concern — namely, **controlling which Bloqs get decomposed**. The goal is to support:
- Decomposing all possible Bloqs, or
- Decomposing a specific list of Bloq types (e.g. QROMs), while preserving others as atomic.

Since `flatten()` always recursively decomposes everything in sight (and only works on `CompositeBloq` nodes), it doesn't quite meet the need for selective, class-based control. That's why this patch includes an explicit traversal and decomposition gate logic.

Hope that makes sense, and sorry if I'm missing another way of approaching this!

mpharrigan:
> Thanks again! Yes, I did try using as_composite_bloq().flatten() early on, but it unfortunately leads to infinite recursion in some cases — especially when the Bloq being flattened internally uses as_composite_bloq() as part of its own decomposition (e.g. GeneralizedQSP).

Oh dear, this sounds like a bug! if you have a little script to reproduce, I'd love to see it; otherwise I'll open an issue and investigate later.

mpharrigan:
Just for your edification, there is indeed a `pred` ("predicate") argument to `flatten` that is called on each bloq instance and any time it returns `False`, the bloq instance will be kept in tact. Note that this will be called with `BloqInstance`, so if you want to test bloq classes, you need to use something like `isinstance(binst.bloq, QROM)`

qualtran/qref_interop/_bloq_to_qref.py
```diff
@@ -128,23 +129,92 @@ def _bloq_instance_name(instance: BloqInstance) -> str:
...
+
+    # BloqInstance → unwrap
+    if isinstance(obj, BloqInstance):
+        return _routine_with_decomposition(obj.bloq, preserve, name=_bloq_instance_name(obj))
+
+    # Plain Bloq → decide
+    bloq = obj
+    if preserve is None or type(bloq) in preserve:
+        try:
+            cb = bloq.decompose_bloq()
```
pqvr, line 210: @mpharrigan in reference to your comment https://github.com/quantumlib/Qualtran/pull/1643#issuecomment-2868113477, changing this line from `cb = bloq.decompose_bloq()` to `cb = bloq.decompose_bloq().flatten()` will cause `RecursionError: maximum recursion depth exceeded`.

pqvr:
> Oh dear, this sounds like a bug! if you have a little script to reproduce, I'd love to see it; otherwise I'll open an issue and investigate later.

I might be abusing the `flatten()` function or using it in a way that you don't condone, but I referenced the location in the diff where this was causing infinite recursion https://github.com/quantumlib/Qualtran/pull/1643/files#r2083115715

qualtran/qref_interop/_bloq_to_qref.py
```diff
@@ -128,23 +128,93 @@ def _bloq_instance_name(instance: BloqInstance) -> str:
...
+    # BloqInstance → unwrap
+    if isinstance(obj, BloqInstance):
+        return _routine_with_decomposition(obj.bloq, preserve, name=_bloq_instance_name(obj))
+
+    # Plain Bloq → decide
+    bloq = obj
+    if preserve is None or type(bloq) in preserve:
+        try:
+            #cb = bloq.decompose_bloq()
+            cb = bloq.as_composite_bloq().flatten()
```
mpharrigan, line 210: did you mean to commit this?

pqvr: Ah, shoot, nice catch. Actually, I was experimenting with the infinite recursion we were discussing. I must have left the broken experiment in. Let me change that back, and thank you for catching!

mpharrigan:
@mstechly can you give a verbal LGTM

mstechly:
LGTM!
(I've been monitoring the discussion silently, but had nothing to add :) )

mpharrigan (approved):

mpharrigan:
I think this introduced a flaky test in `test_importing_qualtran_object_gives_expected_routine_object` where the ordering of sub-elements causes a comparison to fail. I don't have a reliable way to reproduce, but it failed our nightly CI and failed once on my machine

pqvr:
> I think this introduced a flaky test in `test_importing_qualtran_object_gives_expected_routine_object` where the ordering of sub-elements causes a comparison to fail. I don't have a reliable way to reproduce, but it failed our nightly CI and failed once on my machine

Ah, that's quite annoying. Thanks for mentioning that, @mpharrigan. If there's no repro (as I'm not able to repro it locally) do you think it makes sense to remove the test? Thoughts?

pqvr:
> To reproduce, you seemingly need to just run it a bunch of times. I would guess that you're asserting a collection to be equal when its elements aren't guaranteed to be in the same order. This has been showing up in our CI sporadically, so I'd like to see a fix !

Same, and thanks for writing that issue up!

PR #1645, opened 2025-05-08
[2025-05] Bump dependencies

mpharrigan:
- regularly scheduled maintenence 
- propagate changes from https://github.com/quantumlib/Qualtran/pull/1638 to dev.env.txt
- mypy gets its own .env.txt and only runs on the actual source code. mypy checks for dev_tools/ moved to a unified dev-tools CI job, which has the full dev env  
- mypy/graphviz false alarms from update


mhucka:
It seems like the Python code changes ought to be separated from the dependency file changes. Is that possible in this case?

mpharrigan:
the code changes (which are just type annotation things -- mostly ignores) are only required with the new dependencies

mhucka (approved):
LGTM

PR #1652, opened 2025-05-28
Correct lint issues exposed with update to pylint-3

### LINKED ISSUE #1440: Correct lint issues exposed with update to pylint-3
pylint-3 has some default rules which currently fail.
These were temporarily disabled in [.pylintrc](https://github.com/quantumlib/Qualtran/blob/main/dev_tools/conf/.pylintrc) to pass the CI for #1439.

Consider enabling them back with associated code fixups (see TODO in .pylintrc).

micpap25:
Fixes #1440 

Re-enables the following pylint issues which were disabled when bumping to pylint-3.

- [ ] abstract-class-instantiated
- [X] deprecated-class
- [ ] possibly-used-before-assignment
- [ ] used-before-assignment

### DIFF
--- a/dev_tools/conf/.pylintrc
+++ b/dev_tools/conf/.pylintrc
@@ -8,7 +8,6 @@ reports=no
 py-version=3.9
 disable=
     abstract-class-instantiated,        # TODO: #1440 - enable and fix
-    deprecated-class,                   # TODO: #1440 - enable and fix
     possibly-used-before-assignment,    # TODO: #1440 - enable and fix
     used-before-assignment,             # TODO: #1440 - enable and fix
     C,
--- a/qualtran/_infra/composite_bloq.py
+++ b/qualtran/_infra/composite_bloq.py
@@ -13,6 +13,7 @@
 #  limitations under the License.
 
 """Classes for building and manipulating `CompositeBloq`."""
+from collections.abc import Hashable
 from functools import cached_property
 from typing import (
     Callable,
@@ -32,7 +33,6 @@ from typing import (
     TypeVar,
     Union,
 )
-from collections.abc import Hashable
 
 import attrs
 import networkx as nx
--- a/qualtran/_infra/composite_bloq.py
+++ b/qualtran/_infra/composite_bloq.py
@@ -19,7 +19,6 @@ from typing import (
     cast,
     Dict,
     FrozenSet,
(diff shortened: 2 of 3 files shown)


qualtran/bloqs/reflections/reflection_using_prepare_test.py
```diff
@@ -133,7 +133,7 @@ def test_reflection_using_prepare(num_ones, eps, global_phase):
     prepared_state = result.final_state_vector.reshape(2 ** len(selection), -1).sum(axis=1)
     if np.sign(global_phase) == 1:
         signs = '-' * num_ones + '+' * (9 - num_ones)
-    elif np.sign(global_phase) == -1:
+    else:  # np.sign(global_phase) == -1:
```
mpharrigan, line 136: can we keep this as an elif and add an else clause that raises a ValueError or AssertionError?

micpap25:
@mpharrigan Pushed the reversion. My thought process is that the current state isn't good code since we are using abstraction purely for `isinstance` checks which seems convoluted and could be handled by one variable.

mpharrigan:
The distinction shows up in calls to the constructor, with the unique class names helping to make the code more self-documenting.

Should we keep `abstract-class-instantiated` as a global skip? Am I correct in saying that they were all false-positives? And that now any time someone uses one of these classes they would have to add an additional "skip" comment?

micpap25:
I can definitely confirm that last part, that a skip would be necessary to use these classes. 

Parsing through the code I do think it's all false positives. I don't believe any abstract classes are getting instantiated. It would be nice to be able to keep the warning for real cases, but I'm not sure it's worth it with all the false positives.

micpap25:
Not sure why anything is failing out now since all that has changed between now and the last succeeding CI was comments after the reversion...?

mpharrigan:
Yes, I think we should keep `abstract-class-instantiated` as a global skip and add a note to the pylint config that it produces too many false positives with our usage of `attrs`.

If we instantiate an abstract class, it will get caught at runtime (e.g. during a unit test)

mpharrigan:
There's also a known flaky test at the moment #1657 , which is the `pytest` failure. I can re-run the CI in that case.

mpharrigan (approved):
Nice, thanks for getting these three checks fixed and enabled!

PR #1658, opened 2025-06-02
pretty msds v1

### LINKED ISSUE #853: Add a pretty formatter for rotation angles so we can have nicer musical score diagrams
When plotting musical score diagrams for `QvrZPow` with a symbolic rotation angle `sympy.Symbol('Y')`, I want the output to have nicely formatted rotation angles instead of very small floating point values. For example:

When running the code
```python
qvr_zpow = QvrZPow(Register('x', QFxp(12, 6, False)), gamma = sympy.Symbol('Y'))
show_bloq(qvr_zpow, 'dtype')
show_bloq(qvr_zpow.decompose_bloq(), 'musical_score')
```

I want the output to be like 

image

instead of  (current main)

image

or what you get if you update the pretty name of `ZPowGate` bloq to return `f'Z^{self.exponent}'`

image

To generate the original diagram, I implemented a very hacky `format` function locally as follows:

```python
def format(x: SymbolicFloat) -> str:
    if isinstance(x, float) or x.is_constant():
        if x.is_integer:
            return f'{int(x)}'
        for i in range(1, 20):
            if x * (2**i) == 1:
                return f'/2^{i}'
        return f'{x:.2g}'
    if isinstance(x, sympy.Symbol):
        return str(x)
    if isinstance(x, sympy.Expr):
        return '*'.join(format(x) for x in x.args[::-1])
    return str(x)
```

The goal of this issue would be to add a similar production quality format function that can be used to get nice string representations of exponents represented as a `SymbolicFloat`

cc https://github.com/quantumlib/Qualtran/issues/791

jbolns:
The changes in this PR enable pretty print of musical score diagrams. 

The changes in `qualtran\drawing\_show_funcs.py` aim to be enable pretty formatting in a way that is independent from internal logic (to allow for changes) but flexible (to easily adapt to any changes):
- MSD data is captured on the way to the diagram
- Inspected, re-shuffled, evaluated, and simplified using incremental RegEx and SymPy steps
- Re-inserted for final rendered.

If the internal logic changes and labels change (I estimate minimally to moderately), the approach can be adapted by modifying the foundational RegEx capture of symbols and/or the locals for simpify.

&

PR also include a small change to `qualtran\bloqs\rotations\quantum_variable_rotation.py`. Perhaps I missed something when I set things up since I am submitting this PR as part of unitaryHack and there is limited time to get familiar with the codebase. That said, the `offset` was causing errors.

Closes #853 

Maybe.

Ps. I tend to use more space and comments, but I tried my best to stick to the style in the target files. Happy to expand.

### DIFF
--- a/qualtran/bloqs/rotations/quantum_variable_rotation.py
+++ b/qualtran/bloqs/rotations/quantum_variable_rotation.py
@@ -191,6 +191,7 @@ class QvrZPow(QvrInterface):
         if self.cost_dtype.signed:
             out[0] = bb.add(ZPowGate(exponent=1, eps=eps), q=out[0])
         offset = 1 + self.cost_dtype.num_frac - self.num_frac_rotations
+        offset = offset if isinstance(offset, int) else 1
         for i in range(num_rotations):
             power_of_two = i - self.num_frac_rotations
             exp = (2**power_of_two) * self.gamma * 2
--- a/qualtran/drawing/_show_funcs.py
+++ b/qualtran/drawing/_show_funcs.py
@@ -15,7 +15,9 @@
 """Convenience functions for showing rich displays in Jupyter notebook."""
 
 import os
-from typing import Dict, Optional, overload, Sequence, TYPE_CHECKING, Union
+import re
+import sympy
+from typing import Dict, Optional, Text, overload, Sequence, TYPE_CHECKING, Union
 
 import IPython.display
 import ipywidgets
@@ -25,12 +27,132 @@ from qualtran import Bloq
 from .bloq_counts_graph import format_counts_sigma, GraphvizCallGraph
 from .flame_graph import get_flame_graph_svg_data
 from .graphviz import PrettyGraphDrawer, TypedGraphDrawer
-from .musical_score import draw_musical_score, get_musical_score_data
+from .musical_score import MusicalScoreData, TextBox, draw_musical_score, get_musical_score_data
 from .qpic_diagram import qpic_diagram_for_bloq
 
 if TYPE_CHECKING:
     import networkx as nx
-    import sympy
+
+
+def pretty_format_msd(msd: MusicalScoreData) -> MusicalScoreData:
+    """
+    Beautifies MSD to enable pretty diagrams.
+    
+    Args:
+        msd: The original MusicalScoreData that wants to be diagramised.
+
+    Returns:
+        pretty_msd: A beautified MSD
+
+    """
+
+    def symbolic_abs(n_or_s: Union[int, float, 'sympy.Symbol']) -> int:
+        """
+        Handles absolute value operations that may or may not include a symbol
+
+        Args:
+            n_or_s: The number or symbol to process.
+
+        Returns:
+            The absolute value of a numeric input or 1 (integer) if n_or_s is symbol.
+
+        """
+
+        if isinstance(n_or_s, sympy.Symbol):
+            return 1
+        return sympy.Abs(n_or_s)
+
+    sympify_locals = {
+        "expression": {    
+            "Y": sympy.Symbol("Y"),
+            "Abs": symbolic_abs,
+        },
+        "exponent": {
+            "Min": sympy.Min,
+            "ceiling": sympy.ceiling,
+            "log2": lambda x: sympy.log(x, 2),
+            "Abs": symbolic_abs,
+            "Y": sympy.Symbol("Y"),
+        }
+    }
+
+    new_soqs = []
+    for soq_data in msd.soqs:
+        if isinstance(soq_data.symb, (TextBox, Text)):
+            original_lbl = soq_data.symb.text
+            abs_pattern = r"Abs\((?P<symbol_value>[a-zA-Z])\)"
+            match = re.search(abs_pattern, original_lbl)
+
+            if match:
+                symbol_value = match.group("symbol_value")
+                removals = r"\*" + re.escape(symbol_value) + r"$"
+                modified_lbl = re.sub(removals, "", original_lbl)
+                split_text = re.split(r"\^", modified_lbl, 1)
+                gate_clean = split_text[0]
+
+                if len(split_text) > 1:
+                    lbl_raw = split_text[1].strip()
+                    lbl_clean = lbl_raw
+                    base, exponent_raw = lbl_raw.split("**")
+                    try:
+                        exponent_sympified = sympy.sympify(exponent_raw, locals=sympify_locals["exponent"], evaluate=True)
+                        exponent_clean = str(sympy.Integer(exponent_sympified.evalf(chop=True)))
+                        lbl_reconstructed = base + "**" + exponent_clean
+                        try:
+                            lbl_reconstructed_sympified = sympy.sympify(lbl_reconstructed, locals=sympify_locals["expression"], evaluate=True)
+                            lbl_to_clean = lbl_reconstructed_sympified
+                            if not lbl_reconstructed_sympified.is_number:
+                                symbols_to_substitute = lbl_reconstructed_sympified.free_symbols
+                                subs_map = {s: sympy.Integer(1) for s in symbols_to_substitute}
+                                lbl_to_clean = lbl_reconstructed_sympified.subs(subs_map)
+                            lbl_clean = str(sympy.simplify(lbl_to_clean))
+                            try:
+                                int_val = int(float(lbl_clean))
+                                if int_val == float(lbl_clean):
+                                    lbl_clean = str(int_val)
+                            except ValueError:
+                                pass
+                        except (sympy.SympifyError, NameError, TypeError, ValueError) as e:
+                            lbl_clean = lbl_reconstructed
+                    except (sympy.SympifyError, NameError, TypeError, ValueError) as e:
+                        lbl_clean = lbl_raw
+                else:
+                    lbl_clean = ""
+
+                new_lbl = (
+                    f"{gate_clean + '^' if gate_clean else ''}"
+                    f"{symbol_value + '*' if symbol_value else ''}"
+                    f"{lbl_clean if lbl_clean else ''}"
+                )
+
+                new_symb = None
+                if isinstance(soq_data.symb, TextBox):
+                    new_symb = TextBox(text=new_lbl)
+                elif isinstance(soq_data.symb, Text):
+                    new_symb = Text(text=new_lbl, fontsize=soq_data.symb.fontsize)
+
+                new_soq_data = soq_data.__class__(
+                    symb=new_symb,
+                    rpos=soq_data.rpos,
+                    ident=soq_data.ident
+                )
+                new_soqs.append(new_soq_data)
+
+            else:
+                new_soqs.append(soq_data)
+
+        else:
+            new_soqs.append(soq_data)
+
+    pretty_msd = MusicalScoreData(
+        max_x=msd.max_x,
+        max_y=msd.max_y,
+        soqs=new_soqs,
+        hlines=msd.hlines,
+        vlines=msd.vlines
+    )
+
+    return pretty_msd
 
 
 def show_bloq(bloq: 'Bloq', type: str = 'graph'):  # pylint: disable=redefined-builtin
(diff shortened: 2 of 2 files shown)


qualtran/bloqs/rotations/quantum_variable_rotation.py
```diff
@@ -191,6 +191,7 @@ def build_composite_bloq(self, bb: 'BloqBuilder', **soqs: 'SoquetT') -> Dict[str
         if self.cost_dtype.signed:
             out[0] = bb.add(ZPowGate(exponent=1, eps=eps), q=out[0])
         offset = 1 + self.cost_dtype.num_frac - self.num_frac_rotations
+        offset = offset if isinstance(offset, int) else 1
```
mpharrigan: what is the reason for this change? (add a comment?)

qualtran/drawing/_show_funcs.py
```diff
@@ -142,3 +145,81 @@ def show_bloq_via_qpic(bloq: 'Bloq', width: int = 1000, height: int = 400):
 
     IPython.display.display(Image(output_file_path, width=width, height=height))
     os.remove(output_file_path)
+
+
+def pretty_format_msd(msd: MusicalScoreData) -> MusicalScoreData:
```
mpharrigan: this function should live in `musical_score.py`

qualtran/drawing/_show_funcs.py
```diff
@@ -142,3 +145,81 @@ def show_bloq_via_qpic(bloq: 'Bloq', width: int = 1000, height: int = 400):
...
+    """
+    Beautifies MSD to enable pretty diagrams
```
mpharrigan: explain what transformations are done

qualtran/drawing/_show_funcs.py
```diff
@@ -142,3 +145,81 @@ def show_bloq_via_qpic(bloq: 'Bloq', width: int = 1000, height: int = 400):
...
+                gate, base, exponent = sum([p.split("**", 1) for p in lbl_no_symbols.split("^")], [])
+
+                if len(base.split("*", 1)) > 1:
+                    mult_str, base = base.rsplit("*", 1)
+                    mult = sympy.sympify(mult_str, evaluate=True)
+
+                exponent = sympy.sympify(exponent, locals=simpify_locals, evaluate=True)
+                expression = str(base) + "**" + str(exponent)
+                expression = sympy.sympify(expression, locals=simpify_locals, evaluate=True)
+                new_lbl = str(gate) + "^" + symbol + "*" + str(expression * mult)
```
mpharrigan: I'm not sure what this code aims to do, but it seems rather brittle. Why are we replacing parts of the label with the character `"1"`?

qualtran/bloqs/rotations/quantum_variable_rotation.py
jbolns: I noted this in the issue as a concern/question.

The instructions in the issue do not lead to completion. There is an indexing error in one of the iterations in the loop below this line I added.

```
qvr_zpow = QvrZPow(Register('x', QFxp(12, 6, False)), gamma = sympy.Symbol('Y'))
show_bloq(qvr_zpow, 'dtype')
show_bloq(qvr_zpow.decompose_bloq(), 'musical_score')
```

Perhaps I'm missing a step or something. Else, it is a pre-existing bug that might be related to the small angle rotations changes in 38db1aff2c4b37944acbb817350b5effffade114 

In the next commit, this line will move to inside the loop to minimise its involvement to only when strictly necessary. 

```
for i in range(num_rotations):
    power_of_two = i - self.num_frac_rotations
    exp = (2**power_of_two) * self.gamma * 2
    offset = offset if isinstance(offset, int) else 1  # offset -> 1 if not int to avoid indexing errors  #HERE
    out[-(i + offset)] = bb.add(ZPowGate(exponent=exp, eps=eps), q=out[-(i + offset)])
return {self.cost_reg.name: bb.join(out, self.cost_dtype)}
```

However, this is only to ensure things proceed as needed for testing of the feature in this issue. 

If the instructions to produce the foundational diagram were incomplete, please let me know. 

If it is a bug, it should be checked separately.

qualtran/drawing/_show_funcs.py
jbolns: Moving it. I can also use a smaller function for each soq rather than the entire msd. 
Will show in next commit. 
Will also add a test in the corresponding file (now possible since its a soq-level function).

qualtran/drawing/_show_funcs.py
jbolns: Will do. Shows in next commit.

qualtran/drawing/_show_funcs.py
jbolns: Right, yes! You are completely right.  

So, any errors fallback to the original label, which ensures completion, but I can make it more robust in the next commit. 

More importantly, the mathematical concern is undeniably valid.

It's clear I took the diagram examples in the issue too literally, perhaps assuming there was some quantum magic happening somewhere that escaped me. 

Thinking about it as a standard symbolic equation, there seems to be a need to define a preference:
- semi-pretty labels: some simplification is possible if Abs(Y) is not evaluated with some sort of substitution (which would affect other Y values), but the pretty version won't be entirely pretty. (better than nothing I guess)
- arbitrary replacement: pi, 2*pi, some random value? (no likey, not symbolic anymore)
- user-input placeholder: user chooses what value to use for evaluation (same)
- a panel of diagrams: several rotation angles rendered as a panel of diagrams rather than a single panel diagram (quite a deviation from original concept so I'm hoping there is another approach I'm not considering but, somewhat applicable).

I'll work a little on the next commit but it would be cool if you could tell me your views on the matter and potential alternatives not considered, to avoid the need for many additional commits.

jbolns: Actually, after running a few tests if seems the combination of factors is such that only very very very small angles would lead to a difference, in which case, might as well use pi to evaluate the Abs value in the expression. Will explain thoroughly in comments.

jbolns:
@mpharrigan 
I might be coming across a bit brisk. Apologies. I'd never done hackathons and apparently it gets to my head. :smiley:
Nonetheless, I value the feedback thus far and the importance of good code. High quality outcome > fast outcome.

jbolns:
Done, @ACE07-Sev.
Thanks for the heads up. Totally missed that one.

mhucka:
For reference, I'm copying here the image from [a comment by the original author](https://github.com/quantumlib/Qualtran/issues/853#issuecomment-2927824630) in issue #853 that shows a screenshot of the output produced by the code in this PR:

<div align="center">

</div>

mhucka:
@jbolns With respect to the output produced by your code currently (c.f. previous comment above), let's step back a bit and review some conventions of these "musical score diagrams":

- the horizontal axis = time, with time progressing from the left (earlier operations) to the right (later operations)
- the vertical axis = qubits, registers of qubits, or (sometimes) classical bits, with each horizontal line representing a separate qubit/register/classical bit
- boxes on the horizontal lines represent gates, bloqs, operations, or other such constructs operating on the qubit/register/etc

Now, with this in mind, comparing the [original description in issue #853](https://github.com/quantumlib/Qualtran/issues/853#issue-2223690812) where @tanujkhattar showed a screenshot illustrating the format of the diagrams, a discrepancy becomes apparent. The original always has the bloq/gate/operation name in the box, and parameters above the box:

<div align="center">
image
</div><br>

By contrast, in your image, the entire text is inside the box:

<div align="center">
image
</div><br>

So, this is the first thing I suggest be addressed in the PR: move the parameters or arguments (in this case, rotation angles) out and above the boxes.

jbolns:
I'm very happy for this feedback and would like to investigate a solution for it. Having said that, I do believe it would be best to merge this PR first and then address this formatting matter as part of a new broader issue revisiting the entire MSD workflow. 

The reasoning is as follows. 

Currently, **without anything in this PR (except the one-line fix in `quantum_variable_rotation.py`)**, Qualtran is printing the following diagrams (with full expression and all details inside the box):

default-print

If you didn't change the placement of labels intentionally, then it means the change went unnoticed. That's already the second such situation, the first being the bug that made it necessary for me to include the one-line fix in `quantum_variable_rotation.py`. 

Things like this tend to go unnoticed when they are collateral damage from changes to other (more important) features. As such, I am in a professional obligation to consider that if I change this back on a rush, to meet the deadline of the hackathon, I might undo changes introduced for reasons more important than MSDs.

To do this appropriately, thus, I would need to take a deep dive into past commits to learn what happened and how can it be fixed without additional collateral damage.

This takes time and goes far beyond adding a simple pretty formatter.

Ergo, I suggest merging the pretty formatter in this PR and then open a broader issue for "revising the MSD workflow" with a description along the lines of "PR #1658 introduced a pretty formatter for MSDs and found a number of inconsistencies with the MSD workflow that appear to result from pre-existing changes to the codebase. It would be ideal to review the MSD workflow to ensure it is working efficiently and gives users the option to print full or pretty results using the changes introduce in 1658". 

This is really only a suggestion made from my very limited point of view. It is, of course, your repository, so I will respect your choice. 

Either way, I can only work more on this starting end of week.

jbolns:
I have added a new commit. As seen in the image below, it recovers the content (though not the formatting) of labels as illustrated on the original issue.

If both the content and formatting can be recovered, prettification should be waaaaaaay easier. To begin, since the expression is already given fairly simplified, I do not need to worry about potential variation in mathematical patterns. The whole sympify situationship might not even be necessary.  

simplified_by_default

**Why? / What? / Where? / When?**
Using a combination of `git bisect` and `git rebase` (**on local branches** that I haven't pushed even to my fork to avoid accidents, as these are quite serious operations), I found out the following. 

*A.*
Commit 38db1aff2c4b37944acbb817350b5effffade114 broke MSDs in two ways. 
- `qualtran/bloqs/rotations/quantum_variable_rotation.py`: Changing line [169](https://github.com/quantumlib/Qualtran/commit/38db1aff2c4b37944acbb817350b5effffade114#diff-ad6833d36c51e14d082981a8f713ed939011a1e643b855fe1139399581dfc0e5L169) for line [200](https://github.com/quantumlib/Qualtran/commit/38db1aff2c4b37944acbb817350b5effffade114#diff-ad6833d36c51e14d082981a8f713ed939011a1e643b855fe1139399581dfc0e5R200) causes expression not to simplify automatically, reason why labels now render the full mathematical expression.
  - I reverted back to the original line. 
- `qualtran/bloqs/rotations/quantum_variable_rotation.py`: The original version used an integer for indexing, in lines [170-173](https://github.com/quantumlib/Qualtran/commit/38db1aff2c4b37944acbb817350b5effffade114#diff-ad6833d36c51e14d082981a8f713ed939011a1e643b855fe1139399581dfc0e5L170). The changed to an offset, in line [202](https://github.com/quantumlib/Qualtran/commit/38db1aff2c4b37944acbb817350b5effffade114#diff-ad6833d36c51e14d082981a8f713ed939011a1e643b855fe1139399581dfc0e5R202), causes indexing errors in the loop.
  - I had already addressed this by exchanging the offset for an integer when the offset itself is not an integer.

The fix in this last commit keeps everything in the newest version except the misbehaving lines.

To get a sense of whether this combined version could be problematic, before changing the lines in this PR, I created a test local branch to rebase the code all the way from 38db1aff2c4b37944acbb817350b5effffade114.

There was only one conflict that required solving:
- `qualtran/bloqs/rotations/quantum_variable_rotation.py`: In line [198](https://github.com/quantumlib/Qualtran/commit/511b77645e4ebf97973397227e3dfefcde302444#diff-ad6833d36c51e14d082981a8f713ed939011a1e643b855fe1139399581dfc0e5L198) of commit 38db1aff2c4b37944acbb817350b5effffade114, there is a change from `return {self.cost_reg.name: bb.join(out, self.cost_reg.dtype)}` to `return {self.cost_reg.name: bb.join(out, self.cost_dtype)}`.
- This seemed simply a git conflict, because the latest version seems to work fine. So, I have kept the latest version.
- It is yet to be seen if this causes a mess with tests in other parts of Qualtran (I purposely skipped check/pytest so you can see yourselves if any and which tests fail).

*B.*
Commit db953b9c48dc18146cf91e3164dfcce1207e6b61 broke the formatting of the MSDs.
- Before this commit, MSDs rendered with labels on top of the boxes with the gates, bloqs, etc constructs.
- After this commit, MSDs render only with labels containing mathematical expression inside the box. 

It is a significant commit, so I will explore it at a different point in time. 

@mhucka

jbolns:
I checked on B (last comment) and found what I think is a somewhat smart solution.

The key was realising that the current behaviour in main is not the desired/ideal point of departure.

For the same reason, the final solution is different to anything in this PR.

For the same reason, I made a separate PR, #1669 .

For the same reason, this PR can be closed.

@mhucka

Issue #1657, opened 2025-06-02
Flaky test in `test_importing_qualtran_object_gives_expected_routine_object` 


mpharrigan:
#1643 introduced a flaky test in `test_importing_qualtran_object_gives_expected_routine_object` where the ordering of sub-elements causes a comparison to fail. I don't have a reliable way to reproduce, but it crops up in the nightly and PR continuous integeration builds

_Originally posted by @mpharrigan in https://github.com/quantumlib/Qualtran/issues/1643#issuecomment-2881043907_

mpharrigan:
cc @pqvr

mstechly:
@mpharrigan please take a look at: https://github.com/quantumlib/Qualtran/pull/1664

PR #1664, opened 2025-06-17
fix: fixes flakiness in bloq->qref conversion

mstechly:
This should fix: https://github.com/quantumlib/Qualtran/issues/1657 .

From what I could say from running it locally, I stopped getting the intermittent failures in Bloq -> QREF conversion.
But you know how it is with flaky tests – in case it re-occurs, please ping me!


mpharrigan (approved):
thank you!

Issue #1646, opened 2025-05-09
Free sympy version


mpharrigan:
Following https://github.com/quantumlib/Qualtran/pull/1167, we restrict sympy for mypy type checking issues. Need to investigate

mpharrigan:
The issues seem to stem from mypy treating later versions of `sympy.Expr` as `typing.Any` -- this seems to be particularly accute when combined with  the `TypeIs` annotation in `is_symbolic`. I found this by using `typing_extensions.reveal_type`

PR #1674, opened 2025-07-10
Upgrade sympy to 1.4

dstrain115:
- Upgrades sympy and fixes typechecking issues that result.

Fixes: #1646


dev_tools/conf/mypy.ini
```diff
@@ -25,6 +25,8 @@ ignore_missing_imports = true
 follow_imports = silent
 ignore_missing_imports = true
 
+[mypy-sympy.*]
+follow_untyped_imports = True
```
mpharrigan, line 29: did you notice any effect on the typechecking runtime?

dev_tools/requirements/envs/dev.env.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.10
+# This file is autogenerated by pip-compile with Python 3.12
```
mpharrigan: since the pinned requirements are very delicate w.r.t. exactly which python environment you run them from, we use the `..from_docker.sh` script to bump the dependencies from the included docker image. As an example, you can see that your name is now forever immortalized in the requirements comments. 

I'll run the script and push to this pr to save you some time

mpharrigan:
wooooooooooo

mpharrigan (approved):

dev_tools/conf/mypy.ini
dstrain115: I saw less mypy errors (it was like 12 versus 24 without this).  I didn't really investigate it though.

dev_tools/requirements/envs/dev.env.txt
dstrain115: Thanks!  I just ran the update deps script in this directory.  Thanks for running it correctly for me!

PR #1677, opened 2025-07-16
Update type annotations throughout the code base

### LINKED ISSUE #1653: replace deprecated `typing` aliases with those from `collections.abc`
A follow-up on #1652.
The deprecated classes in question were aliases from `typing` which are recommended not to be used according to https://docs.python.org/3/library/typing.html.
The default python types, or typings from `collections.abc`, should be used instead.

micpap25:
Fixes #1653 and in addition, updates all type annotations to modern syntax and removes no-longer-needed `# type: ignore` comments where possible. Takes advantage of constructs available in Python 3.12 (which is the current minimum for Qualtran).

This work was started by @micpap25, with subsequent additional work by @mhucka and assistance from Gemini CLI.

General summary:

- Require Mypy versions ≥ 2.0
- Add missing type annotations
- Use unquoted type names
- Use `type` to define types instead of `TypeAlias`
- Use built-in types like `list` and `dict` instead of `typing`'s `List` and `Dict`
- Replace `Union` and `Optional` with shorthand `|` notation
- Remove as many `# type: ignore` comments as possible
- Remove `ignore_errors` from Mypy configuration
- Add `warn_unused_ignores = true` and `warn_redundant_casts = true` to Mypy config


mhucka (changes requested):
Thanks for doing this!

This is a monster PR and it's hard to do individual line-oriented comments, so I'll have to summarize change requests and ask you to find the individual instances. Overall it looks good! There are just a few repeated patterns of changes that can be improved:

* Some of the new `Union` types can be rewritten to simpler forms. There are two cases I see:

  * A type declaration of the form `Union[X, None]` can be rewritten as `X | None`, but it is safest to do this only if the `X` is _not_ a quoted type name. For example, the case of
    ```python
    call_graph_example: Union[BloqExample, None] = field()
    ```
    can be rewritten as
    ```python
    call_graph_example: BloqExample | None = field()
    ```

  * Unions of the form `Union[X, None]` are also equivalent to `Optional[X]`, and this is safer to use this variant if the `X` is a quoted type name. So for example,
    ```python
    Union['cirq.Operation', None]
    ```
    can be converted to
    ```python
    Optional['cirq.Operation']
    ```

* Some combinations of Optional and Union can be further simplified. Here is an example:
    ```python
    target_bitsizes: Optional[Union[SymbolicInt, tuple[SymbolicInt, ...]]] = None
    ```
    can be
    ```python
    target_bitsizes: SymbolicInt | tuple[SymbolicInt, ...] | None = None
    ```

* There are some remaining uses of `typing.Type`. I'm looking at `qualtran/bloqs/data_loading/qroam_clean.py` but are probably others. Basically, things like `cls: Type['QROAMCleanAdjoint']` can be `cls: type['QROAMCleanAdjoint']` and the import of `Type` from `typing` can be dropped.

That's all I have for now.

micpap25:
Hi @mhucka Thank you for the detailed feedback, just to confirm that this is in addition to the changes to `typing.Set` and `annotations` that were discussed in the smaller PR?

mhucka:
> Hi @mhucka Thank you for the detailed feedback, just to confirm that this is in addition to the changes to `typing.Set` and `annotations` that were discussed in the smaller PR?

Hi @micpap25 – thanks for coming back to this. Yes, I think that, basically, if there are cases of upper case Set, it should be possible to replace them with the lower-case variant. The tricky cases are what we discussed before about quoted type names.

micpap25:
@mhucka What about using `collections.abc.Set` in place of `typing.Set`? Does that work in the trickier cases?

micpap25:
@mhucka It seems there are some issues with the testing, in particular Pylint seems to have issues with the changes to `Union` and `Optional` you described.

Once we resolve this, I'm happy to continue the work on `Union` and `Optional` though I think it might constitute a new PR.

micpap25:
@mhucka Bumping to confirm you've seen the above comments

mhucka:
@micpap25 I've started working on resolving the type problems. The type annotations need more work, but at this point I think it'll be easier if I tackle remaining type work, and push changes to your branch. The work is in progress.

mhucka:
@micpap25 I have a set of changes to the type annnotations. At your earliest convenience, can you update the branch and resolve the merge conflicts? Then I can apply my changes on top of the result.

micpap25:
@mhucka Should be good, I changed some instances of `typing.Type` to `type` which referred to types with quotes around them, so hopefully that doesn't break anything like `set` and works nicely like `dict`.

mhucka:
@micpap25 I finally pushed the changes to your branch. There were a lot, but this should hopefully be complete.

mhucka (approved):

pyproject.toml
```diff
@@ -113,7 +113,7 @@ test = [
 ]
 
 typing = [
-    "mypy~=1.14",
+    "mypy~=2.0",
```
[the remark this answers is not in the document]
mhucka: Wow, seriously GCA?

dev_tools/bloq-method-overrides-report.py
```diff
@@ -43,10 +38,13 @@ def _call_graph(bc: type[Bloq]):
...
         )
-    if annot['ssa'] != 'SympySymbolAllocator':
+    if 'SympySymbolAllocator' not in str(annot.get('ssa', '')):
         print(f"{bc}.build_call_graph `ssa: 'SympySymbolAllocator'`")
-    if annot['return'] != Set['BloqCountT']:  # type: ignore[misc]
-        print(f"{bc}.build_call_graph -> 'BloqCountT'")
+    ret_str = str(annot.get('return', ''))
+    if not any(
+        sub in ret_str for sub in ('BloqCountDictT', 'set[BloqCountT]', 'Mapping[', 'BloqCountT')
+    ):
```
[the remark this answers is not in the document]
mhucka: This is actually a good catch.

dev_tools/qualtran_dev_tools/bloq_report_card.py
```diff
@@ -133,5 +136,5 @@ def summarize_results(report_card: pd.DataFrame) -> pd.DataFrame:
         .fillna(0)
         .astype(int)
     )
-    summary.columns = [v.name.lower() for v in summary.columns]
+    summary.columns = [getattr(v, 'name', str(v)).lower() for v in summary.columns]
```
mpharrigan, line 139: ??

dev_tools/bloq-method-overrides-report.py
```diff
@@ -43,19 +38,20 @@ def _call_graph(bc: type[Bloq]):
...
             f'and a return type annotation'
         )
-    if annot['ssa'] != 'SympySymbolAllocator':
+    if 'SympySymbolAllocator' not in str(annot.get('ssa', '')):
         print(f"{bc}.build_call_graph `ssa: 'SympySymbolAllocator'`")
-    if annot['return'] != Set['BloqCountT']:  # type: ignore[misc]
-        print(f"{bc}.build_call_graph -> 'BloqCountT'")
+    ret_str = str(annot.get('return', ''))
+    if not any(sub in ret_str for sub in ('BloqCountDictT', 'set[BloqCountT]', 'Mapping[')):
+        print(f"{bc}.build_call_graph -> {ret_str!r}")
```
mpharrigan, line 45: this goes against the spirit of using the script to check that the annotations are (1) there and (2) the idiomatic ones. I think success for any `Mapping[` is particularly loose

qualtran/_infra/controlled.py
```diff
@@ -563,14 +550,12 @@ def as_cirq_op(
...
-    def _circuit_diagram_info_(
-        self, args: 'cirq.CircuitDiagramInfoArgs'
-    ) -> 'cirq.CircuitDiagramInfo':
+    def _circuit_diagram_info_(self, args: cirq.CircuitDiagramInfoArgs) -> cirq.CircuitDiagramInfo:
         import cirq
 
         from qualtran.cirq_interop._bloq_to_cirq import _wire_symbol_to_cirq_diagram_info
 
-        if isinstance(self.subbloq, cirq.Gate):
+        if isinstance(self.subbloq, GateWithRegisters):
```
mpharrigan, line 558: ?

qualtran/_infra/quantum_graph.py
```diff
@@ -114,14 +116,24 @@ def _check_idx(self, attribute, value):
...
-    def shape(self) -> Tuple[int, ...]:
+    def shape(self) -> tuple[int, ...]:
         return ()
 
-    def item(self, *args) -> '_Soquet':
+    def __getitem__(self, item) -> _Soquet:
+        raise TypeError("Tried to index into a single soquet.")
+
+    @property
+    def bb(self) -> Any:
```
mpharrigan, line 130: where does this come up?

qualtran/_infra/registers.py
```diff
@@ -449,10 +452,12 @@ def __iter__(self) -> Iterator[Register]:
...
         return len(self._registers)
 
-    def __hash__(self):
+    def __hash__(self) -> int:
         return hash(self._registers)
 
-    def __eq__(self, other) -> bool:
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, Signature):
+            return False
```
mpharrigan, line 460: this is probably right, but it's hard to review this PR which mostly contains respellings of type annotations with sporadic logic changes such as this

mpharrigan:
I started skimming through this PR, but it's very difficult to review. It's 99% "respellings" with no runtime changes whatsoever; but sprinkled with some logic changes that I'd actually like to look at. In particular, adding default values and fallbacks is something I've found to be particularly pernicious in the age of agentic coding when the bots sometimes use it to swallow real errors

PR #1683, opened 2025-08-09
Add bloq for Measurement in the X basis

dstrain115:
- MeasX measures a qubit in the X basis.
- If simulated classically, this returns a classical distribution of 0 and 1 with a possible
phase if 1 is measured.

Original work done by mpharrigan in #1678

### DIFF
--- a/qualtran/bloqs/basic_gates/x_basis_test.py
+++ b/qualtran/bloqs/basic_gates/x_basis_test.py
@@ -141,10 +141,8 @@ def test_meas_x_classical_sim() -> None:
     with pytest.raises(ValueError, match='Invalid classical value'):
         _ = bloq.on_classical_vals(q=2)
 
-    results = [
-        do_phased_classical_simulation(bloq, {'q': 0}, rng=np.random.default_rng())
-        for _ in range(100)
-    ]
+    rng = np.random.default_rng(seed=12345)
+    results = [do_phased_classical_simulation(bloq, {'q': 0}, rng=rng) for _ in range(100)]
 
     # Assert measurements are random
     assert all(c[0]['c'] in {0, 1} for c in results)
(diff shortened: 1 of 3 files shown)


qualtran/bloqs/basic_gates/x_basis_test.py
```diff
@@ -119,3 +130,78 @@ def _keep_and(b):
     bloq = XGate().controlled(CtrlSpec(qdtypes=QUInt(n), cvs=1))
     _, sigma = bloq.call_graph(keep=_keep_and)
     assert sigma == {And(): n - 1, CNOT(): 1, And().adjoint(): n - 1, XGate(): 4 * (n - 1)}
+
+
+def test_meas_x_classical_sim() -> None:
+    bloq = MeasX()
+
+    with pytest.raises(ValueError, match='MeasX imparts a phase'):
+        _ = bloq.call_classically(q=0)
```
mpharrigan, line 139: 👍

qualtran/bloqs/basic_gates/x_basis_test.py
```diff
@@ -119,3 +130,78 @@ def _keep_and(b):
...
+
+    with pytest.raises(ValueError, match='Invalid classical value'):
+        _ = bloq.on_classical_vals(q=2)
+
+    results = [
+        do_phased_classical_simulation(bloq, {'q': 0}, rng=np.random.default_rng())
```
mpharrigan: what do you think about seeding our random generator in tests? usually we do; and theoretically line 151 or 152 could fail

qualtran/bloqs/basic_gates/x_basis_test.py
```diff
@@ -119,3 +130,78 @@ def _keep_and(b):
...
+    np.testing.assert_allclose(rho, should_be, atol=1e-8)
+
+    # Plus -> mixture
+    bb = BloqBuilder()
+    q = bb.add(PlusState())
+    c = bb.add(MeasX(), q=q)
+    cbloq = bb.finalize(c=c)
+    rho = cbloq.tensor_contract(superoperator=True)
+    should_be = np.asarray([[1, 0], [0, 0]])
+    np.testing.assert_allclose(rho, should_be, atol=1e-8)
```
mpharrigan: I think these tests are correct, but the comments are not

Measuring the zero ket in the x basis gives fully-mixed state (the identity matrix); as does measuring the one ket.
Measuring the plus state in the x basis gives you a deterministic "0" result |0><0| = [[1,0],[0,0]]

qualtran/bloqs/basic_gates/x_basis_test.py
dstrain115: Added seed.

qualtran/bloqs/basic_gates/x_basis_test.py
dstrain115: Ha, yeah, copied from Z-basis test and fixed the operators but not the comments.  Fixed now.

PR #1689, opened 2025-08-12
Add issue templates

mhucka:
This adds templates based on other Quantumlib projects. This follows the practices of other projects in providing a more guided workflow for people to report bugs, ask questions, or take note of tasks to be done.


mpharrigan:
I find these to not add much value over freeform input. We can revisit if there are common missing items from bug reports

PR #1695, opened 2025-08-24
Add bloq for the Guided Hamiltonian Problem

anurudhp:
Implement the algorithm from Sec 4.2 of https://arxiv.org/abs/2406.19378v1


anurudhp:
@tanujkhattar ptal

PR #1698, opened 2025-08-26
[open-systems] Testing Measurement-Based Uncomputation notebook

mpharrigan:
Quantum information cannot be destroyed, but during a computation we may produce intermediate values that we wish to discard. We can "uncompute" these values by running the computation in reverse. The ordinary uncomputation strategy requires paying the cost of the computation twice, but [*Halving the cost of quantum addition.* Gidney 2017](https://arxiv.org/abs/1709.06648) shows how measurement in the X basis can effectively discard a bit without expensive uncomputation. The consequence is that the remaining states of the system will pick up phases depending on the random measurement result. [*Verifying Measurement Based Uncomputation.* Gidney 2019](https://algassert.com/post/1903) provides more detail about these phases. It also describes a proceedure for using a phased-classical simulator to "fuzz test" measurement-based uncomputation circuits.

Here, we show how Qualtran can be used to verify measurement based uncomputation circuits following Gidney's proposal.


qualtran/simulation/MBUC.ipynb
```diff
@@ -0,0 +1,375 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "44f40baf-3f87-40c5-9bf1-409ac3f86e68",
+   "metadata": {},
+   "source": [
+    "# Verifying Measurement-Based Uncomputation\n",
+    "\n",
+    "Quantum information cannot be destroyed, but during a computation we may produce intermediate values that we wish to discard. We can \"uncompute\" these values by running the computation in reverse. The ordinary uncomputation strategy requires paying the cost of the computation twice, but [*Halving the cost of quantum addition.* Gidney 2017](https://arxiv.org/abs/1709.06648) shows how measurement in the X basis can effectively discard a bit without expensive uncomputation. The consequence is that the remaining states of the system will pick up phases depending on the random measurement result. [*Verifying Measurement Based Uncomputation.* Gidney 2019](https://algassert.com/post/1903) provides more detail about these phases. It also describes a proceedure for using a phased-classical simulator to \"fuzz test\" measurement-based uncomputation circuits.\n",
```
dstrain115: nit: I think there should be a comma after "during a computation"

qualtran/simulation/MBUC.ipynb
```diff
@@ -0,0 +1,375 @@
...
+   "id": "3dddb0a5-a2b6-44cb-beb9-79b9600aae5e",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import itertools\n",
+    "from qualtran.bloqs.mcmt import And\n",
+    "\n",
+    "and_dag = And(uncompute=True)\n",
+    "for q1, q2 in itertools.product(range(2), repeat=2):\n",
+    "    trg = int(q1==1 and q2 == 1)\n",
```
dstrain115: what is 'trg'?  I would probably use a better name.  
It took me a while to realize this means "target".  I would use that name here and below.

Avoid abbreviations:  https://google.github.io/styleguide/pyguide.html#316-naming

qualtran/simulation/MBUC.ipynb
```diff
@@ -0,0 +1,375 @@
...
+    "    print(f'{q1=}, {q2=}, {trg=}', end='  ')\n",
+    "    (q1o, q2o), = and_dag.call_classically(ctrl=[q1,q2], target=trg)\n",
+    "    assert q1o == q1\n",
```
dstrain115: Avoid abbreviations: https://google.github.io/styleguide/pyguide.html#316-naming

I would use q1_output.

dstrain115 (approved):
LGTM modulo a few nits.

qualtran/simulation/MBUC.ipynb
mpharrigan: done

qualtran/simulation/MBUC.ipynb
mpharrigan: changed here and everywhere

qualtran/simulation/MBUC.ipynb
mpharrigan: done

Issue #1722, opened 2025-09-25
Support numpy 2


mpharrigan:
1. we should note in our runtime requirements that we currently don't support numpy 2
2. we should make the minor changes required to get tests passing for numpy 2

there are new type-checking failures with numpy 2 as well

Issue #1708, opened 2025-09-16
Unpin required cirq version


mpharrigan:
Ideally, we'd support any Cirq version >=1.4 <2.0 (i.e. ~=1.4). The breaking change in https://github.com/quantumlib/Cirq/pull/6960 complicates this

mpharrigan:
we could use `cirq.OP_TREE` which exists in all versions

mpharrigan:
mypy is confusing me with cirq 1.5.0. An example error:

```
qualtran/bloqs/swap_network/cswap_approx.py:75: error: Missing return statement  [return]
qualtran/bloqs/swap_network/cswap_approx.py:82: error: Argument 1 to "list" has incompatible type "Operation | Iterable[OP_TREE]"; expected "Iterable[Operation | Iterable[OP_TREE]]"  [arg-type]
qualtran/bloqs/swap_network/cswap_approx.py:83: error: Argument 1 to "list" has incompatible type "Operation | Iterable[OP_TREE]"; expected "Iterable[Operation | Iterable[OP_TREE]]"  [arg-type]
```

this worked with `OpTree` but fails with `OP_TREE` even though `OP_TREE` is supposed to include `Iterable[OP_TREE]` as part of the union

pavoljuhas:
The mypy should be fixed by #1724.
With respect to the failure in test_add_decomposition I see it as well, for example with
```
pytest --pyargs "qualtran.bloqs.arithmetic.addition_test::test_add_decomposition[0-0-3]"
```
I have bisected that failure to https://github.com/quantumlib/Cirq/pull/7537, specifically the addition of [if-has_unitary-return-unitary block](https://github.com/quantumlib/Cirq/pull/7537/files#diff-c19df60bbf6e2c946bf5dd45816727fcbe0f73b696632f8afd474994e3394772L189) to `cirq.kraus`.  The problem is that the `circuit0` created in the test reports True for `cirq.has_unitary(circuit0)`, but throws TypeError for `cirq.unitary(circuit0)` after line 79 here -

https://github.com/quantumlib/Qualtran/blob/44f95f0927157e1abee38ec2d0f4ba455938882a/qualtran/bloqs/arithmetic/addition_test.py#L70-L79

The exception is the same with Cirq-1.4.0.  Here is a patch to trigger it before any calls to simulator:
```patch
diff --git a/qualtran/bloqs/arithmetic/addition_test.py b/qualtran/bloqs/arithmetic/addition_test.py
[...]

PR #1723, opened 2025-09-25
Support Cirq 1.5

mpharrigan:
towards #1708 -- things are still broken with cirq 1.6

towards #1722  -- I pin numpy and do one easy test fix for this


mhucka (approved):
LGTM. I pulled the branch to a local linux machine for testing, and it passes pytest and other tests.

PR #1724, opened 2025-09-26
Replace cirq.ops.op_tree.OpTree with cirq.OP_TREE

pavoljuhas:
- Replace `cirq.ops.op_tree.OpTree` with `cirq.OP_TREE`
- Fix flagged coefficient type for GlobalPhase

Towards #1708

### DIFF
--- a/qualtran/bloqs/mcmt/and_bloq.py
+++ b/qualtran/bloqs/mcmt/and_bloq.py
@@ -312,7 +312,7 @@ class MultiAnd(Bloq):
         control_values: Tuple[SymbolicInt, ...],
         ancillas: NDArray[cirq.Qid],
         target: cirq.Qid,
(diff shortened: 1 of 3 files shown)


mpharrigan:
I'm going to do a minor edit of your original post so it doesn't close the linked issue. github isn't smart enough to know the difference between "partially resolves" and "resolves"

pavoljuhas:
> thanks! I guess I'm confused why we need `Iterator[...]` explicitly since my reading of the definition of `OP_TREE` permits it and it wasn't needed for `OpTree`

Looking at it again, OpTree was a protocol with `__iter__` method so it was guaranteed to be iterable as is required for output from `yield`.  `OP_TREE` is a union of Operation and Iterable.  It may be non-iterable when Operation, hence not good for yielding as is.

PR #1728, opened 2025-09-29
Update GF2Inverse to use the decomposition from "Concrete quantum cryptanalysisof binary elliptic curves"

NoureldinYosri:
The construction in https://eprint.iacr.org/2020/1296 uses fewer Toffolis and ancillas

I also needed to modify other bloqs to accept a `qgf`


NoureldinYosri:
> Instead of adding a `qgf` field to classes that already have a `bitsize`, please use `_qgf_converter` to maintain backwards incomptability and accept `QGF` for `bitsize` parameter

QGF holds more information than just a bitsize ... the polynomial is used in comparison of QGF instances. the way I do this is backwards compatitible ... the qgf gets `@_qgf.default` which is the old value used

NoureldinYosri:
@tanujkhattar wouldn't this be confusing?

mpharrigan:
sgtm

NoureldinYosri:
@tanujkhattar ptal

tanujkhattar (approved):
Thanks!

qualtran/bloqs/gf_arithmetic/gf2_inverse.py
```diff
@@ -129,90 +149,106 @@ def my_static_costs(self, cost_key: 'CostKey'):
...
+        t = (self.bitsize - 1).bit_count()
+        k1 = bit_length(self.bitsize - 1) - 1
+        k = max(k1 + t - 1, k1 + 1)
+        f = [x] + [None] * k
+        f[k] = bb.allocate(self.bitsize, self.qgf)
+        f = cast(list['Soquet'], f)
+        for i in range(1, k1 + 1):
+            f[i - 1], f[k] = bb.add(GF2Addition(self.qgf), x=f[i - 1], y=f[k])
+            f[k] = bb.add(GF2Square(self.qgf, 2 ** (i - 1)), x=f[k])
+            f[i - 1], f[k], f[i] = bb.add(GF2MulViaKaratsuba(self.qgf), x=f[i - 1], y=f[k])
```
tanujkhattar: One comment - Can you make `GF2MulViaKaratsuba(self.qgf)` a property on the class such that in the decomposition you call `self.gf2_mul` ? This will be helpful because if I need to use a different multiplier, I can just derive from this class and update the `gf_mul` property to return a more optimized multiplier with the same signature.

NoureldinYosri: done

Issue #1713, opened 2025-09-17
MultiControlZ doesn't play nicely with attrs


mpharrigan:
This class abuses inheritance mechanisms which don't play nice with attrs dataclasses. Specifically, the children classes define attributes with init=False so if you dump all the attrs.fields and try to re-create the object, it fails.

(see also #1701 )

daxfohl:
Actually the second option is kind of bad. Say you have a function like
```python
def with_new_target(mcp: MultiControlPauli, new_target: Bloq):
    return attrs.replace(mcp, target_bloq=new_target)
```
Since in the second optioni, `MultiControlZ` inherits from `MultiControlPauli`, it can be passed into `with_new_pauli` and it'll typecheck. But of course it'll fail at runtime. Not the end of the world, and probably better than silently succeeding at runtime, but with the first option trying to pass a `MultiControlZ` into that function would fail to type check, which is what we'd really want.

So I think the preferred option should be the first one: base classes should declare things as readonly properties for the most part, unless it's expected that every subclass will use them as configurable attributes. That way child classes can implement configurable properties as attrs, and fixed properties as properties, and methods like `with_new_target` could be implemented safely and generically.

mpharrigan:
fixed in #1739

PR #1739, opened 2025-10-17
Add MultiControlPauliBase

maxglick:
Add a common base class for MultiControlPauli and its specializations.
This is the recommended approach in the comments of https://github.com/quantumlib/Qualtran/issues/1713.

### DIFF
--- a/qualtran/bloqs/mcmt/multi_control_pauli.py
+++ b/qualtran/bloqs/mcmt/multi_control_pauli.py
@@ -196,10 +196,7 @@ class MultiControlX(MultiControlPauliBase):
 
     cvs: Union[HasLength, Tuple[int, ...]] = field(converter=_to_tuple_or_has_length)
 
-    def __attrs_post_init__(self):
-        pass
-
-    @property
+    @cached_property
     def target_bloq(self) -> 'Bloq':
         return XGate()
 
@@ -243,10 +240,7 @@ class MultiControlZ(MultiControlPauliBase):
 
     cvs: Union[HasLength, Tuple[int, ...]] = field(converter=_to_tuple_or_has_length)
 
-    def __attrs_post_init__(self):
-        pass
-
-    @property
+    @cached_property
     def target_bloq(self) -> 'Bloq':
         return ZGate()
 


maxglick:
Hello.

I am brand new to open source and also not a Python expert.  There are a couple issues here that I know of:
1) Running "pytest qualtran/bloqs/mcmt/multi_control_pauli_test.py", the deprecation warning for MultiControlPauli no longer shows up.
2) I am getting pylint errors of type "unexpected-keyword-arg" for calls to the constructors in this file.

Thanks for your help.

qualtran/bloqs/mcmt/multi_control_pauli.py
```diff
@@ -45,34 +46,27 @@
...
-    clean ancillas using a multi-controlled `AND` gate. Uses the Toffoli ladder
-    construction described in "n−2 Ancilla Bits" section of Ref[1] but uses an
-    $\text{AND} / \text{AND}^\dagger$ ladder instead for computing / uncomputing
-    using clean ancillas instead of the Toffoli ladder. The measurement based
-    uncomputation of $\text{AND}$ does not consume any magic states and thus has
-    better constant factors.
-
-    References:
-        [Constructing Large Controlled Nots](https://algassert.com/circuits/2015/06/05/Constructing-Large-Controlled-Nots.html)
+class MultiControlPauliBase(GateWithRegisters):
```
mpharrigan: I think you're supposed to add `metaclass=abc.ABCMeta` when defining base classes; although I'm not sure if this is strictly required I think it's good practice

qualtran/bloqs/mcmt/multi_control_pauli.py
```diff
@@ -45,34 +46,27 @@
...
+    r"""Abstract base class for MultiControlPauli and its specializations.
     """
 
-    cvs: Union[HasLength, Tuple[int, ...]] = field(converter=_to_tuple_or_has_length)
-    target_bloq: Bloq
-
-    def __attrs_post_init__(self):
-        warnings.warn(
+    def __init__(self):
```
mpharrigan: you could probably leave this as __attrs_post_init__ in `MultiControlPauli`. Maybe the attrs classes don't call the base class init methods, and that's why the deprecation warning goes missing. And I don't actually want to deprecate `MultiControlZ` (although we may change the implementation in the future to what the deprecation warning says to do)

qualtran/bloqs/mcmt/multi_control_pauli.py
maxglick: Done.

qualtran/bloqs/mcmt/multi_control_pauli.py
maxglick: Ah, that makes sense.  Done.

qualtran/bloqs/mcmt/multi_control_pauli.py
```diff
@@ -45,33 +46,16 @@
     from qualtran.resource_counting import BloqCountDictT, SympySymbolAllocator
 
 
-@frozen
```
daxfohl, line 48: I think it might be useful to keep the frozen attribute here. I know there are edge cases where mixing attrs and regular classes in a hierarchy can cause problems, like `super().__post_init__()` can barf sometimes, especially if there's multiple inheritance. WDYT?

qualtran/bloqs/mcmt/multi_control_pauli.py
```diff
@@ -165,7 +149,36 @@ def _has_unitary_(self) -> bool:
...
+    $\text{AND} / \text{AND}^\dagger$ ladder instead for computing / uncomputing
+    using clean ancillas instead of the Toffoli ladder. The measurement based
+    uncomputation of $\text{AND}$ does not consume any magic states and thus has
+    better constant factors.
+
+    References:
+        [Constructing Large Controlled Nots](https://algassert.com/circuits/2015/06/05/Constructing-Large-Controlled-Nots.html)
+    """
+
+    cvs: Union[HasLength, Tuple[int, ...]] = field(converter=_to_tuple_or_has_length)
```
daxfohl, line 167: Should this go in the base class since it's a field in all subclasses?

qualtran/bloqs/mcmt/multi_control_pauli.py
```diff
@@ -181,15 +194,15 @@ class MultiControlX(MultiControlPauli):
...
     """
 
-    target_bloq: Bloq = field(init=False)
-
-    @target_bloq.default
-    def _X(self):
-        return XGate()
+    cvs: Union[HasLength, Tuple[int, ...]] = field(converter=_to_tuple_or_has_length)
 
     def __attrs_post_init__(self):
```
daxfohl: I think this can be removed now. IIUC it was only here to override the warning of `MultiControlPauli`, but that's no longer the base class.

qualtran/bloqs/mcmt/multi_control_pauli.py
```diff
@@ -181,15 +194,15 @@ class MultiControlX(MultiControlPauli):
...
         pass
 
+    @property
```
daxfohl: Should this be `@cached_property`? Or maybe even a class-level instance?

daxfohl (commented):
Nice! A couple comments, though I don't think any of them are necessarily blocking. cc @mpharrigan

qualtran/bloqs/mcmt/multi_control_pauli.py
mpharrigan: no, mixing attrs and inheritance is more trouble than it's worth in general

qualtran/bloqs/mcmt/multi_control_pauli.py
mpharrigan: the base class specifies that it must be present, and this is an implementation; so I think this is good. mixing attrs and inheritance often causes problems and confers no benefit

qualtran/bloqs/mcmt/multi_control_pauli.py
mpharrigan: it could be a cached_property; I think making class variables is tricky to do with attrs

qualtran/bloqs/mcmt/multi_control_pauli.py
maxglick: Done.

qualtran/bloqs/mcmt/multi_control_pauli.py
maxglick: Done (if I understand correctly that we want property->cached_property specifically here and in MultiControlZ).

qualtran/bloqs/mcmt/multi_control_pauli.py
daxfohl: Okay, so basically just use attrs on the leaf classes, and base classes should have no fields? That seems reasonable. 

I was thinking maybe if the whole hierarchy was attrs then that would also be a tractable pattern, but probably harder to enforce, or impossible if something needs to inherit from some external non-attrs class for compatibility (e.g. cirq.Gate). And it can also dork things up in a multiple inheritance case, since `__post_init__` may or may not have a `super().__post_init__()`, depending on the MRO, and if it does, it may or may not have a diamond dependency, so it's hard to ensure you're properly initializing all base classes without double-initializing them.

Admittedly it's still tempting to do this because attrs makes it easy, and there are lots of places where you just want some existing attrs class plus one more field. But there's probably a good reason why record-oriented languages like OCaml don't allow this, so it's probably a good pattern to follow.

(Sorry for stream of consciousness. TLDR: Yeah I think I agree).

qualtran/bloqs/mcmt/multi_control_pauli.py
daxfohl: Tangential to this PR, but asking in general: would it be useful to have representative instances of common bloqs, like how in cirq how `cirq.X` is just an instance of `XPowGate`, to avoid extra allocation for each instance? Or does that not fit right with qualtran's design? @mpharrigan

daxfohl (approved):

qualtran/bloqs/mcmt/multi_control_pauli.py
mpharrigan: I wouldn't use that microoptimization as motivation

PR #1746, opened 2025-10-21
Notebook CI Job -- report times and fix regression

mpharrigan:
- fixes performance regression introduced in #1732 
 - After some experimentation, run on 8 processes with maxtasksperworker=1
 - Report 10 slowest notebooks

### DIFF
--- a/qualtran/bloqs/block_encoding/lcu_block_encoding_test.py
+++ b/qualtran/bloqs/block_encoding/lcu_block_encoding_test.py
@@ -48,7 +48,7 @@ def test_ctrl_lcu_be_cost():
     assert bloq.controlled(CtrlSpec(cvs=0)) == attrs.evolve(bloq, control_val=0)
 
     assert get_cost_value(bloq.controlled(), QECGatesCost()) == GateCounts(
-        cswap=28, and_bloq=77, clifford=438, rotation=16, measurement=77
+        cswap=28, and_bloq=77, clifford=294, rotation=16, measurement=77
     )
 
 
--- a/qualtran/bloqs/chemistry/hubbard_model/qubitization/select_hubbard.py
+++ b/qualtran/bloqs/chemistry/hubbard_model/qubitization/select_hubbard.py
@@ -62,8 +62,9 @@ from qualtran import (
     Signature,
     SoquetT,
 )
-from qualtran.bloqs.basic_gates import CSwap, TwoBitCSwap
+from qualtran.bloqs.basic_gates import CNOT, CSwap, CZ, TwoBitCSwap
 from qualtran.bloqs.bookkeeping import Join2, Split2
+from qualtran.bloqs.mcmt import And
 from qualtran.bloqs.multiplexers.apply_gate_to_lth_target import ApplyGateToLthQubit
 from qualtran.bloqs.multiplexers.select_base import SelectOracle
 from qualtran.bloqs.multiplexers.selected_majorana_fermion import SelectedMajoranaFermion
@@ -277,6 +278,10 @@ class HubbardMajorannaOperator(Bloq):
         return ceil(log2(self.x_dim))
 
     @cached_property
+    def N(self):
+        return 2 * self.x_dim * self.y_dim
+
+    @cached_property
     def control_registers(self) -> Tuple[Register, ...]:
         return () if self.control_val is None else (Register('control', QBit()),)
 
@@ -290,7 +295,7 @@ class HubbardMajorannaOperator(Bloq):
 
     @cached_property
     def target_registers(self) -> Tuple[Register, ...]:
-        return (Register('target', QAny(self.x_dim * self.y_dim * 2)),)
+        return (Register('target', QAny(self.N)),)
 
     @cached_property
     def signature(self) -> Signature:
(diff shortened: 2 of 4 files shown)


mpharrigan:
Results from serial execution on the CI: 24 mins total

```
392.72s /home/runner/work/Qualtran/Qualtran/qualtran/bloqs/phase_estimation/phase_estimation_of_quantum_walk.ipynb
62.10s /home/runner/work/Qualtran/Qualtran/qualtran/bloqs/rotations/quantum_variable_rotation.ipynb
25.70s /home/runner/work/Qualtran/Qualtran/qualtran/bloqs/arithmetic/error_analysis_for_fxp_arithmetic.ipynb
22.08s /home/runner/work/Qualtran/Qualtran/qualtran/bloqs/phase_estimation/kaiser_window_state.ipynb
19.59s /home/runner/work/Qualtran/Qualtran/qualtran/Adjoint.ipynb
18.66s /home/runner/work/Qualtran/Qualtran/qualtran/bloqs/state_preparation/state_preparation_via_rotation_tutorial.ipynb
15.92s /home/runner/work/Qualtran/Qualtran/qualtran/bloqs/optimization/k_xor_sat/kikuchi_guiding_state_tutorial.ipynb
12.80s /home/runner/work/Qualtran/Qualtran/qualtran/bloqs/chemistry/resource_estimation.ipynb
 9.22s /home/runner/work/Qualtran/Qualtran/qualtran/bloqs/arithmetic/t_complexity_of_comparison_gates.ipynb
 7.81s /home/runner/work/Qualtran/Qualtran/qualtran/bloqs/multiplexers/apply_gate_to_lth_target.ipynb
```

mpharrigan:
without maxtasksperchild=1, the total time of the execution step is 6m22s _and doesn't show any of the stdout until the very end_. I'll keep that change in

.github/workflows/pr.yaml
```diff
@@ -76,7 +76,7 @@ jobs:
           pip install -r dev_tools/requirements/envs/pytest.env.txt
           pip install --no-deps -e .
       - run: |
-          python dev_tools/execute-notebooks.py
+          python dev_tools/execute-notebooks.py --n-workers=8
```
mhucka, line 79:
```suggestion
          python dev_tools/execute-notebooks.py --n-workers=$(nproc)
```

According to [GitHub's info about their current standard runners](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories), Linux runners get 4 virtual CPUs. If these notebook executions are primarily doing CPU-bound work and not I/O bound work, then it's _possible_ a value of > 4 here may actually be counter-productive. Also, the characteristics of the GitHub runners may change over time. So, instead of hard-coding this value to 8, what about using `nproc` here to get the number of processing units actually available to the CI job?

dev_tools/execute-notebooks.py
```diff
@@ -22,9 +22,13 @@ def parse_args():
     p.add_argument('--output-nbs', action=argparse.BooleanOptionalAction, default=True)
     p.add_argument('--output-md', action=argparse.BooleanOptionalAction, default=False)
     p.add_argument('--only-out-of-date', action=argparse.BooleanOptionalAction, default=True)
+    p.add_argument('--n-workers', type=int, default=None)
```
mhucka, line 25: Shouldn't the default here be 1, since the type is `int`?

dev_tools/execute-notebooks.py
```diff
@@ -22,9 +22,13 @@ def parse_args():
...
     args = p.parse_args()
     execute_and_export_notebooks(
-        output_nbs=args.output_nbs, output_md=args.output_md, only_out_of_date=args.only_out_of_date
+        output_nbs=args.output_nbs,
+        output_md=args.output_md,
+        only_out_of_date=args.only_out_of_date,
+        n_workers=args.n_workers,
```
mhucka, line 31: Also, optionally, execute-notebooks.py could have a line or two to test the value before calling `execute_and_export_notebooks(…)`. It might be helpful to provide a way for users/developers to indicate they want as many processes as possible without having to pass an actual number. A convention used in some other programs is that a value of `0` means "as many as possible", so one approach here would be to change the line to be

```python
n_workers=(args.n_workers if args.n_workers > 0 else os.cpu_count())
```

dev_tools/qualtran_dev_tools/notebook_execution.py
```diff
@@ -217,23 +233,38 @@ def execute_and_export_notebooks(
         output_md: Whether to save the executed notebooks as markdown
         only_out_of_date: Only re-execute and re-export notebooks whose output files
             are out of date.
+        n_workers: If set to 1, do not use parallelization. Otherwise, use
+            `multiprocessing.Pool(n_workers)` to execute notebooks in parallel.
```
mhucka: The way the argument parser is currently defined, if the user does not supply a value for `--n-workers`, the value will be `None`, which `multiprocessing.Pool(…)` will interpret as "use as many as there are CPU cores on the machine". Although it's true that the docstring above is _technically_ correct, the reader would have to know about the behavior of `multiprocessing.Pool(None)` to deduce how many workers will be used. IMHO a sentence here would be helpful to explain the connection between not providing a value and the resulting number of processes used.

.github/workflows/pr.yaml
mpharrigan: the thing is, these are pretty I/O bound. Usually half the time is spent spinning up the jupyter kernels (which run as little zmq servers) and other overhead

dev_tools/execute-notebooks.py
mpharrigan: the default value of `None` does this, which is how it's done in the Python `multiprocessing` api

.github/workflows/pr.yaml
mhucka: Got it. Yeah, for I/O bound I guess it's fine to exceed the number of CPUs.

dev_tools/execute-notebooks.py
mhucka: Nevermind, in light of the realization that the desired default is "as many as available".

dev_tools/execute-notebooks.py
mhucka: Oops, I actually realized that while writing the comment below (about the docstring), but didn't remember to delete this note.

dev_tools/qualtran_dev_tools/notebook_execution.py
mhucka: I still suggest a note in the docstring to mention the behavior, but it's up to you.

mpharrigan: yep i'll add that

PR #1751, opened 2025-11-04
[Rotation Synthesis] add Tx/Tz/Tx*/Tz* representation and methods to export to Cirq and Quirk

NoureldinYosri:
This PR adds
1. a method to represent an $SU(2)$ matrix as a string of $Tx, Tx^\dagger, Tz, Tz^\dagger$
2. a method that exports the representation as a cirq circuit
3. a method that exports the representatoin as quirk link


mpharrigan:
Can you add a PR description

Issue #1495, opened 2024-11-12
`cirq.unitary` does not work for controlled `QubitizationWalkOperator` operator


anurudhp:
example code:

```py
import cirq
from qualtran.bloqs.for_testing.qubitization_walk_test import get_uniform_pauli_qubitized_walk

_, walk = get_uniform_pauli_qubitized_walk(4)
bloq = walk.controlled()

cbloq = bloq.as_composite_bloq()

sim = cirq.Simulator()
_ = sim.simulate(cbloq.to_cirq_circuit())
```
raises
```
TypeError: Simulator doesn't support Controlled(subbloq=QubitizationWalkOperator(block_encoding=SelectBlockEncoding(select=SelectPauliLCU(selection_bitsize=2, target_bitsize=4, select_unitaries=(cirq.DensePauliString('XYII', coefficient=(1+0j)), cirq.DensePauliString('IYZI', coefficient=(1+0j)), cirq.DensePauliString('IIZX', coefficient=(1+0j)), cirq.DensePauliString('YIIX', coefficient=(1+0j))), control_val=None), prepare=PrepareUniformSuperpositionTest(n=4, cvs=(), qlambda=4.0))), ctrl_spec=CtrlSpec(qdtypes=(QBit(),), cvs=(array(1),))).on(cirq.NamedQubit('ctrl'), cirq.NamedQubit('selection0'), cirq.NamedQubit('selection1'), cirq.NamedQubit('target0'), cirq.NamedQubit('target1'), cirq.NamedQubit('target2'), cirq.NamedQubit('target3'))
```

But decomposing once results in a simulable circuit: 
```py
_ = sim.simulate(cbloq.flatten_once().to_cirq_circuit())
```

I suspect it is some interop issue b/w cirq and GateWithRegisters. Could be related to #1488

PR #1761, opened 2025-11-11
Support Numpy 2 via supporting Cirq 1.5+

mpharrigan:
## Changes for numpy
 - data_types.py -- np.uint8 is stricter now
 - and_bloq.by -- there were np.uint8's sneaking in
 - There are new mypy complaints that I've generally ignored

## Changes for Cirq

We require a more recent version of Cirq to support Numpy 2. 

### ControlledGate and `_has_unitary_`

I've monkeypatched `cirq.ControlledGate`. It was using `cirq.has_unitary` to determine whether something could be controlled. The rest of Cirq now treats "has unitary" as the ability to create a unitary matrix rather than saying whether the gate is reversible.
 - Tracked in #1764 .

### Simulation

There is something fundamentally incompatible between Qualtran's cirq interop functionality and recent Cirq versions. I've taken the "nuclear option", and now when you request the cirq simulation of a `BloqAsCirqGate` it will be done entirely in Qualtran and export a large unitary matrix to the cirq simulator; which can cause performance regressions;
 - see #1763.

In this PR, I've implemented the quick fixes to speed up Qualtran simulation of narrow, deep circuits from my comment on #1336 

Some of the unit tests now cause SIGKILL on my machine. I don't know what's going on here. I've xfailed them.
 - See #1763.

## Related issues

 - #237 
 - #1488 (technically fixes, as written; but see performance regression issue)
 - #1495
 - #1336

### DIFF
--- a/qualtran/bloqs/data_loading/select_swap_qrom.py
+++ b/qualtran/bloqs/data_loading/select_swap_qrom.py
@@ -253,7 +253,7 @@ class SelectSwapQROM(QROMBase, GateWithRegisters):  # type: ignore[misc]
         for i, data in enumerate(self.padded_data):
             for batch_idx in np.ndindex(cast(Tuple[int, ...], self.batched_qrom_shape)):
                 data_idx = [slice(x * k, (x + 1) * k) for x, k in zip(batch_idx, self.block_sizes)]
-                batched_data[i][(*batch_idx, *block_slices)] = data[tuple(data_idx)]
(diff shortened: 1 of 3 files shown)


mpharrigan:
I've done local testing of the full unit test suite (not just the `not slow` tests) on mac and linux with Python 3.10 and newer, as well as with Cirq 1.6. The CI is still on Python 3.10 which caps us at Cirq 1.5. I'll leave bumping the CI Python version to a follow-up. So I think the nightly tests should pass; but we can't know for sure until this is merged. This is an area where the quick tests cannot catch all the errors that seem to happen only on the larger simulation tests.

PR #1776, opened 2025-12-02
make submodules in rotation_synthesis/ private to improve doc view

NoureldinYosri:
For now this is just a minimial change that renames of the files while exporting essential methods/variables at __init__.py files with minimal changes to the the import statements. overtime import statments such as `import xx._yy` should be
 `import xx` unless it is happening within the `xx` directory


mpharrigan (approved):
thanks!

PR #1781, opened 2025-12-12
Fix `Identity`'s ctrl_adder

petim0:
`Identity` is a special bloq: its controlled version is still an identity, but with a larger register, ie. calling `Identity(n).controlled(CtrlSpec(QAny(m)))` gets us `Identity(n+m)`

The old version of `ctrl_adder()` was using `Autopartition` which failed during decomposition as `Autopartition` is not designed to merge registers and pass them to the inner bloq.

This is fixed by this PR by replacing `Autopartition` with `Partition(s)`.

This fix is functional but the resulting `show_bloq()` output is less clean as the Partition steps are visible. An alternative (maybe better) fix is to keep using `Autopartition` but mark the control registers as `Unused()`, but It is weird conceptually to not use Identity(n+m) as the controlled Identity.  

This PR also adds a test covering this bug.

### DIFF
--- a/qualtran/bloqs/basic_gates/identity.py
+++ b/qualtran/bloqs/basic_gates/identity.py
@@ -32,9 +32,9 @@ from qualtran import (
     Signature,
     SoquetT,
 )
+from qualtran.bloqs.bookkeeping import Partition
 from qualtran.drawing import Text, TextBox, WireSymbol
 from qualtran.symbolics import is_symbolic, SymbolicInt
-from qualtran.bloqs.bookkeeping import Partition
 
 if TYPE_CHECKING:
     import cirq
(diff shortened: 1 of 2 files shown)


petim0:
This is the `show_bloq` that is less clean:
Screenshot 2025-12-12 at 16 41 09

mpharrigan (approved):
nice, I agree this is probably the right fix

mpharrigan:
can you run 

```
check/format-incremental --apply
```

to format the code.

There's also a mypy type checking problem in the unit test. Since this is a unit test, you can add a `# type: ignore` comment if it's not obvious how to fix it. You can run the type checker with `check/mypy`.

Issue #1783, opened 2025-12-12
LinearCombinaison bloq has an incorrect check in its __attrs_post_init__()


petim0:
Currently there is this check in `LinearCombinaison`:
```
if self._select is not None and self._select.target_registers != (
            self.signature.get_left("system"),
        ):
            raise ValueError(
                "If given, select oracle must have block encoding `system` register as target."
            )
```
which checks that the name and size of the `Select` `target_registers` is the same as the `LinearCombinaison` `"system"` register.
And the `LinearCombinaison` system register size is defined as: `system=QAny(self.block_encoding.system_bitsize)`

This creates a contradiction: when using the select oracle we connect `Select`'s  `target_register` to `LinearCombinaison`'s system register but also the ancilla register and the resource register. So `Select`'s `target_register` should be bigger than `LinearCombinaison`'s`system` register.

PR #1784, opened 2025-12-12
Fix #1782 and #1783: `LinearCombinaison` adjoint is more general and checks are correct

### LINKED ISSUE #1782: LinearCombination should not be Hermitian
Right now the `adjoint() `of `LinearCombination` is defined as such:
```
def adjoint(self) -> 'LinearCombination':
        return self
```
 
LCU is defined as:
$$LCU = \sum_i \alpha_i U_i$$
And thus:
$$LCU^\dagger = \sum_i \alpha_i^* U_i^\dagger$$

So LCU is hermitian if and only if $\alpha_i \in \mathbb{R}$ (which is required by the class) and all the added unitaries are Hermitian.

We can therefore generalize the `LinearCombination` class and let the Hermicity be implemented by passing Hermitian bloqs to `LinearCombination`.

This is actually what is happening in `LCUBlockEncoding`.

petim0:
This implements solutions to #1783 and #1782, ~~still need to implement some tests~~.

### DIFF
--- a/qualtran/bloqs/basic_gates/identity.py
+++ b/qualtran/bloqs/basic_gates/identity.py
@@ -32,6 +32,7 @@ from qualtran import (
     Signature,
     SoquetT,
 )
+from qualtran.bloqs.bookkeeping import Partition
 from qualtran.drawing import Text, TextBox, WireSymbol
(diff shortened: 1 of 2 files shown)


mpharrigan (approved):
Much cleaner, thanks

PR #1787, opened 2026-01-06
Jan 2026 dependencies bump

mpharrigan:
- bump dependencies to latest versions that form a mutually-compatible set with `pip-compile`.
 - Latest sympy no longer re-exports the symbols we use in `sympy_to_proto.py`, so I added the imports.
 - Mypy/Numpy is getting pickier about `NDArary[cirq.Qid]` and friends. Added `# type: ignore` and opened #1789 to track longer-term fix

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -30,7 +30,7 @@ permissions: read-all
 jobs:
   pytest:
     runs-on: ubuntu-22.04
-    timeout-minutes: 60
+    timeout-minutes: 120
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
       - uses: ts-graphviz/setup-graphviz@c001ccfb5aff62e28bda6a6c39b59a7e061be5b9 # v1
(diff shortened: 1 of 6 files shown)


mpharrigan:
Opened #1789 to track the growing complaints with e.g. `NDArray[cirq.Qid]` in type annotations

mhucka (changes requested):
This looks overall okay, but it seems to mix multiple kinds of changes into one PR:

1. Dependency version updates in the various requirements files
2. Addition of type annotations
3. In the case of `qualtran/serialization/sympy_to_proto.py`, addition of imports

It's not immediately clear whether these are actually related and need to be in the same PR. If they do need to be in the same CL, could the reasons be explained in the PR description?

Conversely, if they don't need to be in the same CL, I would recommend following the principle of small CLs (https://google.github.io/eng-practices/review/developer/small-cls.html) and splitting this up into separate PRs.

(I hope this doesn't come across as too critical. Just trying to do a good review.)

mpharrigan:
should have noted that those changes are in response to new mypy and pylint failures that only occur with the updated dependency versions. will update pr description

PR #1806, opened 2026-02-10
Fixes identified by ruff

mpharrigan:
I ran `ruff` to find style things that I manually verified were relevant, not false positives, and not controversial;  and included the fixes here. I'll commit a ruff config when I've settled on the correct sensitivity-specificity. 

This is preliminary work in support of #1807

### DIFF
--- a/dev_tools/requirements/deps/docs.txt
+++ b/dev_tools/requirements/deps/docs.txt
@@ -7,4 +7,4 @@ pydata-sphinx-theme
 myst-nb
 tensorflow-docs~=2023.5.24
 
-griffe
+griffe~=1.0
--- a/dev_tools/requirements/envs/dev.env.txt
+++ b/dev_tools/requirements/envs/dev.env.txt
@@ -4,7 +4,7 @@
 #
 #    pip-compile --output-file=envs/dev.env.txt deps/dev-tools.txt deps/runtime.txt
 #
-absl-py==2.3.1
+absl-py==2.4.0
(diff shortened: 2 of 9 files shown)


qualtran/bloqs/arithmetic/comparison_test.py
```diff
@@ -202,18 +202,17 @@ def _less_than_equal_expected_t_complexity(gate: LessThanEqual):
...
         return TComplexity(t=8 * n - 4, clifford=46 * n - 21)
+    # When the registers differ in size and `n` is the size of the smaller one and
+    # `d` is the difference in size. The T complexity is the sum of the tree
+    # decomposition as before giving 8n + O(1) and the T complexity of an `And` gate
+    # over `d` registers giving 4d + O(1) totaling 8n + 4d + O(1).
+    # From the decomposition we get that the constant is -4 as well as the clifford counts.
+    elif d == 1:
+        return TComplexity(t=8 * n, clifford=46 * n - 1 + 2 * is_second_longer)
     else:
-        # When the registers differ in size and `n` is the size of the smaller one and
```
fdmalone, line 206: ruff suggested this?

mpharrigan: ruff auto-fixed this, yes

mpharrigan: https://docs.astral.sh/ruff/rules/collapsible-else-if/

PR #1810, opened 2026-02-13
Fix Product's `build_call_graph`

petim0:
The old version of Product had a `build_call_graph` that was incorrect when the bloq encodings had ancillas, as the Multi_controlled_X was not consistent between the decomposition and the `build_call_graph`.

This PR fixes that, I also took the time to change the split/join behavior to control the XGate to an `Autopartition` and I also made the symbolic decomposition work. This PR also adds tests.

Before
[output.pdf](https://github.com/user-attachments/files/25289067/output.pdf)

Now
[outputNow.pdf](https://github.com/user-attachments/files/25289118/outputNow.pdf)

### DIFF
--- a/qualtran/_infra/bloq.py
+++ b/qualtran/_infra/bloq.py
@@ -32,7 +32,6 @@ from typing import (
 if TYPE_CHECKING:
     import cirq
     import networkx as nx
-    import pennylane as qml
     import quimb.tensor as qtn
     import sympy
     from numpy.typing import NDArray
@@ -41,7 +40,6 @@ if TYPE_CHECKING:
 
     from qualtran import (
         AddControlledT,
-        Adjoint,
         BloqBuilder,
         CompositeBloq,
         ConnectionT,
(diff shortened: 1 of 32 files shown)


qualtran/bloqs/block_encoding/product_test.py
```diff
@@ -216,7 +227,19 @@ def test_product_signal_state():
...
+def test_product_symbolic_call_graph_decomposes():
+    from qualtran.bloqs.basic_gates import Hadamard, TGate
+    from qualtran.bloqs.block_encoding.unitary import Unitary
+
+    product_block_encoding = Product(
+        (
+            Unitary(TGate(), ancilla_bitsize=sympy.symbols("anc")),
+            Unitary(Hadamard(), ancilla_bitsize=sympy.symbols("anc")),
+        )
+    )
```
mpharrigan, line 242: this test doesn't do anything and if I add a call to `decompose_bloq` it throws an error

E           qualtran._infra.bloq.DecomposeTypeError: Cannot decompose symbolic self=Product(block_encodings=(Unitary(U=TGate(is_adjoint=False), alpha=1, ancilla_bitsize=anc, resource_bitsize=0, epsilon=0), Unitary(U=Hadamard(), alpha=1, ancilla_bitsize=anc, resource_bitsize=0, epsilon=0)))

PR #1812, opened 2026-02-15
Changes `Partition` to optionally take an input dtype

petim0:
I changed partition to allow to optionally pass a dtype for the input register. 

This change comes from a bug I encountered when partitioning a `QInt` and running `assert_consistent_classical_action`:  
The partition assumes that your input is a `QAny()` and `QAny()` is assumed to have a `_UInt` encoding, therefore when the `_classical_partition` function of `_PartitionBase` uses `self.lumped_dtype.to_bits(x)` and you have a negative number as a input then you get an error you should not have, as the `_UInt` encoding doesn't allow for negative numbers.

Mypy was really not happy with the `match` statement and so I had do do a lot of useless `asserts`. Maybe a better solution is I to just ignore the errors instead.

I also added a `LegacyPartitionWarning` that I then put in the `filterwarnings` of pytest as to not have it everywhere in the code when testing. 

I also choose to have the base dtype of the partition register to be a `QUInt` as to more reflect the fact that we assume a `_Uint` encoding.


mpharrigan:
main philosophical discussion on #1813 

Yes, I think we should let you partition any data type so you can write safe, generic quantum programs.

The constructor for `Partition` has to be backwards compatible and I don't think we should deprecate partitioning a `QAny`. You can monkey with the order and defaults of the constructor to make a backwards-compatible way of accepting a dtype argument or, my preference:

I think this would work better as a new `PartitionT` class that leaves legacy `Partition` alone and uses the fact that the logic is now factored out into a base class to ones advantage. You can add a "see also /  consider instead" to the legacy partition docstring

PR #1851, opened 2026-05-08
`dev-tools` replace deprecated `typing` aliases with those from `collections`

micpap25:
Partial on #1653 
As suggested by @mhucka isolate the changes from #1677 to smaller subsets of the codebase so that the review process is easier.
This PR removes compatibility with pre-3.10 python versions, but Qualtran isn't intended to target those anymore.


qualtran/protos/bloq_pb2.pyi
mhucka, line 1: Just double-checking: @micpap25 how did you (re)generate the .pyi files?

dev_tools/bloq-method-overrides-report.py
```diff
@@ -42,7 +42,7 @@ def _call_graph(bc: Type[Bloq]):
         )
     if annot['ssa'] != 'SympySymbolAllocator':
         print(f"{bc}.build_call_graph `ssa: 'SympySymbolAllocator'`")
-    if annot['return'] != Set[ForwardRef('BloqCountT')]:  # type: ignore[misc]
+    if annot['return'] != set[ForwardRef('BloqCountT')]:  # type: ignore[misc]
```
mhucka: This one is incorrect. The `typing.Set` and the built-in `set` are not considered equal when used in type annotations, even if they contain the same arguments. Here is a small test program to demonstrate:

```python
from typing import ForwardRef, Set

# Original annotation using typing.Set
original = Set[ForwardRef('BloqCountT')]

# PEP 585 annotation using built-in set
pep585 = set[ForwardRef('BloqCountT')]

print(f"Is {original} == {pep585}?")
print(f"Result: {original == pep585}")
```

In the `typing` module, parameterizing a type with a string automatically wraps that string in a `ForwardRef` object. That meant that `typing.Set['BloqCountT']` was identical to `typing.Set[typing.ForwardRef('BloqCountT')]` in the previous version of the code. However, PEP 585's built-in collections do not auto-convert strings to `ForwardRef` objects. When it evaluates `set['BloqCountT']`, Python directly creates a `types.GenericAlias` containing the literal string `'BloqCountT'`.

The fix in this particular case turns out to be easy: get rid of the `ForwardRef` completely and do:

```python
    if annot['return'] != set['BloqCountT']:  # type: ignore[misc]
```

dev_tools/qualtran_dev_tools/reference_docs.py
```diff
@@ -383,7 +383,7 @@ def generate_ref_toc(reporoot: Path):
     page_paths = output_dir.glob('qualtran/**/*.md')
 
     # Group according to module
-    grouped_paths: Dict[Path, List] = defaultdict(list)
+    grouped_paths: dict[Path, list] = defaultdict(list)
```
mhucka: The original type definition could have been a bit more specific in this case, so while the PR doesn't do anything wrong here, we may as well make a small enhancement to the type annotation:

```suggestion
    grouped_paths: dict[Path, list[Path]] = defaultdict(list)
```

dev_tools/qualtran_dev_tools/bloq_report_card.py
mhucka: Not an error in this PR, but while we're at it, this function could use the addition of a return type.

```suggestion
def color_status(v: BloqCheckResult) -> str:
```

mhucka:
```suggestion
def format_status(v: BloqCheckResult) -> str:
```

dev_tools/qualtran_dev_tools/make_reference_docs/_components/render_docstring.py
```diff
@@ -193,7 +193,7 @@ def _write_unknown(f, part: DocstringSection, level: int):
 }
 
 
-def write_docstring_parts(f, parts: List[DocstringSection], level: int):
+def write_docstring_parts(f, parts: list[DocstringSection], level: int):
```
mhucka: This was not in the original code, but could we add a type annotation for `f`?

mhucka (changes requested):
Thanks for this work! It's quite thorough.

I found one place where there's a subtle bug, and a few suggestions for minor tweaks elsewhere, but otherwise this is almost done!

dev_tools/qualtran_dev_tools/bloq_report_card.py
mhucka: (I don't know why GitHub is not showing the original line 40 code here. No matter what I do, it only shows the changed code as an addition. ¯\_(ツ)_/¯)

qualtran/protos/bloq_pb2.pyi
micpap25: I'm not sure why it doesn't show the diffs because they are quite straightforward, but I do just manually change them to remove typing and python 3.8 support like any other file.

dev_tools/bloq-method-overrides-report.py
micpap25: Good to know, because I replace `typing.Set` with `set` throughout the other parts of the code; I could go and fix that as well / remove the `ForwardRef` usage there too? I will make the suggested fix, though.

dev_tools/qualtran_dev_tools/make_reference_docs/_components/render_docstring.py
micpap25: I can add annotations for all the files in the folder, I'm seeing `Writable` and `LinkingWriter` are both types that could be used here? I'm choosing to use the former since it seems more modular but let me know if another type would be better.

dev_tools/bloq-method-overrides-report.py
micpap25: Regarding this bug, it seems that making the change you recommended causes an issue in the `dev-tools` CI about `BloqCountT` being undefined. It seems fixing this for every type `typing.Set` is changed to the built-in `set` will require a lot of `TYPE_CHECKING` imports?

mhucka: Huh. Let me look at it over the weekend.

qualtran/protos/bloq_pb2.pyi
mhucka: Ah. _pb2.pyi files are usually generated by a program, not written manually, and I guess GitHub doesn't make diffs for them by default.

I think this project uses the script in `dev_tools/build-protos.sh` to generate them, but I've never done it myself, so have to defer to @mpharrigan or @tanujkhattar about the exact procedure that is supposed to be used.

(The hand-edited changes look reasonable to my eyes, but in principle, there's a risk that the other type changes elsewhere will affect what `dev_tools/build-protos.sh` will produce. I would rather err on the side of caution than introduce some subtle change that might not be picked up by existing unit test cases yet might affect things like the Qualtran web service.)

dev_tools/qualtran_dev_tools/make_reference_docs/_components/render_docstring.py
mhucka: `Writable` seems to be the correct type, as far as I can tell.

dev_tools/bloq-method-overrides-report.py
mhucka: > Regarding this bug, it seems that making the change you recommended causes an issue in the `dev-tools` CI about `BloqCountT` being undefined. It seems fixing this for every type `typing.Set` is changed to the built-in `set` will require a lot of `TYPE_CHECKING` imports?

I don't have an answer to this yet, but I see all the other comments have been resolved (thank you), so what's left is this question and the .pyi files question above.

micpap25: Just following up to see if you've had a chance to test this? Thanks!

qualtran/protos/bloq_pb2.pyi
mhucka: To be on the safe side, the proto files should be regenerated using the script. However, it's not clear to me that this needs to be done as part of this PR. When I run build-protos.sh, it produces a lot more changes than just the ones related to this PR.

I'll contact the others and try to get this resolved asap.

mhucka (changes requested):
@micpap25 Sorry for the delay in getting back to this. I tested it and mypy and pytest are all happy in my Python 3.12 environment.

To make the code cleaner and avoid having to wrap type annotations in string quotes like `'Writable'`, we should add from `__future__ import annotations` to the top of the files that use Writable.

By putting that as the first instruction in the files, Python postpones evaluation of all type annotations at runtime. This allows you to write unquoted type hints. Illustration:

```python
from __future__ import annotations
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from .._linking_writer import Writable

# Note that Writable now does not need to be quoted.
def _write_text(f: Writable, part: DocstringSectionText, level: int) -> None:
```

dev_tools/bloq-method-overrides-report.py
mhucka: @micpap25 Finally got back to this.

mhucka:
@micpap25 Thank you for your patience with this. I think we have a possible path forward; I'm just waiting for confirmation from the others, which I hope we'll have in 24 hrs.

qualtran/protos/bloq_pb2.pyi
mhucka: > To be on the safe side, the proto files should be regenerated using the script. However, it's not clear to me that this needs to be done as part of this PR. When I run build-protos.sh, it produces a lot more changes than just the ones related to this PR.
> 
> I'll contact the others and try to get this resolved asap.

Update: the conclusion is that we should leave a regeneration of the protos to a separate PR. So, this part is resolved.

mhucka:
Update: we'll do a separate PR to run the proto building script after this one is merged.

dev_tools/bloq-method-overrides-report.py
mhucka: > Regarding this bug, it seems that making the change you recommended causes an issue in the `dev-tools` CI about `BloqCountT` being undefined. It seems fixing this for every type `typing.Set` is changed to the built-in `set` will require a lot of `TYPE_CHECKING` imports?

In the current PR, the change does not seem to be that bad. I did it locally to test:

```diff
diff --git a/dev_tools/bloq-method-overrides-report.py b/dev_tools/bloq-method-overrides-report.py
index 77f51231..836f342a 100644
--- a/dev_tools/bloq-method-overrides-report.py
+++ b/dev_tools/bloq-method-overrides-report.py
@@ -11,7 +11,7 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Set
 
 from qualtran_dev_tools.bloq_finder import get_bloq_classes
 
@@ -45,7 +45,7 @@ def _call_graph(bc: type[Bloq]):
         )
     if annot['ssa'] != 'SympySymbolAllocator':
         print(f"{bc}.build_call_graph `ssa: 'SympySymbolAllocator'`")
-    if annot['return'] != set['BloqCountT']:  # type: ignore[misc]
+    if annot['return'] != Set['BloqCountT']:  # type: ignore[misc]
         print(f"{bc}.build_call_graph -> 'BloqCountT'")
 
 
diff --git a/dev_tools/bloq-quickstarter.html b/dev_tools/bloq-quickstarter.html
index 7ae2240a..d7f16cad 100644
--- a/dev_tools/bloq-quickstarter.html
+++ b/dev_tools/bloq-quickstarter.html
@@ -215,7 +215,7 @@
       signature += "\n        ])"
 
       const template = `from functools import cached_property
-from typing import Optional, Union
+from typing import Optional, Union, Set
 
 from attrs import frozen
 ```

mhucka (approved):

mhucka:
@micpap25 I hope you don't mind; I went ahead and pushed changes directly to your branch with the change to `Set` as well as with the latest changes from the main branch from the upstream Qualtran repo.

If you are satisfied with the changes, I can merge the PR.

micpap25:
Yep, thank you for finishing it off!

mhucka:
Update: PR #1886 is a follow-up to the comment above about doing a separate PR to run the build-protos.sh script.

Issue #1807, opened 2026-02-10
Devops 2026


mpharrigan:
I plan on switching some of the devops tooling for qualtran following the recent release to keep us aligned with best practices in the Python ecosystem
  
 - use uv instead of pip-compile. It's way better and we can remove the clunky docker stuff
 - Bump minimum python to 3.11 following NEP29 and 3.12 in April
 - Consolidate configuration in pyproject.toml. it's currently partially there and partially scattered in dev_tools/conf. 
 - Use ruff instead of black, isort, and pylint. It is nearly drop-in for these but 1) way faster 2) we can add even more lint checks. In practice, there are minor differences in the formatting. The one that affects the most files is ruff likes to maintain a space between the end of the module docstring and the first import line.

PR #1855, opened 2026-05-13
Use `uv` for dependency management in the CI

mpharrigan:
This switches the CI scripts to use uv.lock directly instead of relying on the old-fashioned multiple requirements.txt files

devops 2026 tracker bug #1807


mpharrigan:
Please note that before this change, the flurry of dependabot PRs will have no effect. They update the uv.lock file but understandably don't know to re-generate the frozen requirements.txt files

mpharrigan:
On initial inspection of the CI logs, it was still installing stuff during the "run" phase. The "run" commands also need to be locked down. (nb the reason why it was installing stuff is because it will always install the `dev` group, which we opt out of during the environment setup call)

mpharrigan:
with [e76ea34](https://github.com/quantumlib/Qualtran/pull/1855/commits/e76ea342a608d9219469272188d55aea3acc6b08) all looks well

PR #1857, opened 2026-05-14
Bump black formatter version

mpharrigan:
Resolves a dependabot alert. I tested this locally and there are indeed minor formatting changes introduced with the new version. We'll let those shake out incrementally with `check/format-incremental`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -110,7 +110,7 @@ lint = [
 
 format = [
     "flynt~=0.60",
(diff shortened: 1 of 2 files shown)


mhucka:
xref: https://github.com/quantumlib/Qualtran/pull/1830

mpharrigan:
This subsumes #1830 as we like to keep the version of the formatter pinned to minor version even in the pyproject.toml so we don't get formatting skew

Issue #1763, opened 2025-11-11
Cirq simulation regressions


mpharrigan:
This issue tracks problems that had to be worked around to bump our Cirq version from 1.4 onwards in #1761. Note that many unit tests in the `qualtran.bloqs` quantum standard library use a Cirq simulator for numerical simulation. A way forward might be to re-work these unit tests to use native Qualtran tensor contraction.

### Performance regressions
 - `qualtran.bloqs.hamiltonian_simulation.hamiltonian_simulation_by_gqsp_test.test_hamiltonian_simulation_by_gqsp`
 - `qualtran.bloqs.phase_estimation.qubitization_qpe_test.test_qubitization_phase_estimation_of_walk`

### Breakages
 - ...
 - ...

### Related issues

 - #1488 -- If Cirq is allowed to decompose gates, it often gets down to `And(uncompue=True)` which says that it's not unitary. #1761 "fixes" this by doing tensor contraction in Qualtran and giving a big tensor to Cirq. This causes performance regression
 - #1336 -- The Cirq simulator is often faster for small-qubit-count circuits since it decomposes and simulates in one fell swoop instead of constructing a tensor network and then having quimb contract it

PR #1878, opened 2026-06-18
Speed up hamiltonian_simulation_by_gqsp_test and mark slow tests

mpharrigan:
By using `full_flatten=False`, we avoid contracting a tensor network with hundreds of tensors by short circuiting Controlled as well as SU2Rotation. Per #1336 #1763 -- flattening the tensor network took seconds

I also ran through `--durations=10` on `pytest-quick` running serially and marked two tests as "slow"


mpharrigan:
Marginal improvement on the CI, but on my workstation (with parallelization) it goes from 2:52 to to 1:22

PR #1879, opened 2026-06-19
Mark script export-from-uv-lock.sh as executable

mhucka:
While working on configuring pre-commit for Qualtran, one of the  pre-commit hooks flagged `dev_tools/requirements/export-from-uv-lock.sh` as being a shell script that didn't have the `x` permission set. This is truly the most trivial of PRs, but may as well do it.


mpharrigan (approved):
thanks. Worth noting that we don't use this in the modern CI flow, but perhaps good to have for backwards compatibility or dead-simple, portable requirements.txt file (as briefly explained in the README)

PR #1880, opened 2026-06-19
Add missing quotations around some variables in check/ scripts

mhucka:
Shellcheck reports a few cases of missing quotations around variable expressions. This PR adds them.


mpharrigan:
I get a little antsy when modifying these files that (at one point) were copies of the ones in cirq. @pavoljuhas what do you think: should we just be doing these fixes independently and incrementally?

pavoljuhas:
@mpharrigan - the fixes in this PR were already done for cirq scripts.  Some of these files have slightly diverged as Cirq, for example, moved tool configs to pyproject.toml.  I suppose we should just go on maintaining these scripts independently, that said, cherry-picking improvements one way or another is surely welcome.

check/autogenerate-notebooks
```diff
@@ -15,6 +15,6 @@ if output=$(git status --porcelain) && [ -z "$output" ]; then
   exit 0
 else
   # Uncommitted changes
-  echo $output
+  echo "$output"
```
pavoljuhas, line 18: nice - this keeps the status output as it were instead of joining to one line (or expanding wildcards if there were any).

pavoljuhas (approved):
LGTM for my part.

mhucka:
@mpharrigan Arguably, the best solution would be to have a single set of these common scripts, and have the projects use those instead of keeping separate copies as the projects do now. Echo'ing Pavol's reply, due to the fact that some of the scripts are customized for Qualtran or have diverged from the originals, I'm not sure how feasible that approach would be. It might even require more work than the current approach of occasional ad hoc per-repo housekeeping.

PR #1882, opened 2026-06-19
Fix #1881: update syntax of `pyproject.toml`

### LINKED ISSUE #1881: `pyproject.toml` fails validation
The `pyproject.toml` file has some incorrect keys, possibly due to using an older pyproject.toml syntax. These were found by running [validate-pyproject](https://validate-pyproject.readthedocs.io/en/latest/)

* `author` and `author_email` are not valid under `[project]`. The current accepted format is an array called `authors` (see PEP 621).

* The valid of the `license` field does not conform to PEP 621.

mhucka:
This updates a couple of the fields in `pyproject.toml` to use the currently-required syntax. In addition, this adds a license & copyright header comment.


pyproject.toml
```diff
@@ -1,16 +1,31 @@
...
 requires = ["hatchling"]
 build-backend = "hatchling.build"
 
 [project]
 name = 'qualtran'
 dynamic = ["version"]
-author = "Google Quantum AI"
-author_email = "mpharrigan@google.com"
+authors = [
+    { name = "Google Quantum AI", email = "mpharrigan@google.com" }
```
mpharrigan: do we have a better email for this?

mpharrigan (approved):

pyproject.toml
mhucka: > do we have a better email for this?

Oh yes, we can use quantum-oss-maintainers@google.com if you'd like?

Also, another option is to add a `maintainers` field (to provide both `authors` and `maintainers`),

```toml
maintainers = [
    { name = "Google Quantum AI", email = "quantum-oss-maintainers@google.com" }
]
```

and put individual people in the `authors` field. Or instead of people, you could put something like `The Qualtran Developers` in the authors field, following what we did for qsim. Example:

```toml
authors = [
    { name = "The Qualtran Developers", email = "qualtran-dev@googlegroups.com" }
]
maintainers = [
    { name = "Google Quantum AI", email = "quantum-oss-maintainers@google.com" }
]
```

mpharrigan: whatever you think is best!

mhucka: @mpharrigan Ok, then I would opt for that last option. The latest push has it.

PR #1884, opened 2026-06-19
Use faster runners and uv action caching in `pr.yaml`

mhucka:
This configures the CI workflow to use larger custom GitHub runners. In addition:

* Turns on the cache feature for `uv`
* Updates all GitHub Actions to their latest versions
* Adds an actionlint config file, so that devs can run actionlint locally on the workflow filesl


mpharrigan:
can you explain this:

> Tells actions/setup-python to get the Python version from pyproject.toml

there are still python-version fields as far as I can tell

mpharrigan:
What sort of speed up would you expect from the bigger runners? I haven't done a systematic study but iirc it used to be 6 mins and change for check/pytest-quick and now it's 5m. Notably, I can still run the suite on my macbook air in 1:56 🤷

mhucka:
> can you explain this:
> 
> > Tells actions/setup-python to get the Python version from pyproject.toml
> 
> there are still python-version fields as far as I can tell

Whoops, the description was out of date. I subsequently reverted that part back to explicitly setting the Python version. The description is up-to-date now.

mhucka:
> What sort of speed up would you expect from the bigger runners? I haven't done a systematic study but iirc it used to be 6 mins and change for check/pytest-quick and now it's 5m. Notably, I can still run the suite on my macbook air in 1:56 🤷

Hmmmm … I would have definitely expected a speedup for pytest. In TFQ and Cirq, we definitely saw speedups, IIRC cutting time by 1/2. I don't know why Qualtran's pytest run seems to have stayed the same, even though I see that in `check/pytest-quick`, it does pass the `-n` flag to pytest.

mpharrigan (approved):

PR #1892, opened 2026-07-01
Add a memory monitor to the nightly CI

mhucka:
Add a continuous memory & disk space monitor process in the background to report free memory & disk space every 10 seconds.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -25,12 +25,17 @@ on:
         type: string
         default: ""
 
+concurrency:
+  # Cancel any previously-started but still active runs on the same branch.
+  cancel-in-progress: true
+  group: ${{github.workflow}}-${{github.event.pull_request.number||github.ref}}
+
 # Declare default permissions as read only.
 permissions: read-all
 
 env:
   # Default value for "-n" argument to pytest with pytest-xdist plugin.
-  pytest-n: ${{inputs.processes || 'auto'}}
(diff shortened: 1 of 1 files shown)


fdmalone (approved):
not sure of the full context but would it be better to add the memory monitoring before making the other changes in order to measure their effect? Feel free to disregard though.

mhucka:
> not sure of the full context but would it be better to add the memory monitoring before making the other changes in order to measure their effect? Feel free to disregard though.

Yeah, you're right. I'll remove the extra changes before merging it.

.github/workflows/nightly.yaml
```diff
@@ -25,17 +25,12 @@ on:
...
-  cancel-in-progress: true
-  group: ${{github.workflow}}-${{github.event.pull_request.number||github.ref}}
-
 # Declare default permissions as read only.
 permissions: read-all
 
 env:
   # Default value for "-n" argument to pytest with pytest-xdist plugin.
-  pytest-n: ${{inputs.processes || 'logical'}}
+  pytest-n: ${{inputs.processes || 'auto'}}
```
pavoljuhas, line 38: This sets environment variable named `pytest-n` which cannot be accessed from a shell. Consider renaming to `pytest_n` or just use the context expression as is in the `uv run` command below.

mhucka: It's referenced using `${{env.pytest-n}}`. 

It does seem to work as-is. (Also, note that this env variable is not part of this PR, although the value is being changed.)

PR #1911, opened 2026-07-25
Fix Jacobi-Anger degree search stopping at a Bessel zero crossing

### LINKED ISSUE #1848: `degree_jacobi_anger_approximation` returns wrong degree at moderate `t` with loose `precision`
There is a bug in `qualtran/linalg/polynomial/jacobi_anger_approximations.py`. `|J_n(t)|` for `n < t` oscillates and crosses zero many times before asymptotic decay. The library function assumes monotone behavior in `n`, but it isn't monotonic in this region. As such `bisect_left` lands on the wrong zero.

The bug is in 59–65 of `jacobi_anger_approximations.py`:

```Python
d = 1
while not term_too_small(d):
    d *= 2
d = bisect.bisect_left(range(d), True, key=term_too_small) - 1
```

The doubling loop exits at the first `d` within epsilon, but that can be a zero crossing in the oscillating region. It can be seen by running the test (`jacobi_anger_approximations_test.py`) body with different params:

```Python
import numpy as np
from qualtran.linalg.polynomial.jacobi_anger_approximations import (
    approx_exp_cos_by_jacobi_anger,
    degree_jacobi_anger_approximation,
)

t = 50.0
precision = 1e-2

degree = degree_jacobi_anger_approximation(t, precision=precision)
print(f"degree = {degree}")

random_state = np.random.RandomState(42 + int(t))
P = np.polynomial.Polynomial(approx_exp_cos_by_jacobi_anger(t, degree=degree))
theta = 2 * np.pi * random_state.random(1000)
e_itheta = np.exp(1j * theta)
np.testing.assert_allclose(
    P(e_itheta) * e_itheta ** (-degree),
    np.exp(1j * t * np.cos(theta)),
    rtol=precision * 2,
)
```

The fix would scan monotonically past the oscillating region.

Sanjays2402:
Fixes #1848

`degree_jacobi_anger_approximation` searched for the smallest `n` with `|J_n(t)| <= precision` starting at `n = 1`, but `|J_n(t)|` oscillates through zero for `n < t`, so with a loose precision the doubling loop and bisection could stop at a zero crossing instead of the asymptotic tail. For the issue's case (`t = 50`, `precision = 1e-2`) it returned degree 15 because `|J_16(50)| = 0.005`, and the truncated expansion did not approximate `exp(i t cos(theta))` at all (max error 1.23). It now returns 56.

`|J_n(t)|` decays monotonically only for `n >= t`, so the search starts at `ceil(t)` and treats smaller `n` as not-yet-small, keeping the bisection predicate monotonic. The added test covers `t` in {20, 50} at loose precisions and fails on the current code.


mpharrigan:
thanks, can you address the bot's comments

Sanjays2402:
Both addressed in f19fa72.

`d_min` now uses `abs(float(t))`. The bot is right that `|J_n(-t)| = |J_n(t)|`, so the oscillatory region runs out to `|t|` and the old `float(t)` collapsed the bound to 1 for negative `t`, putting the search right back into the zero crossings.

Added `-20` and `-50` to the new test's parameters. One incidental fix: the seed expression was `42 + int(t)`, which numpy rejects for negative `t` ("Seed must be between 0 and 2**32 - 1"), so it's `42 + abs(int(t))` now.

Verified by reverting just the `abs` on this branch: `test_exp_cos_approximation_loose_precision[0.01--50]` fails, and passes with it. Whole file is 32 passed.

mhucka:
@Sanjays2402 Thank you for your work on this. When you get a chance, could you update the branch on this PR? Then we can examine the Mypy CI checks with the latest code in main.

Sanjays2402:
Updated the branch onto current main and fixed the mypy error introduced by the new negative-t test: the runtime result is narrowed with the same isinstance assertion used by the neighboring tests. The branch is clean at a12b2b8. The focused Jacobi-Anger test file passes (32 passed), targeted mypy passes, and ruff passes.

mhucka (approved):
Thank you for this contribution!

PR #1927, opened 2026-08-08
Update requirements and resolve version incompatibilities

mhucka:
Cloning the repository and doing a `uv sync` produced an environment where some tests failed locally. I suspect some third-party transitive dependencies may have changed recently. Resolving the failures required constraining two package versions in `pyproject.toml`. Following the addition of the constraints and refreshing `uv.lock`, a minor incompatibility arose involving a Matplotlib colormap function used in a notebook.


qualtran/surface_code/flasq/examples/ising_notebook.ipynb
```diff
@@ -2503,7 +2503,7 @@
     "\n",
     "# --- Create a Custom Asymmetric Colormap ---\n",
     "# Sample the standard 'RdBu_r' colormap to get its exact blue, white, and red.\n",
-    "rdbu_r = cm.get_cmap(\"RdBu_r\")\n",
+    "rdbu_r = colormaps[\"RdBu_r\"]\n",
```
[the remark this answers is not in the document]
mhucka: Another parsing error in GCA.

pyproject.toml
```diff
@@ -83,6 +83,10 @@ dependencies = [
     "seaborn",
     "joblib",
     "tqdm",
+
+    # lark 1.3.1 has a TypeError at import time.
+    # (_LexerArgType | None where _LexerArgType is a str)
+    "lark < 1.3.1",
```
mpharrigan: can we add an issue to track this or have some other escape hatch for when they fix it so this constraint doesn't cause problems down the line

mhucka: > can we add an issue to track this or have some other escape hatch for when they fix it so this constraint doesn't cause problems down the line

Good point! I changed it to `lark ! = 1.3.1`, which should be maximally flexible, I hope.

mpharrigan (approved):

mhucka:
@mpharrigan Some further testing locally revealed that the changes to the imports had nothing to do with the requirements version changes, so I took them out of this PR completely and will put them in a separate PR. Mentioning this because I saw that you had just approved this PR, and these changes came after the approval. To be safe, I'm requesting re-review.

mpharrigan (approved):

PR #1929, opened 2026-08-08
Pin CI dependencies via uv lock without re-resolving pyproject.toml

### LINKED ISSUE #1843: The CI will download and install non-pinned versions of dependencies
What's supposed to happen:

 - loose set of versions set in `pyproject.toml` for user installation
 - `uv lock` fixes all dependencies including transitive ones
 - The CI installs the fixed environment of dependencies, then installs qualtran with `--no-deps` so we only get the fixed set of dependencies

What happens:

 - The fixed environment of dependencies includes a dependency on `qualtran` itself, and will happily install non-pinned versions of the dependencies found in `pyproject.toml` even if they aren't in the `uv.lock`.

rosspeili:
Use --no-install-project plus editable --no-deps install in CI workflows, and --no-emit-project in export-from-uv-lock.sh, and also updated contributor and dev dependency docs to reflect the changes.

Fixes #1843.


.agents/rules/python_uv.md
```diff
@@ -6,4 +6,7 @@ description: "Enforce using uv for Python environment management, package instal
 # Python Environment Guidelines
 
 * Always use `uv` for managing Python virtual environments and dependencies.
-* Run Python commands, test suites, and linters via `uv run <command>` (e.g., `uv run check/pytest-quick`
\ No newline at end of file
+* Install the locked environment with `uv sync --frozen [--no-dev] [--group ...]
```
mpharrigan: I don't think this is necessary for agents

dev_tools/requirements/README
```diff
@@ -3,6 +3,12 @@ We've moved to `uv` to manage dependencies for developers and CI.
...
+ - Run `uv sync --frozen --no-dev [--group test/typing/lint/..] --no-install-project` to
+   install the pinned dependencies without installing qualtran itself.
+ - Run `uv pip install --no-deps -e .` to install qualtran in editable mode without
+   re-resolving dependencies from `pyproject.toml`. CI uses this two-step pattern so
+   only versions pinned in `uv.lock` are installed.
  - You can use `export-from-uv-lock.sh` in this directory to output legacy requirements.txt
-   environment files, but they are no longer committed to the repository.
+   environment files, but they are no longer committed to the repository. The script uses
+   `--no-emit-project` so exported files omit qualtran and avoid pulling unpinned
+   dependencies from `pyproject.toml`.
```
mpharrigan: a little superfluous since the legacy requirements.txt never included the project itself and there's the blurb above

CONTRIBUTING.md
```diff
@@ -71,6 +71,12 @@ Guides](https://google.github.io/styleguide/) in your code, particularly the
 
 Do your work and `git commit` your changes to your branch as needed.
 
+We use [`uv`](https://docs.astral.sh/uv/) to manage pinned dependencies for
+development and CI. See [`dev_tools/requirements/README`](./dev_tools/requirements/README)
+for how to install the locked environment and qualtran itself. In brief: use
+`uv sync --frozen [--no-dev] [--group ...] --no-install-project` followed by
+`uv pip install --no-deps -e .` so dependencies come only from `uv.lock`.
```
mpharrigan: again, I think this level of rigor is less important for day-to-day development work. I'd say using `uv` for development is recommended but not required

mpharrigan (commented):
thanks! some nits

rosspeili:
Thanks for the review @mpharrigan, trimmed the rules, CONTRIBUTING, and README per your nits and kept the CI/export changes and the two-step install notes in `dev_tools/requirements/README` since that's where CI behavior is documented. Let me know if anything else looks off. <3

mpharrigan (approved):
thank you!

PR #1932, opened 2026-08-10
Add more checks to pr.yaml

mhucka:
This adds more CI checks:

- a check that files contain a copyright & license header
- a YAML file linter
- a GitHub Actions workflow file linter


mpharrigan:
if we add additional checks, we'll need the ability to run offline for PR authors.

Can you comment on the license header check? we have one as part of the ruff checks

mhucka:
> Can you comment on the license header check? we have one as part of the ruff checks

Oh, I didn't know Ruff could do that. That's better. I'll remove that part from pr.yaml. Thanks for pointing this out; it will be useful to put in place for the other projects.

mhucka:
> if we add additional checks, we'll need the ability to run offline for PR authors.

That makes sense. What form should this ability take?

mhucka:
I should split up the test additions into separate PRs.  I'm going to close this one and redo it that way.

PR #1949, opened 2026-09-10
fix: CompactDataBlock.n_tiles() missing paper's +3 term (#1943)

### LINKED ISSUE #1943: CompactDataBlock uses ceil(1.5*n) instead of the paper's 1.5*n + 3 (arXiv:1808.02892, Fig. 9)
CompactDataBlock's tile-count formula computes `ceil(1.5 * n)`, but the paper it's cited to, Litinski's "A Game of Surface Codes: Large-Scale Quantum Computing with Lattice Surgery" (arXiv:1808.02892), states that the compact data block uses `1.5*n + 3` tiles. This is confirmed in three separate places on page 7 of the paper (the Figure 9 caption, the body text introducing the design, and the summary paragraph), and cross-checked against the ar5iv HTML rendering as well, to rule out a PDF text extraction artifact. All three agree.

For `n_algo_qubits=2` (a small worked example), this drops the tile count from the paper-correct 6 to Qualtran's current output of 3, a difference of exactly 3, which is consistent with a missing additive constant rather than a scaling or rounding error. At `data_d=17` this alone accounts for a 1,734 physical qubit undercount, roughly 38 percent of the paper-correct total for this example.

Repro:

```python
from qualtran.surface_code import CompactDataBlock

block = CompactDataBlock(data_d=17)
block.n_tiles(n_algo_qubits=2)  # returns 3
```

Expected (per cited paper, page 7, Fig. 9): `1.5*2 + 3 = 6`
Actual (code): `ceil(1.5*2) = 3`

For our own downstream use case this changes the reported physical qubit count by about 38 percent (2,880 to 4,614 for this small worked example), so this is not just a rounding scale issue.

The `n_steps_to_consume_a_magic_state=9` constant on the same class is transcribed correctly from the same page, so this does not look like an intentional simplification, just a dropped term.

Source location: `qualtran/surface_code/data_block.py`, `CompactDataBlock.n_tiles()`, lines 169-170
Paper reference: Litinski, arXiv:1808.02892, page 7, Figure 9 caption (also stated in the body text and summary paragraph on the same page)

Happy to open a PR if a fix along these lines is welcome.

kkoci:
Closes #1943.

Main's current `n_tiles()` computes `ceil(1.5 * n_algo_qubits + 2)`. This differs from the cited paper (Litinski, arXiv:1808.02892, page 7, Fig. 9: "A compact block stores n data qubits in 1.5n+3 tiles") due to operator precedence `ceil(1.5*n + 2)` is not the same as `ceil(1.5*(n+2))` for most n (e.g. n=2 gives 5 here, vs. the paper-correct 6).

This adds parentheses to match the paper: `ceil(1.5 * (n_algo_qubits + 2))`, algebraically `1.5n+3`, matching the fix proposed and agreed on in the issue discussion.

Before: n_algo_qubits=2 -> 5 tiles
After:  n_algo_qubits=2 -> 6 tiles (matches the paper)

Added a parametrized test in `data_block_test.py` (`test_compact_block`) covering n=2, 4, 100 against the paper-correct tile count.


mpharrigan (approved):
thanks!

